PR #6, opened 2021-05-28
Maven: integration test config, and new repositories for akkasls

franciscolopezsancho:
updating pom to the latest and greatest


ennru (approved):
LGTM.

PR #8, opened 2021-05-31
Refactor Value Entity to use codegen

octonato:
Not yet to merge since docs are referencing the samples. 

TODO

- [x] make CI build the maven project instead using current SDK snapshot
- [x] refactor integration test to use skeleton from codegen

PR #10, opened 2021-06-01
Base shopping cart Value Entity on codegen

octonato:
Supersedes #8

PR from branch instead of fork so I can send other PRs on top of this.


build.sbt
```diff
@@ -227,27 +226,6 @@ lazy val attachProtobufDescriptorSets = Seq(
...
-    // Akka gRPC only for IntegrationTest
-    Compile / akkaGrpcGeneratedSources := Seq.empty,
-    Compile / javacOptions ++= Seq("-encoding", "UTF-8", "-source", "11", "-target", "11"),
-    testOptions += Tests.Argument(TestFrameworks.JUnit, "-q", "-v", "-a"),
-    IntegrationTest / akkaGrpcGeneratedLanguages := Seq(AkkaGrpc.Java),
-    IntegrationTest / akkaGrpcGeneratedSources := Seq(AkkaGrpc.Client),
-    IntegrationTest / PB.protoSources ++= (Compile / PB.protoSources).value
-  )
-  .settings(attachProtobufDescriptorSets)
-
```
octonato, line 250: Maybe we should keep a minimal sbt build just to run the check headers and javafmt.

octonato: That won't work without some extra tweaks. 

sbt can't compile projects that moved to codegen, so we will need to exclude the projects from compile/test and only have then running the checkers. 

I will defer this since not extremely important.

samples/java-valueentity-shopping-cart/src/main/proto/shoppingcart/shoppingcart_api.proto
```diff
@@ -26,63 +26,62 @@ import "google/api/annotations.proto";
...
 service ShoppingCartService {
-  option (akkaserverless.service) = {
+    option (akkaserverless.service) = {
     type: SERVICE_TYPE_ENTITY
     component: ".domain.ShoppingCart"
   };
 
-  rpc AddItem(AddLineItem) returns (google.protobuf.Empty) {
-    option (google.api.http) = {
-      post: "/ve/cart/{cart_id}/items/add"
```
octonato, line 66: Removing the `/ve` will allow us to use the same UI for both event sourced and value entity samples. 

The protobuf is nearly the same. The Value Entity has an extra `RemoveCart`. UI doesn't have it now since it's currently based on Event Sourcing. 

Later, we can add a `RemoveCart` to the ES sample to show how to implement a tombstone. 

For the record, the UI I'm talking about is this one: https://github.com/cloudstateio/samples-ui-shoppingcart 

This should be ported to akka serverless now.

.circleci/config.yml
```diff
@@ -137,6 +137,29 @@ jobs:
...
+      - checkout-and-merge-to-main
+      - setup_sbt
+      - restore_sbt_cache
+      - run: 
+          name: "Publish local maven"
+          command: | 
+            # tag the head with the hash
+            # dynver requires a version number prefix, ie: v0.0.1-
+            git tag v0.0.1-"$CIRCLE_SHA1"
+            echo 'export SDK_VERSION=0.0.1-${CIRCLE_SHA1}' >> $BASH_ENV
```
pvlugter, line 153: I haven't followed what's been tried, but since this was mentioned on the call.

Instead of tagging, could it use `sbt -q "print sdk/version"` to get the version that will be published locally?

So something like:

```
echo "export SDK_VERSION=$(sbt -q "print sdk/version")" >> $BASH_ENV
```

octonato: That's similar to what I was doing, but my version were more hacky. 

I was even printing the SDK_VERSION surrounded by '' to make sure there were nothing wrong with it. 

Still, when maven was fetching it, it was pulling something like `%1B%5B0m%%1B%5B0m%0.7.0-beta.10-6-1cd5e2ed-SNAPSHOT%1B%5B0m%`

Anyway, it can be related to the way I was trying to pick the version. I will give it another try.

octonato: it works with 

```shell
echo 'export SDK_VERSION='`sbt "print sdk/version" | tail -n 2 | head -n 1` >> $BASH_ENV
```
The output without `print` was probably adding some invisible chars.

samples/java-valueentity-shopping-cart/src/it/java/com/example/valueentity/shoppingcart/domain/ShoppingCartIntegrationTest.java
```diff
@@ -25,18 +25,25 @@
 import java.util.List;
 
 import static org.junit.Assert.assertEquals;
+import static java.util.concurrent.TimeUnit.*;
 
+// Example of an integration test calling our service via the Akka Serverless proxy
+// Run all test classes ending with "IntegrationTest" using `mvn verify -Pfailsafe`
```
ennru, line 31:
```suggestion
// Run all test classes ending with "IntegrationTest" using `mvn verify -Pit`
```
I guess the codegen needs to be fixed...

ennru (approved):
LGTM.

samples/java-valueentity-shopping-cart/src/it/java/com/example/valueentity/shoppingcart/domain/ShoppingCartIntegrationTest.java
ennru: https://github.com/lightbend/akkaserverless-codegen/pull/142

PR #13, opened 2021-06-02
[trial] get version using sbt

octonato:
Just to test a different approach. If it works I will integrate it on original PR.


octonato:
this works. it's now integrated in #10

PR #15, opened 2021-06-04
bump: update akka-grpc to 2.0.0

pvlugter:
I haven't bumped the akka-grpc version in the samples. They fail to build with the java-sdk on beta.10 and akka-grpc plugin on 2.0.0. While it seems fine with newer java-sdk (built against akka-grpc 2.0.0) and the older akka-grpc plugin — and this should be tested in CI as well.

So probably best to wait with bumping the samples until next java-sdk release, and bump both versions together. And also do this in the maven archetypes.

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -10,8 +10,6 @@ object Dependencies {
   // changing the Scala version of the Java SDK affects end users
   val ScalaVersion = "2.13.5"
 
-  // Note: sync with Protobuf version in Akka gRPC and ScalaPB
-  // https://github.com/akka/akka-grpc/blob/v2.0.0/project/Dependencies.scala#L26
   val ProtobufVersion = akka.grpc.gen.BuildInfo.googleProtobufVersion
 
   val AkkaVersion = "2.6.14"
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -12,7 +12,7 @@ object Dependencies {
 
   // Note: sync with Protobuf version in Akka gRPC and ScalaPB
   // https://github.com/akka/akka-grpc/blob/v2.0.0/project/Dependencies.scala#L26
-  val ProtobufVersion = "3.17.1"
+  val ProtobufVersion = akka.grpc.gen.BuildInfo.googleProtobufVersion
 
   val AkkaVersion = "2.6.14"
   val AkkaHttpVersion = "10.2.4" // Note: should at least the Akka HTTP version required by Akka gRPC
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -11,8 +11,8 @@ object Dependencies {
   val ScalaVersion = "2.13.5"
 
   // Note: sync with Protobuf version in Akka gRPC and ScalaPB
-  // https://github.com/akka/akka-grpc/blob/v1.1.1/project/Dependencies.scala#L77
-  val ProtobufVersion = "3.15.1"
+  // https://github.com/akka/akka-grpc/blob/v2.0.0/project/Dependencies.scala#L26
+  val ProtobufVersion = "3.17.1"
 
   val AkkaVersion = "2.6.14"
   val AkkaHttpVersion = "10.2.4" // Note: should at least the Akka HTTP version required by Akka gRPC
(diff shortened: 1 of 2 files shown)


project/Dependencies.scala
```diff
@@ -11,8 +11,8 @@ object Dependencies {
   val ScalaVersion = "2.13.5"
 
   // Note: sync with Protobuf version in Akka gRPC and ScalaPB
-  // https://github.com/akka/akka-grpc/blob/v1.1.1/project/Dependencies.scala#L77
-  val ProtobufVersion = "3.15.1"
+  // https://github.com/akka/akka-grpc/blob/v2.0.0/project/Dependencies.scala#L26
+  val ProtobufVersion = "3.17.1"
```
raboof, line 15: Could be

```suggestion
  val ProtobufVersion = akka.grpc.gen.BuildInfo.googleProtobufVersion
```

raboof (approved):
Change seems good.

So this means the examples in this repo are not built against the SDK in this repo, but against the previously-released SDK? Perhaps we should update our CI setup so that the examples are ran against the 'current' version of the SDK? That would give us a better opportunity to test changes to the SDK (though also mean that if we want to publish a change to the samples without publishing the SDK, we'd need to come up with a workflow for that, e.g. backporting those to some other branch)

octonato:
@raboof we are starting to build the samples with the current (in this branch) SDK version. 
You can see it here: https://github.com/lightbend/akkaserverless-java-sdk/blob/main/.circleci/config.yml#L147-L159

The goal is to that for all the samples.

octonato:
@raboof, I created an issue #17 to discuss the topic you raised here related with versioning and publishing of the samples.

octonato:
@pvlugter, I applied Arnout's suggestion to the branch so we can rollout this sooner.

raboof:
> we are starting to build the samples with the current (in this branch) SDK version.

Ah, that also explains why the integration tests fail on CI then ;). So perhaps we should update akka-grpc in the examples after all, even though it means you will always have to run them with `-Dakkaserverless-sdk.version`?

project/Dependencies.scala
```diff
@@ -11,8 +11,8 @@ object Dependencies {
   val ScalaVersion = "2.13.5"
 
   // Note: sync with Protobuf version in Akka gRPC and ScalaPB
-  // https://github.com/akka/akka-grpc/blob/v1.1.1/project/Dependencies.scala#L77
-  val ProtobufVersion = "3.15.1"
+  // https://github.com/akka/akka-grpc/blob/v2.0.0/project/Dependencies.scala#L26
```
ennru, line 14: As we have it from BuildInfo now, we can remove the sync comments.

project/Dependencies.scala
pvlugter: Good, much better

pvlugter:
CI was failing with maven central download hiccups.

Yes, so CI does run the samples against the current development version. But the samples otherwise have the last release. The current SDK works with the samples on akka-grpc 1.1.1, but not the older beta.10 SDK with akka-grpc 2.0.0. So we can wait with updating the akka-grpc version in the archetypes and the samples until next SDK release, so that the default builds also work for now.

project/Dependencies.scala
pvlugter: Removed the comments

ennru (approved):
LGTM.

pvlugter:
> But why does this succeed when the Maven samples are not updated to use Akka gRPC 2.0.0 for the integration test client?

I probably wasn't clear enough. The combination that does not work is the SDK on old Akka gRPC 1.1.1 and the project protobuf generation on new Akka gRPC 2.0.0 — haven't looked closely, but generation from new plugin with old runtime dependency is likely the problem. So the samples when on old SDK (beta.10) will not work with the Akka gRPC update. While a new SDK (which will pull Akka gRPC 2.0.0 runtime) will still work with the old Akka gRPC 1.1.1 generation. And so the CI sample builds (which overrides to the development version) will work either way. Overriding the runtime dependency could also work for running new Akka gRPC plugin with the old SDK.

My suggestion: first update Akka gRPC to 2.0.0 for the SDK (this PR currently). Release a new SDK version, which should work with either plugin in projects. Then update Akka gRPC in archetypes and samples.

PR #16, opened 2021-06-04
fix: formatting protos

octonato:
Formats Shopping Cart VE and Counter VE protos with indent 2. 

This seems to be tricky when using intellij and antora callouts `// <1>`, as the formatter will align with the callout, like in:

```proto
message IncreaseValue { // <4>
  						string counter_id = 1 [(akkaserverless.field).entity_key = true];
  						int32 value = 2;
}
```

The best would be to have a formatter that does what we want at build time, so we don't run the risk of inadvertently format the protos, miss the bad formatting and messing up with the docs.


ennru (approved):
LGTM.

PR #19, opened 2021-06-04
fix: use com.example as groupId

octonato:
Aligns with the new Counter sample.


ennru (approved):
The group ID isn't as important as the package names. It is common to have them be the same, but not required.

octonato:
> The group ID isn't as important as the package names. It is common to have them be the same, but not required.

Yeah, I know. It's just me and my [OCD](https://en.wikipedia.org/wiki/Obsessive%E2%80%93compulsive_disorder)

Issue #17, opened 2021-06-04
Automatically build and publish sample  imagges


octonato:
In #15, we discussed how we could manage the publishing of samples. 

I don't think we ever discussed a strategy for that. Hence this issu.

octonato:
There is also the case in which we don't want to release a new SDK, but want to bring an update to the image. 

I don't expect it to happen so often, so we may just apply a manual workaround, ie: release it manually, maybe with a tag name that we can recognize as a in-between releases image. 

The manually published image will be the latest and will be available for usage through the console.

PR #21, opened 2021-06-04
fix: adds instruction on how to publish the sample images

octonato:
Explain how to build and publish the sample images.

Related to #17. We still need to decide if we will publish images on release and which version number the should have.


pvlugter (approved):
Yes, we should figure out which versions to use. If we publish these on release, then maybe we should just use the SDK version?

octonato:
> Yes, we should figure out which versions to use. If we publish these on release, then maybe we should just use the SDK version?

I think so and when we want to publish a in between release image, we can append a number to it. I think we can have a more lossy process for these edge cases.

PR #23, opened 2021-06-07
fix: Delay shutdown until proxy has terminated

patriknw:
Ready for review, but keeping as draft until we have released protocol addition.

### DIFF
--- a/sdk/src/main/scala/com/akkaserverless/javasdk/impl/DiscoveryImpl.scala
+++ b/sdk/src/main/scala/com/akkaserverless/javasdk/impl/DiscoveryImpl.scala
@@ -25,17 +25,30 @@ import com.akkaserverless.protocol.discovery.PassivationStrategy.Strategy
 import com.akkaserverless.protocol.discovery._
 import com.google.protobuf.DescriptorProtos
 import java.time.Duration
+import java.util.concurrent.atomic.AtomicReference
 
 import scala.concurrent.Future
+import scala.concurrent.Promise
 import scala.io.Source
 import scala.jdk.CollectionConverters._
 import scala.util.control.NonFatal
 
+import akka.Done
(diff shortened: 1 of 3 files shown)


sdk/src/main/scala/com/akkaserverless/javasdk/impl/DiscoveryImpl.scala
```diff
@@ -67,6 +80,7 @@ class DiscoveryImpl(system: ActorSystem, services: Map[String, Service]) extends
       // only (silently) send service info for hybrid proxy version probe
       Future.successful(Spec(serviceInfo = Some(serviceInfo)))
     } else {
+      proxyTerminatedRef.getAndSet(Promise[Done]()).trySuccess(Done)
```
pvlugter, line 83: Not sure the `getAndSet` and `trySuccess` on the old value is needed, given that it's already set to successful promise, but no harm in doing so I guess.

pvlugter (approved):
LGTM

sdk/src/main/scala/com/akkaserverless/javasdk/impl/DiscoveryImpl.scala
patriknw: yeah, I was just thinking about if the proxy would go havoc and making multiple new discovery calls while the shutdown is in progress (waiting on old promise instance). Doesn't hurt to try to complete the old promise. Anyhow, it's not a big deal if there would be a race because in the end there is a timeout of 5 seconds for the coordinated shutdown phase.

pvlugter: cool, makes sense

patriknw:
updated to released framework `0.7.0-beta.10`. Ready for merge.

PR #26, opened 2021-06-11
sample: add counter using ReplicatedEntity

patriknw:
* based on the valueentity-counter but using ReplicatedEntity
* using codegen like classes, but codegen for ReplicatedEntity doesn't exist yet

I will use this as baseline when trying out SDK API changes for ReplicatedEntity. We can merge this first.


samples/replicatedentity-counter/src/main/java/com/example/domain/CounterImpl.java
```diff
@@ -0,0 +1,58 @@
...
+ */
+@ReplicatedEntity // <1>
+public class CounterImpl extends CounterInterface {
+    @SuppressWarnings("unused")
+    private final String entityId;
+    private final PNCounter counter;
+
+    public CounterImpl(ReplicatedEntityCreationContext context) { // <2>
+        this.entityId = context.entityId();
+        this.counter = context.state(PNCounter.class).orElseGet(context::newPNCounter);
```
patriknw, line 20: By mistake I first used GCounter here, strange that it worked with decrement. Anyway, changed to PNCounter

samples/replicatedentity-counter/dependency-reduced-pom.xml
```diff
@@ -0,0 +1,335 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
pvlugter, line 1: Should the `dependency-reduced-pom.xml` file be committed, or gitignored?

pvlugter (commented):
Cool, LGTM.

Should `GCounter` be removed in this PR though?

samples/replicatedentity-counter/dependency-reduced-pom.xml
patriknw: it's generated by the maven plugin, I removed it again and added to gitignore

patriknw:
> Should GCounter be removed in this PR though?

thanks, that was supposed to be in another PR, reverted

samples/replicatedentity-counter/README.md
```diff
@@ -0,0 +1,78 @@
...
+
+```shell
+mvn compile exec:java
+```
+
+For further details see [Running a service locally](https://developer.lightbend.com/docs/akka-serverless/developing/running-service-locally.html) in the documentation.
+
+## Exercise the service
+
+With both the proxy and your application running, any defined endpoints should be available at `http://localhost:9000`. In addition to the defined gRPC interface, each method has a corresponding HTTP endpoint. Unless configured otherwise (see [Transcoding HTTP](https://docs.lbcs.dev/js-services/proto.html#_transcoding_http)), this endpoint accepts POST requests at the path `/[package].[entity name]/[method]`. For example, using `curl`:
```
pvlugter, line 53: Broken link for HTTP transcoding (will be in the original archetype):

```suggestion
With both the proxy and your application running, any defined endpoints should be available at `http://localhost:9000`. In addition to the defined gRPC interface, each method has a corresponding HTTP endpoint. Unless configured otherwise (see [Transcoding HTTP](https://developer.lightbend.com/docs/akka-serverless/java/proto.html#_transcoding_http)), this endpoint accepts POST requests at the path `/[package].[entity name]/[method]`. For example, using `curl`:
```

pvlugter (approved):

samples/replicatedentity-counter/README.md
pvlugter: PRs for archetypes (#37) and other sample readmes (#38)

PR #31, opened 2021-06-12
changing naming convention from myentity to counter

franciscolopezsancho:


octonato:
Thanks @franciscolopezsancho 

@huntc, I believe it's a good time to change the ES archetype to the `Counter` entity. 

It will take some time before we can publish the archetypes from this branch and since the archetype will need to have a new name because the version misalignment we are rather safe now to change it.

That means that the codegen build won't break because it depends on the old archetype. When we move the codegen-java here, we can then adapt the tests to use the event sourced `Counter` entity.

octonato:
@franciscolopezsancho, this deletion wasn't intentional. That happens automatically when I merged lightbend:rgc/maven-plugin-repo-merge. 

I don't know why that happened. Usually, GitHub would automatically retarget your branch.

PR #32, opened 2021-06-15
chore: rename docs module from java-services to java

pvlugter:
Rename from `java-services` to `java`. Keep a copy at the old location during the transition. Already published to the [`docs/current` branch](https://github.com/lightbend/akkaserverless-java-sdk/tree/docs/current/docs/build/src/managed/modules).


ennru (approved):
LGTM.

PR #34, opened 2021-06-15
cache actionHandlers on creation

jrudolph:
Otherwise, all the `ActionReflection` code (looking up methods and annotations, etc) will be re-executed for every single call.


johanandren (approved):
Good catch!

sdk/src/main/scala/com/akkaserverless/javasdk/impl/action/ActionsImpl.scala
```diff
@@ -57,6 +57,7 @@ final class ActionsImpl(_system: ActorSystem, services: Map[String, ActionServic
   private object creationContext extends ActionCreationContext {
     override def serviceCallFactory(): ServiceCallFactory = rootContext.serviceCallFactory()
   }
+  private val actionHandlers: Map[String, ActionHandler] = services.view.mapValues(_.factory.create(creationContext)).toMap
```
patriknw, line 60: I don't think we say anything about the expected concurrency in the action instances, or have a good reason for creating a new instance for each request, so I think this change is fine.

patriknw (approved):
LGTM

PR #36, opened 2021-06-15
Integrate maven-java into CI build

octonato:


.circleci/config.yml
```diff
@@ -95,62 +106,94 @@ jobs:
...
+            mvn -B archetype:generate -DgroupId=com.example -DartifactId=counter-value-entity -DarchetypeGroupId=com.lightbend -DarchetypeArtifactId=maven-archetype-akkasls -DarchetypeVersion=$SDK_VERSION
+            cd counter-value-entity
+            mvn -B -Dakkaserverless-sdk.version=$SDK_VERSION compile       
+      - run: 
+          name: Run integration tests for Event Sourced Entity Archetype
+          command: |
+            cd maven-java
+            mvn -B archetype:generate  -DgroupId=com.example -DartifactId=counter-event-sourced-entity -DarchetypeGroupId=com.lightbend -DarchetypeArtifactId=maven-archetype-akkasls-event-sourced-entity -DarchetypeVersion=$SDK_VERSION
+            cd counter-event-sourced-entity
+            mvn -B -Dakkaserverless-sdk.version=$SDK_VERSION compile
```
octonato, line 183: I'm taking a different approach here than the original build GH Action. The original build was deploying the proxy, the user service and then hitting it with curl. 

I confess, that the main reason that I'm not reproducing it here is because it's a little more complicated to do it in CircleCI. I could have tried harder, but at this stage I'm taking a shortcut. I prefer to move faster with the merge of the projects. 

Also, I believe we will be better served if we use the integration test testkit with a fully implement application and this is something I would like to add once we move the codegen under the SDK project.

ennru (approved):
LGTM.

PR #45, opened 2021-06-17
fix: use com.akkaserverless groupId and renamed artifacts

octonato:


.circleci/config.yml
```diff
@@ -171,14 +171,14 @@ jobs:
           name: Run integration tests for Value Entity Archetype
           command: |
             cd maven-java
-            mvn -B archetype:generate -DgroupId=com.example -DartifactId=counter-value-entity -DarchetypeGroupId=com.lightbend -DarchetypeArtifactId=maven-archetype-akkasls -DarchetypeVersion=$SDK_VERSION
+            mvn -B archetype:generate -DgroupId=com.example -DartifactId=counter-value-entity -DarchetypeGroupId=com.lightbend -DarchetypeArtifactId=akkaserverless-maven-archetype -DarchetypeVersion=$SDK_VERSION
```
ennru, line 174:
```suggestion
            mvn -B archetype:generate -DgroupId=com.example -DartifactId=counter-value-entity -DarchetypeGroupId=com.akkaserverless -DarchetypeArtifactId=akkaserverless-maven-archetype -DarchetypeVersion=$SDK_VERSION
```

.circleci/config.yml
```diff
@@ -171,14 +171,14 @@ jobs:
...
             cd counter-value-entity
             mvn -B -Dakkaserverless-sdk.version=$SDK_VERSION compile       
       - run: 
           name: Run integration tests for Event Sourced Entity Archetype
           command: |
             cd maven-java
-            mvn -B archetype:generate  -DgroupId=com.example -DartifactId=counter-event-sourced-entity -DarchetypeGroupId=com.lightbend -DarchetypeArtifactId=maven-archetype-akkasls-event-sourced-entity -DarchetypeVersion=$SDK_VERSION
+            mvn -B archetype:generate  -DgroupId=com.example -DartifactId=counter-event-sourced-entity -DarchetypeGroupId=com.lightbend -DarchetypeArtifactId=akkaserverless-maven-archetype-event-sourced-entity -DarchetypeVersion=$SDK_VERSION
```
ennru, line 181:
```suggestion
            mvn -B archetype:generate  -DgroupId=com.example -DartifactId=counter-event-sourced-entity -DarchetypeGroupId=com.akkaserverless -DarchetypeArtifactId=akkaserverless-maven-archetype-event-sourced-entity -DarchetypeVersion=$SDK_VERSION
```

maven-java/akkaserverless-maven-archetype-event-sourced-entity/pom.xml
```diff
@@ -1,11 +1,11 @@
 <project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd">
     <modelVersion>4.0.0</modelVersion>
 
-    <artifactId>maven-archetype-akkasls-event-sourced-entity</artifactId>
+    <artifactId>akkaserverless-maven-archetype-event-sourced-entity</artifactId>
```
ennru, line 4:
```suggestion
    <artifactId>maven-archetype-event-sourced-entity</artifactId>
```
As we get `akkaserverless` in the group ID, we could drop it from the artifact ID.

ennru (commented):
Drop `akkaserverless-` from the module and artefact names?

maven-java/akkaserverless-maven-archetype-event-sourced-entity/pom.xml
octonato: Yeah, we said it on the call, but when changing it I found it would be better to keep it. Mainly because of the value-entity one. It feels a bit naked if only `maven-archetype` for the default one. 

But I agree that we are being repetitive when saying `com.akkaserverless:akkaserverless-maven-archetype`

octonato: An alternative could be the following:

`starter` for the the default one
(uses value-entity)

```
 mvn archetype:generate \                                                                                                                                                                         <rgc/rename-artifacts-for-publishing>
  -DarchetypeGroupId=com.akkaserverless \
  -DarchetypeArtifactId=maven-project-starter \
  -DarchetypeVersion=LATEST
```

and then...

```
 mvn archetype:generate \                                                                                                                                                                         <rgc/rename-artifacts-for-publishing>
  -DarchetypeGroupId=com.akkaserverless \
  -DarchetypeArtifactId=maven-project-event-sourced-entity \
  -DarchetypeVersion=LATEST
```

Then, later on, if we decide that the default is without any proto file, we can keep the name `maven-project-starter` and we add another one called `maven-project-value-entity`.

ennru: Let's stick to `archetype` as that is Maven-speak.

We should not introduce another thing like "starter". "kickstart" would align with what we use in the docs right now, but it has been questioned if that is a good word.

I don't see a problem with the naked "maven-archetype", actually.

octonato: shall we also remove `akkaserverless` from `akkaserverless-maven-plugin`?

that will make: 

```
com.akkaserverless % maven-plugin
com.akkaserverless % maven-archetype
com.akkaserverless % maven-archetype-event-sourced-entity
```

ennru: Yes, that should appeal to your need for consistency. ;-)

octonato: well, unfortunately that won't work

```
Artifact Ids of the format maven-___-plugin are reserved for 
plugins in the Group Id org.apache.maven.plugins
Please change your artifactId to the format ___-maven-plugin
In the future this error will break the build.
```

plugins starting with `maven-plugin` are reserved

ennru (approved):
LGTM.

PR #57, opened 2021-06-18
first iteration, test passing

franciscolopezsancho:
I created a Makefile where you can` run-all` to verify all is in sync. Right now passes which I'm not sure is completely right. I'll double check now.


franciscolopezsancho:
@johanandren  we already have in place in CI verification that codegen, maven-java and samples(valueentity counter) are in sync. This draft was about achieving that so I reckon better close it. WDYT?

raboof:
> This draft was about achieving that so I reckon better close it. WDYT?

I think this also covers the archetype, right? That seems like it'd be good to have.

johanandren:
I think we can close this since we already have ci coverage for generating project out of (repo version of) archetype and compiling that.

PR #59, opened 2021-06-22
Implementation of user function health check method

johanandren:
Impl of new health check method in discovery service, depends on https://github.com/lightbend/akkaserverless-framework/pull/759


johanandren:
I think in the future it could make sense to do more actual health checking, but for now let's just go with checking if the user function server can respond.

patriknw:
This was included in https://github.com/lightbend/akkaserverless-java-sdk/pull/104

Issue #58, opened 2021-06-21
Build maven plugin and archetype using sbt


octonato:
When migrating the maven plugin and archetypes here, I keep it as maven project. But it requires some hacking to build it using the same version as the sbt project. 

Lagom has a maven plugin and archetype that are build with sbt. I think we should do the same here. 

Then we don't need to hack things around and when publish we only need to deal with sbt release. That will make our life easier.

raboof:
> Lagom has a maven plugin and archetype that are build with sbt. I think we should do the same here

Akka gRPC does the same thing. This requires some massaging since we can't simply use the maven-plugin-tools anymore, and instead have to introduce our own `project/SbtMavenPlugin.scala`. I wonder if this is really worth it...

octonato:
Is `SbtMavenPlugin.scala` the only thing we need? 

The advantage of doing everything from the sbt is that we don't need extra tweaks to align the versions and the publishing gets simplified.

raboof:
> Is `SbtMavenPlugin.scala` the only thing we need?

it seems we also need to 'manually' write the plugin.xml (https://github.com/lagom/lagom/blob/master/dev/maven-plugin/src/main/maven/plugin.xml) instead of having it generated for us from the Mojo code and comments. Seems tricky to keep in sync, unless I'm missing something?

octonato:
@raboof, only seeing your reaction now. Yes indeed, we will need to keep it aligned, but I don't think that's tricky. Some extra work, but feasible. 

On the other hand, we do have a solution now that works. A little hacky, but it works. So, I'm closing this issue. I don't think that the situation we have now is such a terrible hack that would justify refactoring it. 

We can re-assess if that really bites us in the future.

Issue #27, opened 2021-06-11
[epic] Bring in maven plugin, archetype and java codegen


octonato:
As discussed, we plan to merge into this project the maven plugin, the maven archetypes and the java codegen. 

The execution plan is as follow: 

## Phase 1:

- [x] **move**: bring `maven-java` under `java-sdk` (preserving commit history) #28 
- [x] **build**: must be integrated in CircleCI (similar to what we are doing to the samples already) https://github.com/lightbend/akkaserverless-java-sdk/pull/36
- [x] **versioning**: we want the maven plugin and archetypes to have the same version of the java-sdk. https://github.com/lightbend/akkaserverless-java-sdk/pull/65
- [x] **publishing**: archetypes and plugin must be published under a new name because of version misalignment https://github.com/lightbend/akkaserverless-java-sdk/pull/45 and https://github.com/lightbend/akkaserverless-java-sdk/pull/65
- [x] **docs**: documentation must be adapted since users won't need to change anything in their `settings.xml` file and we will have a new artifact name #96 
- [x] **github issues**: move all open issues from `maven-java` to `java-sdk`

## Phase 2:

First we need to split codegen into codegen-core, codegen-java and codegen-js (preserving commit history) See https://github.com/lightbend/akkaserverless-codegen/issues/147

- [x] move `codegen-java` under `java-sdk` (preserving commit history)

Please, feel free to add here any other point of action that you believe we will need.

ennru:
@octonato Common believe is that everything is covered now. It would be good if you'd check and try a release.

octonato:
Yep, I think we have everything.

octonato:
For the record, the publishing of artifacts didn't work. The reason is that we call `publishM2` a few times in the build and then we get a message that artifact is already published (in local).

That brings us back to the cache issue. We should call `publishM2` only once in the build, but we need to get to understand why the m2 cache is not working as expected.

octonato:
After a few build tweaks and burned release numbers we have v0.7.0-beta.15 released including the codegen, the maven plugin, the archetypes and the SDK all together. 

We can close this issue! 🎉

PR #60, opened 2021-06-22
Build and deploy maven-plugin with sbt

raboof:
Refs #58 in preparation for #27


raboof:
abandoning this in favor of #65, so we can keep leveraging more of the maven tooling to generate the plugin.xml from source

PR #61, opened 2021-06-22
feat: brings in the java codegen + codegen core

octonato:
To accelerate things we are bringing in the codegen core as well. We will later decide if it spins-off as another library or if we copy that same code for JS. 

This is not yet ready. We need to integrate build into the sdk build. I'm keeping it as draft for now.

Part of #27

TODOs
- [ ] integrate into the sdk sbt build
- [ ] add back the integration tests - may re-think then because now we have SDK and maven archetype on same place
- [ ] rename the package and groupId to `com.akkaserverless`


octonato:
I will start over after the some open PRs in codegen are merged.

Issue #50, opened 2021-06-17
Manage state and use Effect type


patriknw:
One thing that calls attention is how different are the APIs for Value Entity, Event Sourced Entity, View, ReplicatedEntity.

This has been discussed in https://github.com/lightbend/akkaserverless/discussions/4306#discussioncomment-872700 and several other places so I will not repeat the motivation here.

Actions:
- [ ] Manage state for EventSourcedEntity instead of instance variables in user code
- [ ] Remove `@Snapshot` / `@SnapshotHandler`
- [ ] Access current state via method parameter instead of context
- [ ] Emit event and ValueEnitty state update via returned Effect instead of via context
- [ ] Remove context.failure since returning `Effect.failure` covers that

This also means that the current `Reply` type will be renamed to `Effect`, and it is also used for defining the reply, forward and failure.

PR #62, opened 2021-06-23
feat: Spike of API changes, #50

patriknw:
API proposal that is introducing the Effect return type and changes to state management.

This is a "compile only" spike until we agree on the API.

Probably easies to look at the changes of the examples:
* valueentity-counter ... CounterImpl.java
* java-eventsourced-shopping-cart ... ShoppingCartImpl.java

Refs #50


sdk/src/main/java/com/akkaserverless/javasdk/eventsourcedentity/EventSourcedEntityEffect.java
```diff
@@ -0,0 +1,130 @@
...
+ * A return type to allow returning forwards or failures, and attaching effects to messages.
+ *
+ * @param <T> The type of the message that must be returned by this call.
+ */
+public abstract class EventSourcedEntityEffect<T> implements Effect<T> {
+
+  public static class Builder<T, S> {
+    private Builder() {}
+
+    public Builder2<T, S> emitEvent(Object event) {
```
patriknw, line 43: `Object` here is not great, but there is no interface that all events would implement.

sdk/src/main/java/com/akkaserverless/javasdk/eventsourcedentity/EventSourcedEntityEffect.java
```diff
@@ -0,0 +1,130 @@
...
+      throw new UnsupportedOperationException("Not implemented yet"); // FIXME
+    }
+
+    /**
+     * Create a message reply.
+     *
+     * @param payload The payload of the reply.
+     * @return A message reply.
+     */
+    public MessageReply<T> message(T payload) { // FIXME rename to reply? Note thenReply
```
patriknw, line 53: yes, this should be renamed to `reply(T message)`

samples/java-eventsourced-shopping-cart/src/main/java/com/example/shoppingcart/domain/ShoppingCartImpl.java
```diff
@@ -16,94 +16,111 @@
...
     this.entityId = entityId;
   }
 
   @Override
-  public ShoppingCartDomain.Cart snapshot() {
-    return ShoppingCartDomain.Cart.newBuilder()
-        .addAllItems(cart.values().stream().map(this::convert).collect(Collectors.toList()))
-        .build();
+  protected ShoppingCartDomain.Cart emptyState() {
+    return ShoppingCartDomain.Cart.getDefaultInstance();
```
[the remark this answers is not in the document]
sean-walsh: I like how explicit this is.

sdk/src/main/java/com/akkaserverless/javasdk/eventsourcedentity/EventSourcedEntityEffect.java
sean-walsh: A marker interface here?

patriknw: it's typically the generated proto messages that are the events

sean-walsh: Ah yes, thanks.

samples/java-eventsourced-shopping-cart/src/main/java/com/example/shoppingcart/domain/ShoppingCartImpl.java
```diff
@@ -16,94 +16,111 @@
...
-    if (item == null) {
-      item = convert(itemAdded.getItem());
-    } else {
-      item =
-          item.toBuilder()
-              .setQuantity(item.getQuantity() + itemAdded.getItem().getQuantity())
-              .build();
-    }
-    cart.put(item.getProductId(), item);
+  protected ShoppingCartDomain.Cart itemAdded(ShoppingCartDomain.ItemAdded itemAdded, ShoppingCartDomain.Cart currentState) {
```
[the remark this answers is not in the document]
octonato: Yes, I think it reads better with state is first. In my brain I read it as: given this `State`, what should happen if we apply this `Command`. 

Which also reads as, given this `State` what is the command handler (`Cmd => Effect`) we have for it.

Same reasoning applies event handlers. 

So, yeah. State first all the way down.

samples/java-eventsourced-shopping-cart/src/main/java/com/example/shoppingcart/domain/ShoppingCartImpl.java
```diff
@@ -16,94 +16,111 @@
...
   @Override
-  protected Empty removeItem(ShoppingCartApi.RemoveLineItem item, CommandContext ctx) {
-    if (!cart.containsKey(item.getProductId())) {
-      throw ctx.fail(
-          "Cannot remove item " + item.getProductId() + " because it is not in the cart.");
+  public Effect<Empty> removeItem(
+          ShoppingCartApi.RemoveLineItem command,
+          ShoppingCartDomain.Cart currentState,
+          EventSourcedEntityEffect.Builder<Empty, ShoppingCartDomain.Cart> effectBuilder,
+          CommandContext ctx) {
```
octonato, line 78: EventSourcedEntityEffect.Builder + CommandContext is looking heavy. 

Not fully clear why we need it. I have to checkout the code and feel it under my fingers. 

In any case, worth experiment more to reduce it.

samples/java-eventsourced-shopping-cart/src/main/java/com/example/shoppingcart/domain/ShoppingCartInterface2.java
```diff
@@ -0,0 +1,51 @@
...
+            ShoppingCartDomain.Cart currentState,
+            EventSourcedEntityEffect.Builder<Empty, ShoppingCartDomain.Cart> effectBuilder,
+            CommandContext ctx);
+
+    @CommandHandler(name = "RemoveItem")
+    public abstract Effect<Empty> removeItem(
+            ShoppingCartApi.RemoveLineItem command,
+            ShoppingCartDomain.Cart currentState,
+            EventSourcedEntityEffect.Builder<Empty, ShoppingCartDomain.Cart> effectBuilder,
+            CommandContext ctx);
```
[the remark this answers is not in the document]
octonato: what will be left in the command context? 

Otherwise, we could have a `CallContext<State>` that we use to get command context and build the reply. 

(CallContext sucks, I just need to pick a name.)

octonato (commented):
I did a quick pass. I will checkout and look more in detail later.

samples/java-eventsourced-shopping-cart/src/main/java/com/example/shoppingcart/domain/ShoppingCartInterface2.java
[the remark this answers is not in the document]
pvlugter: Combining into one context / effect builder would be good. And a simpler name too. Does look too heavy currently.

samples/java-eventsourced-shopping-cart/src/main/java/com/example/shoppingcart/domain/ShoppingCartImpl.java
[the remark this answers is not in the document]
pvlugter: State first would probably read better. JS SDK has command, state, context. Probably so that handlers can just have the command if current state is not needed.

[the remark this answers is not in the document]
patriknw: I have changed the order, state first: 9017436

samples/java-eventsourced-shopping-cart/src/main/java/com/example/shoppingcart/domain/ShoppingCartInterface2.java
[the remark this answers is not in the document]
patriknw: I have included the effect builder in the CommandContext: fc680a3

samples/java-eventsourced-shopping-cart/src/main/java/com/example/shoppingcart/domain/ShoppingCartImpl.java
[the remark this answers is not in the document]
patriknw: If we have `emptyState` we have to include make that accessible from the SDK also, not only in generated code like this. We could have an interface or abstract class that the impl has to implement.

samples/java-eventsourced-shopping-cart/src/main/java/com/example/shoppingcart/domain/ShoppingCartImpl.java
```diff
@@ -16,94 +16,109 @@
...
 
-import java.util.LinkedHashMap;
-import java.util.Map;
+import java.util.Comparator;
+import java.util.List;
+import java.util.Optional;
+import java.util.function.Predicate;
 import java.util.stream.Collectors;
 
 @EventSourcedEntity(entityType = "eventsourced-shopping-cart")
```
patriknw, line 32: Where should the entityType be defined? I think it belongs to the registration, similar to how the viewId is defined in the registration of views.

sdk/src/main/java/com/akkaserverless/javasdk/eventsourcedentity/EventSourcedEntityEffect.java
```diff
@@ -0,0 +1,137 @@
...
+      return new ForwardReplyImpl<>(serviceCall);
+    }
+
+    /**
+     * Create a failure reply.
+     *
+     * @param description The description of the failure.
+     * @return A failure reply.
+     */
+    public FailureReply<T> failure(String description) {
```
patriknw, line 91: Shall we rename this to `error`? https://www.reactivemanifesto.org/glossary#Failure

```
public ErrorReply<T> error(String description)
```

samples/java-eventsourced-shopping-cart/src/main/java/com/example/shoppingcart/domain/ShoppingCartImpl.java
pvlugter: In the registration sounds good. There may be an option for that already? (I remember this being discussed before)

sdk/src/main/java/com/akkaserverless/javasdk/eventsourcedentity/EventSourcedEntityEffect.java
pvlugter: Makes sense

samples/java-eventsourced-shopping-cart/src/main/java/com/example/shoppingcart/domain/ShoppingCartInterface2.java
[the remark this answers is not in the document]
pvlugter: 👍🏼 looks better behind the CommandContext.

Small thing, but I wonder if it would be cleaner as just `effects` rather than `effectBuilder`:

```java
return context.effects()
        .emitEvent(event)
        .thenReply(newState -> Empty.getDefaultInstance());
```

but maybe it's clearer to say that it's a builder explicitly.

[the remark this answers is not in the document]
patriknw: I like the suggestion of `context.effects()`.

On that topic. Effect is typically well know terminology for those that know some functional programming, but is it known for the target Java developers? Do we have any alternative naming suggestion?

Action? but that would be confusing since we already have that for Action components.

Just looking at the english dictionary for Effect, it might be just fine.

samples/java-eventsourced-shopping-cart/src/main/java/com/example/shoppingcart/domain/ShoppingCartImpl.java
octonato: since we are touching all this, it might be good to rename this to `entityTypeHint`. Maybe on a follow-up PR.

samples/java-eventsourced-shopping-cart/src/main/java/com/example/shoppingcart/domain/ShoppingCartImpl.java
```diff
@@ -16,94 +16,111 @@
...
-public class ShoppingCartImpl extends ShoppingCartInterface {
+public class ShoppingCartImpl extends ShoppingCartInterface2 {
   @SuppressWarnings("unused")
   private final String entityId;
 
-  private final Map<String, ShoppingCartApi.LineItem> cart = new LinkedHashMap<>();
-
   public ShoppingCartImpl(@EntityId String entityId) {
```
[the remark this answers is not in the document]
octonato: Absolutely, one thing that we might one to inject are clients to other services (local component or any other service)

samples/java-eventsourced-shopping-cart/src/main/java/com/example/shoppingcart/domain/ShoppingCartImpl.java
```diff
@@ -16,94 +16,109 @@
...
+                    ShoppingCartDomain.LineItem.newBuilder()
+                            .setProductId(command.getProductId())
+                            .setName(command.getName())
+                            .setQuantity(command.getQuantity())
+                            .build())
+            .build();
+
+    return context.effectBuilder()
+            .emitEvent(event)
+            .thenReply(newState -> Empty.getDefaultInstance());
```
octonato, line 66: This is looking much better. 

I wonder what we could do to remove the context at all. 

For instance, in Akka Typed, the entity needs to implement a given type. Could we do the same here? 

```java
abstract class ShoppingCartInterface2 implements EventSourcedEntity<Cart>
class ShoppingCartImpl extends ShoppingCartInterface2
```

`EventSourcedEntity` would have the effect builders already typed on `Cart` and will also have the `commandHandler` and `eventHandler` entry points to be implemented.

`ShoppingCartInterface2` would implement `commandHandler` and `eventHandler` and do the dispatch to the methods the user needs to implement. And we make handler final in the interface. The user only touch the method to dispatch to.

For the return type, I don't think the builder needs to have a type argument, because the compiler will force the user to finish it with the right type, ie: `.thenReply(newState -> "I'm done");` won't compile.

We are left with the data that is provide by the context, command name, command id and seqNum (for event sourcing). 

Would that be a good idea to wrap the command instead? 

```java
class ShoppingCartImpl extends implements EventSourcedEntity<Cart> {
  
  public Effect<Empty> addItem(
    ShoppingCartDomain.Cart currentState,
    Command<ShoppingCartApi.AddLineItem> input) {
	  
    // use wrap to access command and other call related info
    input.command() // returns ShoppingCartApi.AddLineItem
    input.commandName()
    input.commandId()
	    
    return effectBuilder() // already typed on Cart because of interface
             .emitEvent(event)
             // compile enforces return type
             .thenReply(newState -> Empty.getDefaultInstance()); 

	  }
}
```

samples/valueentity-counter/src/main/java/com/example/domain/CounterImpl.java
```diff
@@ -21,45 +26,64 @@ public CounterImpl(@EntityId String entityId) { // <2>
...
-        CounterDomain.CounterState state = ctx.getState() // <2>
+        CounterDomain.CounterState state = currentState // <2>
                 .orElseGet(() -> CounterDomain.CounterState.newBuilder().build()); // <3>
         CounterDomain.CounterState newState =  // <4>
                 state.toBuilder().setValue(state.getValue() + command.getValue()).build();
-        ctx.updateState(newState); // <5>
-        return Empty.getDefaultInstance();
+        return context.effectBuilder()
+                .updateState(newState) // <5>
+                .thenReply(Empty.getDefaultInstance()); // FIXME add convenience shortcut for reply Empty?
```
octonato, line 42: Yep, I was having the same thought. 

We should consider is if it realistic to think that one day we will support other formats than protobuf. If not (most likely not), it's fine to bound ourselves to `Empty`. 

An alternative is to have our own type (eg: `Done`) to replace `Empty`. And then promote the idea that if you just want to return an ack, you can use `Done` and the builder has a convenience for it. 

Anyway, not a big deal if we are sure about having a long term relationship with proto's `Empty`

sdk/src/main/java/com/akkaserverless/javasdk/SideEffect.java
```diff
@@ -0,0 +1,50 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.akkaserverless.javasdk;
+
+import com.akkaserverless.javasdk.impl.reply.EffectImpl;
+
+/** A side effect. */
+public interface SideEffect {
```
octonato, line 22: 👍 Good to call it by its name!

octonato (commented):
Looking good @patriknw 

I drop some new ideas about the effect builders and commands.

samples/java-eventsourced-shopping-cart/src/main/java/com/example/shoppingcart/domain/ShoppingCartImpl.java
patriknw: I think we need the `implements EventSourcedEntity<Cart>` anyway, for other reasons. I think I mentioned that in relation to the `emptyState`.

I like the idea of having the `effectBuilder()` (or `effects()` as was suggested elsewhere) in that base interface. I have to try if that really works with the reply type as you say.

I'm skeptical to the `Command` wrapping. One less parameter, but it becomes more indirection for the most common usage, which is accessing the command message. The context parameter can often be ignored.

samples/valueentity-counter/src/main/java/com/example/domain/CounterImpl.java
patriknw: We already support json. I agree that this is just a convenience that can be added if needed.

samples/java-eventsourced-shopping-cart/src/main/java/com/example/shoppingcart/domain/ShoppingCartImpl.java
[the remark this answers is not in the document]
patriknw: @octonato not that I would like to bring back memories to our old discussion about emptyState 😄 , but what do you think we should do here? `emptyState` or `Optional` parameters everywhere? Same in ValueEntity or is the difference motivated?

samples/java-eventsourced-shopping-cart/src/main/java/com/example/shoppingcart/domain/ShoppingCartInterface2.java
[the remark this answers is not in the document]
pvlugter: I think Effect is fine, and could be useful having it aligned with Akka.

Agree that Action would be confusing. Alternative could be something like Operation.

samples/java-eventsourced-shopping-cart/src/main/java/com/example/shoppingcart/domain/ShoppingCartImpl.java
pvlugter: Having the interface typed looks good, and for empty / initial state, and with the effects builder in the base rather than context.

samples/valueentity-counter/src/main/java/com/example/domain/CounterImpl.java
pvlugter: Rather than tying this to protobuf `Empty` specifically, it could be nicer if it returned a default instance of whatever the gRPC service method's return type is. I think that's supported in the JS SDK. Not sure how easy with the new approach, but possible if it has a handle on the method descriptor?

samples/java-eventsourced-shopping-cart/src/main/java/com/example/shoppingcart/domain/ShoppingCartImpl.java
octonato: > I'm skeptical to the Command wrapping. One less parameter, but it becomes more indirection for the most common usage, which is accessing the command message. The context parameter can often be ignored.

Yeah, you may be right. I just dislike having params that are not used. Specially if they are often ignored. They are always available and often ignored. Just noise. 

I wish we didn't have that at all. Do we really need a `commandName` if we have already the command? 

And what is a `commandId` exactly? 
If that's something generated by the framework, why do I need such a thing in my domain? And if it's something that I need in my domain than that should be a field in my command.

patriknw: I don't think we will be able to fully get rid of the context. There are several metadata related things that should be accessible, such as
```
CommandContext.commandName
CommandContext.commandId
MetadataContext.metadata

EntityContext.entityId
Context.serviceCallFactory

EventContext.sequenceNumber


ActionContext.eventSubject

UpdateHandlerContext.eventSubject
UpdateHandlerContext.commandName
ViewContext.viewId
```

Let me try the reply type first...

samples/java-eventsourced-shopping-cart/src/main/java/com/example/shoppingcart/domain/ShoppingCartImpl.java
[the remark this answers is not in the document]
octonato: Thanks for bring back those memories! :-)

My whole idea in the past is that I wanted to avoid exposing users to `Option` because I believe there is a clear point in time where the journal is empty and therefore we start with `None`. So, first event moves from `None` to `Some`. All subsequent events can only work on a `Some`. 

From that observation, we could say there are two possible event handlers (and command handlers): `Event => State` and `(State, Event) => State`. And we can then hide the `Option` from the user. 

Anyway, that just to remind my opinion about it since you triggered my memories. :-)

Back to the present...

I think we don't want to introduce two kinds of command and event handlers because that would require the user to annotate the methods in the proto file. Event handlers would even be more challenging because there are not even defined in the proto file. 

So, that to say, my opinion is that we should not bother users with an `Optional` type if most of the time it will be a `Some`. Only the first command handler and event handler will see it as whatever the user defines as empty state. 

And most probably, java developers will use a `null`. 

For the `ValueEntity`, I think we can apply the same reasoning. The `emtpyState` will only be used when there is no row (including after deleting a row and starting over again).

samples/java-eventsourced-shopping-cart/src/main/java/com/example/shoppingcart/domain/ShoppingCartImpl.java
octonato: Ok, let' go by pieces. 

I believe that some of then can be a constructor parameter or deemed obsolete (or not crucial). We will see..

samples/valueentity-counter/src/main/java/com/example/domain/CounterImpl.java
octonato: That would be interesting, but I don't know if I understand what you mean with a handle on the method descriptor. 

If a method returns `Foo`, we need to return an `Effect<Foo>`. If the builder is not already typed on `Foo` we will need something like: `replyWith(Foo.class)` and later behind the scenes build a `Foo.getDefaultInstance()` on behalf of the user. 

Between ``replyWith(Foo.class)` and `thenReply(Foo.getDefaultInstance())`, we save some characters but we increase API surface for something that is not so hard for the users to do.

samples/java-eventsourced-shopping-cart/src/main/java/com/example/shoppingcart/domain/ShoppingCartImpl.java
patriknw: Moved to the base class approach: 23c7fb9
Looking good.

I'll give the context some more thought. We could always give access to it via the base class instead of a parameter.

samples/java-eventsourced-shopping-cart/src/main/java/com/example/shoppingcart/domain/ShoppingCartImpl.java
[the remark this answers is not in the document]
patriknw: I agree. Then I introduce `emptyState` in ValueEntity also to make it consistent.

`null` or the proto empty instance can be used as starting point.

samples/valueentity-counter/src/main/java/com/example/domain/CounterImpl.java
patriknw: I think it would work with `return effects().replyWithEmpty()`. The `Foo` can be inferred from the return type of the method. We would create an `EmptyReply<Foo>`, which we will translate to the `Foo.getDefaultInstance()`.

patriknw: and `Foo.getDefaultInstance()` would be constructed via the proto method from the gRPC service method that Peter mentioned.

samples/java-eventsourced-shopping-cart/src/main/java/com/example/shoppingcart/domain/ShoppingCartImpl.java
patriknw: I removed the context parameter, and it's accessed via the base class instead. Some runtime caveats but that is worth it. 05239d2

samples/java-eventsourced-shopping-cart/src/main/java/com/example/shoppingcart/domain/ShoppingCartImpl.java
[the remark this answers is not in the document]
patriknw: Added emptyState for ValueEntity also: bd4aff8

sdk/src/main/java/com/akkaserverless/javasdk/eventsourcedentity/EventSourcedEntityEffect.java
patriknw: error it is: b5cea76

samples/java-eventsourced-shopping-cart/src/main/java/com/example/shoppingcart/domain/ShoppingCartImpl.java
```diff
@@ -16,94 +16,102 @@
...
-  public ShoppingCartDomain.Cart snapshot() {
-    return ShoppingCartDomain.Cart.newBuilder()
-        .addAllItems(cart.values().stream().map(this::convert).collect(Collectors.toList()))
-        .build();
+  protected ShoppingCartDomain.Cart emptyState() {
+    return ShoppingCartDomain.Cart.getDefaultInstance();
   }
 
   @Override
-  public void handleSnapshot(ShoppingCartDomain.Cart cart) {
```
patriknw, line 48: One thing we can consider adding would be conversion methods to/from storage representation, so the business logic in command and event handlers can use a more convenient state representation than the proto objects (or json). That class would be somewhat difficult to define in the proto codegen options though, unless it's defined by it's fully qualified java class name.

sdk/src/main/java/com/akkaserverless/javasdk/eventsourcedentity/EventSourcedEntityBase.java
```diff
@@ -0,0 +1,168 @@
...
+
+  /**
+   * Implement by returning the initial empty state object. This object will be passed into the
+   * command and event handlers, until a new state replaces it.
+   *
+   * <p>Also known as "zero state" or "neutral state".
+   *
+   * <p><code>null</code> is an allowed value.
+   */
+  protected abstract S emptyState();
```
raboof, line 45: I wonder if we should pass the `entityId` to the `emptyState`?

patriknw: entityId will be passed to the factory lambda (constructor) via the CreationContext, so it's possible to keep that in an instance variable and accessed here

patriknw:
We are starting a feature branch `sdk-codegen-dev` and this PR is now squashed and targeting that.

patriknw:
Merging into feature branch

samples/valueentity-counter/src/main/java/com/example/domain/CounterImpl.java
pvlugter: Yes, I was thinking of just `replyWithEmpty()` or `replyWithDefault()` or similar and the actual default instance created using the gRPC method descriptor.

samples/java-eventsourced-shopping-cart/src/main/java/com/example/shoppingcart/domain/ShoppingCartImpl.java
pvlugter: Yeah, working with protobuf objects directly is more awkward. With going to and from builders, changing anything deeper. Would be interesting to see what could be improved, whether that's conversions, or wrappers that make it easier to mutate, in the way lenses are used for this.

patriknw: created issue https://github.com/lightbend/akkaserverless-java-sdk/issues/102 for this idea

PR #65, opened 2021-06-24
build publish maven plugin from CI to sonatype

raboof:
towards #27

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -338,7 +338,7 @@ workflows:
             - samples-tests    
             - tck-tests
             - validate-docs   
-      - publish-maven-plugin:
+      - publish-maven:
           filters: # version tags only
             tags:
               only: /^v.*/
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -240,8 +240,19 @@ jobs:
           name: Publish maven plugin and archetype
           command: |
             cd maven-java
-          mvn -B versions:set -DnewVersion=$SDK_VERSSION
-          mvn -B -pl akkaserverless-maven-plugin deploy
+            cat >~/.m2/settings.xml <<EOF;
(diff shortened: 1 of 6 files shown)


maven-java/pom.xml
```diff
@@ -46,34 +46,89 @@
...
+      <plugins>
+        <plugin>
+          <groupId>org.sonatype.plugins</groupId>
+          <artifactId>nexus-staging-maven-plugin</artifactId>
+          <version>1.6.7</version>
+	  <extensions>true</extensions>
+          <configuration>
+            <serverId>ossrh</serverId>
+            <nexusUrl>https://s01.oss.sonatype.org</nexusUrl>
+            <autoReleaseAfterClose>false</autoReleaseAfterClose>
```
raboof, line 66: we could set this to 'true' eventually, but for the first release perhaps let's give ourselves a chance to check the artifacts?

.circleci/config.yml
```diff
@@ -230,6 +230,46 @@ jobs:
...
+    steps:
+      - checkout
+      - restore_deps_cache
+      - set-sdk-version
+      - run:
+          name: Publish maven plugin and archetype
+          command: |
+            cd maven-java
+            echo $PGP_SECRET | base64 -d | gpg --import --batch
+            # It seems ~/.m2 is not actually restored from the cache?
```
octonato, line 245: We need to investigate it further. We may be doing something wrong. 

When I was bringing the maven code here, I was surprised to see that I have to call sbt publishM2 more then once in the same CI build. I was planning to call only once and reuse the local artifacts on the next jobs. Didn't work as expected.

octonato (approved):
LGTM

octonato:
hmm, I could have squashed before merging. 🤦

PR #68, opened 2021-06-25
fix: merge maven-java project into sdk project

octonato:
First phase of #27 

I believe this branch can already get into `main`. The codegen will follow.


pvlugter (commented):
I'll make the same suggestion as for the merge into the JavaScript SDK. Rather than having all the earlier commits, then moving to a subdirectory in another commit, we can rewrite the history with something like `git-filter-repo --path-rename :maven-java` — so that it's as if it was always under a subdirectory. That could make the history a lot easier to follow.

octonato:
@pvlugter 

> we can rewrite the history with something like git-filter-repo --path-rename :maven-java/ — so that it's as if it was always under a subdirectory

Ok, now I understand what you said in the stand-up today. 

I'm fine with that. We can do it later no? I mean. We can merge as such and re-write the history in the `main` branch.

.circleci/config.yml
```diff
@@ -14,7 +14,7 @@ commands:
             git config user.email "circleci@example.com"
             git config user.name "CircleCI"
       - build-tools/merge-with-parent:
-          parent: main
+          parent: rgc/feature-branch-maven-java
```
patriknw, line 17: should this be reverted to `main` when we merge it?

patriknw (approved):
LGTM, with a question about the ci config?

.circleci/config.yml
octonato: absolutely, create you spotted it.

octonato:
> We can merge as such and re-write the history in the main branch. 

One other reason that should re-write the history later is that we will soon bring the codegen here and we will have the same issue. We should re-write it on everything is on the SDK.

.circleci/config.yml
octonato:
```suggestion
          parent: main
```

patriknw:
> We can merge as such and re-write the history in the main branch.

Rewriting history in the main branch doesn't sound right to me.

octonato:
> Rewriting history in the main branch doesn't sound right to me.

If we run `git-filter-repo --path-rename :maven-java/`, we re-write the whole history. If we are annoyed by it or by the strange history we get when we merge `maven-java`, then we should simply squash it and throw away the history from `maven-java` (and then the same for the codegen.

raboof:
> > Rewriting history in the main branch doesn't sound right to me.
> 
> If we run `git-filter-repo --path-rename :maven-java/`, we re-write the whole history. If we are annoyed by it or by the strange history we get when we merge `maven-java`, then we should simply squash it and throw away the history from `maven-java` (and then the same for the codegen).

Would it be possible to rewrite only the history of `maven-java` (and `codegen`), and not of the existing code? If that's the case then it'd be attractive to do it before merging.

patriknw:
I thought the purpose of this PR, or the feature branch was to rewrite the history before merging to the stable `main` branch. Anyway, we must move forward without struggling with the history.

octonato:
The `maven-java` history is already re-written. The commit messages got the issues and PRs references removed, but the authors were preserved. 

When I run `git-filter-repo --path-rename :maven-java/` it didn't work and I need to add a `--force`. That re-wrote the whole history.

I will try again and see if there are other options I can use.

octonato:
I completely missed Peter's point I think. The goal was to run `git-filter-repo --path-rename :maven-java/` in the maven-java project before bring it here. 

I will see what I can do about it, but not clear if it has a lot of value. Might want to catch-up with him later today when he's back online.

octonato:
superseded by #72

pvlugter:
Sorry, I wasn't clear about the changes. Looks like it's all been sorted now. Yes, run all the `git-filter-repo` commands on the original repo before bringing it in. Just as notes, for any future merges, history cleanup that I did for JS SDK included:

If only wanting to retain some parts: `git-filter-repo --path keep/path/ --path keep-another-path/`
Rewrite all history to a subdirectory: `git-filter-repo --path-rename :base-path/`
Rebase everything on the first commit to remove merge commits: `git rebase (git rev-list --max-parents=0 HEAD)`
Add prefix to commits with: `git-filter-repo --message-callback 'return b"prefix: " + message'`
Rewrite emails as needed with: `git-filter-repo --email-callback 'return email.replace(b"old@email", b"new@email")'`

Then pull into the new repo with a `git fetch --no-tags ...` and rebase the branch on main to bring into the current history. I also then reset the committer details (from the rebase) back to the original author (couldn't remember if there was a better way to keep that clean):
```
git filter-branch --env-filter 'GIT_COMMITTER_EMAIL="$GIT_AUTHOR_EMAIL"; GIT_COMMITTER_NAME="$GIT_AUTHOR_NAME"; export GIT_COMMITTER_DATE="$GIT_AUTHOR_DATE"' -- main..HEAD
```

PR #70, opened 2021-06-30
Add doc sources for Java SDK from main docs

pvlugter:
Copy the Java SDK docs from the main Akka Serverless docs, so they can be maintained here.

The partials from the developing module are still partials here (but under java module so they don't conflict), in case we want to move these to a common repo for all SDK docs later on.

Cross-references to the main docs are all converted to external links to the published production docs, so the docs here can be used and checked separately. These are full links in the asciidoc sources, so they get validated by the asciidoc external link checker, which doesn't support variables.

There's a simple setup for building and viewing the docs locally — uses a default theme, as the Akka Serverless doc theme is defined as supplemental ui in a private repo currently.

Kept as separate commits in this PR, in case we want to make any changes to the setup before merging.

**Note**: need to coordinate with removing these sources in the main docs, otherwise there will be conflicts with duplicate files, so marking this as draft.


ennru (approved):
Looks good.
I believe duplicating the partials can be a price we can afford.

docs/src/modules/java/pages/kickstart.adoc
```diff
@@ -0,0 +1,239 @@
...
+
+* In a `settings.xml` file located in your home directory in a sub-directory named `.m2`. If you have an existing `settings.xml` file, either in your home directory or in `\{MAVEN_HOME}/conf`, add the `akkaserverless` profile element shown in the full example below and list it as `activeProfile` in the existing file.
+* Your organization may have configured a proxy to be used to download Maven artifacts, such as Jfrog Artifactory. If you have difficulties updating `settings.xml` for a proxy, please consult your system administrator.
+
+The following shows a complete `settings.xml` file configured to use the Lightbend repository:
+
+[source,xml]
+----
+include::example$settings.xml[]
+----
```
octonato, line 43: If #72 gets merged first, we can already remove that section from the docs.

docs/src/modules/java/pages/kickstart.adoc
```diff
@@ -0,0 +1,239 @@
...
+--
+Windows 10+::
++
+--
+[source,command window, subs="attributes"]
+----
+mvn archetype:generate ^
+  -DarchetypeGroupId=com.lightbend ^
+  -DarchetypeArtifactId=maven-archetype-akkasls ^
+  -DarchetypeVersion=LATEST
```
octonato, line 77: Similar, if #72 gets merged first, we can update the groupId and artifactId

octonato (approved):
Looks good.

docs/src/modules/java/pages/kickstart.adoc
pvlugter: Sounds good. We'll end up publishing these docs to make the switch over to the sources being here, so we may want to release the new archetypes first.

pvlugter:
> I believe duplicating the partials can be a price we can afford.

And we can still leave them as partials if we want, to make updating them across repos easier.

pvlugter:
To be able to make the switch to the doc sources being here, I'll first publish these under a different branch `docs/tmp/current` instead of `docs/current`. To avoid the main docs breaking on duplicate sources during the transition.

pvlugter:
Published these sources to `docs/tmp/current` with `make deploy`. Sources removed from the main docs in https://github.com/lightbend/akkaserverless-docs/pull/746. Once that's merged, we can publish to `docs/current` again and merge here.

pvlugter:
Will merge this now, and go through with the switch.

PR #73, opened 2021-07-01
first pass using codegen instead of reflection

raboof:
Manually create a `CustomerValueEntityHandler` that would be generated as part of codegen.

Added basic code to hook it into `Main`, but that API would still change and partly be generated.

Reading the codegen code, it's clear we're losing some features that used to be available with codegen, marked them with TODO's - we should consider whether we're OK with losing them or want to introduce new mechanisms.


samples/java-customer-registry/src/main/java/customer/CustomerValueEntityHandler.java
```diff
@@ -0,0 +1,107 @@
...
+  public Reply<Any> handleCommand(Any command, CommandContext<Any> context) throws Throwable {
+    AdaptedCommandContext<CustomerDomain.CustomerState> adaptedContext =
+        new AdaptedCommandContext(context, anySupport);
+    try {
+      // TODO we used to support passing in the command as Jackson-parsed model object
+      // or as ScalaPB class as well. With this change we tie ourselves to Java protobuf.
+      Effect<? extends GeneratedMessageV3> effect = invoke(command, adaptedContext);
+      // TODO we used to support accepting ScalaPB and Jackson objects as responses as well.
+      //  Are we OK with losing that? I guess we could have separate methods for that on the
+      //  builder, though it's not obvious how to achieve type-safety for those then.
```
patriknw, line 55: yes, I think we still want the json support. Could perhaps be an additional proto option on the entity (or even rpc) and we would generate different conversion.

samples/java-customer-registry/src/main/java/customer/CustomerValueEntityHandler.java
```diff
@@ -0,0 +1,107 @@
...
+
+      // FIXME probably ValueEntityHandler should be adapted to expect an `Effect<? extends
+      // GenerateMessageV3>`?
+      return null;
+    } catch (Throwable t) {
```
patriknw, line 60: we don't want to catch fatal errors. Enough with `Exception`?

samples/java-customer-registry/src/main/java/customer/Main.java
```diff
@@ -37,6 +53,24 @@ public static void main(String[] args) throws Exception {
...
               CustomerDomain.getDescriptor())
           // end::register[]
+          .lowLevel()
+          .registerValueEntity(
+              // TODO this factory probably can disappear, we can pass in the handler directly?
+              context ->
+                  new CustomerValueEntityHandler(
+                      // TODO I guess construction the Entity should somehow remain part of the
+                      // customer API, so they can pass
+                      // in  anything they like in the constructor, including the context?
```
patriknw, line 63: yes, we should change the registration methods to take a factory lambda `CreationContext -> ValueEntity`

patriknw (commented):
this is looking very promising

samples/java-customer-registry/src/main/java/customer/CustomerValueEntityHandler.java
johanandren: Isn't the json support some specific proto type that we could identify here (and turn to some json model) or how did that actually work?

(Somewhat confusing when there is also transcoding from json to proto in the proxy)

samples/java-customer-registry/src/main/java/customer/CustomerValueEntityHandler.java
```diff
@@ -0,0 +1,133 @@
...
+  @Override
+  public ValueEntityBase.Effect<Any> handleCommand(
+      Any command, Any state, CommandContext<Any> context) throws Throwable {
+    AdaptedCommandContext<CustomerDomain.CustomerState> adaptedContext =
+        new AdaptedCommandContext(context, anySupport);
+    try {
+      CustomerDomain.CustomerState parsedState =
+          CustomerDomain.CustomerState.parseFrom(state.getValue());
+      // TODO we used to support passing in the command as Jackson-parsed model object
+      // or as ScalaPB class as well. With this change we tie ourselves to Java protobuf.
```
johanandren, line 67: I think we can solve the ScalaPB part when we get to doing a Scala SDK?

samples/java-eventsourced-shopping-cart/src/main/java/com/example/shoppingcart/domain/ShoppingCartImpl.java
```diff
@@ -25,7 +25,7 @@
 import java.util.function.Predicate;
 import java.util.stream.Collectors;
 
-public class ShoppingCartImpl extends AbstractShoppingCart {
+public class ShoppingCartImpl extends ShoppingCartInterface2 {
```
johanandren, line 28: Ups, my bad!

samples/java-eventing-shopping-cart/src/main/proto/cart/shopping_cart_view_model.proto
```diff
@@ -23,6 +23,10 @@ import "cart/shopping_cart_domain.proto";
 import "google/api/annotations.proto";
 
 service ShoppingCartViewService {
+  option (akkaserverless.service) = {
+    type : SERVICE_TYPE_VIEW
+  };
```
octonato, line 28: We don't need this for now. This is only needed when using the codegen and the eventing sample doesn't use it. Anyway, I will remove it for now since I will be pushing to this PR (soon).

samples/java-customer-registry/src/main/java/customer/CustomerValueEntityHandler.java
```diff
@@ -0,0 +1,133 @@
...
+      ValueEntityBase.Effect<? extends GeneratedMessageV3> effect =
+          invoke(command, parsedState, adaptedContext);
+      // TODO we used to support accepting ScalaPB and Jackson objects as responses as well.
+      //  Are we OK with losing that? I guess we could have separate methods for that on the
+      //  builder, though it's not obvious how to achieve type-safety for those then.
+
+      // FIXME probably ValueEntityHandler should be adapted to expect an `Effect<? extends
+      // GenerateMessageV3>`?
+      return null;
```
raboof, line 76: I think this can now be implemented?

octonato (approved):
I think we should merge this one for now. 

I have done some experiments for the handler, but better to add on a follow up PR.

samples/java-customer-registry/src/main/java/customer/CustomerValueEntityHandler.java
octonato: I will do it in a follow-up PR.

PR #74, opened 2021-07-01
Update to latest akka-http / akka-grpc snapshots

jrudolph:
We can probably wait with merging until we have released a milestone proper for akka-grpc but let's see what PR validation has to say.

### DIFF
--- a/project/Common.scala
+++ b/project/Common.scala
@@ -36,8 +36,6 @@ object CommonSettings extends AutoPlugin {
       run / javaOptions ++= {
         sys.props.collect { case (key, value) if key.startsWith("akka") => s"-D$key=$value" }(breakOut)
       },
-      // to get latest versions
-      resolvers += "akka-http-snapshot-repository" at "https://oss.sonatype.org/content/repositories/snapshots",
     )
 
   override def projectSettings = Seq(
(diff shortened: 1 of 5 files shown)


sdk/src/main/resources/reference.conf
```diff
@@ -50,6 +50,8 @@ akkaserverless {
         # that means this is the upper cap on how many entities of each type can live in one node
         # and have concurrent connections from the proxy
         http2.max-concurrent-streams = 30000
+
+        http2.min-collect-strict-entity-size = 1
```
patriknw, line 54: 👍

project/Common.scala
```diff
@@ -35,7 +35,9 @@ object CommonSettings extends AutoPlugin {
       scalaVersion := Dependencies.ScalaVersion,
       run / javaOptions ++= {
         sys.props.collect { case (key, value) if key.startsWith("akka") => s"-D$key=$value" }(breakOut)
-      }
+      },
+      // to get latest versions
+      resolvers += "akka-http-snapshot-repository" at "https://oss.sonatype.org/content/repositories/snapshots",
```
patriknw, line 40: I don't think we can use sonatype snapshots in the sdk, because existing builds (pom.xml) doesn't have that snapshot repo configured. We probably need a real milestone on maven central.

Or we could wait with merging/releasing the SDK.

```
Failed to execute goal on project valueentity-counter: Could not resolve dependencies for project com.example:valueentity-counter:jar:0.1: Could not find artifact com.lightbend.akka.grpc:akka-grpc-runtime_2.13:jar:2.0.0-30-fa341ccd-SNAPSHOT
```

patriknw: ah, that was what you said

jrudolph:
Now using released milestones.

pvlugter (approved):

raboof (approved):

PR #75, opened 2021-07-01
feat: brings in the java codegen + codegen core

octonato:
We can quick merge this one, it's the whole codegen code minus the javascript. 

Build is not integrated, but we can add it in upcoming PRs (we did so for the javascript). 

Part of #27


patriknw (approved):
LGTM

PR #95, opened 2021-07-02
fix: integrate codegen build in sdk build

octonato:
* Code is not formatted with the settings from the SDK
* Headers adapted as well
* build somewhat aligned with other builds we have, eg: Dependencies defined in Dependencies.scala


build.sbt
```diff
@@ -41,26 +43,6 @@ lazy val sdk = project
...
-        "com.akkaserverless" % "akkaserverless-proxy-protocol" % AkkaServerless.FrameworkVersion % "protobuf-src",
-        "com.akkaserverless" % "akkaserverless-sdk-protocol" % AkkaServerless.FrameworkVersion % "compile;protobuf-src",
-        "org.scalatest" %% "scalatest" % ScalaTestVersion % Test,
-        akkaDependency("akka-testkit") % Test,
-        akkaDependency("akka-stream-testkit") % Test,
-        akkaHttpDependency("akka-http-testkit") % Test,
-        "ch.qos.logback" % "logback-classic" % LogbackVersion % "test;provided",
-        "ch.qos.logback.contrib" % "logback-json-classic" % LogbackContribVersion % Provided,
-        "com.fasterxml.jackson.core" % "jackson-databind" % JacksonDatabindVersion
-      ),
```
octonato, line 63: moved into Dependencies.scala

build.sbt
```diff
@@ -89,33 +72,47 @@ lazy val testkit = project
         "scalaVersion" -> scalaVersion.value
       ),
     buildInfoPackage := "com.akkaserverless.javasdk.testkit",
-    libraryDependencies ++= Seq(
-        "org.testcontainers" % "testcontainers" % TestContainersVersion,
-        "junit" % "junit" % JUnitVersion % Provided,
-        "org.junit.jupiter" % "junit-jupiter" % JUnitJupiterVersion % Provided
-      ),
```
octonato, line 96: moved into Dependencies.scala

build.sbt
```diff
@@ -89,33 +72,47 @@ lazy val testkit = project
...
 
 lazy val tck = project
   .in(file("tck"))
   .dependsOn(sdk, testkit % Test)
   .enablePlugins(AkkaGrpcPlugin, PublicDockerImage)
   .settings(
     name := "akkaserverless-tck-java-sdk",
-    libraryDependencies ++= Seq(
-        "com.akkaserverless" % "akkaserverless-tck-protocol" % AkkaServerless.FrameworkVersion % "protobuf-src"
-      ),
```
octonato, line 112: moved into Dependencies.scala

octonato (commented):
I removed the integration tests from the codegen. 

I don't it makes sense to keep anything that was in there. The setup was maybe a little weird. The maven archetype includes the codegen, but the codegen was testing the maven archetype. 

Anyway, now that we have all together here we can re-think how we want to build integration tests.

ennru (approved):
LGTM.

.circleci/config.yml
```diff
@@ -76,12 +76,15 @@ commands:
...
+      - run: 
+          name: publish m2
+          command: sbt codegenJava/publishM2
       - run:
           name: "Build maven-java"
           command: |
             cd maven-java
             echo "Running maven-java with SDK version: '$SDK_VERSION'"
-            mvn versions:set -DnewVersion=$SDK_VERSION
+            mvn versions:set -DnewVersion=$SDK_VERSION -Dakkaserverless-codegen.version=$SDK_VERSION
```
octonato, line 87: Interested to see how this will work out. 

Before building the plugin we need to build the codegen and we must change the codegen version in the pom file with `-Dakkaserverless-codegen.version`. 

I'm curious to see what will be the release pom. Will that have the original `akkaserverless-codegen.version` value? If so, the build will have to effectively modify the file before publishing.

maven-java/akkaserverless-maven-plugin/pom.xml
```diff
@@ -16,6 +16,11 @@
 
   <name>Akka Serverless Maven Plugin</name>
 
+  <properties>
+    <!-- this dummy 0.0.0 version is replaced during the by the current build version -->
+    <akkaserverless-codegen.version>0.0.0</akkaserverless-codegen.version>
```
octonato, line 21: this don't feel good :-(

PR #98, opened 2021-07-02
sdk codegen: remove 'setState' from value entity context, always use effects

raboof:
Since we now have the effect system. This now also requires
returning an effect rather than a value even for the
annotation-based implementation.

Haven't updated the sample projects yet.

### DIFF
--- a/tck/src/main/java/com/akkaserverless/javasdk/tck/model/valueentity/ValueEntityTckModelEntity.java
+++ b/tck/src/main/java/com/akkaserverless/javasdk/tck/model/valueentity/ValueEntityTckModelEntity.java
@@ -47,30 +47,43 @@ public class ValueEntityTckModelEntity extends ValueEntityBase<Persisted> {
   @CommandHandler
   public Effect<Response> process(Request request, CommandContext<Persisted> context) {
     // FIXME the effect API doesn't support all combinations, and that might be fine?
-    //    Effect<Response> reply = null;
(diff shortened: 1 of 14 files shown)


tck/src/main/java/com/akkaserverless/javasdk/tck/model/valueentity/ValueEntityTckModelEntity.java
```diff
@@ -44,38 +45,40 @@ public ValueEntityTckModelEntity(Context context) {
...
+    //          break;
+    //        case EFFECT:
+    //          com.akkaserverless.tck.model.ValueEntity.Effect effect = action.getEffect();
+    //          e.add(SideEffect.of(serviceTwoRequest(effect.getId()), effect.getSynchronous()));
+    //          break;
+    //        case FAIL:
+    //          return effects().error(action.getFail().getMessage());
+    //      }
+    //    }
+    //    effects().updateState(Persisted.newBuilder().setValue(state).build());
```
johanandren, line 72: I remember this was tricky with reply in the JS sdk as well, perhaps what I did there to only keep the last reply/effect around could be useful to look at as inspiration? https://github.com/lightbend/akkaserverless-javascript-sdk/blob/main/tck/value-entity.js#L57

johanandren (commented):
Looks promising, didn't find any real issues

johanandren (approved):

Issue #42, opened 2021-05-12
Different name for the abstract base class


ennru:
Right now, the abstract base class for the Event Sourced Entity gets the suffix `Interface` which is misleading.

octonato:
I also got that feeling that it should be called `AbsctractSomething`. 

And we can call the implementation `Something` instead of `SomethingImpl`.

franciscolopezsancho:
fixed on https://github.com/lightbend/akkaserverless-java-sdk/pull/99

franciscolopezsancho:
#109

PR #99, opened 2021-07-02
changing Interface class name to Abstract

franciscolopezsancho:
tested on
- codegen
- maven-java
- single application calling to codegen through maven-java-plugin (maybe unnecessary : )

### DIFF
--- a/codegen/java-gen/src/main/scala/com/lightbend/akkasls/codegen/java/ActionServiceSourceGenerator.scala
+++ b/codegen/java-gen/src/main/scala/com/lightbend/akkasls/codegen/java/ActionServiceSourceGenerator.scala
@@ -41,7 +41,7 @@ object ActionServiceSourceGenerator {
     val implSourcePath =
       sourceDirectory.resolve(packagePath.resolve(implClassName + ".java"))
 
-    val interfaceClassName = className + "Interface"
+    val interfaceClassName = "Abstract" + className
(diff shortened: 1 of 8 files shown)


franciscolopezsancho:
I think we some dependency issue. The `maven-java-test` that is failing is because they are using the previous codegen which creates `[info] Generated: target/generated-sources/akkaserverless/java/com/example/shoppingcart/domain/ShoppingCartInterface.java
` . Part of this PR is to make sure the maven-java now expects `target/generated-sources/akkaserverless/java/com/example/shoppingcart/domain/AbstractShopping.java`. 

Shall I do this in two steps? Meaning first a PR only to the codegen and later to maven-java?

Or you might prefer to change the CI so it creates first the codegen and uses it?

LMKWYT

franciscolopezsancho:
forget about my last comment, please. Everything should now work, I added the required git dependencies.

ennru:
Fixes #42

ennru (approved):
LGTM.
A next step could be to remove the `Impl` postfix on the class created for the user.

johanandren (approved):

franciscolopezsancho:
you mean to remove the `Impl` in this PR @ennru ?

ennru:
> you mean to remove the `Impl` in this PR

No, let's do that in another PR.

PR #105, opened 2021-07-05
Include codegen in build continued

johanandren:
Continuing on @octonato s work in #95 since he is on vacation


johanandren:
After updating the samples to compile with the repo version of the codegen, it no longer works running them with the samples, not sure if we should just disable the integration tests for now until we land or what is the next step.

Example needed signature to compile:
```
@Override
  public Reply<Empty> addItem(ShoppingCartApi.AddLineItem command, CommandContext<ShoppingCartDomain.Cart> ctx) 
```

But then running the integration test fails with:
```
com.akkaserverless.javasdk.impl.EntityExceptions$EntityException: No command handler found for command [AddItem] on class com.example.shoppingcart.domain.ShoppingCartImp
```

johanandren:
I guess I'll start with rebasing on those conflicts tomorrow.

codegen/java-gen/src/it/scala/com/lightbend/akkasls/codegen/java/JavaCodegenIntegrationSuite.scala
```diff
@@ -1,582 +0,0 @@
...
-  def retryUntil[A](condition: A => Boolean, limit: Integer = 20, delay: Long = 1000)(
-      function: => A
-  ): A =
-    Some(function).filter(condition).getOrElse {
-      if (limit > 0) {
-        Thread.sleep(delay)
-        retryUntil(condition, limit - 1, delay)(function)
-      } else throw new Exception("Retry limit reached")
-    }
-}
```
raboof, line 582: I guess all this will be replaced with something based on #57 ?

raboof (approved):
Nice, good general improvements as well. Weird that the `Reply` and `CommandHandler` changes got crosswired, but I guess that's the point of moving it to one repo :)

ennru (approved):
LGTM.

patriknw:
Green! Great!. Note that this should be merged, not squashed. Johan might want to squash a few of the latest commits to cleanup those iterations.

johanandren:
I'll do some squashysquash now.

codegen/java-gen/src/it/scala/com/lightbend/akkasls/codegen/java/JavaCodegenIntegrationSuite.scala
johanandren: Three of the samples now are part of the CI-build with the built codegen, and at least two of those run integration tests with actual proxy, so perhaps a tricky containery test like this is no longer needed. Not quite sure.

PR #108, opened 2021-07-05
chore: configure release drafter

ennru:


.github/release-drafter.yml
```diff
@@ -0,0 +1,45 @@
...
+categories:
+  - title: '🚀 Productboard Features'
+    labels:
+      - 'Feature'
+  - title: ':toolbox: Chore'
+    labels:
+      - 'chore'
+  - title: 'Documentation'
+    labels:
+      - 'Documentation'
```
patriknw, line 13: How does it work for things marked with more than one label, for example Documentation and chore?

I'm thinking it might be better to base the categories solely on the conventional commit prefix rather than also looking for file locations.

.github/release-drafter.yml
```diff
@@ -0,0 +1,45 @@
+# Config for https://github.com/release-drafter/release-drafter
+name-template: 'Akka Serverless Java SDK $NEXT_PATCH_VERSION'
+tag-template: 'v$NEXT_PATCH_VERSION'
+categories:
+  - title: '🚀 Productboard Features'
```
pvlugter, line 5: Just `Features` (without `Productboard`) for SDK repos?

.github/release-drafter.yml
```diff
@@ -0,0 +1,45 @@
...
+  # java-sdk specific
+  - label: 'Maven'
+    files:
+      - 'maven-java/*'
+  - label: 'Documentation'
+    files:
+      - 'docs/*'
+  - label: 'tck'
+    files:
+      - 'tck/*'
```
pvlugter, line 45: Also add some of the other conventional commit types that we expect to use? `fix` in particular?

ennru:
PRs with multiple labels that are configured in the release drafter will be duplicated.
My thinking today is to not sort do anything about the conventional commit message at all and only use file location for automatic grouping as the commit message makes it obvious enough. Easily glancing over all PRs that e.g. touched the Proxy is more useful. 
I'll update based on this thinking.

.github/release-drafter.yml
patriknw: I agree, just Features is better since this is open source

patriknw (approved):
LGTM

PR #111, opened 2021-07-06
Removing "Impl" Postfix to codegen classes

franciscolopezsancho:
From codegen and docs

also https://github.com/lightbend/akkaserverless-java-sdk/issues/42


ennru (approved):
This makes it a lot nicer.
LGTM.

PR #120, opened 2021-07-08
Generate value entity effects interface

raboof:


codegen/java-gen/src/main/scala/com/lightbend/akkasls/codegen/java/EntityServiceSourceGenerator.scala
```diff
@@ -143,31 +143,35 @@ object EntityServiceSourceGenerator {
       className: String,
       interfaceClassName: String,
       entityType: String
+  ): Document = {
```
raboof, line 146: (this *looks* like a lot of changes, but that's mainly because I split out the value entity and event sourced entity cases to their own methods)

johanandren (approved):

Issue #52, opened 2021-06-17
Replace reflection with generated code


patriknw:
Since the codegen is the selected approach we should remove the Java annotation based alternative and its reflection based implementation.

* [x] Value Entities: Effect API (no implementation) #62
* [x] Value Entities: Implementation of Effect API #91
* [x] Value Entities: Update codegen to produce Effect-based interface, initial implementation and tests #120
* [x] Value Entities: 'manual' prototype of 'handler' code to replace reflection-based implementation #73
* [x] Value Entities: Update codegen to produce 'handler' code to replace reflection-based implementation #148 and #243
* [x] Event Sourced Entities: Effect API (no implementation) #62 
* [x] Event Sourced Entities: Implementation of Effect API https://github.com/lightbend/akkaserverless-java-sdk/pull/91
* [x] Event Sourced Entities: Update codegen to produce Effect-based interface, initial implementation and tests https://github.com/lightbend/akkaserverless-java-sdk/pull/132
* [x] Event Sourced Entities: Update codegen to produce 'handler' code to replace reflection-based implementation #163 (potential additional cleanup in #150), actual codegen of handler in #231
* [x] Views: Effect API (no implementation) #230
* [x] Views: Implementation of Effect API #240
* [x] Views: Update codegen to produce Effect-based interface, initial implementation and tests #240
* [x] Views: Update codegen to produce 'handler' code to replace reflection-based implementation #240
* [x] Actions: Effect API (no implementation) #258 
* [x] Actions: Implementation of Effect API #258 
* [x] Actions: Update codegen to produce Effect-based interface, initial implementation and tests #258 
* [x] Actions: Update codegen to produce 'handler' code to replace reflection-based implementation #258 
* [x] Protobuf option to support JSON API's that convert to method calls with Jackson-serialized objects
* [x] Remove annotation-based implementation
* [x] Re-enable all tests/samples/CI

raboof:
The goal here is to simplify things by having 'one way' to do things, and since via the 'codegen' we already have all the information on where things go, the annotations are not really needed anymore. So we want to replace the `AkkaServerless().registerXxx().registerYyy().start()` with something that is generated as part of the codegen and has the same effect, but without relying on the annotations.

The general approach to taking this on could be:
* Take #62 as the starting point
* Start converting the java-customer-registry sample, replacing `AkkaServerless().registerXxx().registerYyy().start()` with a hand-written version of the to-be-generated code
[...]

PR #132, opened 2021-07-09
Generate event sourced effects base class and impl

johanandren:
Refs #52


johanandren:
Stumbled upon https://github.com/lightbend/akkaserverless-java-sdk/pull/62#discussion_r662220158 and realise the context is available as a field and should not be a parameter. Saw some other comment by Patrik elsewhere that pushed me in this direction, will remove again before this PR is done.

codegen/java-gen/src/main/scala/com/lightbend/akkasls/codegen/java/EntityServiceSourceGenerator.scala
```diff
@@ -206,75 +207,59 @@ object EntityServiceSourceGenerator {
...
-              "return Reply.failure" <> parens(notImplementedError("command", command.fqn)) <> semi
-            }
-          },
-          line <> line
-        ) <>
-        (entity match {
-          case ModelBuilder.EventSourcedEntity(_, _, _, events) =>
+            )(
+              "return" <+> qualifiedType(state.fqn) <> ".getDefaultInstance();"
+            ) <>
```
octonato, line 215: Maybe better to leave it as abstract method? 

Agree that protos have their default (empty) instance, but still I think this should be an user decision.

codegen/java-gen/src/main/scala/com/lightbend/akkasls/codegen/java/EntityServiceSourceGenerator.scala
```diff
@@ -206,75 +207,59 @@ object EntityServiceSourceGenerator {
...
                 method(
                   "public",
-                  "void",
-                  lowerFirst(event.fqn.name),
+                  "Effect" <> angles(qualifiedType(command.outputType)),
+                  lowerFirst(command.fqn.name),
                   List(
-                    qualifiedType(event.fqn) <+> "event"
+                    qualifiedType(state.fqn) <+> "currentState",
+                    qualifiedType(command.inputType) <+> "command"
```
octonato, line 228: Could we use the command name instead of "command"?

codegen/java-gen/src/main/scala/com/lightbend/akkasls/codegen/java/EntityServiceSourceGenerator.scala
```diff
@@ -206,75 +207,59 @@ object EntityServiceSourceGenerator {
...
                   ),
                   emptyDoc
                 ) {
-                  "throw new RuntimeException" <> parens(
-                    notImplementedError("event", event.fqn)
-                  ) <> semi
+                  "return effects().error" <> parens(notImplementedError("command", command.fqn)) <> semi
```
octonato, line 232: same here, command name instead of "command". Though, not clear what "command" means here.

codegen/java-gen/src/main/scala/com/lightbend/akkasls/codegen/java/EntityServiceSourceGenerator.scala
```diff
@@ -206,75 +207,59 @@ object EntityServiceSourceGenerator {
...
+                  events.toSeq.map { event =>
+                    "@Override" <>
+                    line <>
+                    method(
+                      "public",
+                      qualifiedType(state.fqn),
+                      lowerFirst(event.fqn.name),
+                      List(
+                        qualifiedType(state.fqn) <+> "currentState",
+                        qualifiedType(event.fqn) <+> "event"
```
octonato, line 249: event name?

codegen/java-gen/src/main/scala/com/lightbend/akkasls/codegen/java/EntityServiceSourceGenerator.scala
```diff
@@ -206,75 +207,59 @@ object EntityServiceSourceGenerator {
...
+                      ),
+                      emptyDoc
+                    ) {
+                      "throw new RuntimeException" <> parens(
+                        notImplementedError("event", event.fqn)
```
octonato, line 254: same

octonato (approved):
Looks good with some minor suggestions

build.sbt
```diff
@@ -163,6 +164,31 @@ lazy val `java-customer-registry` = project
...
+      "net.aichler" % "jupiter-interface" % JupiterKeys.jupiterVersion.value % IntegrationTest
+    ),
+    Compile / akkaGrpcGeneratedLanguages := Seq(AkkaGrpc.Java),
+    Compile / javacOptions ++= Seq("-encoding", "UTF-8", "-source", "11", "-target", "11"),
+    testOptions += Tests.Argument(jupiterTestFramework, "-q", "-v"),
+    inConfig(IntegrationTest)(JupiterPlugin.scopedSettings),
+    IntegrationTest / akkaGrpcGeneratedSources := Seq(AkkaGrpc.Client),
+    IntegrationTest / PB.protoSources ++= (Compile / PB.protoSources).value
+  )
+ */
```
octonato, line 190: I didn't get the intention of this. Was that for local tests? 

In any case, we don't want the project in sbt but in maven and that's already setup to use the current codegen and sdk. So, maven tests should work.

codegen/java-gen/src/main/scala/com/lightbend/akkasls/codegen/java/EntityServiceSourceGenerator.scala
johanandren: I think you are right.

codegen/java-gen/src/main/scala/com/lightbend/akkasls/codegen/java/EntityServiceSourceGenerator.scala
johanandren: Sure

codegen/java-gen/src/main/scala/com/lightbend/akkasls/codegen/java/EntityServiceSourceGenerator.scala
johanandren: Should have been event I think, copy pasta.

build.sbt
johanandren: I added this to get faster feedback when working on it by including it in the sbt build, but then that applies all header and formatting and whatnot so it turned out it wasn't a good idea.

johanandren:
I'll merge this to make continuing the other work on top easier, and do the small fixes in another PR.

PR #137, opened 2021-07-12
doc: fail on warnings or errors in doc generation

pvlugter:
With the new doc builder using antora 3 (alpha pre-release), the build can be failed on warnings or errors. Previously these would only be logged and the build succeed. This will now fail the docs when includes are not found. 

Uncovered an existing error in the docs (missed file rename from #111):

```
[01:23:08.651] ERROR (asciidoctor): target of include not found: example$archetype/first-service/src/main/java/com/example/domain/Counter.java
    file: /antora/docs/src/modules/java/pages/kickstart.adoc:119
    source: /antora (refname: log-failure-levels <worktree>, start path: docs/src)
```

Passed the build before, will now fail.


johanandren (approved):
Great!

PR #141, opened 2021-07-13
run publishM2 only once

octonato:


raboof (approved):
I'm happy if CI is :laughing:

octonato:
CleanShot 2021-07-13 at 15 48 11


CI beautifully shows the job dependency graph, but build fails because artifact not found. 

Anyway, I think this is the structure we want. We now need to find out why the cache is not working as expected.

marcospereira:
> build fails because artifact not found.

That is because jobs don't share artifacts. You can try to use the same cache across different jobs, but I not sure if this is a proper solution.

octonato:
> That is because jobs don't share artifacts. You can try to use the same cache across different jobs, but I not sure if this is a proper solution.

That's the part of the puzzle we are not understanding. We are caching the artifacts. We have been doing so in an attempt to reduce build time, but now it seems that we never re-use the cache between jobs which makes it useless.

octonato:
I believe the issue is related with concurrent cache writes. Jobs restore a cache and save it back a the end. I can then override a cache containing the artifact that we just published. The next job will then fail with artifact not found error.

octonato:
@marcospereira, I'm not sure how to understand the CircleCI docs, but maybe you know more about it. Is that so that a cache is re-used between the same job (over different workflow runs), but not shared by two jobs in the same workflow?  

I'm trying now a different approach. I'm calling `publishM2` on each job that needs it, but then only once. 

Previously, a single job was calling `publishM2` more than once and that's was clearly wrong. 

Anyway, CI is building now.

pvlugter:
CircleCI has a feature called workspaces, where you can have a filesystem shared between jobs in the same workflow.

https://circleci.com/docs/2.0/persist-data/

We've used it before to only build docker images once and pass these across jobs.

.circleci/config.yml
```diff
@@ -308,83 +300,66 @@ jobs:
 workflows:
   CI:
     jobs:
-      - checks:
-          filters: # required since publish jobs have tag filters and dependencies
-            tags:
-              only: /.*/
```
octonato, line 314: As discussed during stand-up, we won't run the tests when tagging. That's the default behavior in CircleCI, hence the removal of this config settings.

.circleci/config.yml
```diff
@@ -308,83 +300,66 @@ jobs:
...
+      - validate-docs
+      - checks
       - tests:
-          filters: # required since publish jobs have tag filters and dependencies
-            tags:
-              only: /.*/
+          requires: 
+            - checks
+            - validate-docs
```
octonato, line 308: only starting the tests if the basic validation passes

.circleci/config.yml
```diff
@@ -308,83 +300,66 @@ jobs:
...
               ignore: /.*/
           requires:
-            - checks
-            - tests
-            - integration-tests
-            - maven-java-tests
-            - samples-tests    
-            - tck-tests
-            - validate-docs   
+            - publish
```
octonato, line 344: only publish maven if main artifacts got published

.circleci/config.yml
```diff
@@ -308,83 +300,66 @@ jobs:
...
               only: /^v.*/
             branches:
               ignore: /.*/
           requires:
-            - checks
-            - tests
-            - integration-tests
-            - tck-tests
-            - validate-docs
+            - publish
```
octonato, line 353: I'm wondering why the tck is an apart publishing step. Shouldn't that be just part of the main artifacts? 

Anyway, we can change on follow-up PR.

octonato (commented):
This is ready for another review. 

I added the build dependencies changes we discussed during stand-up.

.circleci/config.yml
octonato: Nevermind, I forgot that the TCK involves the publishing of a docker image. So, makes sense to have it's own step.

octonato:
> LGTM but I don't know circle-ci enough to know if there is still some issue with shared cache whatnot that you discussed during standup.

that's the surprise effect, we will only see if after we merge and try to release :-)

PR #142, opened 2021-07-13
Implement event sourced effects and managed state.

johanandren:
A bit invasive as I gave up on not breaking the old reply style APIs after struggling with that a for too long.


johanandren:
Woho, all tests green, now I'll see if I can clean this up a bit and encapsulate the command handling a bit so we can re-use in the testkit.

johanandren:
This is still a bit messy, but it'd be nice to merge and work from here rather than do more in this PR given that I only have tomorrow left before vacation.

codegen/java-gen/src/main/scala/com/lightbend/akkasls/codegen/java/EntityServiceSourceGenerator.scala
```diff
@@ -211,7 +211,9 @@ object EntityServiceSourceGenerator {
               Nil,
               emptyDoc
             )(
-              "return" <+> qualifiedType(state.fqn) <> ".getDefaultInstance();"
+              "throw new UnsupportedOperationException" <> parens(
+                dquotes("Not implemented yet, replace with your empty entity state")
+              ) <> semi
```
johanandren, line 216: These are codegen changes requested in #132: 
 * dont' generate a working default empty state
 * name the command and events after their type instead of `command` and `event`

samples/java-eventsourced-shopping-cart/src/main/java/com/example/shoppingcart/domain/ShoppingCartImpl.java
```diff
@@ -39,7 +39,7 @@ public ShoppingCartImpl(String entityId) {
   }
 
   @Override
-  protected ShoppingCartDomain.Cart emptyState() {
+  public ShoppingCartDomain.Cart emptyState() {
```
johanandren, line 42: Now public because we need to call it from the outside.

sdk/src/main/java/com/akkaserverless/javasdk/eventsourcedentity/EventSourcedEntityBase.java
```diff
@@ -23,13 +23,17 @@
...
+import java.util.Optional;
 import java.util.function.Function;
 
 // FIXME rename to EventSourcedEntity when the old annotation is removed
 
 /** @param <S> The type of the state for this entity. */
 public abstract class EventSourcedEntityBase<S> {
 
+  private Optional<CommandContext> commandContext = Optional.empty();
+  private Optional<EventContext> eventContext = Optional.empty();
```
johanandren, line 35: We toggle these between empty and actually containing a context for the respective calls, we don't expect that they will be used in the general case anymore since they only contain access to metadata etc. which usually isn't needed.

sdk/src/main/java/com/akkaserverless/javasdk/eventsourcedentity/EventSourcedEntityBase.java
```diff
@@ -38,27 +42,40 @@
    *
    * <p><code>null</code> is an allowed value.
    */
-  protected abstract S emptyState();
+  public abstract S emptyState();
```
johanandren, line 45: Now public because we need to call it from the outside.

sdk/src/main/java/com/akkaserverless/javasdk/eventsourcedentity/EventSourcedEntityBase.java
```diff
@@ -79,6 +96,8 @@ protected EventContext eventContext() {
 
       OnSuccessBuilder<S> emitEvent(Object event);
 
+      OnSuccessBuilder<S> emitEvents(Collection<? extends Object> event);
```
johanandren, line 99: Emitting multiple events needed to live up to TCK test.

sdk/src/main/java/com/akkaserverless/javasdk/lowlevel/EventSourcedEntityHandler.java
```diff
@@ -49,20 +45,15 @@
    * @param context The command context.
    * @return The reply to the command, if the command isn't being forwarded elsewhere.
    */
-  Reply<Any> handleCommand(Any command, CommandContext context);
+  EventSourcedEntityBase.Effect<Any> handleCommand(Any command, CommandContext context);
```
johanandren, line 48: Note that this effect is only half done as we haven't applied the events yet, and therefore the final reply message has not been created. Actual interpretation of this effect is in the `EventSourcedEntitiesImpl` stream (for now)

sdk/src/main/scala/com/akkaserverless/javasdk/impl/AnySupport.scala
```diff
@@ -389,6 +389,11 @@ class AnySupport(descriptors: Array[Descriptors.FileDescriptor],
           scalaPbMessage.toByteString
         )
 
+      case null =>
+        throw SerializationException(
+          s"Don't know how to serialize object of type null. Try passing a protobuf, using a primitive type, or using a type annotated with @Jsonable."
+        )
```
johanandren, line 395: Got here when reflect-calling old void `@EventHandlers` so won't happen in the future but for good measure.

sdk/src/main/scala/com/akkaserverless/javasdk/impl/ReflectionHelper.scala
```diff
@@ -197,12 +199,12 @@ private[impl] object ReflectionHelper {
     handlers.asInstanceOf[Array[ParameterHandler[M, C]]]
   }
 
-  def verifyAtMostOneMainArgument[M, C <: Context](name: String,
-                                                   method: Method,
-                                                   parameters: Array[ParameterHandler[M, C]]) =
-    if (parameters.count(_.isInstanceOf[MainArgumentParameterHandler[_, _]]) > 1) {
+  def verifyAtMostTwoMainArguments[M, C <: Context](name: String,
```
johanandren, line 202: This messes with other component reflection, but we will anyway drop support for that so I didn't care so much.

sdk/src/main/scala/com/akkaserverless/javasdk/impl/ReflectionHelper.scala
```diff
@@ -275,14 +277,15 @@ private[impl] object ReflectionHelper {
...
-    val mainArgumentDecoder: JavaPbAny => AnyRef = parameters
-      .collectFirst {
-        case MainArgumentParameterHandler(inClass) =>
-          getMainArgumentDecoder(name, inClass, serviceMethod.inputType)
-      }
-      .getOrElse(_ => NotUsed)
+    val mainArgumentDecoder: JavaPbAny => AnyRef =
+      parameters
+        .collectFirst {
+          case MainArgumentParameterHandler(inClass) if inClass == serviceMethod.inputType.typeClass =>
```
johanandren, line 285: First "main argument parameter"  is now the state :/

sdk/src/main/scala/com/akkaserverless/javasdk/impl/ReflectionHelper.scala
```diff
@@ -315,6 +318,18 @@ private[impl] object ReflectionHelper {
...
       verifyOutputType(getFirstParameter(method.getGenericReturnType))
 
+      result =>
+        result.asInstanceOf[ValueEntityEffectImpl[_]].secondaryEffect match {
+          case ErrorReplyImpl(description, _) => Reply.failure(description)
+          case ForwardReplyImpl(serviceCall, _) => Reply.forward(serviceCall)
+          case MessageReplyImpl(message, metadata, _) => Reply.message(serialize(message), metadata)
+          case NoReply(_) => Reply.noReply()
+          case NoSecondaryEffectImpl => Reply.noReply()
+        }
```
johanandren, line 328: Not entirely sure this is actually ever executed

sdk/src/main/scala/com/akkaserverless/javasdk/impl/eventsourcedentity/EventSourcedEntitiesImpl.scala
```diff
@@ -177,53 +183,84 @@ final class EventSourcedEntitiesImpl(_system: ActorSystem,
...
+                // FIXME currently snapshotting final state after applying all events even if trigger was mid-event stream?
+                val snapshot =
+                  if (shouldSnapshot) Option(ScalaPbAny.fromJavaProto(handler.currentState()))
+                  else None
+                (scalaPbEvents.result(),
+                 commandEffect.secondaryEffect(service.anySupport.decode(handler.currentState())),
+                 snapshot)
+              case NoPrimaryEffect =>
+                (Vector.empty, commandEffect.secondaryEffect(handler.currentState()), None)
+            }
```
johanandren, line 213: This is the interpretation "engine" that we might want to share with the testkit (but sans the back and forth with `PbAny`s) - invoke command handler, apply emitted events, figure out end state, potential snapshot and side effects (such as reply to the user)

sdk/src/main/scala/com/akkaserverless/javasdk/impl/eventsourcedentity/EventSourcedEntitiesImpl.scala
```diff
@@ -177,53 +183,84 @@ final class EventSourcedEntitiesImpl(_system: ActorSystem,
...
+               ))
+
+            case ok =>
+              (endSequenceNumber,
+               Some(
+                 OutReply(
+                   EventSourcedReply(
+                     command.id,
+                     clientAction,
+                     context.sideEffects ++ EffectSupport.sideEffectsFrom(serializedSecondaryEffect),
```
johanandren, line 258: We should remove the side effects gathering from the context completely

sdk/src/main/scala/com/akkaserverless/javasdk/impl/eventsourcedentity/EventSourcedEntitiesImpl.scala
```diff
@@ -254,18 +291,8 @@ final class EventSourcedEntitiesImpl(_system: ActorSystem,
...
     final var performSnapshot: Boolean = false
 
-    override def emit(event: AnyRef): Unit = {
-      checkActive()
-      val encoded = anySupport.encodeScala(event)
-      val nextSequenceNumber = sequenceNumber + events.size + 1
-      handler.handleEvent(ScalaPbAny.toJavaProto(encoded), new EventContextImpl(entityId, nextSequenceNumber))
-      events :+= encoded
-      performSnapshot = (snapshotEvery > 0) && (performSnapshot || (nextSequenceNumber % snapshotEvery == 0))
-    }
```
johanandren, line 267: This is the old "engine", handling an event immediately when the imperative context.emit was called so that the next line of the command handler after calling this would already see the updated state.

sdk/src/test/scala/com/akkaserverless/javasdk/impl/eventsourcedentity/EventSourcedEntitiesImplSpec.scala
```diff
@@ -132,28 +139,6 @@ class EventSourcedEntitiesImplSpec extends AnyWordSpec with Matchers with Before
...
-    }
-
-    "fail when snapshot handler throws exception" in {
-      service.expectLogError("Terminating entity due to unexpected failure") {
-        val entity = protocol.eventSourced.connect()
-        entity.send(init(ShoppingCart.Name, "cart", snapshot(42, cartSnapshot())))
-        entity.expect(failure("Unexpected failure: Boom: no items"))
-        entity.expectClosed()
-      }
-    }
```
johanandren, line 155: No more snapshot handler (for now at least, but we might want something like it, but optional, moving forward turning an internal inmem state (some convenient type) to and from the stored representation (protobuf)?

sdk/src/test/scala/com/akkaserverless/javasdk/impl/eventsourcedentity/EventSourcedEntitiesImplSpec.scala
```diff
@@ -200,23 +185,6 @@ class EventSourcedEntitiesImplSpec extends AnyWordSpec with Matchers with Before
...
-        entity.send(command(1, "cart", "AddItem", addItem("foo", "bar", -42)))
-        entity.expect(actionFailure(1, "Cannot add negative quantity of item [foo]", restart = true))
-        entity.passivate()
-        val reactivated = protocol.eventSourced.connect()
-        reactivated.send(init(ShoppingCart.Name, "cart"))
-        reactivated.send(command(1, "cart", "GetCart"))
-        reactivated.expect(reply(1, EmptyCart))
-        reactivated.passivate()
-      }
-    }
```
johanandren, line 218: Not possible anymore with effect api

sdk/src/test/scala/com/akkaserverless/javasdk/impl/eventsourcedentity/EventSourcedEntitiesImplSpec.scala
```diff
@@ -287,50 +255,64 @@ object EventSourcedEntitiesImplSpec {
...
-        Empty.getDefaultInstance
+      def removeItem(currentState: ShoppingCartDomain.Cart, item: ShoppingCartApi.RemoveLineItem): Effect[Empty] = {
+        throw new RuntimeException("Boom: " + item.getProductId) // always fail for testing
       }
 
-      @Snapshot
-      def snapshot: ShoppingCartDomain.Cart = Protocol.cartSnapshot(cart.values.toSeq: _*)
+      private def domainCartToMap(cart: ShoppingCartDomain.Cart): Map[String, Item] = {
+        cart.getItemsList.asScala.map(li => li.getProductId -> Item(li.getProductId, li.getName, li.getQuantity)).toMap
+      }
```
johanandren, line 301: This is basically that stored-type to and from convenient type transformation

sdk/src/main/java/com/akkaserverless/javasdk/eventsourcedentity/EventSourcedEntityBase.java
```diff
@@ -38,27 +42,40 @@
...
    * <p>It will throw an exception if accessed from constructor or event handler.
    */
-  protected CommandContext commandContext() {
-    throw new UnsupportedOperationException("Not implemented yet"); // FIXME
+  protected final CommandContext commandContext() {
+    return commandContext.orElseThrow(
+        () ->
+            new IllegalStateException("CommandContext is only available when handling a command."));
+  }
+
```
octonato, line 57: That's good!

sdk/src/main/java/com/akkaserverless/javasdk/eventsourcedentity/EventSourcedEntityBase.java
```diff
@@ -38,27 +42,40 @@
...
   /**
    * Additional context and meta data for an event handler.
    *
    * <p>It will throw an exception if accessed from constructor or command handler.
    */
-  protected EventContext eventContext() {
-    throw new UnsupportedOperationException("Not implemented yet"); // FIXME
+  protected final EventContext eventContext() {
+    return eventContext.orElseThrow(
+        () -> new IllegalStateException("EventContext is only available when handling an event."));
```
octonato, line 70: 👍

sdk/src/main/java/com/akkaserverless/javasdk/eventsourcedentity/EventSourcedEntityBase.java
octonato: Could have some more variants and avoid `Colleciton`? 

```java
emitEvent(Object event);
emitEvent(Object event1, Object event2);
emitEvent(Object event1, Object event2, Object event3);
emitEvent(Object event1, Object event2, Object event3, Object* otherEvents);
```

octonato (commented):
@johanandren, I did a first pass. 

Will continue later.

sdk/src/test/scala/com/akkaserverless/javasdk/impl/eventsourcedentity/EventSourcedEntitiesImplSpec.scala
octonato: The generated classes will be tailored to work with the state that is declared in the proto, so I don't think we can do anything here. 

Users can always transform it to something more easier to use (with setter and getters if they prefer so), but ultimately they will get proto in and need to return proto, right?

sdk/src/test/scala/com/akkaserverless/javasdk/impl/eventsourcedentity/EventSourcedEntitiesImplSpec.scala
```diff
@@ -287,50 +255,64 @@ object EventSourcedEntitiesImplSpec {
...
+        if (updatedItem.getName == "FAIL") throw new RuntimeException("Boom: name is FAIL") // fail for testing
+        val currentItems = domainCartToMap(currentState)
+        val updatedItems = currentItems.updatedWith(
+          updatedItem.getProductId
+        ) {
+          case Some(existing) =>
+            Some(Item(updatedItem.getProductId, updatedItem.getName, existing.quantity + updatedItem.getQuantity))
+          case None => Some(Item(command.getItem.getProductId, command.getItem.getName, command.getItem.getQuantity))
+        }
+        domainCartFromMap(updatedItems)
```
octonato, line 291: `(State, Command) => Effect` and `(State, Event) => State` looks so much better. 

IMO, it makes very clear what is the expected implementation. Especially for the event handler.

`itemAdded(itemAdded: ShoppingCartDomain.ItemAdded): Unit` says nothing. Could be anything.

octonato (approved):
Looks good. 

Is there anything else you think we should add on this PR? Or are we good for merge?

sdk/src/test/scala/com/akkaserverless/javasdk/impl/eventsourcedentity/EventSourcedEntitiesImplSpec.scala
johanandren: I agree 👍

sdk/src/test/scala/com/akkaserverless/javasdk/impl/eventsourcedentity/EventSourcedEntitiesImplSpec.scala
johanandren: Yes and no, that's what we generate, but if handcoding an entity, I can actually even have a primitive as state or something `@Jsonable` and that is automagically turned into proto by the serialization (https://github.com/lightbend/akkaserverless-java-sdk/blob/main/sdk/src/main/scala/com/akkaserverless/javasdk/impl/AnySupport.scala#L392). That's pretty neat, and would be nice if we could get to some way doing that with the static/codegen design as well (but not a priority right now).

sdk/src/main/java/com/akkaserverless/javasdk/eventsourcedentity/EventSourcedEntityBase.java
johanandren: I think I agree, but I don't want to change it here, so let's revisit that later.

johanandren: That was wrt persist vs emit. The vararg overload I added now.

PR #145, opened 2021-07-14
fix: build maven project using jdk11

octonato:
For unknown reason, we were building the maven projects using jdk8. It turns out that since we first need to publish the codegen locally and the [sbt is configured](https://github.com/lightbend/akkaserverless-java-sdk/blob/main/build.sbt#L48-L49) with `--release` flag, we get the following error:

```java
-release is only supported on Java 9 and higher
```
(https://app.circleci.com/pipelines/github/lightbend/akkaserverless-java-sdk/503/workflows/fcbd35cb-21e1-44a9-a9fe-a36ef028d7f2/jobs/3088)

For the record, when I say for unknown reason I mean, I don't know why the old maven-java project was using jdk8. What we have now here is a copy-n-paste from that previous project. Anyway, my understanding is that we only build with jdk11 (not only on this project, but every akka serverless project) therefore align with jdk11 is the way to go.

### DIFF
--- a/maven-java/pom.xml
+++ b/maven-java/pom.xml
@@ -31,6 +31,7 @@
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     <maven.compiler.source>1.8</maven.compiler.source>
     <maven.compiler.target>1.8</maven.compiler.target>
+    <maven.compiler.release>8</maven.compiler.release>
     <maven.version>3.3.9</maven.version>
   </properties>
 
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -233,7 +233,7 @@ jobs:
 
   publish-maven:
     docker:
-      - image: circleci/openjdk:8
+      - image: circleci/openjdk:11
     steps:
       - checkout
       - restore_deps_cache


pvlugter (commented):
Do we have a `-release 8` setting for the maven-java build as well? We still want to target and support JDK 8, even if we're building on JDK 11.

johanandren:
Isn't it safer to compile using Java 8 if that is what we want to support?

octonato:
The maven-java build itself is not using the flag, but the job is configured to use jdk8 and when it comes to build the codegen and the sdk through `publishM2`, it fails because of the flag.

However, the maven build is configured to target jdk8. (see https://github.com/lightbend/akkaserverless-java-sdk/blob/main/maven-java/pom.xml#L30-L35)

I agree with Johan that if the goal is to support JDK8, then we should build with it (as well).

In any case, I would rather unlock this for now and add proper JDK8 support in another PR.

raboof (changes requested):
I'm OK with this if we make sure to also set `maven.compiler.release` for the maven builds - otherwise we might run into problems because we build against the jdk11 bootclasspath even if we generate java8 bytecode.

I think it's hard to say what is "safer": running on jdk8 when building the release avoids forgetting setting the release flag, on the other hand setting the release flag avoids forgetting to run on jdk8 when building :). Perhaps setting the release flag is slightly nicer since it makes building locally more similar to building on CI.

maven-java/pom.xml
```diff
@@ -31,6 +31,7 @@
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     <maven.compiler.source>1.8</maven.compiler.source>
     <maven.compiler.target>1.8</maven.compiler.target>
+    <maven.compiler.release>8</maven.compiler.release>
```
octonato, line 34: Not clear if we should keep source and target. I didn't find clear documentation about it. I may be that `release` is sufficient. In any case, I don't want to fiddle with this right now.

raboof (approved):

johanandren (approved):

Issue #118, opened 2021-07-08
TestKit for unit testing


patriknw:
We need a TestKit for running EventSourcedEntity and ValueEntity implementations in a unit test. The previous Mockito approach of mocking the context will not work for the effects, and it's anyway exposing to much internals to the user.

The goal is to feed the implementation with a command and verify the outcome. Much similar to the [EventSourcedBehaviorTestKit in Akka](https://doc.akka.io/docs/akka/current/typed/persistence-testing.html#unit-testing).

The assertions should not be on the returned effect but on a test specific Result, which is populated with things like:
* emitted events
* optional reply
* optional errors
* optional forward

On the implementation side of this it should probably not take the same approach as EventSourcedBehaviorTestKit. EventSourcedBehaviorTestKit is running the EventSourcedBehavior as a real actor, because that's the only realistic way to make sure that is doing the real thing. In Akka Serverless we can have a synchronous test runner instead of the real EventSourcedEntitiesImpl. It is simple enough to have a second implementation for testing purposes.

patriknw:
One challenge is how to make use of the generated dispatch from rpc method name to invoking the Java command handler method. After discussing with @raboof and @johanandren we came up with an idea that it might be nice to call a client implementation of the gRPC Service from the test. That test implementation could be generated and make use of generated method dispatch logic. Details must be explored. Anyway, this is not the first thing to start with since the codegen is not ready yet.

PR #149, opened 2021-07-15
Wip unit testkit

franciscolopezsancho:
creating an specific test kit for `ShoppingCartImpl`

#118


samples/java-eventsourced-shopping-cart/src/main/java/com/example/shoppingcart/domain/ShoppingCartTestKit.java
```diff
@@ -0,0 +1,62 @@
...
+import com.google.protobuf.Empty;
+import java.util.ArrayList;
+import java.util.List;
+
+// Q Would we need ShoppingCartImpl as parameter constructor? 
+public class ShoppingCartTestKit {
+
+    ShoppingCartDomain.Cart currentState = ShoppingCartDomain.Cart.newBuilder().build();
+
+    // Q What about having many types of events?
```
franciscolopezsancho, line 14: This is the question that concerns me the most by now

samples/java-eventsourced-shopping-cart/src/main/java/com/example/shoppingcart/domain/ShoppingCartTestKit.java
```diff
@@ -0,0 +1,62 @@
...
+import com.example.shoppingcart.ShoppingCartApi;
+import com.google.protobuf.Empty;
+import java.util.ArrayList;
+import java.util.List;
+
+// Q Would we need ShoppingCartImpl as parameter constructor? 
+public class ShoppingCartTestKit {
+
+    ShoppingCartDomain.Cart currentState = ShoppingCartDomain.Cart.newBuilder().build();
+
```
franciscolopezsancho, line 13: This is the question that concerns me the most by now

johanandren: Yeah, we can't know, so that will have to be part of the assertions possible on the result, one suggestion is something like the Akka testkit probes:

```
MyEvent event = result.expectEventOfType(MyEvent.class);
assertEquals(event.something, "some value");
```

This is easy for the normal case of one event, but for multiple it is a bit trickier to provide something convenient, but perhaps `result.expectNthEventOfType(index, MyEvent.class)` or something along those lines?

samples/java-eventsourced-shopping-cart/src/main/java/com/example/shoppingcart/domain/ShoppingCartTestKit.java
```diff
@@ -0,0 +1,62 @@
...
+        return new Result(Empty.getDefaultInstance(), new ArrayList<ShoppingCartDomain.ItemRemoved>(), currentState);
+    };
+
+    // Q or get state?
+    // Q what about the events in this one? what type?
+    public Result getCart(ShoppingCartApi.GetShoppingCart command) {
+        return new Result(Empty.getDefaultInstance(), new ArrayList<ShoppingCartDomain.ItemRemoved>(), currentState);
+    };
+
+    public final class Result<Event> {
```
johanandren, line 37: This won't have to be generated but can be a generic `EventSourcedEntityCommandResult<S>`, the event type we do not know, no common supertype with protobuf so will have to be `List<Object>` and then nice methods to inspect/assert

johanandren (commented):
I like this direction.

samples/java-eventsourced-shopping-cart/src/main/java/com/example/shoppingcart/domain/ShoppingCartTestKit.java
raboof: taking inspiration from the actor testkit, perhaps invoking `expectEventOfType()` multiple times should return the subsequent events?

franciscolopezsancho: `com.google.protobuf.GeneratedMessageV3` seems the supertype

franciscolopezsancho: Good point guys, `expectEventOfType` looks good. I think calling `expectEventOfType()` multiple times is also in tune with the `def event` in `EventSourcedBehaviorTestKitImpl`.

I should create a similar `def event` as well now that I think about it.

samples/java-eventsourced-shopping-cart/src/main/java/com/example/shoppingcart/domain/ShoppingCartTestKit.java
franciscolopezsancho: Good point. Doesn't need to be generated. What about `com.google.protobuf.GeneratedMessageV3` instead of `Object`?

johanandren: I don't think that has much value to the user, it's more of an implementation detail.

franciscolopezsancho: I see what you mean, those nice methods to inspect/assert you mention should suffice. The user doesn't gain any type checking with dealing with `GeneratedMessageV3` but with the methods.

samples/java-eventsourced-shopping-cart/src/main/java/com/example/shoppingcart/domain/ShoppingCartTestKit.java
```diff
@@ -0,0 +1,53 @@
...
+
+// Q Would we need ShoppingCartImpl as parameter constructor? 
+public class ShoppingCartTestKit {
+    
+    ShoppingCartDomain.Cart state;
+
+    public ShoppingCartTestKit(ShoppingCartDomain.Cart state){
+        this.state = state;
+    }
+
```
franciscolopezsancho, line 19: This seems more intuitive for the user, to set an initial state from the beginning, so they know where to `get` it and as @octonato suggested they can initiate the test in some complex scenario without the hassle of running the equivalent commands to reach that state.

octonato (commented):
Looking good @franciscolopezsancho 

I think the next step is to have a working testkit. Not generated yet, but with two parts. A generic one that would live in the SDK and the later-to-be-generated `ShoppingCartTestkit` that should extend the generic one.

Then we need to try to re-use the machinery to interpret the Effects. @raboof was looking into how to make the machinery re-usable (if possible). So maybe good to catch-up with him.

samples/java-eventsourced-shopping-cart/src/main/java/com/example/shoppingcart/domain/ShoppingCartTestKit.java
```diff
@@ -0,0 +1,53 @@
...
+import com.google.protobuf.Empty;
+import java.util.ArrayList;
+import java.util.LinkedList;
+import java.util.Queue;
+import com.akkaserverless.javasdk.testkit.Result;
+
+// Q Would we need ShoppingCartImpl as parameter constructor? 
+public class ShoppingCartTestKit {
+    
+    ShoppingCartDomain.Cart state;
```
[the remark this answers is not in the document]
franciscolopezsancho: Yep, it should be private. Not final, though. I need to reassign.

samples/java-eventsourced-shopping-cart/src/test/java/com/example/shoppingcart/domain/ShoppingCartTest.java
```diff
@@ -0,0 +1,59 @@
...
+        entity = new ShoppingCartImpl(entityId);
+
+        // GIVEN
+        ShoppingCartDomain.Cart initialState = ShoppingCartDomain.Cart.newBuilder().build();
+        ShoppingCartApi.AddLineItem command = ShoppingCartApi.AddLineItem.newBuilder().setProductId("id1")
+                .setName("name").setQuantity(2).build();
+
+        // Q do we want to pass the entity?
+                //before and after, clean?  (not yet)
+        ShoppingCartTestKit shoppingCartTestKit = new ShoppingCartTestKit(initialState);
```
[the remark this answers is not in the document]
franciscolopezsancho: This looks good

samples/java-eventsourced-shopping-cart/src/main/java/com/example/shoppingcart/domain/ShoppingCartTestKit.java
```diff
@@ -0,0 +1,53 @@
...
+
+    public ShoppingCartTestKit(ShoppingCartDomain.Cart state){
+        this.state = state;
+    }
+
+    // Q - better have the state outside always accessible
+    //     than adding another field to Result
+    public ShoppingCartDomain.Cart getState(){
+            return state;
+    }
```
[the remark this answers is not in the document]
franciscolopezsancho: `shoppingCartTestKit.getAllEvents()` seems very useful.  Good idea.

samples/java-eventsourced-shopping-cart/src/main/java/com/example/shoppingcart/domain/ShoppingCartTestKit.java
```diff
@@ -0,0 +1,53 @@
...
+
+import com.example.shoppingcart.domain.ShoppingCartDomain;
+import com.example.shoppingcart.ShoppingCartApi;
+import com.google.protobuf.Empty;
+import java.util.ArrayList;
+import java.util.LinkedList;
+import java.util.Queue;
+import com.akkaserverless.javasdk.testkit.Result;
+
+// Q Would we need ShoppingCartImpl as parameter constructor?
```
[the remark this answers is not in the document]
franciscolopezsancho: true. better do it ourselves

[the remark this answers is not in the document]
raboof: >  Since they are both generated we can just call `new ShoppingCart`

This is the implementation by the user, right? I think it'd make sense to have the user instantiate that: that way they can pass any constructor parameters they like (i.e. settings, ...).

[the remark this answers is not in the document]
octonato: ahah, @raboof already passed here to say that. Thanks!

samples/java-eventsourced-shopping-cart/src/main/java/com/example/shoppingcart/domain/ShoppingCartTestKit.java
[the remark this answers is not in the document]
octonato: yep, correct

samples/java-eventsourced-shopping-cart/src/test/java/com/example/shoppingcart/domain/ShoppingCartTest.java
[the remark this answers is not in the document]
octonato: after the discussion about constructor...

```java
 private final ShoppingCart entity;	 
 private ShoppingCartDomain.Cart state;

 public ShoppingCartTestKit(ShoppingCart entity){
	this.entity = entity;
	this.state = this.entity.emptyState();
 }

 public ShoppingCartTestKit(ShoppingCart entity, ShoppingCartDomain.Cart state){
	this.entity = entity;
	this.state = state;
 }
```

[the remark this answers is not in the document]
franciscolopezsancho: continuing on https://github.com/lightbend/akkaserverless-java-sdk/pull/172

franciscolopezsancho:
continuing on https://github.com/lightbend/akkaserverless-java-sdk/pull/172

PR #150, opened 2021-07-15
[WIP] non reflection event sourced entities

johanandren:
Refs #52 

On top of #142


sdk/src/test/java/com/akkaserverless/javasdk/eventsourcedentity/CartHandler.java
```diff
@@ -0,0 +1,81 @@
...
+  }
+
+  @Override
+  public EventSourcedEntityBase.Effect<?> handleCommand(
+      String commandName, ShoppingCartDomain.Cart state, Any command) {
+    try {
+      switch (commandName) {
+        case "AddItem":
+          // FIXME could parsing to the right type also be pulled out of here?
+          return entity().addItem(state, ShoppingCartApi.AddLineItem.parseFrom(command.getValue()));
```
octonato, line 51: I was looking into that yesterday as well. 

If handleCommand gets an `Object command` instead of `Any`, we could check if it's already parsed or not. 

Each command type would need a generated companion, something like: 

```java
class AddLineItemComp {
  public static AddLineItem cast(Object cmd) {
	if (cmd instance Any) 
		return ShoppingCartApi.AddLineItem.parseFrom(((Any) cmd).getValue()));
	else if (cmd instance AddLineItem) 
		return (AddLineItem) cmd;
	else throw new RuntimeException(...);
  }
}
```

Or maybe something like: 

```java
class ShoppingCartConversions {
  public static AddLineItem toAddLineItem(Object cmd);
}
```

That would also allow use to support other formats, like Jsonable or whatever.

johanandren: I'm not sure we would need the generated companion either, I think that the command contains the right information to deserialize into the right type already, just that the typesystem won't know it. So we should be able to deserialize elsewhere and get the right kind of object (but looks like `AnyRef` to us/the typesystem) then we can expect that to be right here and just cast it to the right type here.

johanandren: So for the record, the point of pulling that elsewhere is that for the testkit we can just send the command object and not first have to do any serialization trixery.

octonato: To pull it elsewhere, we will need to generate another class that parse `Any` to the right type and then pass to the handleCommand, right? 

And if have two variants of handleCommand, one accepting proto Any and another accepting Object? 

Would that be sufficient?

johanandren: I think we should have one logic that works for both, and that it should be to do serialization/deser in the "impl" and then deal with "normal" objects the rest of the way. My impression of having it this close to the actual logic is that it was done like that to support reflection/annotation magics with no parameter, json, primitives and whatnot but now we are aiming for simplicity rather than that ultimate flexibility.

I'm trying out doing this right now, but it turns out the any support doesn't have all types I thought it would. Investigating if I can get around that, hand it more descriptors or something.

johanandren: I got away with doing that, what was needed was passing both (all) the needed proto descriptors to `AnySupport` when creating it, so in most event sourcing cases both the `API` and the `Domain` proto descriptor. Since we are anyway going to codegen that part (I think at least) that shouldn't be a problem.

raboof:
The first part of this has been merged as #163 . I didn't integrate the second commit yet because I wasn't confident this approach would keep working once we add support for jackson, and it isn't clear yet whether it's needed for the testkit.

patriknw:
Approach of second commit incorporated in https://github.com/lightbend/akkaserverless-java-sdk/pull/236

PR #148, opened 2021-07-15
feat: codegen for Value Entity Handler

octonato:

PR #151, opened 2021-07-15
fix: ValueEntityHandler returns Effect of GeneratedMessageV3

octonato:


samples/java-customer-registry/src/main/java/customer/CustomerValueEntityHandler.java
```diff
@@ -56,36 +54,43 @@
...
-      if (e.getClass().isAssignableFrom(InvocationTargetException.class) && e.getCause() != null) {
-        unwrapped = e.getCause();
-      } else {
-        unwrapped = e;
-      }
-      if (unwrapped.getClass().isAssignableFrom(FailInvoked$.class)) {
-        throw unwrapped;
-      } else {
-        throw unwrapped.getCause();
-      }
```
octonato, line 88: I don't think we need extra exception handling here. InvocationTargetException we could get if we were doing some reflection and that's exactly what we are not doing. 

Then, if we get `FailInvoked` we just want to rethrow it. 

Finally, in the last case we throw the cause of the exception we got, but not clear what could that exception be and why we are doing so. I suggest that we just don't do any exception handling here.

samples/java-customer-registry/src/main/java/customer/CustomerValueEntityHandler.java
```diff
@@ -56,36 +54,43 @@
   }
 
   @Override
-  public ValueEntityBase.Effect<Any> handleCommand(
+  public ValueEntityBase.Effect<? extends GeneratedMessageV3> handleCommand(
       Any command, Any state, CommandContext<Any> context) throws Throwable {
+
```
octonato, line 59: this is a leftover from #73

sdk/src/main/java/com/akkaserverless/javasdk/valueentity/ValueEntityBase.java
```diff
@@ -44,7 +44,7 @@
    *
    * <p>It will throw an exception if accessed from constructor.
    */
-  protected CommandContext commandContext() {
+  protected CommandContext<S> commandContext() {
```
octonato, line 47: That was wrong and nobody noticed.

octonato (commented):
@raboof, I'm sending this a separate PR. 

If contains a leftover from #73 and the exception handling part. 

This is also in #148, but I will rebase that other PR on top of this one.

build.sbt
```diff
@@ -123,7 +123,7 @@ lazy val samples = project
     // FIXME include samples again
     // `java-eventsourced-shopping-cart`
     // `java-eventing-shopping-cart`,
-    // `java-customer-registry`
+    `java-customer-registry`
```
johanandren, line 126: Wait, do we want to include in sbt build or not?

samples/java-customer-registry/src/main/java/customer/CustomerValueEntityHandler.java
johanandren: I think this should be just Any/Object, why should it be GeneratedMessageV3?

build.sbt
octonato: I think we want to for now. We are using this project as the API experimentation project, so good to compile it each time.

samples/java-customer-registry/src/main/java/customer/CustomerValueEntityHandler.java
octonato: Because the type that is returned from the generate code is the user defined type, which is a `GeneratedMessageV3`, not an `Any`.

johanandren: As far as I understand they could return anything though, for example some class annotated with @jsonable and the serialization will handle that. Requiring it to be a generated message disallows that?

PR #154, opened 2021-07-16
fix: set current sdk version using @project.version@

octonato:


maven-java/akkaserverless-maven-archetype-event-sourced-entity/src/main/resources/archetype-resources/pom.xml
```diff
@@ -19,7 +19,7 @@
     <jdk.target>11</jdk.target>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
 
-    <akkaserverless-sdk.version>0.7.0-beta.10</akkaserverless-sdk.version>
+    <akkaserverless-sdk.version>@project.version@</akkaserverless-sdk.version>
```
octonato, line 22: Actually we had already a solution for this, but it was not being used for the sdk version, but only for the maven-plugin version. 

We want sdk, testkit and maven-plugin to use `akkaserverless-sdk.version` and it's now set using `@project.version@`.

maven-java/akkaserverless-maven-archetype-value-entity/src/main/resources/archetype-resources/pom.xml
```diff
@@ -206,7 +206,7 @@
       <plugin>
         <groupId>com.akkaserverless</groupId>
         <artifactId>akkaserverless-maven-plugin</artifactId>
-        <version>@project.version@</version>
+        <version>${D}{akkaserverless-sdk.version}</version>
```
johanandren, line 209: `${D}`?

octonato: This is to avoid premature substitution. It's used everywhere in this file whenever we want the archetype to have a ${var}. 

However, it may not be needed. I can check it later, but don't want to change more than necessary here.

octonato: For instance, like here:

https://github.com/lightbend/akkaserverless-java-sdk/blob/main/maven-java/akkaserverless-maven-archetype-value-entity/src/main/resources/archetype-resources/pom.xml#L11-L33

johanandren: Trixy!

johanandren (approved):

PR #156, opened 2021-07-16
feat: samples uses project.version as sdk version

octonato:
Those changes will allows us to:

* change the sdk version in all mvn projects in one shot (changes can be pushed or not)
* samples use its own project.version as sdk version
* remove passing -D in CI, it suffice to use `mvn versions:set`

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -176,14 +176,14 @@ jobs:
             cd maven-java
             mvn -B archetype:generate -DgroupId=com.example -DartifactId=counter-value-entity -DarchetypeGroupId=com.akkaserverless -DarchetypeArtifactId=akkaserverless-maven-archetype -DarchetypeVersion=$SDK_VERSION
             cd counter-value-entity
-            mvn -B -Dakkaserverless-sdk.version=$SDK_VERSION compile
+            mvn -B compile
       - run: 
           name: Run integration tests for Event Sourced Entity Archetype
           command: |
             cd maven-java
             mvn -B archetype:generate  -DgroupId=com.example -DartifactId=counter-event-sourced-entity -DarchetypeGroupId=com.akkaserverless -DarchetypeArtifactId=akkaserverless-maven-archetype-event-sourced-entity -DarchetypeVersion=$SDK_VERSION
             cd counter-event-sourced-entity
-            mvn -B -Dakkaserverless-sdk.version=$SDK_VERSION compile
+            mvn -B compile
       - save_deps_cache
 
   samples-tests:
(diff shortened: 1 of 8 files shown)


.circleci/config.yml
```diff
@@ -176,14 +176,14 @@ jobs:
             cd maven-java
             mvn -B archetype:generate -DgroupId=com.example -DartifactId=counter-value-entity -DarchetypeGroupId=com.akkaserverless -DarchetypeArtifactId=akkaserverless-maven-archetype -DarchetypeVersion=$SDK_VERSION
             cd counter-value-entity
-            mvn -B -Dakkaserverless-sdk.version=$SDK_VERSION compile
```
octonato, line 179: we don't need this anymore since #154, the archetype comes with the right version for the sdk

.circleci/config.yml
```diff
@@ -200,25 +200,25 @@ jobs:
           command: |
             cd samples/java-valueentity-shopping-cart
             echo "Running mvn with SDK version: '$SDK_VERSION'"
-            mvn -Dakkaserverless-sdk.version=$SDK_VERSION -Dakkaserverless-maven-plugin.version=$SDK_VERSION verify -Pit
+            mvn versions:set -DnewVersion=$SDK_VERSION verify -Pit
```
octonato, line 203: because the samples now use ${project.version} as the sdk version, it suffice to call `mvn versions:set`.

samples/java-customer-registry/pom.xml
```diff
@@ -17,8 +17,7 @@
     <akkasls.dockerTag>${project.version}</akkasls.dockerTag>
     <akkasls.mainClass>customer.Main</akkasls.mainClass>
     <jdk.target>11</jdk.target>
-    <akkaserverless-sdk.version>0.7.0-beta.10</akkaserverless-sdk.version>
-    <akkaserverless-maven-plugin.version>0.7.0-beta.10</akkaserverless-maven-plugin.version>
```
octonato, line 21: sdk.version is enough. All artifacts must align to it.

updatePomVersions.sh
```diff
@@ -0,0 +1,33 @@
+# this script updates all maven projects versions (maven-java and samples) to align with the current sdk version
+# if SDK_VERSION env var is defined, it will use it, otherwise it will take the version from sbt
+# after running this script, you may run local tests or simply send a PR with the updates. 
+# the script can also run publishM2 and mvn install to generate the artifacts with the new version.
```
octonato, line 4: With this script we can test locally, but also update the pom after a release. We don't need to do it on each release though.

octonato (commented):
left a few comments to explain the changes

johanandren (approved):
Neat

samples/java-customer-registry/pom.xml
johanandren: I had that on my mental note list but forgot to talk to anyone about it :)

.circleci/config.yml
raboof: OK, so this works because we have `<akkaserverless-sdk.version>${project.version}</akkaserverless-sdk.version>` - which is perhaps OK for us, but would definitely be different for users.

I'm OK with it, but TBQH I don't see the advantage of using `versions:set` over `-D` for the samples. We're not publishing those, are we?

octonato: No, we are not publishing. But we do link them from the samples in the console. And users may checkout such a sample and run locally. In that case, we want the version to more or less aligned. 

Also, if we use `mvn versions:set` we can from the script update the samples to run them locally for our own tests. Otherwise, we need to pass -D to maven, but then, if you want to open such a sample in intellij, you want it to have a fixed sdk version, not something to change with -D. 

> but would definitely be different for users.

That won't be different for users since this is just a sample. When a user starts a new project from the archetype, they get the sdk version fixed and independent from their project version.

raboof: > we want the version to more or less aligned.

We want the SDK version in the sample to be the latest, agreed!

> if you want to open such a sample in intellij, you want it to have a fixed sdk version, not something to change with -D

I can see how that's useful. Using -D means the `pom.xml` don't have to change all the time (marking the git repo dirty), but I guess there's no easy way to have both advantages ;)

> That won't be different for users since this is just a sample.

I meant, it's a bit confusing that the sample is different from what the user actually gets when they start a new project from the archetype. Not too big of a deal though.

octonato: > it's a bit confusing that the sample is different from what the user actually gets when they start a new project from the archetype

Yeah, it's true.
We can try to improve it. We can probably do some `sed` on the pom and fix the version from that same script.

octonato:
@raboof, I will merge this one as such and improve it later one using `sed`. I would like to have this in for now because it's useful for the current dev cycle we are in.

PR #157, opened 2021-07-16
fix: merge main into sdk dev branch

octonato:


docs/src/modules/java/pages/value-entity.adoc
```diff
@@ -64,8 +64,7 @@ As we rely on the code-generation, the class `Counter` gets generated for us, it
 ----
 include::java:example$valueentity-counter/src/main/java/com/example/domain/Counter.java[tag=class]
 ----
-<1> The `@ValueEntity` annotation sets the `entityType` as a unique identifier of the "state storage". The class name may be changed even after data has been created, the `entityType` can't.
-<2> The constructor accepts a parameter marked as the `@EntityId` and we keep the value in a field.
+<1> The constructor accepts a parameter marked as the `@EntityId` and we keep the value in a field.
```
octonato, line 67: In this file I did some adaptations on the docs. Not yet what we need though.

What probably happened here is that we brought in the docs after we spin-off the sdk feature branch. Now, they are together and changes in the include files will immediately impact the docs.

raboof (approved):

PR #158, opened 2021-07-19
fix: preserve execution dir

octonato:
There is a small mistake on the updatePomVersions.sh script. We should not permanently move in `maven-java` directory.

### DIFF
--- a/updatePomVersions.sh
+++ b/updatePomVersions.sh
@@ -12,16 +12,18 @@ if [ $1 ]; then
   sbt publishM2
 fi 
 
-cd maven-java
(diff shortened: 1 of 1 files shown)


octonato:
marking as draft because there is another error in this script

octonato:
> marking as draft because there is another error in this script

fixed in 3427c40b

octonato:
@raboof, once we get this merge, I will integrate it in #157. So please wait with #157 otherwise we will need another `main -> sdk branch` merge.

raboof (approved):

PR #160, opened 2021-07-19
fix: update sample sdk version with sed

octonato:
* set the versions in all pom.xml to be 0.7.0-beta.15 (latest release)
* `updatePomVerions.sh` use `sed` to update `akkaserverless-sdk.version` property

### DIFF
--- a/updatePomVersions.sh
+++ b/updatePomVersions.sh
@@ -27,9 +27,6 @@ fi
 
 for i in samples/*
 do
-  (
-    cd $i
-    mvn versions:set -DnewVersion=$SDK_VERSION
-    rm pom.xml.versionsBackup
-  )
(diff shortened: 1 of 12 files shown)


updatePomVersions.sh
```diff
@@ -27,9 +27,6 @@ fi
...
 do
-  (
-    cd $i
-    mvn versions:set -DnewVersion=$SDK_VERSION
-    rm pom.xml.versionsBackup
-  )
-done
\ No newline at end of file
+  sed  -i .versionsBackup "s/<akkaserverless-sdk.version>\(.*\)<\/akkaserverless-sdk.version>/<akkaserverless-sdk.version>$SDK_VERSION<\/akkaserverless-sdk.version>/" $i/pom.xml
+  rm $i/pom.xml.versionsBackup
```
octonato, line 31: According to sed docs, one could omit the extension when using -i, but it didn't work for me. 

Also, the recommended usage is to create a backup file. We don't really need it, so I delete it as soon as we know that the file was written.

raboof (approved):
You know I like it ;) - thanks!

octonato:
> You know I like it ;) - thanks!

Of course I know it. I did it for you. 😄

Issue #48, opened 2021-06-17
Remove ReplicatedMap from API


patriknw:
Use the more specialized map implementations that are built on top of ReplicatedMap
* ReplicatedCounterMap
* ReplicatedRegisterMap

We should probably add ReplicatedMultiMap to cover ReplicatedMap of ReplicatedSet values.

pvlugter:
So this has ended up with these specialised replicated maps (counter map and register map) being added end-to-end — added to the protocol and using the specialised maps in Akka — so that the proxy sends messages specific to these CRDTs and SDKs can create these directly when an entity is initialised, either initially or after passivation. Otherwise the user needs to wrap an automatically initialised generic map with the specialised map (and for the JS SDK, the replicated entity state can't be set to the specialised map once initialised).

The generic replicated map could be removed now, but it could be useful to keep it as the only way to combine CRDTs currently, while recommending the specialised maps for most use cases.

pvlugter:
Changes merged. I think we can close this now, with ReplicatedMap still available but prefer the specialised maps.

PR #161, opened 2021-07-20
feat: add specialised replicated maps

### LINKED ISSUE #49: Add ReplicatedMultiMap
Add ReplicatedMultiMap to cover ReplicatedMap of ReplicatedSet values.

Note that in Akka (proxy impl) there is a specialized MultiMap, which isn't only ORMap of ORSet.

pvlugter:
Refs #48. Test the specialised replicated maps using the new TCK tests from https://github.com/lightbend/akkaserverless-framework/pull/802 (needs to be published first, draft PR here). Support the new replicated maps from https://github.com/lightbend/akkaserverless-framework/pull/803. Resolves #49.

### DIFF
--- a/tck/src/main/java/com/akkaserverless/javasdk/tck/model/replicatedentity/TckModelReplicatedEntity.java
+++ b/tck/src/main/java/com/akkaserverless/javasdk/tck/model/replicatedentity/TckModelReplicatedEntity.java
@@ -65,6 +65,8 @@ public class TckModelReplicatedEntity {
         return context.state(ReplicatedCounterMap.class);
       case "ReplicatedRegisterMap":
         return context.state(ReplicatedRegisterMap.class);
+      case "ReplicatedMultiMap":
+        return context.state(ReplicatedMultiMap.class);
       case "Vote":
         return context.state(Vote.class);
       default:
@@ -87,6 +89,8 @@ public class TckModelReplicatedEntity {
         return factory.<String>newReplicatedCounterMap();
       case "ReplicatedRegisterMap":
         return factory.<String, String>newReplicatedRegisterMap();
+      case "ReplicatedMultiMap":
+        return factory.<String, String>newReplicatedMultiMap();
       case "Vote":
         return factory.newVote();
       default:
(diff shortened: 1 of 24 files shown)


tck/src/main/java/com/akkaserverless/javasdk/tck/model/replicatedentity/TckModelReplicatedEntity.java
```diff
@@ -199,6 +201,51 @@ private void applyUpdate(ReplicatedData replicatedData, Update update) {
...
+            break;
+          case CLEAR:
+            counterMap.clear();
+            break;
+        }
+        break;
+      case REPLICATED_REGISTER_MAP:
+        @SuppressWarnings("unchecked")
+        ReplicatedRegisterMap<String, String> registerMap =
+            (ReplicatedRegisterMap<String, String>) replicatedData;
```
octonato, line 229: I may be missing something, but why both `ReplicatedCounterMap` and `ReplicatedRegisterMap` have type parameters? 

The proto definitions define the `(K, V)` to be `(String, Long)` and `(String, String)` respectively.

True that I still need to wrap my mind around the whole replicate entity feature and get a better view on how that will be used, but at first sight my understanding is that we are already bound to `(String, Long)` and `(String, String)`

pvlugter: The key can be anything serializable — protobuf message, json, or primitive. Same for register values. The proto definitions here are just for the TCK tests, which are simplified to only string keys and string values for registers.

octonato: I understood this was the TCK, but the map entry in the proto is defined as 

```proto
message ReplicatedRegisterMapEntryUpdate {
  string key = 1;
  string value = 2;
}
```

How could we send anything other then String? Are we sending the bytes encoded as base64?

pvlugter: That's the protocol for the TCK — which sends messages on how the entities should be updated, and only uses strings.

https://github.com/lightbend/akkaserverless-framework/blob/master/protocols/tck/src/main/protobuf/akkaserverless/tck/model/replicatedentity/replicated_entity.proto

The actual proxy protocol for replicated maps uses protobuf Any:

```proto
message ReplicatedMapEntryDelta {
  // The entry key.
  google.protobuf.Any key = 1;
  ReplicatedEntityDelta delta = 2;
}

message ReplicatedRegisterDelta {
  google.protobuf.Any value = 1;
  ReplicatedEntityClock clock = 2;
  int64 custom_clock_value = 3;
}
```

https://github.com/lightbend/akkaserverless-framework/blob/613d873529f8e0785218b0357c281a30a99c9d31/protocols/proxy/src/main/protobuf/akkaserverless/component/replicatedentity/replicated_entity.proto#L132-L136

https://github.com/lightbend/akkaserverless-framework/blob/613d873529f8e0785218b0357c281a30a99c9d31/protocols/proxy/src/main/protobuf/akkaserverless/component/replicatedentity/replicated_entity.proto#L119-L123

octonato: Oh, that explains everything. Thanks for clarifying.

patriknw (approved):
LGTM

pvlugter:
Rebased and updated to framework release 0.7.0-beta.16.

PR #163, opened 2021-07-20
Non-reflection eventsourced entity handler

raboof:
Based on #150 but excluding the move of the deserialization from the
generated code for now: we might still want to do that, but I'm not
entirely confident about it yet - let's see how JSON support and
testkit re-use pan out.


sdk/src/main/scala/com/akkaserverless/javasdk/impl/eventsourcedentity/AbstractEventSourcedEntityHandler.scala
```diff
@@ -0,0 +1,113 @@
...
+}
+
+/**
+ * @tparam S the type of the managed state for the entity
+ * Not for manual user extension or interaction
+ */
+abstract class AbstractEventSourcedEntityHandler[S, E <: EventSourcedEntityBase[S]](protected val entity: E) {
+
+  // ugh, remove
+  var snapshotEvery = 0
```
octonato, line 47: ugh, remove? why? 

If it stays, should be `private var`, no?

sdk/src/main/scala/com/akkaserverless/javasdk/impl/eventsourcedentity/AbstractEventSourcedEntityHandler.scala
```diff
@@ -0,0 +1,113 @@
...
+    case None =>
+      val emptyState = entity.emptyState()
+      require(emptyState != null, "Entity empty state is not allowed to be null")
+      state = Some(emptyState)
+      emptyState
+    case Some(state) => state
+  }
+
+  final protected def setState(newState: S): Unit =
+    state = Some(newState)
```
octonato, line 61: Since `setState` is called with the return of an event handler, we can expect users to return `null`. 


```suggestion
  final protected def setState(newState: S): Unit =
    state = Option(newState)
```

sdk/src/main/scala/com/akkaserverless/javasdk/impl/eventsourcedentity/AbstractEventSourcedEntityHandler.scala
```diff
@@ -0,0 +1,113 @@
...
+  // ugh, remove
+  var snapshotEvery = 0
+
+  private var state: Option[S] = None
+
+  final protected def stateOrEmpty(): S = state match {
+    case None =>
+      val emptyState = entity.emptyState()
+      require(emptyState != null, "Entity empty state is not allowed to be null")
+      state = Some(emptyState)
```
octonato, line 55: I think it should be allowed to be `null`. I would expect java developers to define their `emptyState` with `null`.

sdk/src/main/scala/com/akkaserverless/javasdk/impl/eventsourcedentity/AbstractEventSourcedEntityHandler.scala
```diff
@@ -0,0 +1,113 @@
...
+            entity.setEventContext(Optional.of(eventContextFactory(currentSequence)))
+            val newState = handleEvent(stateOrEmpty(), event)
+            setState(newState)
+          } finally {
+            entity.setEventContext(Optional.empty())
+          }
+          currentSequence += 1
+          shouldSnapshot = shouldSnapshot || (snapshotEvery > 0 && currentSequence % snapshotEvery == 0)
+        }
+        // FIXME currently snapshotting final state after applying all events even if trigger was mid-event stream?
```
octonato, line 99: I think this is ok since snapshotting is just an optimization. We don't need to be precise about what was exactly the event that reached the threshold

sdk/src/main/scala/com/akkaserverless/javasdk/impl/eventsourcedentity/AbstractEventSourcedEntityHandler.scala
```diff
@@ -0,0 +1,113 @@
...
+        val endState = stateOrEmpty()
+        val snapshot =
+          if (shouldSnapshot) Option(endState)
```
octonato, line 102: If we accept that the state can be null, then we don't need to call `stateOrEmpty` and we can just return `state` since it's already an Option.

sdk/src/main/scala/com/akkaserverless/javasdk/impl/eventsourcedentity/AbstractEventSourcedEntityHandler.scala
octonato: I see that this comes from previous code and only moved to here. Still, I do think that the user can always define its state as `null` and at any point of its lifecycle.

octonato (commented):
@raboof, I did a first pass. I will continue tomorrow.

sdk/src/main/scala/com/akkaserverless/javasdk/impl/eventsourcedentity/AbstractEventSourcedEntityHandler.scala
raboof: I think I agree - but let's keep it for a different PR given this is the same as before?

octonato: fine, I will create an issue about this

octonato (approved):
@raboof, there are still some small things to address, like removing the FIXME and making the snapshotEvery private var. 

Feel free to merge and address it in follow-up PRs if you prefer.

PR #164, opened 2021-07-21
entity factory experiment

octonato:
This is just an experiment to see how we could offer a minimal API to register a value entity

### DIFF
--- a/samples/java-customer-registry/src/main/java/customer/CustomerValueEntity.java
+++ b/samples/java-customer-registry/src/main/java/customer/CustomerValueEntity.java
@@ -16,6 +16,7 @@
 
 package customer;
 
+import com.akkaserverless.javasdk.valueentity.ValueEntityContext;
 import com.google.protobuf.Empty;
 import customer.api.CustomerApi;
 import customer.domain.CustomerDomain;
@@ -27,6 +28,12 @@ import customer.domain.CustomerDomain;
  * interface and the users' build tooling should indicate what needs changing.
  */
 public class CustomerValueEntity extends CustomerValueEntityInterface {
+
+  private final String entityId;
+
+  public CustomerValueEntity(ValueEntityContext context) {
+    this.entityId = context.entityId();
+  }
 
   @Override
   public Effect<CustomerApi.Customer> getCustomer(
--- a/samples/java-customer-registry/src/main/java/customer/CustomerValueEntityFactory.java
+++ b/samples/java-customer-registry/src/main/java/customer/CustomerValueEntityFactory.java
@@ -1,0 +1,35 @@
+/*
+ * Copyright 2021 Lightbend Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package customer;
+
+import com.akkaserverless.javasdk.valueentity.ValueEntityContext;
+import com.akkaserverless.javasdk.valueentity.ValueEntityOptions;
+
+// editable, to be generated in src/main/java
+public class CustomerValueEntityFactory extends CustomerValueEntityFactoryBase {
+
+  @Override
+  public CustomerValueEntity newInstance(ValueEntityContext context) {
+    return new CustomerValueEntity(context);
+  }
+
+  @Override
+  public ValueEntityOptions options() {
+    return ValueEntityOptions.defaults();
+  }
+
+}
--- a/samples/java-customer-registry/src/main/java/customer/CustomerValueEntityFactoryBase.java
+++ b/samples/java-customer-registry/src/main/java/customer/CustomerValueEntityFactoryBase.java
@@ -1,0 +1,34 @@
+package customer;
+
+import com.akkaserverless.javasdk.valueentity.ValueEntityContext;
+import com.akkaserverless.javasdk.valueentity.ValueEntityFactory;
+import com.google.protobuf.Descriptors;
+import customer.api.CustomerApi;
+
+// to be generated in generated-source/src/main/java
+public abstract class CustomerValueEntityFactoryBase implements ValueEntityFactory<CustomerValueEntity> {
+
+    // depends on proto file definition therefore final
+    @Override
+    public final Descriptors.ServiceDescriptor serviceDescriptor() {
+        return CustomerApi.getDescriptor().findServiceByName("CustomerService");
+    }
+
+    // depends on proto file definition therefore final
+    @Override
+    public final String entityTypeHint() {
+        return "customers";
+    }
+
+    // depends on proto file definition therefore final
+    @Override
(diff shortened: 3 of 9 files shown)


samples/java-customer-registry/src/main/java/customer/CustomerValueEntity.java
```diff
@@ -28,6 +29,12 @@
  */
 public class CustomerValueEntity extends CustomerValueEntityInterface {
 
+  private final String entityId;
+
+  public CustomerValueEntity(ValueEntityContext context) {
+    this.entityId = context.entityId();
+  }
```
octonato, line 36: users should be able to define the constructor the way they prefer. By default, we generated one getting the `ValueEntityContext` and setting the `entityId`.

samples/java-customer-registry/src/main/java/customer/CustomerValueEntityFactory.java
```diff
@@ -0,0 +1,60 @@
...
+
+package customer;
+
+import com.akkaserverless.javasdk.valueentity.ValueEntityContext;
+import com.akkaserverless.javasdk.valueentity.ValueEntityFactory;
+import com.akkaserverless.javasdk.valueentity.ValueEntityOptions;
+import com.google.protobuf.Descriptors;
+import customer.api.CustomerApi;
+
+public class CustomerValueEntityFactory implements ValueEntityFactory<CustomerValueEntity> {
```
octonato, line 25: This factory brings in all the pieces needed to register and instantiate the entity.

This is generated code, but lives in `src/main/java`. Users can add a constructor and pass in whatever they prefer.

samples/java-customer-registry/src/main/java/customer/CustomerValueEntityFactory.java
```diff
@@ -0,0 +1,60 @@
...
+
+  @Override
+  public CustomerValueEntity newInstance(ValueEntityContext context) {
+    return new CustomerValueEntity(context);
+  }
```
octonato, line 30: Arguments passed to the constructor passed further to the entity. Users can edit both `CustomerValueEntityFactory` and `CustomerValueEntity`

samples/java-customer-registry/src/main/java/customer/CustomerValueEntityFactory.java
```diff
@@ -0,0 +1,60 @@
...
+
+  @Override
+  public ValueEntityOptions options() {
+    return ValueEntityOptions.defaults();
+  }
```
octonato, line 35: Eventually configure the options the way they prefer.

samples/java-customer-registry/src/main/java/customer/CustomerValueEntityFactory.java
```diff
@@ -0,0 +1,60 @@
...
+  @Override
+  public final CustomerValueEntityHandler newHandler(ValueEntityContext context) {
+    return new CustomerValueEntityHandler(newInstance(context));
+  }
+
+  // depends on proto file definition therefore final
+  @Override
+  public final Descriptors.FileDescriptor[] additionalDescriptors() {
+    return new Descriptors.FileDescriptor[0];
+  }
```
octonato, line 59: This is marked final just as an indication that users should preferably not modify these method. However, since it will live in `src/main/java`, they will be able to modify them.

Or maybe we should have another intermediate class that lives under the `generated-sources` dir and defines these method. WDYT?

samples/java-customer-registry/src/main/java/customer/Main.java
```diff
@@ -50,12 +39,7 @@ public static void main(String[] args) throws Exception {
               "customerByName",
               CustomerDomain.getDescriptor())
           // end::register[]
-          .lowLevel()
-          .registerValueEntity(
-              context -> new CustomerValueEntityHandler(new CustomerValueEntity()),
-              CustomerValueEntityHandler.serviceDescriptor,
-              CustomerValueEntityHandler.entityType,
-              ValueEntityOptions.defaults())
+          .registerValueEntity(new CustomerValueEntityFactory())
```
octonato, line 42: That's the main reason for this experiment. 

Because users should be able to inject whatever they want in the entity, they will end up having to touch the registration. With the previous API, they will have to be exposed to all the bits that are not so relevant and will have copy and paste them from the generate code to their main function. 

With the new factory, the registration only requires passing a factory and the user can inject whatever they want in the factory.

sdk/src/main/java/com/akkaserverless/javasdk/lowlevel/ValueEntityHandlerFactory.java
```diff
@@ -25,7 +25,7 @@
  * com.akkaserverless.javasdk.valueentity.ValueEntity @ValueEntity} and similar annotations should
  * be used.
  */
-public interface ValueEntityFactory {
+public interface ValueEntityHandlerFactory {
```
octonato, line 28: This is rather an ValueEntityHandlerFactory. Moreover, I needed to claim that old name. :-)

octonato (commented):
left some comments to explain the changes

samples/java-customer-registry/src/main/java/customer/CustomerValueEntityFactoryBase.java
```diff
@@ -0,0 +1,34 @@
...
+    @Override
+    public final CustomerValueEntityHandler newHandler(ValueEntityContext context) {
+        return new CustomerValueEntityHandler(newInstance(context));
+    }
+
+    // depends on proto file definition therefore final
+    @Override
+    public final Descriptors.FileDescriptor[] additionalDescriptors() {
+        return new Descriptors.FileDescriptor[0];
+    }
```
octonato, line 33: these methods depend on the proto file and should not be modified. 

Maybe `addtionalDescriptors` should be editable and therefore move to `CustomerValueEntityFactory`, but I don't think so. To be decided...

sdk/src/main/java/com/akkaserverless/javasdk/AkkaServerless.java
```diff
@@ -454,6 +451,16 @@ public AkkaServerless registerValueEntity(
         entityClass, descriptor, ValueEntityOptions.defaults(), additionalDescriptors);
   }
 
+  public <T> AkkaServerless registerValueEntity(ValueEntityFactory<T> factory) {
+    return lowLevel()
```
octonato, line 455: I think we should step out of the idea of having a low-level API. 

Users should not be exposed to such a method and ultimately the default API to use will be the one that does not depend on reflection.

raboof: agreed, what is the 'lowlevel' API right now would become the 'codegen-based' API and replace the annotation-based API.

raboof:
(this could use a rebase to make the diff smaller, since it was based on #148 which has since been squash-and-merged)

octonato:
> (this could use a rebase to make the diff smaller, since it was based on #148 which has since been squash-and-merged)

done

samples/java-customer-registry/src/main/java/customer/CustomerValueEntityFactory.java
patriknw: Shouldn't this be in the generated base class, and they can override it here if they need different options?

sdk/src/main/java/com/akkaserverless/javasdk/AkkaServerless.java
patriknw: > I think we should step out of the idea of having a low-level API.

I introduced it, but the intention was never to keep it. It was just too early to remove it and I moved it to the side.

patriknw (commented):
I like this idea

patriknw:
https://github.com/lightbend/akkaserverless-java-sdk/pull/180 is better

PR #165, opened 2021-07-21
feat: add some logging to codegen

raboof:
Notably for the case where there is an entity service,
but no entity component configuration that matches by
the component name - in which case nothing would be
generated, silently.


pvlugter (approved):
👍🏼 good to have this warning. Should this be added to main branch too?

octonato (approved):
Great!

PR #166, opened 2021-07-22
feat: add some logging to codegen

raboof:
Notably for the case where there is an entity service, but no entity
component configuration that matches by the component name - in which case
nothing would be generated, silently.

backport of #165


octonato (approved):
👍

PR #171, opened 2021-07-22
Build valueentity-counter sample with sbt as well

raboof:
So you can run it in sbt without having to publish and switch to maven,
as long as codegen isn't changed.

Right now java-customer-registry is useful for this, but that won't work
anymore when we remove the non-codegen implementation.


octonato (commented):
Not clear what the goal is, but this is what I understand...

`valueentity-counter` and `valueentity-counter-generated` is nearly the same. 

We use `valueentity-counter-generated` to modify the APIs that are ultimately generated. 
We also modify the code that is supposed to be generated in `src/main/java`, like the `Counter` entity. 

When we are done, we copy `valueentity-counter-generated/src/main/java/Counter.java` to `valueentity-counter/src/main/java` (package ommited). Same for the tests.

When we run the mvn compile in `valueentiy-counter`, new code is generated and `Counter.java` should just compile. 

Eventually, we can copy back all the code from `valueentity-counter` back into `valueentity-counter-generated`

Is that the plan? 

Maybe we should call `valueentity-counter-hardcoded` (or -sanbox) instead.

raboof:
> Not clear what the goal is

The goal is to be able to compile a sample from sbt, so you can test some changes without having to go though the whole 'publishM2, invoke maven with just-published version' cycle.

> `valueentity-counter` and `valueentity-counter-generated` is nearly the same.

Not at all, actually: the `valueentity-counter` sbt project refers to both the `valueentity-counter` source directory and the `valueentity-counter-generated` source directory. So the `valueentity-counter` is shared between sbt and maven, and `valueentity-counter-generated` holds just the sources that would be generated by the maven plugin.

> We use `valueentity-counter-generated` to modify the APIs that are ultimately generated.
> We also modify the code that is supposed to be generated in `src/main/java`, like the `Counter` entity.

You could do that during development, but on the `main` (and `sdk-codegen-dev` branch) I think we should always keep the `valueentity-counter-generated` folder in sync with what is actually generated by the codegen on that commit.

pvlugter:
Having the faster loop with an sbt build sounds good. On the committed generated code, noticed that the `CustomerApi` and `CustomerDomain` are generated for both `main` and `test` — could only be generated for `main`? And the Akka gRPC generated code for test includes both the service and the client — should only need to generate the client for tests?

octonato:
@raboof, thanks for the clarification. I missed the fact that the sbt project was still the `valueentity-counter` with additional source folders. 

@pvlugter, if you are referring to the `CounterService.java`, that's because the client implements it. On the other hand, I don't think we need the `CounterServiceHandlerFactory.java`. This is only used on the server side.

octonato (approved):
LGTM, with two suggestions: 

* remove CounterDomain from `test/java/` since already `main/java` (as per Peter's comments)
* remove `CounterServiceHandlerFactory` since not used

raboof:
I completely agree, but I think we should keep `valueentity-counter-generated` consistent with what's actually generated, and fix it there :). Filed #182 and #181 for that.

Issue #173, opened 2021-07-28
archetype does not work


raboof:
On a clean system, the command documented at https://developer.lightbend.com/docs/akka-serverless/java/kickstart.html :

```
mvn archetype:generate -DarchetypeGroupId=com.akkaserverless -DarchetypeArtifactId=akkaserverless-maven-archetype -DarchetypeVersion=LATEST
```

... does not work. The symptoms can be confusing: the logging sometimes looks as if it can't find the LATEST version or it can't find the archetype jar. Looking closer the LATEST version is actually correctly resolved and the archetype found, but the archetype's parent pom was not correctly published.

pvlugter:
Works with the latest (beta.18) release now.

PR #174, opened 2021-07-28
Don't skip deploying the maven-java pom

raboof:
It is required since the other projects have this pom as their parent
pom. Refs #173

### DIFF
--- a/maven-java/pom.xml
+++ b/maven-java/pom.xml
@@ -35,17 +35,6 @@
     <maven.version>3.3.9</maven.version>
   </properties>
 
-  <repositories>
-    <repository>
-      <snapshots>
-        <enabled>false</enabled>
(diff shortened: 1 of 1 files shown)


hseeberger:
@raboof, I only see changes to `maven-java/pom.xml`. There you removed stuff. How does this lead to the desired effect of not skipping deployment of the mavan-java POM?

raboof:
> @raboof, I only see changes to `maven-java/pom.xml`. There you removed stuff. How does this lead to the desired effect of not skipping deployment of the mavan-java POM?

Specifically the removal of

```
        <configuration>
          <skip>true</skip>
        </configuration>
```

from the deploy plugin configuration of the parent pom

hseeberger (approved):
LGTM

PR #179, opened 2021-07-29
Full proto 113

### LINKED ISSUE #113: Need full proto example in Implementing Views
By @mckeeh3 
In Implementing Views for both JS and Java a full view proto file example is needed. The current examples do not include syntax, package, imports, and view specific option details.

Without these details, it is impossible to create views.

rstento:
Fixes #113. Submitting as a draft because I can't build and test this locally.


pvlugter (approved):
LGTM

PR #180, opened 2021-07-30
entity provider experiment (take 2)

octonato:
Similar to #164, but using a provider that receives a function to construct the entity

### DIFF
--- a/samples/java-customer-registry/src/main/java/customer/CustomerValueEntityProvider.java
+++ b/samples/java-customer-registry/src/main/java/customer/CustomerValueEntityProvider.java
@@ -1,0 +1,59 @@
+/*
+ * Copyright 2021 Lightbend Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package customer;
+
+import com.akkaserverless.javasdk.valueentity.ValueEntityContext;
+import com.akkaserverless.javasdk.valueentity.ValueEntityProvider;
+import com.google.protobuf.Descriptors;
+import customer.api.CustomerApi;
+
+import java.util.function.Function;
+
+// editable, to be generated in generated-sources/src/main/java
+public class CustomerValueEntityProvider implements ValueEntityProvider<CustomerValueEntity> {
+
+  private final Function<ValueEntityContext, CustomerValueEntity> entityProviderFunc;
+
+  public CustomerValueEntityProvider(
+      Function<ValueEntityContext, CustomerValueEntity> entityProviderFunc) {
+    this.entityProviderFunc = entityProviderFunc;
+  }
+
+  // depends on proto file definition therefore final
+  @Override
+  public final Descriptors.ServiceDescriptor serviceDescriptor() {
+    return CustomerApi.getDescriptor().findServiceByName("CustomerService");
+  }
+
+  // depends on proto file definition therefore final
+  @Override
+  public final String entityTypeHint() {
+    return "customers";
+  }
+
+  // depends on proto file definition therefore final
+  @Override
+  public final CustomerValueEntityHandler newHandler(ValueEntityContext context) {
+    return new CustomerValueEntityHandler(entityProviderFunc.apply(context));
+  }
+
+  // depends on proto file definition therefore final
+  @Override
+  public final Descriptors.FileDescriptor[] additionalDescriptors() {
+    return new Descriptors.FileDescriptor[0];
+  }
+}
--- a/sdk/src/main/java/com/akkaserverless/javasdk/AkkaServerless.java
+++ b/sdk/src/main/java/com/akkaserverless/javasdk/AkkaServerless.java
@@ -35,14 +35,11 @@ import com.akkaserverless.javasdk.impl.valueentity.AnnotationBasedEntitySupport;
 import com.akkaserverless.javasdk.impl.valueentity.ValueEntityService;
 import com.akkaserverless.javasdk.impl.view.AnnotationBasedViewSupport;
 import com.akkaserverless.javasdk.impl.view.ViewService;
-import com.akkaserverless.javasdk.lowlevel.ActionHandler;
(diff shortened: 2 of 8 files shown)


sdk/src/main/java/com/akkaserverless/javasdk/AkkaServerless.java
```diff
@@ -454,6 +451,20 @@ public AkkaServerless registerValueEntity(
...
+
+  public <T> AkkaServerless register(ValueEntityProvider<T> factory, ValueEntityOptions options) {
+    return lowLevel()
+        .registerValueEntity(
+            context -> factory.newHandler(context),
+            factory.serviceDescriptor(),
+            factory.entityTypeHint(),
+            options,
+            factory.additionalDescriptors());
+  }
```
octonato, line 466: should be `provider`, not `factory`.

samples/java-customer-registry/src/main/java/customer/CustomerValueEntityProvider.java
```diff
@@ -0,0 +1,59 @@
...
+
+import com.akkaserverless.javasdk.valueentity.ValueEntityContext;
+import com.akkaserverless.javasdk.valueentity.ValueEntityProvider;
+import com.google.protobuf.Descriptors;
+import customer.api.CustomerApi;
+
+import java.util.function.Function;
+
+// editable, to be generated in generated-sources/src/main/java
+public class CustomerValueEntityProvider implements ValueEntityProvider<CustomerValueEntity> {
```
octonato, line 27: I'm not so sure how we should call this. 

It's not really a ValueEntityProvider. Instead, it provides the handler and some other required configurations (descriptors and type hint).

samples/java-customer-registry/src/main/java/customer/CustomerValueEntityProvider.java
```diff
@@ -0,0 +1,59 @@
...
+package customer;
+
+import com.akkaserverless.javasdk.valueentity.ValueEntityContext;
+import com.akkaserverless.javasdk.valueentity.ValueEntityProvider;
+import com.google.protobuf.Descriptors;
+import customer.api.CustomerApi;
+
+import java.util.function.Function;
+
+// editable, to be generated in generated-sources/src/main/java
```
raboof, line 26: this no longer needs to be editable, right?

samples/java-customer-registry/src/main/java/customer/CustomerValueEntityProvider.java
raboof: I think ValueEntityProvider is fine: from the users' perspective this is the thing that produces value entity instances, and that this is a 'multiple-step process' with a handler in between is an implementation detail.

sdk/src/main/java/com/akkaserverless/javasdk/valueentity/ValueEntityProvider.java
```diff
@@ -0,0 +1,31 @@
...
+package com.akkaserverless.javasdk.valueentity;
+
+import com.akkaserverless.javasdk.lowlevel.ValueEntityHandler;
+import com.google.protobuf.Descriptors;
+
+public interface ValueEntityProvider<T> {
+
+  Descriptors.ServiceDescriptor serviceDescriptor();
+
+  String entityTypeHint();
```
raboof, line 26: I think we use just `entityType` in Akka Serverless, without `Hint`

samples/java-customer-registry/src/main/java/customer/CustomerValueEntityProvider.java
octonato: Correct, those files will go away.

sdk/src/main/java/com/akkaserverless/javasdk/valueentity/ValueEntityProvider.java
octonato: Yes, and we should stop doing so. This is the type hint and we should name it so. The reason it wasn't done before is that it was a field in the annotation and we would break it. 

With the new providers and codegen we can rename it without breaking anything.

raboof: > This is the type hint and we should name it so

Can you explain why you think this is better? I don't understand what extra information the 'Hint' suffix gives.

raboof: OK, I found https://github.com/lightbend/akkaserverless-framework/pull/670#issuecomment-848234377 - so the reasoning is that this string does not refer to a Java type, but is a string associated with the entity.

raboof: > The reason it wasn't done before is that it was a field in the annotation and we would break it.
>
> With the new providers and codegen we can rename it without breaking anything.

We're still not changing the proto annotation name, though, right? I'm not super convinced adding 'Hint' is helpful here.

octonato: I don't mind to change that on another iteration (when we change the annotation) . 

I do find it's helpful for us to align how we name things. The function of this bit of string is exactly the same as the type hint in Akka so we should name it the same for our own sanity.

samples/java-customer-registry/src/main/java/customer/CustomerValueEntityProvider.java
```diff
@@ -0,0 +1,59 @@
...
+public class CustomerValueEntityProvider implements ValueEntityProvider<CustomerValueEntity> {
+
+  private final Function<ValueEntityContext, CustomerValueEntity> entityProviderFunc;
```
patriknw, line 29: s/entityProviderFunc/entityFactory/

sdk/src/main/java/com/akkaserverless/javasdk/AkkaServerless.java
```diff
@@ -454,6 +451,20 @@ public AkkaServerless registerValueEntity(
         entityClass, descriptor, ValueEntityOptions.defaults(), additionalDescriptors);
   }
 
+  public <T> AkkaServerless register(ValueEntityProvider<T> factory) {
+    return register(factory, ValueEntityOptions.defaults());
+  }
+
+  public <T> AkkaServerless register(ValueEntityProvider<T> factory, ValueEntityOptions options) {
```
patriknw, line 458: An alternative would be that we make ValueEntityProvider a builder, so that options can be specified via the ValueEntityProvider. 

```
ValueEntityProvider.of(context -> new CustomerValueEntity()).withOptions(...)
```

The advantage is that it's easier for us to evolve the builder with more optional properties, than having overloaded register methods.

patriknw (commented):
Good. More lightweight than #164

patriknw:
Continued, for real, in https://github.com/lightbend/akkaserverless-java-sdk/pull/190

PR #183, opened 2021-08-02
towards simplified codegen

octonato:
In this PR: 

* completes the cycle for Value Entity - all generated based on simple interpolation
* all other generators return `String` instead of `Document`


raboof (approved):
The fact that the tests didn't need much changes is a good sign :)

Issue #184, opened 2021-08-02
Add gRPC compound entity keys example


rstento:
@mckeeh3 commented on [Mon Jun 07 2021](https://github.com/lightbend/akkaserverless-docs/issues/637)

In the [Messages](https://docs.akkaserverless.dev/java-services/proto.html#_messages) section of the docs. Please provide a proto file example of a compound multi-field entity key. It is not clear how this is done in proto file syntax.

---

@rstento commented on [Mon Jun 07 2021](https://github.com/lightbend/akkaserverless-docs/issues/637#issuecomment-856284053)

@ennru, could a developer help with this?

---

@ennru commented on [Tue Jun 08 2021](https://github.com/lightbend/akkaserverless-docs/issues/637#issuecomment-856671643)

Thanks for the input. This is not crucial right now.
I believe it is "simply"
```proto
message AddLineItem {
    string user_id = 1 [(akkaserverless.field).entity_key = true];
    string product_id = 2 [(akkaserverless.field).entity_key = true];
```

---

@rstento commented on [Tue Jul 27 2021](https://github.com/lightbend/akkaserverless-docs/issues/637#issuecomment-887719400)

When we are ready to do this, this issue needs to be moved/copied to the Java and the JavaScript SDK repos.

PR #185, opened 2021-08-02
doc: add example of multi part entity key

rstento:


pvlugter (changes requested):
It's good to have an example of a compound entity key. I don't think we should do this with `AddLineItem` in the shopping cart example, as that doesn't actually make sense and could be confusing. All the messages need to have the same entity key to be routed to the same entity instance — while for shopping cart, other messages like `GetShoppingCart` only have a `user_id`, so messages will go to different entity instances. So having the compound entity key for one of the messages just seems confusing.

Since this is already an inline snippet (not tested as part of an actual sample) maybe it should just be an artificial example instead.

rstento:
> It's good to have an example of a compound entity key. I don't think we should do this with `AddLineItem` in the shopping cart example, as that doesn't actually make sense and could be confusing. All the messages need to have the same entity key to be routed to the same entity instance — while for shopping cart, other messages like `GetShoppingCart` only have a `user_id`, so messages will go to different entity instances. So having the compound entity key for one of the messages just seems confusing.
> 
> Since this is already an inline snippet (not tested as part of an actual sample) maybe it should just be an artificial example instead.

How's this @pvlugter?

rstento:
Fixes #184. Updated from review comments, please re-review.

docs/src/modules/java/pages/proto.adoc
```diff
@@ -43,6 +43,19 @@ You define messages that can be sent to or returned from the service. Each messa
 include::example$java-valueentity-shopping-cart/src/main/proto/shoppingcart/shoppingcart_api.proto[tag=proto_messages]
 ----
 
+
+To use a multi-field key, add the `entity_key` notation to each field. For
+example, if the domain proto file `SomeMessage` element defines both
+`first_field` and `second_field` as part of the key, do the same for the service
+message:
```
pvlugter, line 50: The part about defining entity keys for both domain and service messages doesn't make sense here. Entity keys are only defined on incoming messages for the service, so that messages can be routed to the correct entity.

```suggestion
To use a multi-field key, add the `entity_key` notation to each field. For
example, if the `SomeMessage` service message defines both
`first_field` and `second_field` as part of the key:
```

docs/src/modules/java/pages/proto.adoc
```diff
@@ -43,6 +43,19 @@ You define messages that can be sent to or returned from the service. Each messa
...
+
+[source,proto]
+----
+message SomeMessage {
+    string first_field = 1 [(akkaserverless.field).entity_key = true];
+    string second_field = 2 [(akkaserverless.field).entity_key = true];
```
pvlugter, line 56: 2-space indentation for protobuf:

```suggestion
message SomeMessage {
  string first_field = 1 [(akkaserverless.field).entity_key = true];
  string second_field = 2 [(akkaserverless.field).entity_key = true];
```

docs/src/modules/java/pages/proto.adoc
```diff
@@ -96,6 +109,19 @@ include::example$java-valueentity-shopping-cart/src/main/proto/shoppingcart/shop
...
+----
+
+
```
pvlugter, line 124: Entity keys are not defined on domain messages. This part can be removed.

```suggestion
```

docs/src/modules/java/pages/proto.adoc
rstento: Done.

docs/src/modules/java/pages/proto.adoc
rstento: Done, thanks!

rstento:
Hopefully good now, @pvlugter.

docs/src/modules/java/pages/proto.adoc
```diff
@@ -43,6 +43,18 @@ You define messages that can be sent to or returned from the service. Each messa
 include::example$java-valueentity-shopping-cart/src/main/proto/shoppingcart/shoppingcart_api.proto[tag=proto_messages]
 ----
 
+
+To use a multi-field key, add the `entity_key` notation to each field. For
+example, the following `SomeMessage` element defines both
+`first_field` and `second_field` as part of the key:
```
pvlugter, line 49:
```suggestion
To use a multi-field key, add the `entity_key` notation to each field. For example, the following `SomeMessage` element defines both `first_field` and `second_field` as part of the key:
```

pvlugter (approved):

PR #188, opened 2021-08-03
build: compilation test for generated code

octonato:
Adds full cycle for codegen compilation: proto ->  descriptor ->  codegen -> java compiler


build.sbt
```diff
@@ -116,6 +122,35 @@ lazy val codegenJava =
...
+  .in(file("codegen/sbt-plugin"))
+  .enablePlugins(SbtPlugin)
+  .dependsOn(codegenJava)
+  .settings(
+    (publish / skip) := true,
+    name := "akkaserverless-codegen-sbt-plugin",
+    Compile / javacOptions ++= Seq("-encoding", "UTF-8", "-source", "11", "-target", "11")
+  )
+  .settings(crossScalaVersions := Dependencies.ScalaVersionForCodegen,
+            scalaVersion := Dependencies.ScalaVersionForCodegen.head)
```
octonato, line 138: Plugin needs to use Scala 2.12 and therefore `codegenJava` and `codegenCore` as well.

Note that for the maven plugin that will call the codegen it doesn't matter if it's 2.12 or 2.13

codegen/java-gen/src/main/scala/com/lightbend/akkasls/codegen/java/ActionServiceSourceGenerator.scala
```diff
@@ -56,16 +55,16 @@ object ActionServiceSourceGenerator {
     val interfaceSourcePath =
       generatedSourceDirectory.resolve(packagePath.resolve(interfaceClassName + ".java"))
 
-    val _ = interfaceSourcePath.getParent.toFile.mkdirs()
-    val _ = Files.write(
+    interfaceSourcePath.getParent.toFile.mkdirs()
+    Files.write(
```
octonato, line 59: I may be wrong, but I remember to have used `val _ = ` more then once in a Scala 2.12 build. 

It turns out that the second `_` fails because `_` is already defined. 🤷 

Anyway, since the we are not enforcing it during compilation, I thought it would just simpler to not capture the return.

codegen/java-gen/src/main/scala/com/lightbend/akkasls/codegen/java/ActionServiceSourceGenerator.scala
```diff
@@ -107,39 +106,41 @@ object ActionServiceSourceGenerator {
       "package" <+> packageName <> semi <> line <>
       line <>
       ssep(
-        imports.map(pkg => "import" <+> pkg <> semi),
+        imports.to[immutable.Seq].map(pkg => "import" <+> pkg <> semi),
```
octonato, line 109: kiama expected `immutable.Seq` for all methods, but when compiling with 2.12 we get the mutable one. 

Could have changed more code to always return immutable, but since we are stepping out of kiama, I don't think it's worth the effort.

project/ReflectiveCodeGen.scala
```diff
@@ -0,0 +1,71 @@
...
+    import scala.reflect.runtime.universe
+    import scala.tools.reflect.ToolBox
+
+    val tb = universe.runtimeMirror(loader).mkToolBox()
+
+    val source =
+      s"""
+      |import com.lightbend.akkasls.codegen.AkkaServerlessCodegenPlugin
+      |
+      |(protobufDescriptor: java.io.File, genSrcDir: java.io.File) => AkkaServerlessCodegenPlugin.runAkkaServerlessCodegen(protobufDescriptor, genSrcDir)
```
octonato, line 39: I'm now wondering if we can do the same but calling the generator directly and without an intermediate plugin.

octonato (commented):
@raboof, I left some comments here and there. 

For the moment, we can go with this simplified version. Eventually, we can improve it and use in the samples as well.

project/ReflectiveCodeGen.scala
octonato: yep, it works. I removed in the intermediate plugin

build.sbt
```diff
@@ -98,14 +99,18 @@ lazy val tck = project
 lazy val codegenCore =
   project
     .in(file("codegen/core"))
-    .enablePlugins(PublishSonatype)
-    .dependsOn(sdk)
```
raboof, line 102: Nice to remove this dependency :+1:

build.sbt
```diff
@@ -98,14 +99,18 @@ lazy val tck = project
...
       name := "akkaserverless-codegen-core",
       testFrameworks += new TestFramework("munit.Framework"),
       Test / fork := false
     )
     .settings(Dependencies.codegenCore)
+    .settings(
+      Compile / akkaGrpcGeneratedLanguages := Seq(AkkaGrpc.Java)
+    )
+    .settings(crossScalaVersions := Dependencies.ScalaVersionForCodegen,
+              scalaVersion := Dependencies.ScalaVersionForCodegen.head)
```
raboof, line 113: it's a bit unfortunate that we need to force these modules to use Scala 2.12 so they can be used 'reflectively', but OK I suppose ;).

raboof (approved):
I hope the increased complexity with cross-building in the sbt configuration won't come back to haunt us, but the faster feedback on whether generated classes actually compile does seem appealing.

raboof:
I can confirm this is very helpful while writing source generators, but also I noticed some weird problems: it seems sometimes the incremental compiler in sbt gets confused and compiles things in the wrong order, causing 'cannot find symbol' errors for classes that are definitely generated.

That said, I haven't seen that interfere with CI for valid builds yet, so perhaps still a net positive.

patriknw:
ready to merge? (I don't know what it does)

raboof:
> ready to merge?

Yes, I think so

> (I don't know what it does)

It introduces a `codegenJavaCompilationTest` sbt project that runs the codegen for some sample .proto's and checks it compiles as a Java project. This saves having to `publishLocalM2` and test it from Maven.

PR #192, opened 2021-08-04
fix: accumulate protobuf descriptors 

ennru:
The Eventing Shopping Cart sample doesn't work with the current codegen AFAICS.﻿

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -188,7 +188,7 @@ jobs:
 
   samples-tests:
     machine: true
-    description: "Maven Samples Integreation Tests"
+    description: "Maven Samples Integration Tests"
     steps:
       - checkout-and-merge-to-main
       - setup_sbt
@@ -217,6 +217,12 @@ jobs:
           name: Run integration tests for Counter Replicated Entity sample
           command: |
             cd samples/replicatedentity-counter
+            echo "Running mvn with SDK version: '$SDK_VERSION'"
+            mvn -Dakkaserverless-sdk.version=$SDK_VERSION verify -Pit
+      - run:
+          name: Run integration tests for the Eventing Shopping Cart sample
+          command: |
+            cd samples/java-eventing-shopping-cart
             echo "Running mvn with SDK version: '$SDK_VERSION'"
             mvn -Dakkaserverless-sdk.version=$SDK_VERSION verify -Pit
       - save_deps_cache
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -224,7 +224,8 @@ jobs:
           command: |
             cd samples/java-eventing-shopping-cart
             echo "Running mvn with SDK version: '$SDK_VERSION'"
-            mvn -Dakkaserverless-sdk.version=$SDK_VERSION verify -Pit
+            # TODO: enable integration tests with -Pit
+            mvn -Dakkaserverless-sdk.version=$SDK_VERSION verify
       - save_deps_cache
 
   publish:
--- a/codegen/core/src/main/scala/com/lightbend/akkasls/codegen/DescriptorSet.scala
+++ b/codegen/core/src/main/scala/com/lightbend/akkasls/codegen/DescriptorSet.scala
@@ -49,10 +49,8 @@ object DescriptorSet {
   @SuppressWarnings(Array("org.wartremover.warts.Throw"))
   def fileDescriptors(
       file: File
-  ): Either[CannotOpen, Iterable[Either[ReadFailure, Descriptors.FileDescriptor]]] =
-    Using[FileInputStream, Either[CannotOpen, Iterable[
-      Either[ReadFailure, Descriptors.FileDescriptor]
-    ]]](
+  ): Either[CannotOpen, Either[ReadFailure, Iterable[Descriptors.FileDescriptor]]] =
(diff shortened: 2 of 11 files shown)


franciscolopezsancho:
Currently reproducing it with the very same .proto files in a test. This way we can directly test the generation itself. 
This process can also be useful for later tests as an example of how to load and use the proto files from a sample in the `codegen`.

.circleci/config.yml
```diff
@@ -219,6 +219,12 @@ jobs:
             cd samples/replicatedentity-counter
             echo "Running mvn with SDK version: '$SDK_VERSION'"
             mvn -Dakkaserverless-sdk.version=$SDK_VERSION verify -Pit
+      - run:
+          name: Run integration tests for the Eventing Shoppong Cart sample
```
pvlugter, line 223:
```suggestion
          name: Run integration tests for the Eventing Shopping Cart sample
```

codegen/core/src/main/scala/com/lightbend/akkasls/codegen/DescriptorSet.scala
```diff
@@ -49,10 +49,8 @@ object DescriptorSet {
   @SuppressWarnings(Array("org.wartremover.warts.Throw"))
   def fileDescriptors(
       file: File
-  ): Either[CannotOpen, Iterable[Either[ReadFailure, Descriptors.FileDescriptor]]] =
-    Using[FileInputStream, Either[CannotOpen, Iterable[
-      Either[ReadFailure, Descriptors.FileDescriptor]
-    ]]](
+  ): Either[CannotOpen, Either[ReadFailure, Iterable[Descriptors.FileDescriptor]]] =
```
raboof, line 52: So now if we fail to read one file we throw away all results? That might be OK if we're sure we need all of them anyway

codegen/core/src/main/scala/com/lightbend/akkasls/codegen/DescriptorSet.scala
```diff
@@ -83,4 +78,28 @@ object DescriptorSet {
...
+   **/
+  private def accumulatedBuildFrom(
+      reads: Either[ReadFailure, Iterable[Descriptors.FileDescriptor]],
+      file: DescriptorProtos.FileDescriptorProto
+  ): Either[ReadFailure, Iterable[Descriptors.FileDescriptor]] = {
+    reads match {
+      case Left(_) => reads
+      case Right(fileDescriptors) =>
+        try {
+          Right(fileDescriptors ++ List(Descriptors.FileDescriptor.buildFrom(file, fileDescriptors.toArray, true)))
```
raboof, line 97: So we have to pass 'depended-upon' descriptors when parsing 'depending' descriptors - meaning the files must be accumulated in the 'right order'. How do we make sure they are in the right order?

raboof (approved):
Some questions, but if I'm reading this right it LGTM

The `java-eventing-shopping-cart` exercises this scenario, right?

codegen/core/src/main/scala/com/lightbend/akkasls/codegen/DescriptorSet.scala
ennru: I'm not sure they have to be in the right order, but maybe we should `allowUnknownDependencies=false` instead?

codegen/core/src/main/scala/com/lightbend/akkasls/codegen/DescriptorSet.scala
ennru: We bail out on any error now, before it could live with certain descriptors failing.

ennru (commented):
Yes, the initially [failing builds](https://app.circleci.com/pipelines/github/lightbend/akkaserverless-java-sdk/678/workflows/6fbdd362-ae4d-47fd-ab40-04dc5fad5dad/jobs/3997) show how it handled the descriptors in isolation and "invented" placeholders for unresolved references.

codegen/core/src/main/scala/com/lightbend/akkasls/codegen/DescriptorSet.scala
raboof: > I'm not sure they have to be in the right order

Whatever descriptor is the first one will be called with an empty `fileDescriptors` array, so if some of them need things to be available in there then that'd mean the order matters?

> I'm not sure they have to be in the right order, but maybe we should `allowUnknownDependencies=false` instead?

I'm not sure, I guess you could have unknown dependencies (like external proto annotations that we don't care about) that we could allow?

ennru: [This comment](https://github.com/protocolbuffers/protobuf/blob/ae26a81918fa9e16f64ac27b5a2fb2b110b7aa1b/java/core/src/main/java/com/google/protobuf/Descriptors.java#L301-L309) is what I expected, but the API doesn't seem to allow us to first accumulate the descriptors and after that trigger linking. 

I assume `allowUnknownDependencies=false` would make it fail when the order is wrong, but now the placeholders will be linked once the correct descriptor has been added to the list.

ennru:
Rebased as preparation to merge as two commits.

Issue #125, opened 2021-07-08
Add documentation on use of CatchOthers with Views


rstento:
This should be documented for proto and the view impl.

For example, I have a view that only utilizes 2 of many events and the others are ignored. Anything that subscribes to an event sourced entity may need a CatchOthers.

PR #194, opened 2021-08-05
doc: update customer registry sample with ignored event for views

pvlugter:
Refs #125 and #189. Update the customer registry sample used for views with a change address command and address changed event, which is ignored by the views.


ennru (approved):
LGTM.

PR #198, opened 2021-08-09
fix: base on codegen for ShoppingCart in eventing example

ennru:
This PR retrofits use of codegen on the Eventing Shopping Cart example code, which is used for documentation snippets.﻿

Codegen uses different defaults than this example did.

The `ShoppingCartEntity` implementation moves into the `shopping.cart.domain` package and now extends the codgen-generated abstract base class which consistently returns `Reply`.

Similar changes are required for the `ProductPopularityEntity` and the `ShoppingCartView`.


ennru:
#199 must be fixed to use the View codegen in the sample

https://github.com/lightbend/akkaserverless-java-sdk/blob/7944836c27414bd96961245c1d09b86c727c393b/samples/java-eventing-shopping-cart/src/main/java/shopping/cart/ShoppingCartView.java#L34-L38

patriknw (commented):
@raboof Do you know the status here?

raboof:
Not really, but it looks like the next step is to fix #199 . Adding a comment there as well :)

ennru:
This succeeds on CI as the View implementation lives in `shopping.cart.ShoppingCartView` (which uses `UpdateHandlerContext` see #199) whereas codegen creates an empty `shopping.cart.view.ShoppingCartViewServiceView`.

ennru:
Not relevant anymore.

Issue #199, opened 2021-08-10
View codegen must add `UpdateHandler` to method signatures


ennru:
The current codegen for Views creates message signatures for update handlers with the incoming event and the state as `Optional` in the abstract base class. 

Logic that requires access to the `UpdateHandlerContext` can't use codegen.

raboof:
I think the context will move to the abstract superclass as part of #52 anyway. Do we want to fix this on `main` before we finish and merge #52?

patriknw:
This is fixed by new codegen.

PR #200, opened 2021-08-10
feature: rely on some codegen in the Customer Registry sample

ennru:
The current "Customer Registry" contains both a Value Entity and an Event-sourced Entity both are required for doc-snippets.

This PR enables codegen for the Value Entity and for the View and reflects that (a bit) in the documentation.

Similar to #198 

Discovered #199 during this.


patriknw (approved):
LGTM

PR #208, opened 2021-08-16
Avoid name clash between View and Akka gRPC interface

### LINKED ISSUE #207: View class name clashes between Akka gRPC and Akka Serverless
Both Akka Serverless and Akka gRPC generate a class that matches the package and service names of the view from the `.proto` (i.e. `view.CustomerByName`): Akka Serverless for the implementation, and Akka gRPC for the public API.

Akka Serverless does not allow choosing a different class name, because it is referenced from the generated 'registrations'.

You would expect to have the same problem with entities. However, in that case the Akka Serverless implementation name is not derived from the `service` (i.e. `api.CustomerService`) but from the component implementing the service (i.e. `domain.Customer`). Unfortunately, views don't have an equivalent in the `proto`: just the `service` is sufficient in that case.

raboof:
Fixes #207

### DIFF
--- a/codegen/core/src/main/scala/com/lightbend/akkasls/codegen/Packaging.scala
+++ b/codegen/core/src/main/scala/com/lightbend/akkasls/codegen/Packaging.scala
@@ -16,6 +16,8 @@
 
 package com.lightbend.akkasls.codegen
 
+import com.akkaserverless.ServiceOptions.ServiceType
+
 import scala.jdk.CollectionConverters._
 import com.google.protobuf.Descriptors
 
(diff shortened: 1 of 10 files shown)


codegen/core/src/main/scala/com/lightbend/akkasls/codegen/Packaging.scala
```diff
@@ -48,6 +57,12 @@ object FullyQualifiedName {
       } else PackageNaming.from(fileDescriptor)
     FullyQualifiedName(
       descriptor.getName,
+      serviceType match {
+        case ServiceType.SERVICE_TYPE_VIEW =>
+          if (descriptor.getName.endsWith("View")) descriptor.getName + "Impl"
+          else descriptor.getName + "View"
```
patriknw, line 63: 👍

patriknw (approved):
LGTM

PR #209, opened 2021-08-17
First step in adding back replicated entity info

rstento:
This PR adds back the old information on using replicated entities. Please review and update with information on the current implementation.

Also note that I left the includes but right now, it appears as though no example is being included in the doc build. These will need to be adjusted accordingly.

### DIFF
--- a/docs/src/modules/java/pages/replicated-entity.adoc
+++ b/docs/src/modules/java/pages/replicated-entity.adoc
@@ -4,13 +4,13 @@ include::ROOT:partial$include.adoc[]
 
 ifdef::review[REVIEWERS: this is the content from the previous version of replicated entities, please bring it up to date with the newest implementation]
 
-This page documents how to implement  Replicated Entities in Java. For information on what Replicated Entities are, please read how to https://developer.lightbend.com/docs/akka-serverless/designing/state-model.html[choose a state model] and the general xref:developing:replicatedentities.adoc[Replicated Entity] information first.
+This page documents how to implement  Replicated Entities in Java. For information on what Replicated Entities are, please read how to https://developer.lightbend.com/docs/akka-serverless/designing/state-model.html[choose a state model] information first.
 
 A Replicated Entity can be created by annotating it with the link:{attachmentsdir}/api/com/akkaserverless/javasdk/replicatedentity/ReplicatedEntity.html[`@ReplicatedEntity`{tab-icon}, window="new"] annotation.
 
 [source,java,indent=0]
 ----
-include::example$docs/user/replicatedentity/ShoppingCartEntity.java[tag=Entity-class]
+include::example$replicatedentity/ShoppingCartEntity.java[tag=Entity-class]
 ----
 
 == Accessing and creating an Entity's Replicated Entity
@@ -25,7 +25,7 @@ For most use cases, simply injecting the data type directly into the constructor
 
 [source,java,indent=0]
 ----
-include::example$docs/user/replicatedentity/ShoppingCartEntity.java[tag=creation]
+include::example$replicatedentity/ShoppingCartEntity.java[tag=creation]
 ----
 
 In addition to the Replicated Entity, the constructor may accept a link:{attachmentsdir}/api/com/akkaserverless/javasdk/replicatedentity/ReplicatedEntityCreationContext.html[`ReplicatedEntityCreationContext`{tab-icon}, window="new"].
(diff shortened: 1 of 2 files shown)


docs/src/modules/java/pages/replicated-entity.adoc
```diff
@@ -0,0 +1,192 @@
+= Implementing Replicated Entities
+
+include::ROOT:partial$include.adoc[]
+
+ifdef::review[REVIEWERS: this is the content from the previous version of replicated entities, please bring it up to date with the newest implementation]
+
+This page documents how to implement  Replicated Entities in Java. For information on what Replicated Entities are, please read how to https://developer.lightbend.com/docs/akka-serverless/designing/state-model.html[choose a state model] information first.
+
+A Replicated Entity can be created by annotating it with the link:{attachmentsdir}/api/com/akkaserverless/javasdk/replicatedentity/ReplicatedEntity.html[`@ReplicatedEntity`{tab-icon}, window="new"] annotation.
```
patriknw, line 9: I wonder if it's too early to add the documentation for replicated entity? I think the API is not finalized yet.

docs/src/modules/java/pages/replicated-entity.adoc
```diff
@@ -0,0 +1,192 @@
...
+
+Caution should be taken when deleting Replicated Entities--Akka Serverless needs to maintain tombstones for each deleted Replicated Entity. Over time, if many Replicated Entities are created and deleted, this will result in not just running out of memory, but increased network usage as the tombstones still need to be gossipped through the cluster for replication.
+
+== Streamed command handlers
+
+Replicated Entities support streaming over gRPC. Streamed commands can be used to receive and publish updates to the state. If a gRPC service call has a streamed result type, the handler for that call can accept a link:{attachmentsdir}/api/com/akkaserverless/javasdk/replicatedentity/StreamedCommandContext.html[`StreamedCommandContext`{tab-icon}, window="new"], and use that to register callbacks.
+
+
+
+=== Responding to changes
```
patriknw, line 73: this has been removed

docs/src/modules/java/pages/replicated-entity.adoc
```diff
@@ -0,0 +1,192 @@
...
+
+=== Responding to stream cancellation
+
+A streamed command handler may also register an link:{attachmentsdir}/api/com/akkaserverless/javasdk/replicatedentity/StreamedCommandContext.html#onCancel-java.util.function.Consumer-[`onCancel`{tab-icon}, window="new"] callback to be notified when the stream is cancelled. The cancellation callback handler may update the Replicated Entity. This is useful if the Replicated Entity is being used to track connections, for example, when using link:{attachmentsdir}/api/com/akkaserverless/javasdk/replicatedentity/Vote.html[`Vote`{tab-icon}, window="new"] Replicated Entities to track a user's online status.
+
+== Replicated Entity data types
+
+The Akka Serverless Java SDK offers Java classes for each of the data types available.
+
+=== Counters and flags
```
patriknw, line 105: there are many name changes here

patriknw:
I think we (engineering) should take a first stab at adding back the documentation when the API is ready for that. Then @rstento can edit.

cc @pvlugter

docs/src/modules/java/pages/replicated-entity.adoc
pvlugter: Yes, the whole `Streamed command handlers` section should be removed.

docs/src/modules/java/pages/replicated-entity.adoc
pvlugter: We can either add docs for the current API, with annotations, and release that; or wait until this is all switched over to new codegen approach.

patriknw:
I asked Leon in a meeting how we should proceed with releasing/documenting replicated entities. Since we are close to completing the new code generation based api for other components we should do that for replicated entities too before announcing to the public audience. Meaning that we hold this documentation effort until we have the final APIs and code generation in place. It is scheduled for next sprint.

rstento:
> I asked Leon in a meeting how we should proceed with releasing/documenting replicated entities. Since we are close to completing the new code generation based api for other components we should do that for replicated entities too before announcing to the public audience. Meaning that we hold this documentation effort until we have the final APIs and code generation in place. It is scheduled for next sprint.

Sounds good to me. I'll close this ticket.

PR #217, opened 2021-08-19
Make generated ValueEntityProvider a builder

patriknw:
* include options in ValueEntityProvider instead of having overloaded register
  methods
* this will make it easier for us to add more optional parameters when needed

Follow up on https://github.com/lightbend/akkaserverless-java-sdk/pull/190/files#r687726999

### DIFF
--- a/codegen/java-gen/src/main/scala/com/lightbend/akkasls/codegen/java/SourceGenerator.scala
+++ b/codegen/java-gen/src/main/scala/com/lightbend/akkasls/codegen/java/SourceGenerator.scala
@@ -203,7 +203,7 @@ object SourceGenerator extends PrettyPrinter {
                 |)""".stripMargin
 
           case entity: ModelBuilder.ValueEntity =>
-            s".register(new ${entity.fqn.name}Provider(create${entity.fqn.name}))"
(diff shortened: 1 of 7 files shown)


samples/java-customer-registry/src/main/java/customer/CustomerValueEntityProvider.java
```diff
@@ -17,20 +17,37 @@
...
 import com.akkaserverless.javasdk.valueentity.ValueEntityContext;
+import com.akkaserverless.javasdk.valueentity.ValueEntityOptions;
 import com.akkaserverless.javasdk.valueentity.ValueEntityProvider;
 import com.google.protobuf.Descriptors;
 import customer.api.CustomerApi;
 
 import java.util.function.Function;
 
-// to be generated in generated-sources/src/main/java
+// FIXME to be generated in generated-sources/src/main/java
```
patriknw, line 27: Seems like generation is not used in this sample yet. I guess the protos don't have the codegen annotations yet. I'll follow up on that in separate PR.

patriknw:
I'll continue by adding similar provider registration for event sourced

johanandren (approved):

octonato (approved):
Looks much better with the builder approach. 👍

PR #230, opened 2021-08-23
Update handler effects API for views 

johanandren:
Refs #52


samples/java-eventing-shopping-cart/src/main/java/shopping/cart/ShoppingCartView2.java
```diff
@@ -0,0 +1,66 @@
...
+import java.util.Optional;
+
+// user view impl
+public class ShoppingCartView2 extends AbstractShoppingCartView {
+
+  private static Logger LOG = LoggerFactory.getLogger(ShoppingCartView2.class);
+
+  public UpdateEffect<ShoppingCartViewModel.CartViewState> processItemAdded(
+      ShoppingCartDomain.ItemAdded event,
+      Optional<ShoppingCartViewModel.CartViewState> state) {
```
octonato, line 33: we may need to re-think if we want to pass an Optional here. Currently, we support a few variants and use reflection to find out how to call the method. When we generate it, we will need to pick one style. 

To keep it similar to the entities API, we could introduce an empty state for the view. If we don't have anything on the DB, we pass the emptyState to the user. User can decide to use `null` or whatever the profo default instance (or whatever they prefer). 


And we should swap the param: `(State, Command) => Effect`

sdk/src/main/java/com/akkaserverless/javasdk/view/ViewBase.java
```diff
@@ -0,0 +1,75 @@
...
+public abstract class ViewBase<S> {
+
+  private Optional<UpdateHandlerContext> updateHandlerContext = Optional.empty();
+
+  /**
+   * Additional context and metadata for an update handler.
+   *
+   * <p>It will throw an exception if accessed from constructor.
+   */
+  protected final UpdateHandlerContext updateHandlerContext() {
```
octonato, line 36: We send commands to entities and therefore we have a `CommandContext`. For views, we could have an `UpdateContext`. 

Having `handler` in the name sound weird.

octonato (commented):
left some comments and suggestions

samples/java-eventing-shopping-cart/src/main/java/shopping/cart/ShoppingCartView2.java
```diff
@@ -0,0 +1,66 @@
...
+import shopping.cart.view.ShoppingCartViewModel;
+
+import java.util.Optional;
+
+// user view impl
+public class ShoppingCartView2 extends AbstractShoppingCartView {
+
+  private static Logger LOG = LoggerFactory.getLogger(ShoppingCartView2.class);
+
+  public UpdateEffect<ShoppingCartViewModel.CartViewState> processItemAdded(
```
patriknw, line 31: `@Override`

samples/java-eventing-shopping-cart/src/main/java/shopping/cart/ShoppingCartView2.java
patriknw: sounds good to me, consistency is an important goal

sdk/src/main/java/com/akkaserverless/javasdk/view/ViewBase.java
patriknw: UpdateContext sounds good to me

sdk/src/main/java/com/akkaserverless/javasdk/view/ViewBase.java
```diff
@@ -0,0 +1,75 @@
...
+
+    interface Builder<S> {
+
+      UpdateEffect<S> updateState(S newState);
+
+      /** Ignore this event (and continue to process the next). */
+      UpdateEffect<S> ignore();
+
+      /**
+       * Trigger an error for the event. FIXME what happens on error, retry of the same event?
```
patriknw, line 68: yes, offset is not stored, projection is restarted and same is tried again. I assume `error` should have the same semantics as throwing an exception.

patriknw (commented):
looking good, this will also make it easier to add the more advanced update features

PR #231, opened 2021-08-23
EventSourcedEntityHandler generation and registration via provider

patriknw:
* codegen of EventSourcedEntityHandler (no reflection)
* registration via EventSourcedEntityProvider
* also split of java-customer-registry sampel into two;
  one for value entity and another for event sourced

Refs #52


samples/java-eventsourced-customer-registry/src/main/java/customer/domain/CustomerEntity.java
```diff
@@ -0,0 +1,123 @@
...
+import com.akkaserverless.javasdk.eventsourcedentity.CommandHandler;
+import com.akkaserverless.javasdk.eventsourcedentity.EventHandler;
+import com.akkaserverless.javasdk.eventsourcedentity.EventSourcedContext;
+import com.akkaserverless.javasdk.eventsourcedentity.EventSourcedEntity;
+import com.akkaserverless.javasdk.eventsourcedentity.EventSourcedEntityBase;
+import com.akkaserverless.javasdk.eventsourcedentity.EventSourcedEntityBase.Effect;
+import com.google.protobuf.Empty;
+import customer.api.CustomerApi;
+
+/** An event sourced entity. */
```
johanandren, line 18: Not necessarily for this PR but one thing that would be nice would be if we could include what protobuf it corresponds to here (like we have manually in some of the serverless framework test classes).

sdk/src/main/java/com/akkaserverless/javasdk/eventsourcedentity/EventSourcedEntityProvider.java
```diff
@@ -0,0 +1,33 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.akkaserverless.javasdk.eventsourcedentity;
+
+import com.akkaserverless.javasdk.impl.eventsourcedentity.EventSourcedEntityHandler;
+import com.google.protobuf.Descriptors;
+
+public interface EventSourcedEntityProvider<S, E extends EventSourcedEntityBase<S>> {
```
johanandren, line 22: Should we add some "Not for user extension" here or do we consider it actual API for user extension as well as common (internal) interface for generated code?

johanandren (approved):
LGTM

maven-java/akkaserverless-maven-archetype-event-sourced-entity/pom.xml
```diff
@@ -2,12 +2,12 @@
     <modelVersion>4.0.0</modelVersion>
 
     <artifactId>akkaserverless-maven-archetype-event-sourced-entity</artifactId>
-    <version>0.7.0-beta.18</version>
+    <version>0.7.0-beta.18-61-1b3ddaec-SNAPSHOT</version>
```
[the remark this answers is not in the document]
patriknw: yeah, I shouldn't have committed `maven-java`. I'll revert those.

sdk/src/main/java/com/akkaserverless/javasdk/eventsourcedentity/EventSourcedEntityProvider.java
patriknw: I'll add a comment that the concrete implementation is generated, to start with.

octonato: Good question. 

Let's be restrictive and eventually relax the rules if needed. We have all the interest in keep it 'private'.

patriknw: we can't make it completely internal because it's the parameter to `register`

patriknw:
This is ready. I'll continue in a new PR by looking at the differences between event sourced and value entities and try to align. Part of that also look at the serialization second step in https://github.com/lightbend/akkaserverless-java-sdk/pull/150

codegen/java-gen/src/main/scala/com/lightbend/akkasls/codegen/java/EntityServiceSourceGenerator.scala
```diff
@@ -316,7 +482,7 @@ object EntityServiceSourceGenerator {
                                       className: String): Option[String] = {
     entity match {
       case eventSourcedEntity: ModelBuilder.EventSourcedEntity =>
-        None
+        Some(eventSourcedEntityProvider(service, eventSourcedEntity, packageName, className))
```
octonato, line 485: I made it an option (here and above) because we didn't have a provider and a handler for ES. 

We can remove the Option now.

codegen/java-gen/src/main/scala/com/lightbend/akkasls/codegen/java/EntityServiceSourceGenerator.scala
```diff
@@ -197,20 +367,16 @@ object EntityServiceSourceGenerator {
...
-      <> line <>
       `class`("public", s"$className extends $interfaceClassName") {
         "@SuppressWarnings" <> parens(dquotes("unused")) <> line <>
         "private" <+> "final" <+> "String" <+> "entityId" <> semi <> line <>
         line <>
         constructor(
           "public",
           className,
-          List("@EntityId" <+> "String" <+> "entityId")
+          List("EventSourcedContext" <+> "context")
```
octonato, line 377: Value entity is only passing the `String entityId`. We need to align this.

The context is not so useful right now. It only has the entityId and the `serviceCallFactory`, but the `serviceCallFactory` is already available from the `CommandContext`. 

We probably want to have something better than the exiting (untyped) `ServiceCallFactory` and then we want it to be injected in the constructor so users can instantiate clients at construction time (and not per command using the CommandContext). 

So, passing the context is probably something we may need to keep (and introduce in VE), but not yet clear how we will really use it. 

On the other hand, passing context make unit testing harder. It will be better to not have a mock context of any kind.

sdk/src/main/java/com/akkaserverless/javasdk/eventsourcedentity/EventSourcedEntityProvider.java
octonato: what I mean by 'private' is that it's not real private, but marked as "Not for user extension"

octonato (approved):
Looking good. 

I left some comments and suggestions. Don't need to be impl in this PR, so approving already. We can eventually add in follow-up PRs.

codegen/java-gen/src/main/scala/com/lightbend/akkasls/codegen/java/ValueEntitySourceGenerator.scala
```diff
@@ -23,6 +23,7 @@ import org.bitbucket.inkytonik.kiama.output.PrettyPrinterTypes.Document
 
 object ValueEntitySourceGenerator {
   import SourceGenerator._
+  import EntityServiceSourceGenerator.generateImports
```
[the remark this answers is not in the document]
patriknw: it's used by both ValueEntitySourceGenerator and EntityServiceSourceGenerator, should probably be in some third utility class but EntityServiceSourceGenerator anyway looked like it wasn't only about EventSourced

codegen/java-gen/src/main/scala/com/lightbend/akkasls/codegen/java/EntityServiceSourceGenerator.scala
patriknw: ok, I'll remove that in next round

codegen/java-gen/src/main/scala/com/lightbend/akkasls/codegen/java/EntityServiceSourceGenerator.scala
patriknw: ok, there are more things that are different, I'll align them

codegen/java-gen/src/main/scala/com/lightbend/akkasls/codegen/java/ValueEntitySourceGenerator.scala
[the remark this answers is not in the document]
octonato: Correct, the code for value entity and event sourced entity was previously in EntityServiceSourceGenerator and we started to split it into two different generators. 

At some point, we spined-off `ValueEntitySourceGenerator` and `EntityServiceSourceGenerator` should have been renamed to `EntitySourcedServiceSourceGenerator`, but we had other PRs in the making and it was better to postpone the rename to avoid conflicts.

PR #234, opened 2021-08-24
regenerate valueentity-counter projects

octonato:
The value entity counter sample was using some non-generated classes and was not compatible with the new registration.


samples/valueentity-counter/src/main/java/com/example/Main.java
```diff
@@ -1,20 +1,29 @@
...
-}
+
+  private static final Logger LOG = LoggerFactory.getLogger(Main.class);
+
+  public static final AkkaServerless SERVICE =
+    // This withGeneratedComponentsAdded wrapper automatically registers any generated Actions, Views or Entities,
+    // and is kept up-to-date with any changes in your protobuf definitions.
+    // If you prefer, you may remove this wrapper and manually register these components.
+    registerAll(new AkkaServerless(),
+      Counter::new);
```
octonato, line 23: I think we could simplify this with

```java
registerAll(Counter::new);
```

and let registerAll create an instance of `AkkaServerless` and return it.

samples/valueentity-counter/src/main/java/com/example/domain/Counter.java
```diff
@@ -18,8 +19,8 @@
     @SuppressWarnings("unused")
     private final String entityId;
 
-    public Counter(String entityId) { // <1>
-        this.entityId = entityId;
+    public Counter(ValueEntityContext context) { // <1>
+        this.entityId = context.entityId();
```
octonato, line 23: Apparently we have already aligned this. Value Entity get the context.

samples/valueentity-counter-generated/src/main/java/com/example/domain/CounterProvider.java
```diff
@@ -0,0 +1,66 @@
...
+import com.akkaserverless.javasdk.valueentity.ValueEntityProvider;
+import com.example.CounterApi;
+import com.example.domain.CounterDomain;
+import com.google.protobuf.Descriptors;
+import com.google.protobuf.Empty;
+import com.google.protobuf.EmptyProto;
+import java.util.function.Function;
+
+/** A value entity provider */
+public class CounterProvider implements ValueEntityProvider {
```
patriknw, line 18: Why is this in `src/main/java`?

patriknw: what is `value-entity-counter-generated` compared to `value-entity-counter`? something we use for testing?

patriknw (commented):
looking good, just a question

samples/valueentity-counter-generated/src/main/java/com/example/domain/CounterProvider.java
octonato: Value entity counter has code that is generated using the maven plugin and we can't generate it when building with sbt. So, instead we do the following: 

* publish artifact locally
* generate code using the mvn plugin (cd samples/project && mvn compile) 
* the generated code is copied into `valueentity-counter-generated` in `src/main/java`. 
* the sbt project takes `value-entity-counter` and adds the code in `value-entity-counter-generated` as extra source. 

Now, when building with sbt, we can build `valueentity-counter` without needing the plugin. 
(you can find more info here: #171)

If it sounds complicated, is because it is. We have been trying a different things in an attempt to streamline the mixed build (sbt/mvn), but we didn't yet reach an optimal setup. 

The main advantage of this approach is that we can use IntelliJ to navigate through `valueentity-counter`, but the maintenance is high. I struggled today to get all things in place and got lost a few times.

octonato: The ideal solution would be to have a sbt plugin for the codegen. The samples would be a first-class project in sbt and not the hacky solution we have right now. 

We will need such a plugin for the Scala SDK anyway. The plugin will need to be able to emit java as well (like the grpc plugin).

docs/src/modules/java/pages/value-entity.adoc
```diff
@@ -107,29 +107,3 @@ As our example relies on the code generation, the registration gets automaticall
 ----
 include::java:example$valueentity-counter/src/main/java/com/example/Main.java[]
 ----
-
-== Testing the Entity
```
octonato, line 111: I needed to remove this part because the codegen is temporarily not generating the unit test. 

This section will need to be re-written anyway because it should cover the upcoming testkit.

samples/valueentity-counter-generated/src/main/java/com/example/domain/CounterProvider.java
patriknw: Wow, that sounds indeed complicated. Nothing we need to change or improve now.

I haven't used that approach and didn't find it too difficult to try out new codegen in sample. I first used the ordinary tests `codegenJava/test` and then the instructions in https://github.com/lightbend/akkaserverless-java-sdk/blob/main/CONTRIBUTING.md

I compile/run the sample with mvn, and have an intellij maven project for the sample (have to modify the sdk version in pom to import project).

patriknw (approved):
LGTM

PR #235, opened 2021-08-24
include views to java-gen-compilation-tests

octonato:
This include a view file in `codegen/java-gen-compilation-tests`.

For the record, the `codegen/java-gen-compilation-tests` module runs the codegen against proto files, generate code and verify that it can be compiled.

### DIFF
--- a/codegen/java-gen/src/main/scala/com/lightbend/akkasls/codegen/java/SourceGenerator.scala
+++ b/codegen/java-gen/src/main/scala/com/lightbend/akkasls/codegen/java/SourceGenerator.scala
@@ -203,12 +203,15 @@ object SourceGenerator extends PrettyPrinter {
             cmd.inputType :: cmd.outputType :: Nil
           }
 
+        val otherDescriptors = collectRelevantTypeDescriptors(relevantTypes, service.fqn)
+        val sep = if (otherDescriptors.isEmpty) "" else ","
+
         List(
           ".registerView(\n" ++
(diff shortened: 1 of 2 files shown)


octonato:
@johanandren, this is interesting for your current work. As you move with the codegen for the views  that will help you verify that the generate code is compilable. 

Note, the added proto file is quite simple, we may want to have a more complex one exercising more cases.

I will send a PR on top of this to add an action proto. However, I found a blocker for now. I think we never got a sample exercising the codegen and the actions API.

codegen/java-gen/src/main/scala/com/lightbend/akkasls/codegen/java/SourceGenerator.scala
```diff
@@ -203,12 +203,15 @@ object SourceGenerator extends PrettyPrinter {
...
+        val sep = if (otherDescriptors.isEmpty) "" else ","
+
         List(
           ".registerView(\n" ++
           (if (service.transformedUpdates.nonEmpty) s"  ${service.fqn.name}.class,\n" else "") ++
           s"""|  ${service.fqn.parent.javaOuterClassname}.getDescriptor().findServiceByName("${service.fqn.protoName}"),
-              |  "${service.viewId}",
-              |  ${Syntax.indent(collectRelevantTypeDescriptors(relevantTypes, service.fqn), 4)}
+              |  "${service.viewId}"$sep
+              |  ${Syntax.indent(otherDescriptors, 4)}
```
octonato, line 214: we may not have extra descriptors to add, so we need the separator may be an empty string or a `,`.

johanandren (approved):

PR #240, opened 2021-08-25
View Effects implementation

johanandren:
On top of #230 

Refs #52

Dropped annotations, handcoded `Provider` and `Handler`s in one sample and the TCK (which passes)


johanandren:
One thing left now, I think: how to deal with views without transform updates. I'm thinking if we should generate a separate `Provider` for those that can be used with `.register` without needing a handler/actual view etc. Alternative could be that we generate an empty `View` implementation (no update handlers) and everything else looks the same but then there'd still be a dummy `emptyState` etc.

codegen/java-gen/src/main/scala/com/lightbend/akkasls/codegen/java/ViewServiceSourceGenerator.scala
```diff
@@ -53,146 +57,226 @@ object ViewServiceSourceGenerator {
...
+       |
+       |$imports
+       |
+       |public class ${view.className} extends ${view.interfaceName} {
+       |
+       |  public ${view.className}(ViewContext context) {}
+       |
+       |  @Override
+       |  public ${qualifiedType(view.state.fqn)} emptyState() {
+       |    throw new RuntimeException("Empty state for '${view.className}' not implemented yet");
```
patriknw, line 244: in the entities we have:
```
throw new UnsupportedOperationException("Not implemented yet, replace with your empty entity state");
```

patriknw: Regarding generating a default empty state or not. I think this is fine, also for transform_updates=false because it will not be called in that case.

patriknw (commented):
looking good

patriknw:
> Alternative could be that we generate an empty View implementation (no update handlers) and everything else looks the same but then there'd still be a dummy emptyState etc.

I like this alternative. Would be good also for the upcoming transform_results combination.
emptyState can be as it is with UnsupportedOperationException since it will not be called.

codegen/java-gen/src/main/scala/com/lightbend/akkasls/codegen/java/ViewServiceSourceGenerator.scala
johanandren: Was just arriving at the same thought.

codegen/core/src/main/scala/com/lightbend/akkasls/codegen/ModelBuilder.scala
```diff
@@ -109,7 +109,19 @@ object ModelBuilder {
...
+      else fqn.name + "View"
+
+    val abstractViewName = "Abstract" + viewClassName
+    val handlerName = viewClassName + "Handler"
+    val providerName = viewClassName + "Provider"
+
+    val classNameQualified = s"${fqn.parent.javaPackage}.$viewClassName"
+    val providerNameQualified = s"${fqn.parent.javaPackage}.$providerName"
+
+    val state = State(commands.head.outputType)
```
johanandren, line 124: I think just picking the first command output type as the state type is not quite safe, one scenario where it will break that I can think of is a view with no transform and only a query that "projects" the result into some other type (unless I remember wrong that we support that).

patriknw: we have a validation for the return types, if that is the case you think about: AS-00109 All View update methods must have the same return type

johanandren: Hmm, maybe I'm misinterpreting that the `commands` here will include both queries and update method - was thinking that the first command (not sure how they are ordered) could be the query, and that could be a projection into some non-state type.

patriknw: ah, I understand, you are right if commands includes both update methods and query methods. it should look for a update method

johanandren:
I found one issue that is messed up but that I don't think surfaces with the current enabled samples compiled/tested and that is the de-duplication of names where it tries to add `Impl` or such to the name to avoid it clashing with the classes generated by protobuf. 

Right now that turns `service ShoppingCartViewService` into `ShoppingCartViewServiceView` in the java-eventing-shopping-cart sample, and if I rename it to `ShoppingCartView` which makes sense, I get `ShoppingCartViewImplView` generated which is even weirder.

patriknw:
related to the View/Impl naming is https://github.com/lightbend/akkaserverless-java-sdk/issues/238

sdk/src/main/java/com/akkaserverless/javasdk/view/View.java
```diff
@@ -16,22 +16,64 @@
...
+   * <p>It will throw an exception if accessed from constructor.
+   */
+  protected final UpdateContext updateContext() {
+    return updateContext.orElseThrow(
+        () ->
+            new IllegalStateException("UpdateContext is only available when handling an update."));
+  }
+
+  /** INTERNAL API */
+  public final void setUpdateContext(Optional<UpdateContext> context) {
```
patriknw, line 40: While you are at it, could you rename this to `_internalSetUpdateContext`. I did similar rename  in https://github.com/lightbend/akkaserverless-java-sdk/pull/255 but didn't touch this to avoid conflicts.

I'd like to "hide" it because it's public but internal.

sdk/src/main/scala/com/akkaserverless/javasdk/impl/view/ViewHandler.scala
```diff
@@ -0,0 +1,48 @@
...
+
+package com.akkaserverless.javasdk.impl.view
+
+import com.akkaserverless.javasdk.view.{UpdateContext, View}
+
+import java.util.Optional
+
+abstract class ViewHandler[S, V <: View[S]](protected val view: V) {
+
+  final def handleUpdate(state: Option[Any], event: Any, context: UpdateContext): View.UpdateEffect[_] = {
```
patriknw, line 25: rename to `_internalHandleUpdate` and mark with INTERNAL API

codegen/core/src/main/scala/com/lightbend/akkasls/codegen/ModelBuilder.scala
johanandren: Did something about that now, looking only at the first update method (with some test coverage). Not sure if there may still be cases I have not thought about.

johanandren:
Ready for final review. `java-eventing-shopping-cart` is broken, but it also is regardless of this PR, and not part of what's validated by CI for the `sdk-codegen-dev`-branch so let's fix that later all at once.

patriknw (approved):
LGTM

PR #251, opened 2021-08-26
non-reflective API for Actions

octonato:
Still in draft, but first review round welcome. 

TODO:
- [x] re-enable codegen (need to disable because lost of old code was remove. ie: annotation, odl register)
- [ ] solution for hacky Reply
- [x] add proper sample using action, not half hacked in customer registry


johanandren (approved):
Feels like quite a bit of stuff in progress here, but let's merge so that I can continue with views on top instead of conflicting work that we notice later.

PR #256, opened 2021-08-31
Test emit of several events, #116

### LINKED ISSUE #116: emit several events
When I added the initial effects I only added support for emitting one event. We should also have support for many. `emitEvents` taking a List or vararg parameter. Might not be possible to have overload with both List and and vararg so might be enough with List.

patriknw:
It was already implemented, but I added a test.

Fixes #116

### DIFF
--- a/sdk/src/main/java/com/akkaserverless/javasdk/eventsourcedentity/EventSourcedEntity.java
+++ b/sdk/src/main/java/com/akkaserverless/javasdk/eventsourcedentity/EventSourcedEntity.java
@@ -20,9 +20,9 @@ import com.akkaserverless.javasdk.impl.eventsourcedentity.EventSourcedEntityEffe
 import com.akkaserverless.javasdk.Metadata;
 import com.akkaserverless.javasdk.ServiceCall;
 import com.akkaserverless.javasdk.SideEffect;
-import com.akkaserverless.javasdk.reply.MessageReply;
 
 import java.util.Collection;
+import java.util.List;
 import java.util.Optional;
 import java.util.function.Function;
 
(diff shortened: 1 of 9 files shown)


sdk/src/main/java/com/akkaserverless/javasdk/eventsourcedentity/EventSourcedEntity.java
```diff
@@ -94,9 +94,7 @@ public final void _internalSetEventContext(Optional<EventContext> context) {
 
       OnSuccessBuilder<S> emitEvent(Object event);
 
-      OnSuccessBuilder<S> emitEvents(Object event1, Object... additionalEvents);
-
-      OnSuccessBuilder<S> emitEvents(Collection<? extends Object> event);
+      OnSuccessBuilder<S> emitEvents(List<?> event);
```
patriknw, line 97: Not good to allow unordered things here, such as `Set`.

Also, I think there is a problem with overloads with vararg Object like that.

sdk/src/main/scala/com/akkaserverless/javasdk/impl/eventsourcedentity/EventSourcedEntityHandler.scala
```diff
@@ -116,7 +116,7 @@ abstract class EventSourcedEntityHandler[S, E <: EventSourcedEntity[S]](protecte
           currentSequence += 1
           shouldSnapshot = shouldSnapshot || (snapshotEvery > 0 && currentSequence % snapshotEvery == 0)
         }
-        // FIXME currently snapshotting final state after applying all events even if trigger was mid-event stream?
+        // snapshotting final state since that is the "atomic" write
```
patriknw, line 119: included test of this

johanandren (approved):

Issue #263, opened 2021-09-01
Bring back JSON support


patriknw:
and enable the eventing TCK tests again

see FIXME in EventSourcedEntityTwoHandler

PR #269, opened 2021-09-02
Bring back JSON support to codegen

johanandren:
First stab at a possible new JSON support solution #263

More explicit and less magic (and dangerous), no new protobuf annotations needed, user code instead says exactly type to try to deserialize to.

If we agree on something like this I think what is missing is support for decoding heterogenous json types based on the "jsonType" in the type_url (since it is a single handler for all possible json subtypes). Perhaps something like `.deserializeJsonAndHandle(subType, ..., decoded -> return effect for that type)`.

I think we should also drop the `@Jsonable` annotation if we go with this because all places we expect to serialize/deserialize is explicit so the annotation becomes superfluous.


sdk/src/main/scala/com/akkaserverless/javasdk/impl/AnySupport.scala
```diff
@@ -338,31 +338,27 @@ class AnySupport(descriptors: Array[Descriptors.FileDescriptor],
   private def resolveTypeUrl(typeName: String): Option[ResolvedType[_]] =
     allTypes.get(typeName).map(resolveTypeDescriptor)
 
-  private def decodeJson(typeUrl: String, bytes: ByteString) = {
-    val jsonType = typeUrl.substring(AkkaServerlessJson.length)
+  def decodeJson[T](allowedClass: Class[T], typeUrl: String, bytes: ByteString): T = {
+    if (!typeUrl.startsWith(AkkaServerlessJson)) {
+      throw new IllegalArgumentException(
+        s"Protobuf bytes with type url [$typeUrl] cannot be decoded as JSON, must start with ${AkkaServerlessJson}"
```
patriknw, line 344:
```suggestion
        s"Protobuf bytes with type url [$typeUrl] cannot be decoded as JSON, must start with [${AkkaServerlessJson}]"
```

sdk/src/main/java/com/akkaserverless/javasdk/action/ActionContext.java
```diff
@@ -41,4 +42,14 @@
...
+
+  /**
+   * Try to deserialize the protobuf any message to the given class annotated
+   *
+   * @param jsonClass The class to deserialize to.
+   * @param any A protobuf Any message with the type `json.akkaserverless.com/JsonValue`.
+   * @param <T> The type of the deserialized message
+   * @return The deserialized message
+   */
+  <T> T deserializeJson(Class<T> jsonClass, Any any);
```
patriknw, line 54: since this is more like a utility function we don't have to provide it via the context, right? Place it in some helper class? JsonConverter? JsonSupport? JsonDecoder?

tck/src/main/java/com/akkaserverless/javasdk/tck/model/localpersistenceeventing/LocalPersistenceSubscriber.java
```diff
@@ -18,63 +18,73 @@
...
-        .setMessage(jsonMessage.message)
-        .build();
+  public Effect<LocalPersistenceEventing.Response> processAnyValue(Any any) {
+    JsonMessage jsonMessage = actionContext().deserializeJson(JsonMessage.class, any);
+    return effects()
+        .reply(
+            LocalPersistenceEventing.Response.newBuilder()
+                .setId(actionContext().eventSubject().orElse(""))
+                .setMessage(jsonMessage.message)
+                .build());
```
patriknw, line 72: did we have support for encoding JSON also? Like publishing a JSON message to a topic?

Would that be similar with `Any` as the response type and then explicit conversion?

tck/src/main/java/com/akkaserverless/javasdk/tck/model/localpersistenceeventing/LocalPersistenceSubscriber.java
```diff
@@ -18,63 +18,73 @@
...
   }
 
-  public LocalPersistenceEventing.Response processAnyEvent(
-      JsonMessage jsonMessage, CloudEvent cloudEvent) {
-    return LocalPersistenceEventing.Response.newBuilder()
-        .setId(cloudEvent.subject().orElse(""))
-        .setMessage(jsonMessage.message)
-        .build();
+  public Action.Effect<LocalPersistenceEventing.Response> processAnyEvent(Any any) {
+    JsonMessage jsonMessage = actionContext().deserializeJson(JsonMessage.class, any);
```
patriknw, line 44: I like this explicit approach but I don't know how it was intended to work before and if there was some strong opinions about that it should be automagic and not be part of user's business logic code.

sdk/src/main/java/com/akkaserverless/javasdk/action/ActionContext.java
johanandren: It currently needs access to the `AnySupport` instance that comes from the service to do the actual json deserialization. But perhaps it could be extracted from there, I'll look into that.

tck/src/main/java/com/akkaserverless/javasdk/tck/model/localpersistenceeventing/LocalPersistenceSubscriber.java
johanandren: We already have `emitJson`-effects in the value and event sourced entities, so I guess we can align with that. Not entirely sure I understand how publishing to a topic works and if that fits though, will look into that.

tck/src/main/java/com/akkaserverless/javasdk/tck/model/localpersistenceeventing/LocalPersistenceSubscriber.java
johanandren: The alternative as I see it is with reflection out of the picture is to introduce protobuf annotations describing the possible Java classes that the json could be deserialized to and then generate methods for each of those types as well as logic to try to deserialize the right type in the handler. That would add more complexity here, require bigger changes all the way back to the protocol/annotations adding another case of stringly typed references to the Java codebase in the protobuf.

johanandren:
Digging a bit deeper:

The JSON support currently actually documented are the following two aspects as far as I can see:

1. Subscribing to a topic, defining the input type to `google.protobuf.Any`. This is needed since the producer could be non-akka-serverless and want to produce json.

https://developer.lightbend.com/docs/akka-serverless/java/topic-eventing.html#_receiving_json_messages

This is pretty much what I address in this PR, and since the format of the published data is in the control of the producer I'd expect that it could also be some other format than json, which means that an action/entity will need access to the raw bytes to be able to use a third party library to deserialize to the internal data format of the app.

2. Arbitrary use of JSON in place of probuf, no real actual use cases described in docs or samples. The TCK uses this to emit json events from an event sourced entity and a value entity to be consumed by in a topic subscriber and verify that the subscriber part works (but not actually replaying/applying json events.

> Akka Serverless uses Jackson to serialize JSON. Any classes that are annotated with `@Jsonable` will be serialized to and from JSON using Jackson.

https://developer.lightbend.com/docs/akka-serverless/java/serialization.html#_json

I guess this part is more about thinking that Java devs want to avoid having to deal with the protobuf builder apis etc and work with pojos, but I'm not convinced it is a great idea to actually support - it essentially means sugar for an escape hatch using proto `Any` everywhere to store non-schema data/whatever.

Maybe I'm missing some strong motivation for this but I'd wish we take a step back from that idea and reconsider if it is worth it. Or for someone telling me how useful it is so that I change my mind.

sdk/src/main/java/com/akkaserverless/javasdk/action/ActionContext.java
patriknw: ah, I missed that part. I think AnySupport is per component so makes sense to add it to the context (or the Action base class)

tck/src/main/java/com/akkaserverless/javasdk/tck/model/localpersistenceeventing/LocalPersistenceSubscriber.java
patriknw: publishing to a topic is the returned message when proto defined with eventing.out

patriknw:
> The TCK uses this to emit json events from an event sourced entity and a value entity to be consumed by in a topic subscriber

We don't support eventing.out (publish to topic) directly from an EventSourcedEntity or ValueEntity. It must go via a projection to an action. Therefore I don't see the need for emitting/storing events in json.

tck/src/main/java/com/akkaserverless/javasdk/tck/model/localpersistenceeventing/LocalPersistenceSubscriber.java
patriknw: Unless I miss something we should focus on the json support for pub-sub. Since that is done via actions and the main responsibility of such actions is to translate between internal/external protocols I think it's a good fit to handle json there as well. Explicitly.

tck/src/main/java/com/akkaserverless/javasdk/tck/model/localpersistenceeventing/LocalPersistenceSubscriber.java
johanandren: Sorry, I was wrong about that, `emitJson` was TCK action methods. After some more investigation any parameter or return value where you instead used a non proto Java class annotated with `@Jsonable` would be automagically serialized/deserialized into a proto any with the akka serverless json type_url.

tck/src/main/java/com/akkaserverless/javasdk/tck/model/localpersistenceeventing/LocalPersistenceSubscriber.java
octonato: Ok, I may be missing the whole json support story here, but it isn't clear to me how that will look like for a generated action. The problem with this example from the TCK is that it's fully hand-coded. 

For illustration, let's take the following protobuf fragment

```
rpc SayHello(Person) returns (Hello) {}
```

this will generate a method as

```java
public Action.Effect<Hello> sayHello(Person person)
```

How would we know that when generating this Action, we will need to pass a String or Any and that the user will have to manually deserialize it to the right type? 

So, the only way out of this to me is to say, if you want to pass json to an action, you have to encode it by hand yourself, but that means that user will have to generate a Provider and a Handler. And that bring us to the next question, how we will tell Akka Serverless how such an Action needs to be exposed? What are the endpoints? 

It seems to me that whatever we do, a proto service definition is always needed and it's in the proto file that we will need to tell if the types will be ultimately serialized to json.

johanandren:
Ok, I didn't find any docs on publishing to a topic so that's why it is a bit unclear to me. (Section "Publishing and subscribing to topics on a broker" only covers subscribing)

octonato:
> Subscribing to a topic, defining the input type to google.protobuf.Any. This is needed since the producer could be non-akka-serverless and want to produce json.

@johanandren, regarding this and what I commented [here](https://github.com/lightbend/akkaserverless-java-sdk/pull/269#discussion_r701856228). It seems that the path here is to create a proto file that accepts `google.protobuf.Any` and subscribe to some topic. As an implementer of such an Action, I will need to know what will be encoded inside the Any and deserialize it myself. Correct?

tck/src/main/java/com/akkaserverless/javasdk/tck/model/localpersistenceeventing/LocalPersistenceSubscriber.java
johanandren: So if a user is building an action that subscribes to a topic, they will need to _know_ what serialization format the messages are, and type their action input parameter as `google.protobuf.Any` (just like how it is now), then do an explicit deserialization to their expected internal type. We provide Jackson parsing json support out of the box to deserialize to pojos since that might be common enough.

If a user is publishing to a topic and do not want to use protobuf they will instead use `google.protobuf.Any` as return value, and have explicit logic to turn their object into the right serialized format. Again we provide explicit support to turn a pojo into json and then that into a `google.protobuf.Any` with the akka serverless json type-url.

If a user wants to produce or consume some other event encoding with an arbitrary third party library for serialization/deserialization they have access to the raw bytes from the `google.protobuf.Any` and can do that (not sure about how/if the proxy deals with that currently though).

johanandren:
@octonato yes, that is my suggestion at least https://github.com/lightbend/akkaserverless-java-sdk/pull/269#discussion_r701868060

patriknw:
> Section "Publishing and subscribing to topics on a broker" only covers subscribing

That is issue https://github.com/lightbend/akkaserverless-java-sdk/issues/122

johanandren:
To sum up what I'd like to do, after discussing this in this PR:

In addition to the json/grpc-web transcoding in front of the proxy already present for regular HTTP/json calls from clients
we will only support json to and from pub/sub-topics since that is potential communication with the outside world where a different serialized encoding than Protobuf is needed (and potentially out of our control if it is about consuming a topic). 

It is likely useful that users can also support an arbitrary format here (say XML, YAML, Avro, CSV, CBOR) with some third party library, which means access to the raw bytes is likely good.

Because of that we drop all other automagic encoding to json in favour of explicit methods for dealing with transformation from and to `google.protobuf.Any` instances with the json string and the `type_url` `json.akkaserverless.com/...`.

When deserializing consumed messages it is not likely that the outside knows/should know about the internal Java classes of the service, so trying to parse the specific class out of the type_url suffix is not useful, however, it will be useful to be able to use that suffix to discern heterogenous types (when it is not possible to have a common Java super type that Jackson understands for all different types of event in one topic).

When serializing we can use the class name for this suffix, but since knowing the fully qualified name of a class in the service introduces tight coupling, it would make more sense if this was something symbolic/defined by the user where serialization happens.

Since we know that a handler on an action is for consuming/publishing to a topic and is specified to use `com.protobuf.Any` we can generate some hints about the explicit transformation api in the generated code for such action handlers to guide users to get it right.

Since the TCK relies on projections to test JSON support, I think we'll need to change the TCK to instead just call a few topic-subscriber methods directly with json to verify that it works.

@pvlugter, @retgits would be good with your feedback here if I'm missing something important around the previous json support.

tck/src/main/java/com/akkaserverless/javasdk/tck/model/localpersistenceeventing/LocalPersistenceSubscriber.java
pvlugter: The explicit deserialisation looks good to me too. And it feels like it will be clearer that the protobuf rpc method and the action implementation method both take Any.

sdk/src/main/java/com/akkaserverless/javasdk/action/ActionContext.java
johanandren: I actually think the any support state is not needed, I'll make it a utility helper instead.

johanandren: Done, now in JsonSupport.encode/decode

johanandren:
Ready for final review.

codegen/java-gen/src/main/scala/com/lightbend/akkasls/codegen/java/ActionServiceSourceGenerator.scala
```diff
@@ -114,10 +114,21 @@ object ActionServiceSourceGenerator {
       val outputType = cmd.outputType.fullName
 
       if (isUnary(cmd)) {
+        val jsonTopicHint = {
+          // note: the somewhat funky indenting is on purpose to lf+indent only if comment present
+          if (cmd.inFromTopic && cmd.inputType.fullQualifiedName == "com.google.protobuf.Any")
+            """|// JSON input from a topic can be decoded using JsonSupport.decodeJson(MyClass.class, any)
+               |  """.stripMargin
```
patriknw, line 121: 👍

sdk/src/main/java/com/akkaserverless/javasdk/JsonSupport.java
```diff
@@ -0,0 +1,118 @@
...
+import com.google.protobuf.UnsafeByteOperations;
+
+import java.io.IOException;
+import java.util.Optional;
+
+public final class JsonSupport {
+
+  public static final String AKKA_SERVERLESS_JSON = "json.akkaserverless.com/";
+
+  private static final ObjectMapper objectMapper = new ObjectMapper();
```
patriknw, line 32: oh, we should probably have a better configured ObjectMapper by default. Maybe even use some subset of Akka's configuration mechanism for Jackson. I'll create an issue.

sdk/src/main/java/com/akkaserverless/javasdk/JsonSupport.java
```diff
@@ -0,0 +1,118 @@
...
+   *     JSON, useful for example when multiple different objects are passed through a pub/sub
+   *     topic.
+   * @throws IllegalArgumentException if the given value cannot be turned into JSON
+   */
+  public static <T> Any encodeJson(T value, String jsonType) {
+    try {
+      ByteString json = UnsafeByteOperations.unsafeWrap(objectMapper.writeValueAsBytes(value));
+      return Any.newBuilder().setTypeUrl(AKKA_SERVERLESS_JSON + jsonType).setValue(json).build();
+    } catch (JsonProcessingException ex) {
+      throw new IllegalArgumentException("Could not encode [" + value.getClass() + "] as JSON", ex);
```
patriknw, line 68: `value.getClass().getName()` ?

tck/src/test/java/com/akkaserverless/javasdk/tck/RunTck.java
```diff
@@ -38,9 +38,6 @@ public static void main(String[] args) throws Exception {
     try {
       new GenericContainer<>(DockerImageName.parse(TCK_IMAGE).withTag(TCK_VERSION))
           .withEnv("TCK_SERVICE_HOST", "host.testcontainers.internal")
-          .withCommand(
-              "-Dakkaserverless.tck.ignore-tests.0=eventing") // FIXME disabled because of lack of
-          // JSON support
```
patriknw, line 43: !

patriknw (approved):
LGTM, something minor...

Issue #262, opened 2021-09-01
Final cleanup


patriknw:
Once the new api and codegen for ReplicatedEntity and Actions have been completed we should make sure to cleanup remaining obsolete stuff, such as:
- lowlevel package
- EntityFactory
- `@EntityId`, `@AkkaServerlessAnnotation`
- reply package
- ReflectionHelper

Issue #83, opened 2021-06-15
New API and codegen for Replicated Entities


ennru:
We're reviewing and simplifying the APIs for replicated entities 
https://github.com/lightbend/akkaserverless-framework/issues/698
to make them available for users during the open beta phase.

How can we support using Replicated Entities with codegen?

pvlugter:
This will be switched to the new approach first, without annotations or reflection.

This will also need codegen annotations added to the protocol in framework, including a way to specify the replicated data types (with key/value types where appropriate).

patriknw:
> including a way to specify the replicated data types (with key/value types where appropriate)

that might become rather verbose? is there a way to require that is specified in the concrete ReplicatedEntity class instead? the downside of that suggestion might be that the first generated class will not compile when those types are unknown

pvlugter:
> > including a way to specify the replicated data types (with key/value types where appropriate)
> 
> that might become rather verbose? is there a way to require that is specified in the concrete ReplicatedEntity class instead? the downside of that suggestion might be that the first generated class will not compile when those types are unknown

Yeah, the alternative would be generate something without the types. Could be confusing either way — specifying the extra types, or having something that only partially generates.

So I was expecting the code to have the replicated data type: so that it's `ReplicatedEntity<ReplicatedCounter>` or for shopping cart map just as a test example might be `ReplicatedEntity<ReplicatedRegisterMap<String, ShoppingCartDomain.LineItem>>`.

So where the codegen annotations have this for Event Sourced Entity:

```
option (akkaserverless.file).event_sourced_entity = {
  name: "ShoppingCart"
  entity_type: "eventsourced-shopping-cart"
  state: "Cart"
  events: ["ItemAdded", "ItemRemoved"]
};
```

That could maybe be something like:

```
option (akkaserverless.file).replicated_entity = {
  name: "ShoppingCart"
  entity_type: "replicated-shopping-cart"
  replicated_register_map: {
    key: "String"
    value: "LineItem"
  }
};
```

patriknw:
That looks pretty good. I agree that it's good to refer to the proto value (LineItem) like that, since that is similar to how events and state are defined.

pvlugter:
Not sure what the replicated data types that don't need any extra settings would be: `replicated_counter: {}` or `replicated_counter: ""` or something is a bit awkward. Could also do something where there are optional key and value for when these are used, with an enum for the data type. Enums are required by convention to have prefix though. Then it would be:

[...]

PR #270, opened 2021-09-03
New API approach for replicated entity

pvlugter:
Refs #83. Add the new non-annotation based approach for replicated entity, bringing things in line with the other entities, including tests.

Update the TCK to use the new approach — this still has the one entity that can have different types of replicated data, based on the entity id. It could be useful to update the TCK replicated entity tests to require separate, specialised entities for each data type.

Update the replicated entity counter sample to use the new approach. There's no codegen yet (and we need the annotations in the protocol first) but it's set up with the to-be-generated files manually created. Enabled in CI again.

Also clean up most of the now obsolete code for #262. `@AkkaServerlessAnnotation` is still used for other annotations (like `@Jsonable` and test annotations). Otherwise the unused code should be removed now.


sdk/src/test/java/com/akkaserverless/javasdk/replicatedentity/CartEntity.java
```diff
@@ -0,0 +1,105 @@
...
+
+  @Override
+  public Effect<Empty> addItem(
+      ReplicatedRegisterMap<String, ShoppingCartDomain.LineItem> currentData,
+      ShoppingCartApi.AddLineItem addLineItem) {
+    if (addLineItem.getQuantity() <= 0) {
+      return effects().error("Cannot add negative quantity to item " + addLineItem.getProductId());
+    }
+
+    currentData.setValue(addLineItem.getProductId(), updateItem(addLineItem, currentData));
```
johanandren, line 50: I didn't follow the API discussions on replicated, so maybe I missed something, but I was a bit surprised that it is based on mutating the structure and not returning it as a part of the effect like for the value/es entities.

sdk/src/test/proto/replicatedentity/shoppingcart/shoppingcart_api.proto
```diff
@@ -0,0 +1,82 @@
...
+  string product_id = 1;
+  string name = 2;
+  int32 quantity = 3;
+}
+
+message Cart {
+  repeated LineItem items = 1;
+}
+
+service ShoppingCartService {
```
johanandren, line 57: I guess for continuing with codegen there will be a `option (akkaserverless.service) = { type : ... };` here as well?

johanandren (approved):
LGTM with some questions

sdk/src/test/proto/replicatedentity/shoppingcart/shoppingcart_api.proto
johanandren: Ah, now I saw the "we need the annotations in the protocol first". NVM!

sdk/src/main/scala/com/akkaserverless/javasdk/impl/replicatedentity/ReplicatedEntitiesImpl.scala
```diff
@@ -0,0 +1,267 @@
...
+    Flow[ReplicatedEntityStreamIn]
+      .mapConcat { in =>
+        in.message match {
+          case In.Command(command) =>
+            List(runner.handleCommand(command))
+          case In.Delta(delta) =>
+            runner.handleDelta(delta)
+            Nil
+          case In.Delete(_) =>
+            // ???
```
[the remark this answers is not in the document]
pvlugter: This is from the original:

https://github.com/lightbend/akkaserverless-java-sdk/blob/ab151acf8c3014384f9a6358d728c644c99bcf36/sdk/src/main/scala/com/akkaserverless/javasdk/impl/replicatedentity/ReplicatedEntityImpl.scala#L123-L125

I haven't checked this yet, but on the proxy side it should be stopping the entity in the user function by closing the stream. There's probably never this incoming `Delete` message. I can tidy up the protocol when making some other changes on the framework side.

sdk/src/test/java/com/akkaserverless/javasdk/replicatedentity/CartEntity.java
pvlugter: Right, I haven't changed that part of the API in this PR. Just removing the annotation/reflection/context-based approach for the entity as a whole. No discussions yet, as far as I know, around what an immutable or effect-based API for the CRDTs themselves should be.

We could make the replicated data immutable, where the new structure is passed through to an update effect; or we could have commands for effects (specific to the replicated data type), like `Increment` or `SetValue`.

Is there a particular approach that you're thinking of already?

sdk/src/test/proto/replicatedentity/shoppingcart/shoppingcart_api.proto
pvlugter: There are some commented out annotations in the replicated entity counter sample. And some discussion in #83.

sdk/src/test/java/com/akkaserverless/javasdk/replicatedentity/CartEntity.java
johanandren: I think a "managed" immutable data structure would better match the value/es entities and be less confusing vs a mutable object passed to the method to mutate to achieve the change. Even with a mutable structure I think having an effect saying "update this" could make sense for clarity.

pvlugter: Yeah, the immutable data structure would be a step towards something more similar.

Another approach would be to rework things completely, so rather than having a replicated entity that contains replicated data object, there are specialised replicated entities. It's now `ReplicatedEntity<ReplicatedCounter>`, while it could as well be `ReplicatedCounterEntity`. So currently, an immutable-style API would be along the lines of:

```java
class MyCounter extends AbstractCounter { // extends ReplicatedEntity<ReplicatedCounter>
  @Override public ReplicatedCounter emptyData(ReplicatedDataFactory factory) {
    return factory.newCounter();
  }

  @Override public Effect<Response> someMethod(ReplicatedCounter counter, Request request) {
    return effects()
        .update(counter.increment(request.getValue()))
        .thenReply(Response.newBuilder...);
  }
}
```

While something that's a specialised entity for each replicated data type could instead be closer to event sourced or value entities:

```java
class MyCounter extends AbstractCounter { // extends ReplicatedCounterEntity
  // no need for empty data — counter empty state is 0 already
  // register can have an empty state
  
  // pass in the current replicated data value — Long, Set, Map, etc
  @Override public Effect<Response> someMethod(long currentValue, Request request) {
    return effects()
        .increment(request.getValue()) // specific effects for each replicated data type
        .thenReply(Response.newBuilder...);
  }
}
```

johanandren: I think that looks very nice with the specialized entity for the data type.

samples/replicatedentity-counter/src/main/java/com/example/counter/domain/Counter.java
```diff
@@ -0,0 +1,47 @@
...
+public class Counter extends AbstractCounter {
+  @SuppressWarnings("unused")
+  private final String entityId;
+
+  public Counter(ReplicatedEntityContext context) {
+    this.entityId = context.entityId();
+  }
+
+  @Override
+  public ReplicatedCounter emptyData(ReplicatedDataFactory factory) {
```
patriknw, line 16: that's a nice way to give access to the factory

samples/replicatedentity-counter/src/main/java/com/example/counter/domain/Counter.java
```diff
@@ -0,0 +1,47 @@
...
+
+    return effects().reply(Empty.getDefaultInstance());
+  }
+
+  @Override
+  public Effect<CounterApi.CurrentCounter> getCurrentCounter(ReplicatedCounter counter, CounterApi.GetCounter command) {
+    CounterApi.CurrentCounter current = CounterApi.CurrentCounter.newBuilder().setValue(counter.getValue()).build();
+    return effects().reply(current);
+  }
+}
```
patriknw, line 47: this looks nice!

sdk/src/test/java/com/akkaserverless/javasdk/replicatedentity/CartEntity.java
patriknw: That looks good, but how much do we have to duplicate/specialize in the different parts of codegen and abstract classes?

The immutable + effects.update isn't too bad either.

patriknw (commented):
LGTM, might be best to merge this first and continue with the alternative API in separate PR?

sdk/src/test/java/com/akkaserverless/javasdk/replicatedentity/CartEntity.java
pvlugter: > how much do we have to duplicate/specialize in the different parts of codegen and abstract classes

I think this will be similar for any of the approaches, where codegen will need to be specific to the replicated data type. It would just be the difference between using the right type parameter or the right base class.

I'll move the API discussions for the replicated data level over to #83.

pvlugter:
I'll merge this first step, removing annotation-based approach at the entity-level. Then continue with discussion on changing the replicated data APIs in #83, and changes in a new PR.

PR #272, opened 2021-09-03
fix: regenerate from Maven archetype

ennru:
This updates the Maven archetype generated sources to the current state of the Java SDK.
(They are currently out of sync as we improved codegen to consistently use the `Reply` type.)


johanandren (approved):

ennru:
And yes, I'm aware the Java SDK and codegen is about to fundamentally change in September 2021.

Issue #223, opened 2021-08-20
Update samples to use the new codgen


patriknw:
https://github.com/lightbend/akkaserverless-java-sdk/tree/sdk-codegen-dev/samples:
- [x] #302
- [x] #285
- [x] #277
- [x] #278
- [x] #276
- [x] #273
- [x] replicatedentity-counter
- [x] valueentity-counter

lightbend-labs:
- [ ] #366 
- [ ] #328
- [ ] https://github.com/lightbend-labs/akkaserverless-java-car-metrics

performance testing:
- [x] #304 

Some of those are also used by the documentation, see https://github.com/lightbend/akkaserverless-java-sdk/issues/222

patriknw:
I think we'll close this. We can help with remaining samples (if any) if requested.

Issue #222, opened 2021-08-20
Update documentation to use the new codgen


patriknw:
- [x] actions.adoc (@octonato)
- [x] api.adoc (@patriknw) 
- [x] entity-eventing.adoc (@octonato) - replaced by "Publishing and Subscribing with Actions"
- [x] eventsourced.adoc (@patriknw)
- [x] forwarding.adoc (@octonato) - replaced by "Actions as Controllers" and "Running Side Effects"
- [x] kickstart.adoc (@patriknw)
- [x] proto.adoc (@patriknw)
- [x] serialization.adoc (@patriknw)
- [x] topic-eventing.adoc (@octonato) - replaced by "Publishing and Subscribing with Actions"
- [x] value-entity.adoc (@patriknw)
- [x] views.adoc (@patriknw)
- [x] development-process-java.adoc (@patriknw)
- [x] archetype / first-service (@patriknw)
- [x] Quickstart https://github.com/lightbend/akkaserverless-docs/tree/master/docs/modules/quickstart/pages (@patriknw)

See https://github.com/lightbend/akkaserverless-java-sdk/issues/52

octonato:
All subtasks are completed, closing this.

PR #281, opened 2021-09-06
Update value-entity.adoc and valueentity-counter sample, #222

patriknw:
Refs #222
Refs #223

### DIFF
--- a/docs/src/modules/java/pages/value-entity.adoc
+++ b/docs/src/modules/java/pages/value-entity.adoc
@@ -1,11 +1,9 @@
 = Implementing Value Entities in Java
 include::ROOT:partial$include.adoc[]
 
-https://developer.lightbend.com/docs/akka-serverless/reference/glossary.html#value_entity[Value Entities] persist state on every change and Akka Serverless needs to serialize that data to send it to the underlying data store. The most straightforward way to persist the state is with Protocol Buffers using `protobuf` types. Akka Serverless automatically detects if an updated state is `protobuf`, and serializes it using `protobuf`. For other serialization options, including JSON, see xref:java:serialization.adoc[Serialization options for Java services].
+https://developer.lightbend.com/docs/akka-serverless/reference/glossary.html#value_entity[Value Entities] persist state on every change and Akka Serverless needs to serialize that data to send it to the underlying data store. The most straightforward way to persist the state is with Protocol Buffers using `protobuf` types.
 
-While Protocol Buffers are the recommended format for persisting state, we recommend that you do not persist your service's `protobuf` messages. While this may introduce some overhead to convert from one type to the other, it allows the service public interface logic to evolve independently of the data storage format, which should be private.
-
-NOTE: For value entities, modify the state and then trigger a save of that state. Both the JavaScript and the Java SDK have a `context.updateState(newState)` method for this purpose. If you change the state but do not call `updateState`, that state change is lost.
+While Protocol Buffers are the recommended format for persisting state, we recommend that you do not persist your service's public `protobuf` messages. While this may introduce some overhead to convert from one type to the other, it allows the service public interface logic to evolve independently of the data storage format, which should be private.
 
 The steps necessary to implement a Value Entity include:
 
@@ -33,8 +31,8 @@ include::example$valueentity-counter/src/main/proto/value-entities/counter_domai
 <3> Let the messages declared in this protobuf file be inner classes to the Java class `CounterDomain`.
 <4> The protobuf option `(akkaserverless.file).value_entity` is specific to code-generation as provided by the Akka Serverless Maven plugin.
 <5> `name` denotes the base name for the Value entity, the code-generation will create initial sources `Counter`, `CounterTest` and `CounterIntegrationTest`. Once these files exist, they are not overwritten, so you can freely add logic to them.
-<6> `enity_type` is a unique identifier of the "state storage". The entity name may be changed even after data has been created, the `entity_type` can't. This value shows in the `@ValueEnity` annotation of your entity implementation.
-<7> `state` points to the protobuf message representing the Value entity's state which is kept by Akka Serverless
+<6> `enity_type` is a unique identifier of the "state storage". The entity name may be changed even after data has been created, the `entity_type` can't.
+<7> `state` points to the protobuf message representing the Value entity's state which is kept by Akka Serverless.
 <8> The `CounterState` protobuf message is what Akka Serverless stores for this entity.
 
 The `counter_api.proto` file defines the commands we can send to the Counter service to manipulate or access the Counter`s state. They make up the service API:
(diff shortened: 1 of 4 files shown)


docs/src/modules/java/pages/value-entity.adoc
```diff
@@ -1,11 +1,9 @@
...
 
-While Protocol Buffers are the recommended format for persisting state, we recommend that you do not persist your service's `protobuf` messages. While this may introduce some overhead to convert from one type to the other, it allows the service public interface logic to evolve independently of the data storage format, which should be private.
-
-NOTE: For value entities, modify the state and then trigger a save of that state. Both the JavaScript and the Java SDK have a `context.updateState(newState)` method for this purpose. If you change the state but do not call `updateState`, that state change is lost.
+While Protocol Buffers are the recommended format for persisting state, we recommend that you do not persist your service's public `protobuf` messages. While this may introduce some overhead to convert from one type to the other, it allows the service public interface logic to evolve independently of the data storage format, which should be private.
```
johanandren, line 6:
```suggestion
While Protocol Buffers are the recommended format for persisting state, we recommend that you do not persist your service's public `protobuf` messages. This may introduce some overhead to convert from one type to the other, but allows the service public interface logic to evolve independently of the data storage format, which should be private.
```

docs/src/modules/java/pages/value-entity.adoc
```diff
@@ -33,8 +31,8 @@ include::example$valueentity-counter/src/main/proto/value-entities/counter_domai
 <3> Let the messages declared in this protobuf file be inner classes to the Java class `CounterDomain`.
 <4> The protobuf option `(akkaserverless.file).value_entity` is specific to code-generation as provided by the Akka Serverless Maven plugin.
 <5> `name` denotes the base name for the Value entity, the code-generation will create initial sources `Counter`, `CounterTest` and `CounterIntegrationTest`. Once these files exist, they are not overwritten, so you can freely add logic to them.
-<6> `enity_type` is a unique identifier of the "state storage". The entity name may be changed even after data has been created, the `entity_type` can't. This value shows in the `@ValueEnity` annotation of your entity implementation.
-<7> `state` points to the protobuf message representing the Value entity's state which is kept by Akka Serverless
+<6> `enity_type` is a unique identifier of the "state storage". The entity name may be changed even after data has been created, the `entity_type` can't.
```
johanandren, line 34:
```suggestion
<6> `entity_type` is a unique identifier of the "state storage". The entity name may be changed even after data has been created, the `entity_type` can't.
```

docs/src/modules/java/pages/value-entity.adoc
```diff
@@ -55,35 +53,38 @@ include::example$valueentity-counter/src/main/proto/value-entities/counter_api.p
...
-
-As we rely on the code-generation, the class `Counter` gets generated for us, it extends the generated class `CounterInterface` which we're not supposed to change as it gets regenerated in case we update the protobuf descriptors.
+A Value Entity implementation is a Java class where you define how each command is implemented. The class `Counter` gets generated for us based on the `counter_api.proto` and `counter_domain.proto` definitions. Once the `Counter.java` file exist, it is not overwritten, so you can freely add logic to it. `Counter` extends the generated class `AbstractCounter` which we're not supposed to change as it gets regenerated in case we update the protobuf descriptors. AbstractCounter` contains all method signatures corresponding to the API of the service. If you change the API you will see compilation errors in the `Counter` class and have to implement the methods required by `AbstractCounter`.
```
johanandren, line 56:
```suggestion
A Value Entity implementation is a Java class where you define how each command is implemented. The class `Counter` gets generated for us based on the `counter_api.proto` and `counter_domain.proto` definitions. Once the `Counter.java` file exist, it is not overwritten, so you can freely add logic to it. `Counter` extends the generated class `AbstractCounter` which we're not supposed to change as it gets regenerated in case we update the protobuf descriptors. `AbstractCounter` contains all method signatures corresponding to the API of the service. If you change the API you will see compilation errors in the `Counter` class and have to implement the methods required by `AbstractCounter`.
```

docs/src/modules/java/pages/value-entity.adoc
```diff
@@ -55,35 +53,38 @@ include::example$valueentity-counter/src/main/proto/value-entities/counter_api.p
...
 
 [source,java]
 .src/main/java/com/example/domain/Counter.java
 ----
 include::java:example$valueentity-counter/src/main/java/com/example/domain/Counter.java[tag=class]
 ----
-<1> The constructor accepts a parameter marked as the `@EntityId` and we keep the value in a field.
+<1> Extends the generated `AbstractCounter`, which extends link:{attachmentsdir}/api/com/akkaserverless/javasdk/valueentity/ValueEntity.html[`ValueEntity` {tab-icon}, window="new"].
+<2> By default, the generated constructor has a `ValueEntityContext` parameter, but you can change this to accept other parameters.
```
johanandren, line 64: I think that if we should mention this, we should also mention that you'd then have to inject them when handing off the constructor reference in `Main`, but that would be noisy/long to describe here, so perhaps that should be something detailed in a later section?

Or we don't mention at all and clever users will figure that out...

docs/src/modules/java/pages/value-entity.adoc
```diff
@@ -95,15 +96,19 @@ The following example shows the implementation of the `GetCurrentCounter` comman
 include::java:example$valueentity-counter/src/main/java/com/example/domain/Counter.java[tag=getCurrentCounter]
 ----
 <1> The current state is passed to the method
-<2> We use its value to create the `CurrentCounter` return value.
+<2> We use its value to create the `CurrentCounter`  value that is sent as a reply by returning an `Effect` with `effects().reply`.
```
johanandren, line 99:
```suggestion
<2> We use its value to create the `CurrentCounter` value that is sent as a reply by returning an `Effect` with `effects().reply`.
```

samples/valueentity-counter/pom.xml
```diff
@@ -19,7 +19,7 @@
     <jdk.target>11</jdk.target>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
 
-    <akkaserverless-sdk.version>0.7.0-beta.18</akkaserverless-sdk.version>
+    <akkaserverless-sdk.version>0.7.0-beta.18-102-aceb450c-SNAPSHOT</akkaserverless-sdk.version>
```
johanandren, line 22:
```suggestion
    <akkaserverless-sdk.version>0.7.0-beta.18</akkaserverless-sdk.version>
```

johanandren (approved):
Mostly LGTM

docs/src/modules/java/pages/value-entity.adoc
patriknw: It's mentioned later where Main is described. We could skip it here and only have it in the Main description? Ok to have it at both places, somewhat brief here? I don't think this should forward reference to Main.

johanandren: I'd vote for skip here.

patriknw: done: https://github.com/lightbend/akkaserverless-java-sdk/pull/281/commits/a66a34a030f5e9149ddac7f8c650528dbd1eaca4

johanandren (approved):

PR #287, opened 2021-09-07
Add required update effect for replicated entities

pvlugter:
This is an in-between step for #83. Adds a required `update` effect for replicated entities, but keeps the current mutable APIs for replicated data. It does this by making a (deep) copy of the replicated data before passing to the handler method, so that the entity is only updated when `effects().update(data)` is used, otherwise it will still have the old data without changes.

It's probably a little awkward mixing approaches, but it's simpler than converting the data structures to immutable straight away, and sets things up so that could be done next — or add the data-type-specific effects under the specialised entity classes, with the underlying replicated data as mutable but the user-facing interface as read-only. And other than the TCK, there's no test coverage for the replicated data structures themselves.

### DIFF
--- a/samples/replicatedentity-counter/src/main/java/com/example/counter/domain/Counter.java
+++ b/samples/replicatedentity-counter/src/main/java/com/example/counter/domain/Counter.java
@@ -13,11 +13,6 @@ public class Counter extends AbstractCounter {
   }
 
   @Override
-  public ReplicatedCounter emptyData(ReplicatedDataFactory factory) {
-    return factory.newCounter();
-  }
-
-  @Override
   public Effect<Empty> increase(ReplicatedCounter counter, CounterApi.IncreaseValue command) {
     if (command.getValue() < 0) {
       return effects().error("Increase requires a positive value. It was [" + command.getValue() + "].");
@@ -25,7 +20,7 @@ public class Counter extends AbstractCounter {
 
     counter.increment(command.getValue());
 
-    return effects().reply(Empty.getDefaultInstance());
+    return effects().update(counter).thenReply(Empty.getDefaultInstance());
   }
 
   @Override
@@ -36,7 +31,7 @@ public class Counter extends AbstractCounter {
 
     counter.decrement(command.getValue());
 
-    return effects().reply(Empty.getDefaultInstance());
+    return effects().update(counter).thenReply(Empty.getDefaultInstance());
   }
 
   @Override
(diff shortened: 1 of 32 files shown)


samples/replicatedentity-counter/src/main/java/com/example/counter/domain/Counter.java
```diff
@@ -12,11 +12,6 @@ public Counter(ReplicatedEntityContext context) {
     this.entityId = context.entityId();
   }
 
-  @Override
-  public ReplicatedCounter emptyData(ReplicatedDataFactory factory) {
-    return factory.newCounter();
```
pvlugter, line 17: Simpler with default empty data now provided by the specialised entity classes. Only the register will require an empty value now.

samples/replicatedentity-counter/src/main/java/com/example/counter/domain/Counter.java
```diff
@@ -25,7 +20,7 @@ public ReplicatedCounter emptyData(ReplicatedDataFactory factory) {
 
     counter.increment(command.getValue());
 
-    return effects().reply(Empty.getDefaultInstance());
+    return effects().update(counter).thenReply(Empty.getDefaultInstance());
```
pvlugter, line 23: Counter will only be updated if passed to `update`

sdk/src/main/scala/com/akkaserverless/javasdk/impl/Contexts.scala
```diff
@@ -17,14 +17,9 @@
...
 
 private[impl] trait ActivatableContext extends Context {
   private final var active = true
   final def deactivate(): Unit = active = false
   final def checkActive(): Unit = if (!active) throw new IllegalStateException("Context no longer active!")
 }
-
-object FailInvoked extends RuntimeException with NoStackTrace {
-  override def toString: String = "CommandContext.fail(…) invoked"
-}
```
pvlugter, line 30: Another leftover that could be removed now that context.fail doesn't exist

johanandren (approved):
LGTM

sdk/src/main/java/com/akkaserverless/javasdk/replicatedentity/ReplicatedRegisterEntity.java
```diff
@@ -0,0 +1,31 @@
...
+
+package com.akkaserverless.javasdk.replicatedentity;
+
+public abstract class ReplicatedRegisterEntity<T> extends ReplicatedEntity<ReplicatedRegister<T>> {
+  /**
+   * Implement to set the default empty value for the register.
+   *
+   * @return the default empty value
+   */
+  public abstract T emptyValue();
```
patriknw, line 25: clever

patriknw (approved):
LGTM

PR #291, opened 2021-09-07
fix: view 'implementation' is generated but not managed

raboof:
Add the right comment


johanandren (approved):

johanandren:
Will need fix(es) in the codegen tests as well.

patriknw (approved):
LGTM

PR #293, opened 2021-09-07
update eventsourced.adoc to new API and codgen, #222

patriknw:
* align with value-entity.adoc
* switch to using the java-eventsourced-shopping-cart
  instead of the java-eventing-shopping-cart
  
Refs #222

### DIFF
--- a/docs/src/modules/java/pages/eventsourced.adoc
+++ b/docs/src/modules/java/pages/eventsourced.adoc
@@ -69,7 +69,7 @@ We need to implement all methods our Event Sourced Entity offers as https://deve
 
 The code-generation will generate an implementation class with an initial empty implementation which we'll discuss below.
 
-Command handlers show in the implementation class as methods that override abstract methods from `AbstractShoppingCart`. The methods take the current state as the first parameter and the request message as the second parameter. They return an `Effect`, which describes next processing actions, such as emitting events and sending a reply.
+Command handlers are implemented in the implementation class as methods that override abstract methods from `AbstractShoppingCart`. The methods take the current state as the first parameter and the request message as the second parameter. They return an `Effect`, which describes next processing actions, such as emitting events and sending a reply.
 
 === Updating state
 
@@ -120,7 +120,7 @@ From the code-generation, the registration gets automatically inserted in the ge
 include::example$java-eventsourced-shopping-cart/src/main/java/com/example/shoppingcart/Main.java[]
 ----
 
-By default, the generated constructor has a `EventSourcedEntityContext` parameter, but you can change this to accept other parameters. If you change the constructor of the `ShoppingCart` class you will see a compilation error here, and you have to adjust the factory function that is passed to `AkkaServerlessFactory.withComponents`.
+By default, the generated constructor has an `EventSourcedEntityContext` parameter, but you can change this to accept other parameters. If you change the constructor of the `ShoppingCart` class you will see a compilation error here, and you have to adjust the factory function that is passed to `AkkaServerlessFactory.withComponents`.
 
 When more components are added the `AkkaServerlessFactory` is regenerated and you have to adjust the registration from the `Main` class.
 
--- a/docs/src/modules/java/pages/eventsourced.adoc
+++ b/docs/src/modules/java/pages/eventsourced.adoc
@@ -3,108 +3,137 @@
 include::ROOT:partial$include.adoc[]
 include::partial$eventsourced.adoc[]
 
-== Creating an Event Sourced Entity
+https://developer.lightbend.com/docs/akka-serverless/reference/glossary.html#event_sourced_entity[Event Sourced Entities] persist changes as events and snapshots. Akka Serverless needs to serialize that data to send it to the underlying data store. The most straightforward way to persist the state is with Protocol Buffers using `protobuf` types.
 
-Create an Event Sourced Entity by annotating it with the link:{attachmentsdir}/api/com/akkaserverless/javasdk/eventsourcedentity/EventSourcedEntity.html[`@EventSourcedEntity` {tab-icon}, window="new"] annotation.
+While Protocol Buffers are the recommended format for persisting state, we recommend that you do not persist your service's public `protobuf` messages. This may introduce some overhead to convert from one type to the other, but allows the service public interface logic to evolve independently of the data storage format, which should be private.
+
+The steps necessary to implement an Event Sourced Entity include:
+
+. Defining the API and domain objects in `.proto` files.
+. Implementing behavior in command and event handlers.
+. Creating and initializing the Entity.
+
+
+The sections on this page walk through these steps using a shopping cart service as an example.
+
+== Defining the `proto` files
+
+NOTE: Our Event Sourced Entity example is a shopping cart service.
+
+The following `shoppingcart_domain.proto` file defines our "Shopping" Event Sourced Entity. The entity manages line items of a cart and stores events `ItemAdded` and `ItemRemoved` to represent changes to the cart. Real world entities store much more data -- often structured data --  they represent an _Entity_ in the https://en.wikipedia.org/wiki/Domain-driven_design[domain-driven design] sense of the term.
+
+[source,proto]
+.src/main/proto/shoppingcart/shoppingcart_domain.proto
+----
+include::example$java-eventsourced-shopping-cart/src/main/proto/shoppingcart/shoppingcart_domain.proto[]
+----
+
+<1> Any classes generated from this protobuf file will be in the Java package `com.example.shoppingcart.domain`.
+<2> Import the Akka Serverless protobuf annotations, or options.
+<3> Let the messages declared in this protobuf file be inner classes to the Java class `ShoppingCartDomain`.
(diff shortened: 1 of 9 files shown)


docs/src/modules/java/pages/eventsourced.adoc
```diff
@@ -3,108 +3,137 @@
 include::ROOT:partial$include.adoc[]
 include::partial$eventsourced.adoc[]
 
-== Creating an Event Sourced Entity
+https://developer.lightbend.com/docs/akka-serverless/reference/glossary.html#event_sourced_entity[Event Sourced Entities] persist changes as events and snapshots. Akka Serverless needs to serialize that data to send it to the underlying data store. The most straightforward way to persist the state is with Protocol Buffers using `protobuf` types.
```
johanandren, line 6: At least the only way we should talk about in the docs, so maybe?
```suggestion
https://developer.lightbend.com/docs/akka-serverless/reference/glossary.html#event_sourced_entity[Event Sourced Entities] persist changes as events and snapshots. Akka Serverless needs to serialize that data to send it to the underlying data store, this is done with Protocol Buffers using `protobuf` types.
```

docs/src/modules/java/pages/eventsourced.adoc
```diff
@@ -3,108 +3,137 @@
...
 ----
+<1> Extends the generated `AbstractShoppingCart`, which extends link:{attachmentsdir}/api/com/akkaserverless/javasdk/eventsourcedentity/EventSourcedEntity.html[`EventSourcedEntity` {tab-icon}, window="new"].
+<2> Defines the initial, empty, state that is used before any updates.
+
+We need to implement all methods our Event Sourced Entity offers as https://developer.lightbend.com/docs/akka-serverless/reference/glossary.html#command_handler[_command handlers_].
+
+The code-generation will generate an implementation class with an initial empty implementation which we'll discuss below.
 
-== Handling commands
+Command handlers show in the implementation class as methods that override abstract methods from `AbstractShoppingCart`. The methods take the current state as the first parameter and the request message as the second parameter. They return an `Effect`, which describes next processing actions, such as emitting event and sending a reply.
```
johanandren, line 72:
```suggestion
Command handlers show in the implementation class as methods that override abstract methods from `AbstractShoppingCart`. The methods take the current state as the first parameter and the request message as the second parameter. They return an `Effect`, which describes next processing actions, such as emitting events and sending a reply.
```

docs/src/modules/java/pages/eventsourced.adoc
```diff
@@ -3,108 +3,137 @@
...
+=== Updating state
 
-The command handler also can take the gRPC service call input type as a parameter to receive the command message. This is optional. For example, our `GetCart` service call doesn't need any information from the message, since it's just returning the current state. Meanwhile, the `AddItem` service call does need information from the message, since it needs to know the product id, description and quantity to add to the cart. The return type of the command handler must be the output type for the gRPC service call, this will be sent as the reply.
+In the example below, the `AddItem` service call uses the request message `AddLineItem`. It returns an `Effect` to emit an event and send a reply. The state is updated by the event handler.
```
johanandren, line 76: Worth saying something about order of things?
```suggestion
In the example below, the `AddItem` service call uses the request message `AddLineItem`. It returns an `Effect` to emit an event and then send a reply once the event is stored successfully. The state is updated by the event handler.
```

docs/src/modules/java/pages/eventsourced.adoc
```diff
@@ -3,108 +3,137 @@
...
+== Snapshots
 
-== Registering the Entity
+Snapshots are an important optimization for Event Sourced Entities that emit many events. Rather than reading the entire journal upon loading or restart, Akka Serverless can initiate them from a snapshot.
 
-Once you've created your Entity, you can register it with the link:{attachmentsdir}/api/com/akkaserverless/javasdk/AkkaServerless.html[`AkkaServerless` {tab-icon}, window="new"] server by invoking the `registerEventSourcedEntity` method. In addition to passing your Event Sourced Entity class and protobuf service descriptor, you also need to pass any descriptors that you use for persisting events, for example, `ShoppingCart` as defined in the `shopping-cart-domain.proto` descriptor.
+Snapshots are stored and handled automatically by Akka Serverless without any specific code required. Snapshots are stored after a configured number of events:
```
johanandren, line 131: Should we mention the actual default value? (every 100 events)

Here we describe the config based way which is global to all event sourced entities in the project, should we mention the programmatic/per entity type way as well? (Also, should we recommend one over the other?)

johanandren (approved):
LGTM but with some ideas/opinions

docs/src/modules/java/pages/eventsourced.adoc
patriknw: I actually implemented a snippet for the programatic but then though it was much easier to use config. Since an Akka Serverless Service typically has only one entity type the config should work in most cases.

> Should we mention the actual default value? (every 100 events)

I'm a bit allergic to mentioning default values in docs. The snippet there shows the config with 100.

If we want more in depth reference docs we should show (parts of) reference.conf.

docs/src/modules/java/pages/eventsourced.adoc
```diff
@@ -3,108 +3,137 @@
...
-[source,java,indent=0]
+[source,java]
+.src/main/java/com/example/shoppingcart/Main.java
 ----
-include::example$java-eventing-shopping-cart/src/main/java/shopping/cart/domain/ShoppingCartEntity.java[tag=itemAdded]
+include::example$java-eventsourced-shopping-cart/src/main/java/com/example/shoppingcart/Main.java[]
 ----
 
-== Snapshots
+By default, the generated constructor has a `EventSourcedEntityContext` parameter, but you can change this to accept other parameters. If you change the constructor of the `ShoppingCart` class you will see a compilation error here, and you have to adjust the factory function that is passed to `AkkaServerlessFactory.withComponents`.
```
pvlugter, line 123:
```suggestion
By default, the generated constructor has an `EventSourcedEntityContext` parameter, but you can change this to accept other parameters. If you change the constructor of the `ShoppingCart` class you will see a compilation error here, and you have to adjust the factory function that is passed to `AkkaServerlessFactory.withComponents`.
```

docs/src/modules/java/pages/value-entity.adoc
```diff
@@ -16,7 +16,7 @@ The sections on this page walk through these steps using a counter service as an
 
 == Defining the `proto` files
 
-NOTE: Our Value entity example starts with the "Counter" service as included in xref:kickstart.adoc[].
+NOTE: Our Value Entity example starts with the "Counter" service as included in xref:kickstart.adoc[].
```
pvlugter, line 19: 👍🏼 good to be consistent with case

samples/java-eventsourced-shopping-cart/src/main/resources/application.conf
```diff
@@ -0,0 +1,4 @@
+
+# //tag::snapshot-every[]
+akkaserverless.event-sourced-entity.event-sourced-entity.snapshot-every = 100
```
pvlugter, line 3: Double `event-sourced-entity`

```suggestion
akkaserverless.event-sourced-entity.snapshot-every = 100
```

docs/src/modules/java/pages/eventsourced.adoc
```diff
@@ -3,108 +3,137 @@
...
 ----
+<1> Extends the generated `AbstractShoppingCart`, which extends link:{attachmentsdir}/api/com/akkaserverless/javasdk/eventsourcedentity/EventSourcedEntity.html[`EventSourcedEntity` {tab-icon}, window="new"].
+<2> Defines the initial, empty, state that is used before any updates.
+
+We need to implement all methods our Event Sourced Entity offers as https://developer.lightbend.com/docs/akka-serverless/reference/glossary.html#command_handler[_command handlers_].
+
+The code-generation will generate an implementation class with an initial empty implementation which we'll discuss below.
 
-== Handling commands
+Command handlers show in the implementation class as methods that override abstract methods from `AbstractShoppingCart`. The methods take the current state as the first parameter and the request message as the second parameter. They return an `Effect`, which describes next processing actions, such as emitting events and sending a reply.
```
pvlugter, line 72: Not sure `show` in `Command handlers show in the implementation class as methods ...` reads that well. Maybe `are implemented` is more direct, even with `implementation class`. Or could be `are required` or similar.

```suggestion
Command handlers are implemented in the implementation class as methods that override abstract methods from `AbstractShoppingCart`. The methods take the current state as the first parameter and the request message as the second parameter. They return an `Effect`, which describes next processing actions, such as emitting events and sending a reply.
```

pvlugter (approved):
Looks good. One fix and a couple of small suggestions.

docs/src/modules/java/pages/eventsourced.adoc
patriknw: I agree, must be similar in value entity docs. "implemented in the implementation" is somewhat funny. I'll see if I can rephrase that.

docs/src/modules/java/pages/eventsourced.adoc
johanandren: If I put my user hat on, I think I'd find it somewhat important to know the the default, so perhaps worth pulling in reference.conf then (or at least mentioning where they can find the default value)?

PR #294, opened 2021-09-07
Enable it test for eventing sample

patriknw:


.circleci/config.yml
```diff
@@ -244,9 +244,9 @@ jobs:
           name: Eventing "Shopping Cart" sample
           command: |
             cd samples/java-eventing-shopping-cart
-            echo "Running mvn with SDK version: '$SDK_VERSION' - FIXME disabled for codegen/API work"
-            # TODO: enable integration tests with -Pit
-            # mvn -Dakkaserverless-sdk.version=$SDK_VERSION verify
+            echo "Running mvn with SDK version: '$SDK_VERSION'"
+            docker-compose -f .circleci/google-pubsub-emulator-docker-compose.yml up -d
```
patriknw, line 248: is this the wrong place for this?
there is a FIXME in `integration-tests`, but would that be active also for this step?

octonato: I don't think we need docker-compose here. The integration tests used test containers and is supposed to start whatever is needed.

patriknw: ok, then it's something missing in that integration test.

I couldn't run it locally either without starting the emulator on the side.

patriknw: I don't know anything about that so leaving that for someone else. Created issue https://github.com/lightbend/akkaserverless-java-sdk/issues/297

PR #300, opened 2021-09-08
Add codegen for replicated entities

### LINKED ISSUE #83: New API and codegen for Replicated Entities

pvlugter:
Resolves #83 (with follow up issues created for other parts of the discussion in that issue).

Add codegen for replicated entities. Use the codegen in the replicated entity counter sample now, and add a replicated entity shopping cart sample.


patriknw (approved):
LGTM, it's so nice to see how all components follow the same pattern now.

PR #303, opened 2021-09-08
Some minor fixes for quickstart example.

RayRoestenburg:
Refs #302


RayRoestenburg:
The sample was already working in sdk-codegen-dev, if anything is missing in doc, I'll run into that later or maybe build will fail..

raboof (approved):

RayRoestenburg:
(Just a comment, the sample is almost identical to the java-valueentity-customer-registry, though that specific examples adds some more views and an action, there is quite some duplication there.)

Issue #41, opened 2021-05-31
Use JUnit5 as default


ennru:
Update: we decided to not offer it as an option, but pick juni5.

(keeping previous description for the record)

--------------------------------------

Our Java SDK testkit supports both JUnit 4 and JUnit 5.
The current codegen generates JUnit 4 code.

We'd possibly need a configuration to allow selecting the desired JUnit style (and support for it in codegen).

octonato:
Should we really have this as an option? 

I think it's fine to use JUnit 5 by default. If users prefer to use JUnit4 they can change their dependencies and adapt.

ennru:
I agree with a JUnit 5 default.
The codegen and Maven archetype do JUnit 4 right now...

ennru:
As a first step, our examples should be switched to use JUnit 5.

Java/Protobuf SDK
- [ ] java-protobuf-customer-registry-kafka-quickstart
- [ ] java-protobuf-customer-registry-quickstart
- [ ] java-protobuf-customer-registry-views-quickstart
- [ ] java-protobuf-doc-snippets
- [ ] java-protobuf-eventsourced-counter
- [ ] java-protobuf-eventsourced-customer-registry
- [ ] java-protobuf-eventsourced-customer-registry-subscriber
- [ ] java-protobuf-eventsourced-shopping-cart
- [ ] java-protobuf-fibonacci-action
- [ ] java-protobuf-first-service
- [ ] java-protobuf-reliable-timers
- [ ] java-protobuf-replicatedentity-examples
- [ ] java-protobuf-replicatedentity-shopping-cart
- [ ] java-protobuf-shopping-cart-quickstart
- [ ] java-protobuf-valueentity-counter
- [ ] java-protobuf-valueentity-counter-spring-client
- [ ] java-protobuf-valueentity-customer-registry
- [ ] java-protobuf-valueentity-shopping-cart
- [ ] java-protobuf-view-store
- [ ] java-protobuf-web-resources

Java SDK
- [ ] java-spring-customer-registry-quickstart
- [ ] java-spring-customer-registry-views-quickstart
- [ ] java-spring-doc-snippets
- [ ] java-spring-eventsourced-counter
- [ ] java-spring-eventsourced-customer-registry
- [ ] java-spring-eventsourced-customer-registry-subscriber
- [ ] java-spring-eventsourced-shopping-cart
- [ ] java-spring-fibonacci-action
- [ ] java-spring-reliable-timers
[...]

PR #305, opened 2021-09-08
Update kickstart.adoc and pom files

patriknw:
Refs #222

### DIFF
--- a/docs/src/modules/java/pages/kickstart.adoc
+++ b/docs/src/modules/java/pages/kickstart.adoc
@@ -107,7 +107,7 @@ For more information on descriptors, see xref:java:proto.adoc[].
 
 === Component implementation
 
-For the default service description in the archetype, the plugin creates an abstract base class (eg. `CounterInterface`) which always reflects the latest service description.
+For the default service description in the archetype, the plugin creates an abstract base class (eg. `AbstractCounter`) which always reflects the latest service description.
 
 IMPORTANT: Do not modify the base class as it is regenerated on each invocation of `mvn compile`
 
@@ -133,7 +133,7 @@ For more details see xref:java:value-entity.adoc[].
 
 === Unit and integration tests
 
-The Akka Serverless Maven plugin creates a unit test stub for each class. Use this stub as a starting point to test the logic in your implementation. The Akka Serverless Java SDK test kit supports both link:https://junit.org/junit4/[JUnit 4] and link:https://junit.org/junit5/[JUnit 5].
+The Akka Serverless Maven plugin creates a unit test stub for Entity. Use this stub as a starting point to test the logic in your implementation. The Akka Serverless Java SDK test kit supports both link:https://junit.org/junit4/[JUnit 4] and link:https://junit.org/junit5/[JUnit 5].
 
 .src/test/java/com/example/domain/CounterTest.java
 [source,java]
--- a/docs/src/modules/java/pages/kickstart.adoc
+++ b/docs/src/modules/java/pages/kickstart.adoc
@@ -133,7 +133,7 @@ For more details see xref:java:value-entity.adoc[].
 
 === Unit and integration tests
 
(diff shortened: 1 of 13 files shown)


docs/src/modules/java/pages/kickstart.adoc
```diff
@@ -133,7 +133,7 @@ For more details see xref:java:value-entity.adoc[].
 
 === Unit and integration tests
 
-The Akka Serverless Maven plugin creates a unit test stub for each class. Use this stub as a starting point to test the logic in your implementation. The Akka Serverless Java SDK test kit supports both link:https://junit.org/junit4/[JUnit 4] and link:https://junit.org/junit5/[JUnit 5].
+The Akka Serverless Maven plugin creates a unit test stub for Entity. Use this stub as a starting point to test the logic in your implementation. The Akka Serverless Java SDK test kit supports both link:https://junit.org/junit4/[JUnit 4] and link:https://junit.org/junit5/[JUnit 5].
```
johanandren, line 136:
```suggestion
The Akka Serverless Maven plugin creates a unit test stub for the Entity. Use this stub as a starting point to test the logic in your implementation. The Akka Serverless Java SDK test kit supports both link:https://junit.org/junit4/[JUnit 4] and link:https://junit.org/junit5/[JUnit 5].
```

johanandren (approved):

docs/src/modules/java/pages/kickstart.adoc
octonato: We once decided to stick with junit5. 
See https://github.com/lightbend/akkaserverless-java-sdk/issues/41

I don't remember from where the idea comes from 'supporting'. Is it possible that the codegen was previously generating something for one or other style of tests? 

Anyway, I don't think we should change that in this PR, but we probably want to address #41 before we release a new SDK.

octonato (approved):
LGTM

maven-java/akkaserverless-maven-archetype-event-sourced-entity/src/main/resources/archetype-resources/pom.xml
```diff
@@ -299,18 +297,6 @@
...
-    <dependencies>
-      <dependency>
-        <groupId>com.typesafe.akka</groupId>
-        <artifactId>akka-bom_${scala.binary.version}</artifactId>
-        <version>${akka-streams.version}</version>
-        <type>pom</type>
-        <scope>import</scope>
-      </dependency>
-    </dependencies>
-  </dependencyManagement>
```
raboof, line 312: don't we want to promote this anymore?

raboof (approved):

maven-java/akkaserverless-maven-archetype-event-sourced-entity/src/main/resources/archetype-resources/pom.xml
patriknw: We didn't have it in any of our samples. I don't know when or why it was introduced? It should work with transitive dependencies from sdk as long as user isn't starting to add Akka dependencies to their build.

raboof: > I don't know when or why it was introduced?

Not sure either, looks like it's been there for ages (be8f8d6ea)

> It should work with transitive dependencies from sdk as long as user isn't starting to add Akka dependencies to their build.

This archetype *is* adding an Akka dependency to the build, though (akka-stream) - perhaps that can be removed?

patriknw: This *is* the archetype and and I'm removing the akka-stream dependency here.

raboof: > I'm removing the akka-stream dependency here.

Aaah gotcha I didn't put those together yet :+1:

PR #308, opened 2021-09-08
Generate View emptyState depending on transform_updates, #290

### LINKED ISSUE #290: Should View class really be generated for automatic view
We implemented it so that a view without any `transform_updates: true` methods will anyway generate all classes that a view with update transformations will. This means there is a generated default `emptyState` that throws, but since the view is automatic it is actually never invoked. 

We did it like that to make it easy to enable transform in the protobuf after having generated things, and avoiding to have different code paths in the codegen (for example only generating a `Provider`).

After a discussion with @RayRoestenburg I wonder if this isn't more confusing than we thought when going down that path - could we make it less confusing - either by doing it in a different way or generating some explanatory comment for such cases?

patriknw:
Fixes #290


johanandren (approved):

raboof (approved):
Nice!

PR #310, opened 2021-09-08
Generate test starting points

johanandren:
For event sourced and value entities.


codegen/java-gen/src/main/scala/com/lightbend/akkasls/codegen/java/EventSourcedEntityTestKitGenerator.scala
```diff
@@ -74,13 +77,21 @@ object EventSourcedEntityTestKitGenerator {
...
           |
+          |  public static $testkitClassName of(Function<EventSourcedEntityContext, $entityClassName> entityFactory) {
+          |    return of("testkit-entity-id", entityFactory);
+          |  }
+          |
+          |  public static $testkitClassName of(String entityId, Function<EventSourcedEntityContext, $entityClassName> entityFactory) {
+          |    return new $testkitClassName(entityFactory.apply(new TestKitEventSourcedEntityContext(entityId)));
+          |  }
+          |
           |  public ${testkitClassName}(${entityClassName} entity) {
```
patriknw, line 95: Shall we make the constructor `private` to guide towards using the `of` factory methods (and not provide alternatives)?

testkit/src/main/java/com/akkaserverless/javasdk/testkit/TestKitEventSourcedEntityContext.java
```diff
@@ -0,0 +1,40 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.akkaserverless.javasdk.testkit;
+
+import com.akkaserverless.javasdk.ServiceCallFactory;
+import com.akkaserverless.javasdk.eventsourcedentity.EventSourcedEntityContext;
+
+public final class TestKitEventSourcedEntityContext implements EventSourcedEntityContext {
```
patriknw, line 22: This doesn't have to be public. Can be in `impl` package and written in Scala.

johanandren:
Here's a sample generated ES test, feedback much welcome:

```
public class CounterTest {

  @Test
  public void exampleTest() {
    CounterTestKit testKit = CounterTestKit.of(Counter::new);
    // use the testkit to execute a command
    // of events emitted, or a final updated state:
    // EventSourcedResult<SomeResponse> result = testKit.someOperation(SomeRequest);
    // verify the emitted events
    // ExpectedEvent actualEvent = result.getNextEventOfType(ExpectedEvent.class);
    // assertEquals(expectedEvent, actualEvent)
    // verify the final state after applying the events
    // assertEquals(expectedState, testKit.getState());
    // verify the response
    // SomeResponse actualResponse = result.getReply();
    // assertEquals(expectedResponse, actualResponse);
  }

  @Test
  public void increaseTest() {
    CounterTestKit testKit = CounterTestKit.of(Counter::new);
    // EventSourcedResult<Empty> result = testKit.increase(IncreaseValue.newBuilder()...build());
  }


  @Test
  public void decreaseTest() {
    CounterTestKit testKit = CounterTestKit.of(Counter::new);
    // EventSourcedResult<Empty> result = testKit.decrease(DecreaseValue.newBuilder()...build());
  }


  @Test
  public void resetTest() {
    CounterTestKit testKit = CounterTestKit.of(Counter::new);
    // EventSourcedResult<Empty> result = testKit.reset(ResetValue.newBuilder()...build());
  }


  @Test
  public void getCurrentCounterTest() {
    CounterTestKit testKit = CounterTestKit.of(Counter::new);
    // EventSourcedResult<CurrentCounter> result = testKit.getCurrentCounter(GetCounter.newBuilder()...build());
  }

}
```

patriknw:
Example is looking good. Rename `ExpectedEvent` to `SomeEvent`?

patriknw:
`expectedState` is missing, verify it at least once with compiler

johanandren:
That's true for `expectedEvent` and `expectedResponse` as well, so more placeholder for the user building up something to compare to than just uncomment and go.

codegen/java-gen/src/main/scala/com/lightbend/akkasls/codegen/java/EventSourcedEntityTestKitGenerator.scala
johanandren: I did not because I was thinking it is worth the flexibility where users have more constructor parameters etc, but I guess they could still do that with only the `.of` factories.

testkit/src/main/java/com/akkaserverless/javasdk/testkit/TestKitEventSourcedEntityContext.java
johanandren: I was thinking users _might_ want to create their entity with some specific/custom entity id, but I guess we can provide factories for that as well, from the testkit class itself for example.

codegen/java-gen/src/main/scala/com/lightbend/akkasls/codegen/java/EventSourcedEntityTestKitGenerator.scala
patriknw: I'm thinking that the Function with context is anyway something they need to understand in the Main. If we want convenience without the Function I think it should be another `of` factory method rather than mixing constructor and factory methods.

testkit/src/main/java/com/akkaserverless/javasdk/testkit/TestKitEventSourcedEntityContext.java
patriknw: you have already defined a second `of` with entityId 😄

johanandren: 🤦 problem solved then, I'll make it private!

patriknw (approved):
LGTM

PR #314, opened 2021-09-08
New docs for Actions

octonato:
I changed the example to cover the most common use case, a pure stateless function. 

My goal is to extend the Actions documentation to expand on the other use cases:

* request transformation / forward
* consuming events from journal
* consuming events from PubSub

I will send the other PRs on top of this one.


docs/src/modules/java/partials/actions.adoc
```diff
@@ -1,25 +1,19 @@
 include::ROOT:partial$include.adoc[]
 
-Actions are stateless functions that can be triggered in multiple ways. For example, by
+Actions are stateless functions that can be used to implement different uses cases, such as:
+
+- a pure function
+- request conversion - you can use Actions to convert incoming data into a different format before forwarding a call to a different component.
+- subscribing to events within the same service
```
patriknw, line 7: some mention that it can publish events (to a different service)

patriknw (approved):
LGTM, just one minor suggestion

docs/src/modules/java/partials/actions.adoc
octonato: Good point, that makes me wondering if we should simply say: Publish and Subscribe to events without making a distinction between internal and external events. 

Then, when move to the "Publish and Subscribe" page, we can explain that we can subscribe to local journal or to topic, push events to another service or to another topic, etc. 

At the end, the use cases around publishing and subscribing are the same except the source of the subscriptions. 

So it may make sense to explain them together.

patriknw: makes sense in this overview

docs/src/modules/java/pages/actions.adoc
```diff
@@ -3,59 +3,85 @@
...
+include::java:example$fibonacci-action/src/main/java/com/example/Main.java[]
 ----
-<1> Use `Effect<Empty>` as return type
-<2> Create the parameter the target method accepts
-<3> Use the fully-qualified gRPC service name of the target method
-<4> Specify the Protobuf rpc method name
-<5> Pass the target methods expected parameter type
-<6> Return a forward to the call with the required parameter
+
+By default, the generated constructor has a `ValueEntityContext` parameter, but you can change this to accept other parameters.
```
johanandren, line 83:
```suggestion
By default, the generated constructor has a `ActionCreationContext` parameter, but you can change this to accept other parameters. 
```

PR #318, opened 2021-09-09
Add docs for replicated entities

pvlugter:
Add docs for replicated entities, in line with the other entity docs. And then add details and examples for each of the replicated data types.


docs/src/modules/java/pages/replicated-entity.adoc
```diff
@@ -0,0 +1,466 @@
...
+<7> The service descriptor shows the API of the entity. It lists the methods a client can use to issue Commands to the entity.
+<8> The protobuf option `(akkaserverless.service)`  is specific to code-generation as provided by the Akka Serverless Maven plugin and points to the protobuf definition `ShoppingCart` we've seen above (in the `com.example.shoppingcart.domain` package).
+
+
+== Implementing behavior
+
+A Replicated Entity implementation is a Java class where you define how each command is implemented. The class `ShoppingCart` gets generated for us based on the `shoppingcart_api.proto` and `shoppingcart_domain.proto` definitions. Once the `ShoppingCart.java` file exist, it is not overwritten, so you can freely add logic to it. `ShoppingCart` extends the generated class `AbstractShoppingCart` which we're not supposed to change as it gets regenerated in case we update the protobuf descriptors. `AbstractShoppingCart` contains all method signatures corresponding to the API of the service. If you change the API you will see compilation errors in the `ShoppingCart` class and you have to implement the methods required by `AbstractShoppingCart`.
```
johanandren, line 62: implementation + implemented so close to each other in same sentence, what about:
```suggestion
A Replicated Entity implementation is a Java class where you define how each command is handled. The class `ShoppingCart` gets generated for us based on the `shoppingcart_api.proto` and `shoppingcart_domain.proto` definitions. Once the `ShoppingCart.java` file exist, it is not overwritten, so you can freely add logic to it. `ShoppingCart` extends the generated class `AbstractShoppingCart` which we're not supposed to change as it gets regenerated in case we update the protobuf descriptors. `AbstractShoppingCart` contains all method signatures corresponding to the API of the service. If you change the API you will see compilation errors in the `ShoppingCart` class and you have to implement the methods required by `AbstractShoppingCart`.
```

johanandren (approved):

patriknw (approved):
LGTM. Great work, Peter.

docs/src/modules/java/pages/replicated-entity.adoc
pvlugter: Good. I think this was just adapted from the other entity docs, so we may want to update for the other entities too.

pvlugter: Updated others in #338

pvlugter:
Note that I didn't document Vote, as it's difficult to explain its usefulness currently. Created an issue for discussion: https://github.com/lightbend/akkaserverless-framework/issues/859

PR #319, opened 2021-09-09
conditionally include NotUsed and Source imports

octonato:
Noticed that Actions have unused imports. Some imports are only need if the proto has a stream method


patriknw (approved):
LGTM

raboof (approved):
Almost starts feeling attractive to introduce a `Snippet` data type that has imports and code, so concatenating snippets could merge their imports.

Test expectations need to be updated, otherwise LGTM!

octonato:
> Test expectations need to be updated, otherwise LGTM!

Of course, I don't know where I was with my head.

PR #325, opened 2021-09-09
added example of Action as a Controller

octonato:
This PR covers the Action as Controller use case:

It temporarily removes the navigation link to `forwarding.adoc`, `entity-eventing.adoc` and `java:topic-eventing.adoc`. They are quite out-of-date now. 

The two eventing pages will be addressed in upcoming PR.

Parts of what is covered in the forwarding page is now covered in the "Action as Controller". I think this page should become only about running side effects. 

About forwarding from entities, we should either remove this feature because we now have Actions or document it on their respective sections (and with a clear use case why someone would do such a thing).

If we keep the feature, we can delay any documentation to the moment we build a replacement for the ServiceCall. As of today, the forward is an obscure feature that we only explain in the context of Actions, so I think it's acceptable to ignore it for the entities for now.


docs/src/modules/java/pages/actions-as-controller.adoc
```diff
@@ -0,0 +1,52 @@
+= Actions as Controller
+
+Actions can be used to implement link:https://en.wikipedia.org/wiki/Model%E2%80%93view%E2%80%93controller[MVC Controllers] by
+acting as the external interface of a service, receiving requests, operating over the requests values and forwarding the call
+to other components.
+
+== Defining the `proto` file
+
+To illustrate how you can use an Action as a Controller, we will build on top of the xref:value-entity.adoc[Value Entity Counter] example and add
+an Action that will take the incoming `Increase` command, double its value and forward to the request to `Counter` entity.
```
patriknw, line 10:
```suggestion
an Action that will take the incoming `Increase` command, double its value and forward the request to a `Counter` entity.
```

docs/src/modules/java/pages/actions-as-controller.adoc
```diff
@@ -0,0 +1,52 @@
...
+
+Note that in this example we are adding an Action to an existing service, the Value Entity Counter service.
```
patriknw, line 12: that example is used in value-entity.adoc so would be good with a link

samples/valueentity-counter-with-action-controller/pom.xml
```diff
@@ -0,0 +1,349 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+
+  <groupId>com.example</groupId>
+  <artifactId>valueentity-counter</artifactId>
```
patriknw, line 7: Do we need a full new sample for this? can't it be added to the existing `valueentity-counter` example? The only thing we might need to "fake" for the docs is the registration in Main.

samples/valueentity-counter-with-action-controller/src/main/java/com/example/actions/DoubleCounterAction.java
```diff
@@ -0,0 +1,41 @@
...
+    int doubled = increaseValue.getValue() * 2;
+    CounterApi.IncreaseValue increaseValueDoubled =
+        increaseValue.toBuilder().setValue(doubled).build(); // <2>
+
+    ServiceCallRef<CounterApi.IncreaseValue> callRef = // <3>
+        creationContext.serviceCallFactory()
+          .lookup(
+              "com.example.CounterService",
+              "Increase",
+              CounterApi.IncreaseValue.class);
```
patriknw, line 36: can this be done in the constructor instead? for actions it is no runtime advantage since we create a new instance for each request, but isn't it conceptually right to initialize the ServiceCallRef once and then use it for each call?

patriknw (commented):
good example, but I wonder if we can include it in the existing instead of adding a new?

samples/valueentity-counter-with-action-controller/pom.xml
octonato: Exactly, that's where I got blocked and what triggered me to copy the sample. 

I see a few options: 
* fake the registration in the value entity docs
* copy like here - I could have copied less code, just enough to make it compile. 
* copy and add a script to keep things in sync, like we do in the platform guide

Another option is change the registration. Instead of one method `withComponents`, we could have a method per component eg: `registerCounter`, `registerDoubleCounterAction`. 

We will lose the the compile time check in the main. When adding a new component, nothing will tell the user that they need to go there fix it. 

But I also realised that the `withComponents` may be annoying in some situations. For instance, if you need to temporarily disable a component, you will need to remove the proto file and all generated classes. That can be very annoying during development cycle.

samples/valueentity-counter-with-action-controller/src/main/java/com/example/actions/DoubleCounterAction.java
octonato: I had this in mind when I started, but I changed my mind when I realised that the lookup is per method, not per service. 

I don't know if it's conceptually right. It felt to me easier to put it in the method so we don't need to explain why we are creating it in the constructor. Anyway, I don't mind much. Hopefully this goes away soon.

octonato: > since we create a new instance for each request

That's actually our solution for the `ActionContext` and the async boundaries issue https://github.com/lightbend/akkaserverless-java-sdk/issues/268

We should have only one context passed to the constructor.

docs/src/modules/java/pages/actions-as-controller.adoc
octonato: I linked in the beginning of page. Will link again.

samples/valueentity-counter-with-action-controller/pom.xml
octonato: I found a little hack that reduce the duplication by a lot

samples/valueentity-counter/src/main/java/com/example/Main.java
```diff
@@ -6,25 +6,42 @@
...
+
+    // end::registration[]
+    /* the comment hack bellow is need to show only the Counter::new in the value entity docs
+    // tag::registration-value-entity[]
     return AkkaServerlessFactory.withComponents(
             Counter::new);
+    // end::registration-value-entity[]
+     */
+    // tag::registration-value-entity[]
+    // tag::registration[]
```
octonato, line 38: This is the hack I found to reuse the same project but only show the Counter when presenting the Value Entity. 

We do have a bit of manual maintenance, but that's ok. If we ever touch the protos, we will have to touch that file as well.

patriknw: that's fine

samples/valueentity-counter/pom.xml
```diff
@@ -19,7 +19,7 @@
     <jdk.target>11</jdk.target>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
 
-    <akkaserverless-sdk.version>0.7.0-beta.18</akkaserverless-sdk.version>
+    <akkaserverless-sdk.version>0.7.0-beta.18-117-35889025-SNAPSHOT</akkaserverless-sdk.version>
```
patriknw, line 22:
```suggestion
    <akkaserverless-sdk.version>0.7.0-beta.18</akkaserverless-sdk.version>
```

patriknw (approved):
LGTM

johanandren (approved):

Issue #322, opened 2021-09-09
Testkit support for asserting all different effects


johanandren:
Currently the Result type returned expects a reply message, we should support capturing the other types of effects and asserting those as well.

johanandren:
Merged, but one thing that is still missing is some way to inspect/assert side effects (I think we can add that later).

patriknw:
@johanandren @franciscolopezsancho this ticket is still open. Do we have more work planned for this ticket or can we close it?

franciscolopezsancho:
It felt out of my radar, sorry. We need it for Event Sourced Entities and Value Entities, right?

johanandren:
Side effect access still missing in ValueEntityResult and EventSourcedResult yes.

Replicated entities we don't have any testkit for (yet) but that should be tracked in a different separate issue.

franciscolopezsancho:
this is blocked by #722

franciscolopezsancho:
it's done now for Event Sourced Entities and Actions. Still missing Value entities.

PR #326, opened 2021-09-09
More effect inspections for the testkits

johanandren:
Refs #322

Pausing it for now, will continue doing the same for ES testkit given time before the release tomorrow.


patriknw (approved):
LGTM!

testkit/src/main/java/com/akkaserverless/javasdk/testkit/ValueEntityResult.java
```diff
@@ -20,23 +20,43 @@
...
+  boolean isError();
+
+  /** The error description. If the result was not an error an exception is thrown */
+  String getError();
+
+  /** @return true if the call had a noReply effect, false if not */
+  boolean isNoReply();
+
+  /** @return true if the call updated the entity state */
+  boolean stateWasUpdated();
```
octonato, line 55: Since this is asking a question, I think it will read better if: `wasStateUpdated()`

testkit/src/main/java/com/akkaserverless/javasdk/testkit/EventSourcedResult.java
```diff
@@ -26,55 +26,52 @@
...
+  /**
+   * @return The updated state. If the state was not updated (no events emitted) an exeption is
+   *     thrown
    */
-  public EventSourcedResult(R reply, List<Object> events) {
-    this.reply = reply;
-    this.events = events;
-    this.eventsIterator = events.iterator();
-  }
+  Object getUpdatedState();
```
octonato, line 62: shall we pair it with a `wasStateUpdated` to align with the ValueEntity?

testkit/src/main/java/com/akkaserverless/javasdk/testkit/EventSourcedResult.java
```diff
@@ -26,55 +26,52 @@
...
 
-  /** The reply object from the handler if there was one. */
-  public R getReply() {
-    return reply;
-  }
+  boolean didEmitEvents();
```
octonato, line 64: 👍 verb when asking questions 😄

octonato (approved):
LGTM with suggestions on method names

testkit/src/main/java/com/akkaserverless/javasdk/testkit/EventSourcedResult.java
johanandren: Here we have the was any events emitted method instead, I think having both does not add much.

testkit/src/main/java/com/akkaserverless/javasdk/testkit/ValueEntityResult.java
johanandren: I started renaming, but after seeing it in use I think i disagree: `if (result.stateWasUpdated()) / assertTrue(result.stateWasUpdated())` vs `if (result.wasStateUpdated()) / assertTrue(result.wasStateUpdated())`

current reads better, no?

testkit/src/main/java/com/akkaserverless/javasdk/testkit/EventSourcedResult.java
johanandren: BTW It's a bit messy on `ValueEntityResult` anyway, since we'd have to add a noisy type param to keep the type of the state which I don't think is worth it, the testkit is typed with the state already and also has a getState of the right type which is likely more useful.

octonato: Emitting events doesn't mean state update.

That could be a mistake, like returning the current state without updating it. Or I could have emitted events and decide to not apply. 

Actually, I think it would make more sense to have `getState` in both APIs and a check like `stateWasUpdated`. `getState` would throw exception if state was deleted in a ValueEntity

testkit/src/main/java/com/akkaserverless/javasdk/testkit/ValueEntityResult.java
octonato: Ok, fine by me

PR #329, opened 2021-09-09
merge main to sdk-codegen-dev

patriknw:
I think we messed up previous `merge` in https://github.com/lightbend/akkaserverless-java-sdk/pull/218 so we had many conflicts.

This is a
```
git merge -s ours main
```

I'll have to verify if this is the right result.


patriknw:
this is not the way to do it

Issue #324, opened 2021-09-09
Wrong reply type_url when using JsonSupport


patriknw:
When I tried to run the Action producing JSON in https://github.com/lightbend/akkaserverless-java-sdk/pull/323 it didn't work.

```
grpcurl -plaintext -d '{"key": "foo", "value": 2}' localhost:9000 com.example.json.MyService/Produce
```

In the SDK it logged:
```
2021-09-09 13:41:41,899 ERROR com.akkaserverless.javasdk.impl.DiscoveryImpl - Error reported from Akka Serverless system:  Warning: com.example.json.MyService.Produce: Expected reply type_url to be [type.googleapis.com/google.protobuf.Any] but was [json.akkaserverless.com/com.example.json.JsonKeyValueMessage
```

Logged:
```
{
	"timestamp": "2021-09-09T11:41:41.891Z",
	"mdc": {
		"akkaAddress": "akka://akkaserverless-proxy@192.168.2.23:25520",
		"sourceThread": "akkaserverless-proxy-akka.actor.default-dispatcher-13",
		"akkaSource": "Serve$(akka://akkaserverless-proxy)",
		"sourceActorSystem": "akkaserverless-proxy",
		"akkaTimestamp": "11:41:41.890UTC"
	},
	"logger": "com.akkaserverless.proxy.Serve$",
	"message": "com.example.json.MyService.Produce: Expected reply type_url to be [type.googleapis.com/google.protobuf.Any] but was [json.akkaserverless.com/com.example.json.JsonKeyValueMessage].",
	"severity": "WARN",
	"thread": "akkaserverless-proxy-akka.actor.default-dispatcher-13"
}

{
	"timestamp": "2021-09-09T11:41:41.897Z",
	"mdc": {
		"akkaAddress": "akka://akkaserverless-proxy@192.168.2.23:25520",
		"sourceThread": "akkaserverless-proxy-akka.actor.default-dispatcher-13",
		"akkaSource": "akka.actor.ActorSystemImpl(akkaserverless-proxy)",
		"sourceActorSystem": "akkaserverless-proxy",
		"akkaTimestamp": "11:41:41.897UTC"
	},
	"logger": "akka.actor.ActorSystemImpl",
	"message": "Unhandled error: [While parsing a protocol message, the input ended unexpectedly in the middle of a field.  This could mean either that the input has been truncated or that an embedded message misreported its own length.].",
[...]

PR #330, opened 2021-09-09
Fix for incorrect json encoding

johanandren:
Refs #324 

Need to verify with actual pub/sub, but pretty sure this is at least a part of the problem.

### DIFF
--- a/samples/doc-snippets/src/main/java/com/example/json/JsonKeyValueMessage.java
+++ b/samples/doc-snippets/src/main/java/com/example/json/JsonKeyValueMessage.java
@@ -1,10 +1,14 @@
 package com.example.json;
 
+import com.fasterxml.jackson.annotation.JsonCreator;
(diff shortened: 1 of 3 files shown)


octonato (approved):
LGTM

I understand that you still want to do some manual test, correct? 

Feel free to merge when ready.

johanandren:
I've done a manual test now, and akkaserverless-json encoded in probuf any to and from topic works fine. 👍 

However, posting manually through an action where the message for the topic also goes back to the grpc client gives the same error. I think that was the same already before, works fine with regular protobuf, and if we want it to work for akkaserverless-json as well we'll need to do something in the proxy. (I cross posted this in the docs PR discussion as well https://github.com/lightbend/akkaserverless-java-sdk/pull/323#discussion_r705488082 )

patriknw:
Thanks for fixing this.

PR #331, opened 2021-09-09
Move lowlevel to impl, #262

patriknw:
Resolves https://github.com/lightbend/akkaserverless-java-sdk/issues/262


octonato (approved):
LGTM, but do we still need to have a low-level class? I think we can just remove that indirection, no?

patriknw:
> but do we still need to have a low-level class? I think we can just remove that indirection, no?

We can probably get rid of those, but that is a bigger refactoring and since that is only internals I didn't want to do that now. For example there is some trick with `Resolved*Factory`.

sdk/src/main/java/com/akkaserverless/javasdk/impl/AkkaServerlessAnnotation.java
```diff
@@ -1,27 +0,0 @@
...
-
-import java.lang.annotation.ElementType;
-import java.lang.annotation.Retention;
-import java.lang.annotation.RetentionPolicy;
-import java.lang.annotation.Target;
-
-/** Mark annotation for all AkkaServerless annotations */
-@Target(ElementType.ANNOTATION_TYPE)
-@Retention(RetentionPolicy.RUNTIME)
-public @interface AkkaServerlessAnnotation {}
```
patriknw, line 27: also removes the last annotation 😄

johanandren (approved):

Issue #321, opened 2021-09-09
Integrations test folder not picked by IntelliJ


octonato:
After importing a sample into IntelliJ, the integration test folder `/src/it/java` needs to be manually added as a test source. 

There is probably something we can do at the pom.xml to hint IntelliJ (and other editors) that this folder contains source files.

johanandren:
I think it can/should be done from the maven plugin: https://github.com/lightbend/akkaserverless-java-sdk/blob/sdk-codegen-dev/maven-java/akkaserverless-maven-plugin/src/main/java/com/akkaserverless/GenerateMojo.java#L122

johanandren:
Oh, wait, that's for the generated test sources, nvm!

octonato:
I'm wondering if the fact that the Mojo adds paths to the project will help IntelliJ to 'see' what are the folders to include as source folders. 

Maybe IntelliJ users some Maven API to get the list of paths to use. If that's the case, if the Mojo adds the integration tests to the maven project, we may solve the issue

aklikic:
one addition: `IntegrationTest` class is generated with `default` maven profile but `*ServiceClient` generation requires `it` profile to be used.

octonato:
Good point @aklikic and I think that is a regression.

The test in CI do run `verify -Pit` so the issue is hidden, but we should have noticed that before in our local machines.

johanandren:
I _think_ the problem is that we add the sources in a profile - maven doesn't have that concept of a whole different dimension of test sources like we can do with sbt. What we need is probably to always add the it-sources and compile those, but then only **run** those tests with the profile enabled.

Hiding them in the profile is probably also what hides them from intellij unless we import the project with the profile enabled.

johanandren:
As a small side note, intellij also understands that code generators may add generated source roots, I think it does this by itself when you build but there is also a separate explicit trigger for it (right click on project and go for maven menu):

<img width="685" alt="Screenshot 2021-09-09 at 18 22 57" src="https://user-images.githubusercontent.com/666915/132724710-7213e48e-9b77-4c11-b8da-d30fe4767caf.png">

PR #332, opened 2021-09-09
Always add integration test sources to maven projects

johanandren:
So that they are compiled and picked up by IDEs even without the it profile enabled

Refs  #321


.circleci/config.yml
```diff
@@ -203,73 +203,86 @@ jobs:
           command: |
             cd samples/customer-registry-quickstart
             echo "Running mvn with SDK version: '$SDK_VERSION'"
+            # must also compile without the profile!
+            mvn -Dakkaserverless-sdk.version=$SDK_VERSION compile
```
patriknw, line 207: should it be `test-compile`?

patriknw (approved):
LGTM

.circleci/config.yml
johanandren: Ah, good idea, let's take it all the way there indeed.

octonato (approved):

PR #333, opened 2021-09-09
document side effect with action controller

octonato:
Add a new page about side effects. 

In retrospect, I think we could keep forward and side effect together.

On the other hand, I think it make more sense to let the use case drive the documentation. So "Action as Controller" deserves it's own page.


johanandren (approved):

octonato:
@johanandren, I will need an extra ✅ since I added one more commit 861b0a2 to fix the build. 

It turns out that there were other pages referencing the old forwarding page. We will need to revisit. The previous and current page explain how to do it with Actions, but the text I just changed refers to forwards from inside entities.

patriknw (approved):
LGTM

johanandren (approved):

PR #335, opened 2021-09-09
Value entity testing docs

johanandren:
Brief but I think this could be enough?


samples/valueentity-counter/src/test/java/com/example/domain/CounterTest.java
```diff
@@ -0,0 +1,80 @@
...
+import com.google.protobuf.Empty;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.NoSuchElementException;
+import org.junit.Test;
+import scala.jdk.javaapi.CollectionConverters;
+
+import static org.junit.Assert.*;
+
+// tag::sample-unit-test[]
```
patriknw, line 20: would be good to show some imports

patriknw (approved):
LGTM, with suggestion to show imports

docs/src/modules/java/pages/value-entity.adoc
```diff
@@ -115,3 +115,35 @@ include::example$valueentity-counter/src/main/java/com/example/Main.java[tag=reg
 By default, the generated constructor has a `ValueEntityContext` parameter, but you can change this to accept other parameters. If you change the constructor of the `Counter` class you will see a compilation error here, and you have to adjust the factory function that is passed to `AkkaServerlessFactory.withComponents`.
 
 When more components are added the `AkkaServerlessFactory` is regenerated, and you have to adjust the registration from the `Main` class.
+
+== Testing the Entity
+
+There are two ways to test a Value Entity, unit tests, which run the Entity class in the same JVM as the test code itself with the help of a test kit and integration tests with the service deployed in a docker container running the entire service and the test interacting over gRPC with it.
```
octonato, line 121: One or another, or something else, but current it hard to see the boundary with the `and`

I personally find the bullets easier to follow. 

```suggestion
There are two ways to test a Value Entity: 
* unit tests, which run the Entity class in the same JVM as the test code itself with the help of a test kit 
* integration tests, with the service deployed in a docker container running the entire service and the test interacting over gRPC with it.
```

```suggestion
There are two ways to test a Value Entity: unit tests, which run the Entity class in the same JVM as the test code itself with the help of a test kit or integration tests with the service deployed in a docker container running the entire service and the test interacting over gRPC with it.
```

octonato (approved):
LGTM 
... with some phrasing suggestions

PR #337, opened 2021-09-09
Rename updateEffects to effects

RayRoestenburg:


patriknw (approved):
LGTM

octonato (approved):
LGTM, but I still think we should call the type `Effect` and later find something else for the query result transformation. 

Anyway, we can revisit it later.

PR #339, opened 2021-09-10
Cherry pick latest commits from main to sdk-codegen-dev

patriknw:
We messed up previous merge in #218 so there are many conflicts even though not  much has changed in main.

This brings in the relevant commits from main.
I think the easiest and safest is if we rename the `sdk-codegen-dev` branch to `main` since it is what we want to continue with. Any other suggestions?

DO NOT SQUASH


johanandren (approved):
Makes sense to cherry pick as far as I can see

PR #342, opened 2021-09-10
Custom viewId via the provider, #311

### LINKED ISSUE #311: Custom viewId
The viewId is the same as the service name in the new codegen. For the steps described in "How to modify a View" in the documentation the service can be renamed to the same as the old service name, but that will not work for the viewId.

Maybe it's enough to add `.withViewId` in the provider and use that with custom registration?

Another thought is to define the `view_id` in proto, similar to what we have for `entity_type`.

patriknw:
Resolves #311


johanandren (approved):

samples/java-eventsourced-customer-registry/src/main/java/customer/MainWithCustomViewId.java
```diff
@@ -0,0 +1,34 @@
...
+
+public final class MainWithCustomViewId {
+
+  private static final Logger LOG = LoggerFactory.getLogger(MainWithCustomViewId.class);
+
+  // tag::register[]
+  public static AkkaServerless createAkkaServerless() {
+    AkkaServerless akkaServerless = new AkkaServerless();
+    akkaServerless.register(CustomerByNameViewProvider.of(CustomerByNameView::new)
+        .withViewId("CustomerByNameV2"));
```
octonato, line 24: since the docs are telling users to keep the old view, we could also show it here, for instance: 

```suggestion
    // TODO: remove this registration on next deployment 
    // and after verifying that new view is working
    akkaServerless.register(CustomerByNameViewProvider.of(CustomerByNameView::new)

    akkaServerless.register(CustomerByNameViewProvider.of(CustomerByNameView::new)
        .withViewId("CustomerByNameV2"));
```

octonato (approved):
LGTM, left minor optional suggestion.

samples/java-eventsourced-customer-registry/src/main/java/customer/MainWithCustomViewId.java
patriknw: Then I would have to add another `CustomerByNameViewV2` to the sample and we should have registration code for both steps of the rollout. As I wrote the steps this code is only referred to in the final step so it's ok.

Could be good to elaborate on that process in more detail, but I'll leave that for later.

octonato: ok, that's fine.

patriknw: created issue https://github.com/lightbend/akkaserverless-java-sdk/issues/348 for that

Issue #317, opened 2021-09-09
view update handler state naming


aklikic:
view update handler uses `state` as a name for the representation of the view row.
This naming could be missleading by relating it to `domain state` and these are two different things.
Even our examples use `doman state` grpc message as a view row representation supporting this relation. 

For example in `java-eventsources-customer-registry` `customer_view.proto`, `domain.CustomerState` is used as a view row representation:
```
 rpc ProcessCustomerCreated(domain.CustomerCreated) returns (domain.CustomerState) {
    option (akkaserverless.method).eventing.in = {
      event_sourced_entity: "customers"
    };
    option (akkaserverless.method).view.update = {
      table: "customers"
      transform_updates: true
    };
  }
```
And in view update handler class `CustomerByNameView` methods use state naming (`emptyState()`,  parametar in `updateEffect` methods).
This all indicates that what we store in the view is a `state`.
In some cases it will be `state` but in others it will be some enriched representation or just a relation.
So chaging the name `state` to something else (like `view`, `viewModel` or `representation`) would be much more descriptive to what it represents. 
Also our view examples should not use `domain state` gprc message but a dedicated gprc message for that view model.

johanandren:
I discussed this with @mckeeh3 a bit recently. It _is_ the state of the view so not sure I agree we should change that naming but I think the second part with the samples re-using the domain representation as the view representation is the actually the cause of the confusion. With a separate type there is no way you'd expect it to be the state of the entity you subscribe to since the types are note the same. It's also bad practice to "leak" the internal domain type through the API, so we shouldn't do that in the samples.

patriknw:
Good point that it's better to have a different state representation for the view, and not reuse the state representation from the entity.

However, for value entities that requires `transform_updates=true` and it was kind of neat to have that automatic.

johanandren:
For samples where `transform_updates=false` there shouldn't be any confusion because then there is no method and no need to implement `emptyState()`.

So the only possible scenario with confusion is in samples with `transform_updates=true` but anyway using the domain types for the view. I think we should update that sample (it's only `java-eventsourced-customer-registry`) to have an API type for the view state instead.

In the future when we have a transformation for the query result as well, that having the domain type in the view will be an ok decision since the result transformation can then turn it into an "API" type.

patriknw:
Sounds good.

[...]

PR #343, opened 2021-09-10
Avoid using the domain state type for the view in sample 

johanandren:
Also a short note in the docs for where we use the same state type.

Refs #317

### DIFF
--- a/docs/src/modules/java/pages/views.adoc
+++ b/docs/src/modules/java/pages/views.adoc
@@ -53,7 +53,7 @@ include::example$java-valueentity-customer-registry/src/main/proto/customer/cust
 <4> The `GetCustomers` method defines the query to retrieve a stream of customers.
 <5> The `(akkaserverless.method).view.query` annotation defines that this method is used as a query of the View.
 
-NOTE: In this sample we use the internal `domain.CustomerState` as the state of the view. This is convenient since it allows automatic updates of the view without any logic but has the draw back that it implicitly makes the `domain.CustomerState` type a part of the public service API. Transforming the state to another type than the incoming update to avoid this can bee seen in <<#event-sourced-entity>>.
+NOTE: In this sample we use the internal `domain.CustomerState` as the state of the view. This is convenient since it allows automatic updates of the view without any logic but has the draw back that it implicitly makes the `domain.CustomerState` type a part of the public service API. Transforming the state to another type than the incoming update to avoid this can be seen in <<#event-sourced-entity>>.
 
 
 If the query should only return one result, remove the `stream` from the return type:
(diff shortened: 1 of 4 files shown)


johanandren:
Needs something more in the event sourced entity view docs!

patriknw (approved):
LGTM

docs/src/modules/java/pages/views.adoc
```diff
@@ -53,6 +53,8 @@ include::example$java-valueentity-customer-registry/src/main/proto/customer/cust
 <4> The `GetCustomers` method defines the query to retrieve a stream of customers.
 <5> The `(akkaserverless.method).view.query` annotation defines that this method is used as a query of the View.
 
+NOTE: In this sample we use the internal `domain.CustomerState` as the state of the view. This is convenient since it allows automatic updates of the view without any logic but has the draw back that it implicitly makes the `domain.CustomerState` type a part of the public service API. Transforming the state to another type than the incoming update to avoid this can bee seen in <<#event-sourced-entity>>.
```
octonato, line 56:
```suggestion
NOTE: In this sample we use the internal `domain.CustomerState` as the state of the view. This is convenient since it allows automatic updates of the view without any logic but has the draw back that it implicitly makes the `domain.CustomerState` type a part of the public service API. Transforming the state to another type than the incoming update to avoid this can be seen in <<#event-sourced-entity>>.
```

samples/java-eventsourced-customer-registry/src/main/java/customer/view/CustomerByNameView.java
```diff
@@ -8,48 +8,69 @@
...
-  public CustomerDomain.CustomerState emptyState() { // <2>
+  public CustomerApi.Customer emptyState() { // <2>
     return null;
   }
 
   @Override // <3>
-  public UpdateEffect<CustomerDomain.CustomerState> processCustomerCreated(
-    CustomerDomain.CustomerState state,
+  public UpdateEffect<CustomerApi.Customer> processCustomerCreated(
+    CustomerApi.Customer state,
```
octonato, line 24: One more thing that we could do to avoid confusion is to call this argument `viewState` to emphasize that this is the persisted view representation.

If that was the view from a value entity and combined with `transform_updates=true`, the method signature would be

```java
processCustomerCreated(CustomerApi.Customer state, CustomerApi.Customer customer)
```

And that can lead to confusion which one is the incoming one and each one is the view state.

octonato (approved):
LGTM

* with typo fix (that I will apply myself)
* suggestion about name that we can address later

docs/src/modules/java/pages/views.adoc
```diff
@@ -53,7 +53,7 @@ include::example$java-valueentity-customer-registry/src/main/proto/customer/cust
 <4> The `GetCustomers` method defines the query to retrieve a stream of customers.
 <5> The `(akkaserverless.method).view.query` annotation defines that this method is used as a query of the View.
 
-NOTE: In this sample we use the internal `domain.CustomerState` as the state of the view. This is convenient since it allows automatic updates of the view without any logic but has the draw back that it implicitly makes the `domain.CustomerState` type a part of the public service API. Transforming the state to another type than the incoming update to avoid this can bee seen in <<#event-sourced-entity>>.
```
johanandren: 🐝

PR #344, opened 2021-09-10
Cleanup json logging docs and don't prettyPrint, #307

### LINKED ISSUE #307: JacksonJsonFormatter prettyPrint in production?
We have added `<prettyPrint>true</prettyPrint>` to many of the samples, including the quickstart.

Will multi-line logging work in production? Otherwise our templates shouldn't use it.

patriknw:
Resolves #307


johanandren (approved):

octonato (approved):

raboof:
Did we learn pretty-printing breaks our logging infra or are we removing it 'just to be safe'?

patriknw:
I have't verified what happens in Google log viewer with multi-line json, but I doubt it would handle that. The Akka Serverless ui has something to combine log messages with same timestamp, but that feels rather fragile to me.

Issue #268, opened 2021-09-02
Stateful action context and async handlers


johanandren:
The ActionContext is now a state on the Action that is set/reset around the call, but the call may do async things and expect to access it in a future/CS callback leading to something much like the good old close over context problem in actors. We should avoid giving users that grief if we can.

_Originally posted by @octonato in https://github.com/lightbend/akkaserverless-java-sdk/pull/258#discussion_r700514689_

patriknw:
isn't that the reason for the `ActivatableContext`?

johanandren:
This is different in that it actually is ok to use the same `ActionContext` instance in callbacks (and you may need that, at least the TCK does), but you can't call `actionContext()` since that has potentially been cleared, so you must make sure to capture it. (The reference to actor context was wrong, I was thinking about `sender()`).

patriknw:
I see what you mean.

Do we create a new Action instance for each request (command)? Otherwise we would have a concurrency bug since the `actionContext` is an instance variable and there can be many Action calls executing at the same time.

Could we make use of that? Not clear the actionContext instance when the call returns. Then it is valid also in callbacks.

johanandren:
I looked into this and we create a new instance of the action class for each incoming command, so while it isn't safe now (potentially accessing a mutable field from other threads) it would be safe to make the `ActionContext` a final field that we do not clear out after the command.

PR #347, opened 2021-09-10
Keep ActionContext in place to be accessed from other threads 

johanandren:
Refs #268


sdk/src/main/scala/com/akkaserverless/javasdk/impl/action/ActionHandler.scala
```diff
@@ -136,8 +136,6 @@ abstract class ActionHandler[A <: Action](protected val action: A) {
         throw new RuntimeException(
           s"No call handler found for call $name on ${action.getClass.getName}"
         )
-    } finally {
-      action._internalSetActionContext(Optional.empty())
```
patriknw, line 140: could be good with a comment here why we don't clear it

johanandren: True, I'll add. I started out with a boolean flag in the handler to detect and throw in case we'd forget and start invoking multiple commands on the same, but reverted that.

patriknw (approved):
LGTM

PR #355, opened 2021-09-10
events should indicate by how much we increase or decrease

octonato:
The events sourced archetype has commands for increase and decrease a counter with a value field meaning that we could increase by 2, 3, etc. 

The events on the other hand, doesn't not hold any value as if we should only increase/decrease by one.

### DIFF
--- a/maven-java/akkaserverless-maven-archetype-event-sourced-entity/src/main/resources/archetype-resources/src/main/proto/counter_domain.proto
+++ b/maven-java/akkaserverless-maven-archetype-event-sourced-entity/src/main/resources/archetype-resources/src/main/proto/counter_domain.proto
@@ -19,10 +19,12 @@ message CounterState {
 
 message ValueIncreased {
     string counter_id = 1 [(akkaserverless.field).entity_key = true];
+    int32 value = 2;
 }
 
 message ValueDecreased {
     string counter_id = 1 [(akkaserverless.field).entity_key = true];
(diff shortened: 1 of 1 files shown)


maven-java/akkaserverless-maven-archetype-event-sourced-entity/src/main/resources/archetype-resources/src/main/proto/counter_domain.proto
```diff
@@ -19,10 +19,12 @@ message CounterState {
 
 message ValueIncreased {
     string counter_id = 1 [(akkaserverless.field).entity_key = true];
```
pvlugter, line 21: Events shouldn't need the counter id and entity key annotation?

pvlugter (approved):

maven-java/akkaserverless-maven-archetype-event-sourced-entity/src/main/resources/archetype-resources/src/main/proto/counter_domain.proto
octonato: correct, I removed in on my the sample #357 but forgot to do the same here. I will fix.

octonato: done

PR #357, opened 2021-09-10
added event sourced counter sample

octonato:
Like the value entity counter, but event sourced. Based on the event sourced archetype. 

The goal is to use it as based for eventing and also, later, as the simplified model for the event sourced documenation.


.circleci/config.yml
```diff
@@ -259,6 +259,13 @@ jobs:
             echo "Running mvn with SDK version: '$SDK_VERSION'"
             mvn -Dakkaserverless-sdk.version=$SDK_VERSION test-compile
             mvn -Dakkaserverless-sdk.version=$SDK_VERSION verify -Pit
+      - run:
+          name: Event Sourced "Counter" sample
+          command: |
+            cd samples/java-eventsourced-counter
```
octonato, line 265: For the record, I'm prefixing it with `java-` like some (but not all) other samples. 

Since we plan to start work on the Scala SDK, it will be good to prefix the projects with the lang.

patriknw (approved):
LGTM

PR #358, opened 2021-09-10
chore(docs): update makefile with latest Java SDK version

retgits:
The minimum supported version should come from the SDK itself, not from the main docs repo. Updating this parameter so we can remove the parameter from the main docs repo.

### DIFF
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -35,7 +35,7 @@ attributes:
 		>> "${managed_partials}/attributes.adoc"
 	echo ":recommended-java-version: 11" \
 		>> "${managed_partials}/attributes.adoc"
-	echo ":java_minimum_sdk_version: 0.7.0-beta.19" \
+	echo ":java_minimum_sdk_version: 0.7.0" \
 		>> "${managed_partials}/attributes.adoc"
 
 apidocs:


docs/Makefile
```diff
@@ -35,7 +35,7 @@ attributes:
 		>> "${managed_partials}/attributes.adoc"
 	echo ":recommended-java-version: 11" \
 		>> "${managed_partials}/attributes.adoc"
-	echo ":java_minimum_sdk_version: 0.7.0-beta.19" \
+	echo ":java_minimum_sdk_version: 0.7.0" \
```
patriknw, line 38: is this something we need to push to `current/docs` to get it out?

patriknw (approved):
LGTM, so many versions all over the place

docs/Makefile
retgits: Let's push it during the next release. Johan (or you) created the PR for the main docs already so we got that covered there. Once this is in the `current/docs` branch, we can make the update in the main docs repo to use this version number rather than what it currently has.

patriknw: I see. We should improve on the Releasing.md so that all steps are clear.

retgits: Great idea! I'll submit a new PR for that.

PR #359, opened 2021-09-10
chore(docs): Update RELEASING.md with docs makefile change

retgits:


retgits:
@patriknw is this the only place it needs to change or are there other places/steps that are involved to trigger the update for docs too?

RELEASING.md
```diff
@@ -3,3 +3,4 @@
 1. Create a `vX.Y.Z` tag and release notes at https://github.com/lightbend/akkaserverless-java-sdk/releases
 2. CircleCI will automatically publish to Sonatype OSSRH (Maven Central) based on the tag.
 3. Update the `akkaserverless-sdk.version` in the `samples/*/pom.xml` files to the released version.
+4. Update the [docs Makefile](https://github.com/lightbend/akkaserverless-java-sdk/blob/main/docs/Makefile#L38) to the released version.
```
patriknw, line 6: I think this should be step 1 (before releasing), because `docs/current` is updated by the release bot when creating the tag

or did I misunderstand something?

retgits: Ah! I didn't know that. I'll change it

RELEASING.md
```diff
@@ -1,5 +1,6 @@
 # Releasing Akka Serverless Java SDK
 
-1. Create a `vX.Y.Z` tag and release notes at https://github.com/lightbend/akkaserverless-java-sdk/releases
-2. CircleCI will automatically publish to Sonatype OSSRH (Maven Central) based on the tag.
-3. Update the `akkaserverless-sdk.version` in the `samples/*/pom.xml` files to the released version.
+1. Before releasing, update the [docs Makefile](https://github.com/lightbend/akkaserverless-java-sdk/blob/main/docs/Makefile#L38) to the new version that will be released.
```
pvlugter, line 3: Shouldn't the minimum supported version only be updated occasionally, not every release, and only when the old version is no longer supported?

If we always update this to the new version on release, what's the point of the minimum version? And we could just use the released SDK version which is automatic.

pvlugter (commented):
Isn't this a minimum supported version?

RELEASING.md
retgits: Do we want to support multiple versions? I can see that after we release v1.0.0 we want to adopt a more formal support policy for our SDKs and support different versions, but up to that I think it should always be the latest version only

pvlugter: Ok, then let's not have a manually edited version, and just use the SDK version that's already added automatically?

pvlugter: The SDK version being released is already written as an attribute:

https://github.com/lightbend/akkaserverless-java-sdk/blob/68b3c6fa43ca734a29b33198f0a6568ea6ae6e5c/docs/Makefile#L32-L33

retgits: Sure! Though I didn’t see how that was done 😔 shall I close this PR?

pvlugter: Yeah, let's not require manually committing the same version before every tagged release.

If we're going to have an actual minimum supported version, then I think it would only be changed when we're a couple of versions beyond.

PR #363, opened 2021-09-13
Add specialised accessors for replicated map

pvlugter:
Since Replicated Maps will usually contain different types of Replicated Data values, add accessor methods for each of the Replicated Data types which take care of the (unchecked) casting and creating default values with the factory.


octonato (approved):
LGTM

PR #364, opened 2021-09-13
chore: Update .scalafmt.conf

patriknw:
Should be a decent time to do this now.


.scalafmt.conf
```diff
@@ -1,24 +1,20 @@
-version = 2.1.0
+version = 2.7.5
```
raboof, line 1: Should we go all the way to 3.0.3?

raboof (approved):
Now seems to be a good time, perhaps we want to go all the way to 3.0.3?

.scalafmt.conf
patriknw: ok, I tried that, see https://github.com/lightbend/akkaserverless-java-sdk/pull/364/commits/f41a9ef1fecb9f9341bc6444773a8de585a95692 and https://github.com/lightbend/akkaserverless-java-sdk/pull/364/commits/0c6df07e9fd6337573fa512393465ec5861a0caa

testkit/src/main/scala/com/akkaserverless/javasdk/testkit/impl/TestKitValueEntityContext.scala
```diff
@@ -20,8 +20,7 @@ import com.akkaserverless.javasdk.ServiceCallFactory
 import com.akkaserverless.javasdk.valueentity.ValueEntityContext
 
 /**
- * INTERNAL API
- * Used by the generated testkit
+ * INTERNAL API Used by the generated testkit
```
raboof, line 23: Hmm, perhaps use an extra newline? Not important though.

raboof (approved):

testkit/src/main/scala/com/akkaserverless/javasdk/testkit/impl/TestKitValueEntityContext.scala
patriknw: yeah, that wasn't so helpful formatting, but not important

octonato (approved):

PR #367, opened 2021-09-13
fix: Use specific archetypeVersion instead of LATEST

patriknw:
* seems like LATEST is not resolving to 0.7.0
* don't know why, but it's better to use the released version

Fixes https://github.com/lightbend/akkaserverless-java-sdk/issues/360

### DIFF
--- a/docs/src/modules/java/pages/developer-tools.adoc
+++ b/docs/src/modules/java/pages/developer-tools.adoc
@@ -18,7 +18,7 @@ mvn \
   archetype:generate \
   -DarchetypeGroupId=com.lightbend \
   -DarchetypeArtifactId=maven-archetype-akkasls \
-  -DarchetypeVersion=LATEST
+  -DarchetypeVersion={akkaserverless-java-sdk-version}
 ----
 
 Windows::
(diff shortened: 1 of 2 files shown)


docs/src/modules/java/pages/developer-tools.adoc
```diff
@@ -18,7 +18,7 @@ mvn \
   archetype:generate \
   -DarchetypeGroupId=com.lightbend \
   -DarchetypeArtifactId=maven-archetype-akkasls \
-  -DarchetypeVersion=LATEST
+  -DarchetypeVersion={akkaserverless-java-sdk-version}
```
patriknw, line 21: note that akkaserverless-java-sdk-version is set via the Makefile

raboof (approved):

andreaTP (approved):

docs/src/modules/java/pages/developer-tools.adoc
```diff
@@ -18,7 +18,7 @@ mvn \
   archetype:generate \
   -DarchetypeGroupId=com.lightbend \
   -DarchetypeArtifactId=maven-archetype-akkasls \
```
pvlugter, line 20: these are the wrong coordinates?

pvlugter: fixed in #375

PR #370, opened 2021-09-13
Remove valueentity-generated

### LINKED ISSUE #245: Remove valueentity-counter-generated
or explain why it adds any value

I think the dev roundtrip described in https://github.com/lightbend/akkaserverless-java-sdk/blob/main/CONTRIBUTING.md works fine.

We will have a sbt plugin for the Scala SDK and then we can probably use that also for Java codegen in some builds.

octonato:
Fixes #245


build.sbt
```diff
@@ -123,36 +123,6 @@ lazy val codegenJavaCompilationTest = project
     name := "akkaserverless-codegen-java-compilation-tests",
     Compile / javacOptions ++= Seq("-encoding", "UTF-8", "-source", "11", "-target", "11"))
 
-lazy val samples = project
-  .in(file("samples"))
-  .aggregate( // samples relying on codegen must use Maven
-    // FIXME consider including this sample again?
-    // `java-eventing-shopping-cart`,
-    // `valueentity-counter`
-  )
```
octonato, line 132: The samples that can be compiled with sbt are already included directly. All the others are build with maven only by CI. We can remove this root project.

raboof: AFAICT the `valueentity-counter` *can* be compiled from sbt when we put the generated code in `valueentity-generated`. That would make it easier to test changes to the runtime that don't affect the codegen (since you could rebuild and run from sbt instead of doing the 'publishLocal' dance). Whether that's worth it is a judgement call, of course.

patriknw: I don't think it's worth keeping it in the sbt build. `publishM2` works fine and then development in the sample projects with their real toolchain.

Note that when developing the version typically don't change (you keep the same sbt session alive) so publishM2 produce the same dev version.

build.sbt
```diff
@@ -2,7 +2,7 @@ import Dependencies._
 
 lazy val `akkaserverless-java-sdk` = project
   .in(file("."))
-  .aggregate(sdk, testkit, tck, codegenCore, codegenJava, codegenJavaCompilationTest, samples)
+  .aggregate(sdk, testkit, tck, codegenCore, codegenJava, codegenJavaCompilationTest)
```
patriknw, line 6: Can more things be removed/simplifed in the build? I think stuff related to this was added in https://github.com/lightbend/akkaserverless-java-sdk/pull/188
e.g. ReflectiveCodeGen.scala

Hopefully the problem with that one has to `clean` after changing `codegenJava` will go away? That is more of a time sink than `publishM2`.

raboof: `codegenJavaCompilationTest` (#188) is separate from `valueentity-generated`. The problem that one has to `clean` after changing `codegenJava` is related to `codegenJavaCompilationTest`, not to `valueentity-generated`.

patriknw: ok, thanks for clarifying

patriknw (approved):
LGTM

Issue #122, opened 2021-07-08
Document how to publish to a topic


rstento:
From @patriknw: Can't see that we describe how to publish to a topic in topic-eventing.adoc?

From @ennru: Correct, that is pending. My idea would be to add an external publish on cart checkout.
That's a good place to show entity id access in a listener.

From @viktorklang: @rstento I want to raise this as a very high priority for Open Beta. Not being able to publish messages makes the Eventing functionality very, very, limited.

octonato:
done in #373

Issue #121, opened 2021-07-08
Document Eventing for Value Entities


rstento:
lightbend/akkaserverless-framework#250 introduced the capability to listen to updates of Value Entities similar to an Event log of Event Sourced entities.

This should be illustrated for publishing to topics.

octonato:
done in #373

Issue #297, opened 2021-09-07
Enable it test for eventing sample


patriknw:
I tried in https://github.com/lightbend/akkaserverless-java-sdk/pull/294 but there is something with pubsub that is not working. Is AkkaServerlessTestkitResource supposed to start the pubsub emulator automatically?

johanandren:
While working on it I only saw them pass locally when manually starting the pub/sub emulator, which is what is mentioned in the README of that sample.

pvlugter:
I think integration testkit just has support for connecting to a running pub/sub emulator. Originally from https://github.com/lightbend/akkaserverless-framework/pull/573.

PR #373, opened 2021-09-13
doc: publishing and subscribing with Actions

octonato:
Major re-write of documentation on publishing and subscribing with Actions. 

Include: 

* subscribing to journal and publishing to topic
* subscribing to value entity state changes
* subscribing to topic with known protobuf type
* subscribing to topic with json message (points to serialization page)
* subscribing to topic support CloudEvents (points to serialization page)
* Ignoring events
* subscribing and forward to somewhere (act upon style)

Refs: #121, #122 and #222


docs/src/modules/java/pages/actions-publishing-subscribing.adoc
```diff
@@ -0,0 +1,46 @@
+= Publishing and Subscribing with Actions
+
+A very common use case when building Microservices is to publish and subscribe to a stream of events. 
+
+With Actions you can: 
+* subscribe to events emitted by an entity within the same service.
```
patriknw, line 6: do we need to clarify that it's also possible to subscribe to state changes from value entities?

docs/Makefile
```diff
@@ -45,28 +47,21 @@ apidocs:
...
-	rsync -a --exclude-from=.examplesignore ../samples/java-eventsourced-shopping-cart/ "${managed_examples}/java-eventsourced-shopping-cart/"
-	rsync -a --exclude-from=.examplesignore ../samples/java-replicatedentity-shopping-cart/ "${managed_examples}/java-replicatedentity-shopping-cart/"
-	rsync -a --exclude-from=.examplesignore ../samples/java-valueentity-shopping-cart/ "${managed_examples}/java-valueentity-shopping-cart/"
-	rsync -a --exclude-from=.examplesignore ../samples/replicatedentity-examples/ "${managed_examples}/replicatedentity-examples/"
-	rsync -a --exclude-from=.examplesignore ../samples/valueentity-counter/ "${managed_examples}/valueentity-counter/"
-	rsync -a --exclude-from=.examplesignore ../samples/fibonacci-action/ "${managed_examples}/fibonacci-action/"
-	rsync -a --exclude-from=.examplesignore ../samples/doc-snippets/ "${managed_examples}/doc-snippets/"
-	rsync -a --exclude-from=.examplesignore ../samples/first-service/ "${managed_examples}/first-service/"
-
+	rsync -a --exclude-from=.examplesignore ../samples/ "${managed_examples}/"
```
octonato, line 50: just sync all the samples

docs/src/modules/java/pages/actions-publishing-subscribing.adoc
```diff
@@ -0,0 +1,46 @@
...
+----
+<1> We convert the incoming domain event `CounterDomain.ValueIncreased` to the outgoing topic API `CounterTopicApi.Increased`.
+<2> We can access the id of the originating entity throught the event subject metadata.
+<3> We use the converted object to reply.
+<4> Similar code should be written for the decrease method. Omitted here for conciseness. 
+<5> Finally, the 'ignoreReset' method is implemented with `effects().noReply()` signaling that it's a noops call.
+
+=== Subscribing and acting
+
+In example above, an Action is used to consume events from the journal, transform them to some external format and publish them to a topic. Another possible usage for Actions is to consume events and act upon. For example, you may consume events from one entity, transform to commands and send to an another entity or an external system. For that purpose, it's enough to add the `(akkaserverless.method).eventing.in` and omit the `(akkaserverless.method).eventing.out`, like we did with the `ignoreReset` method.
```
octonato, line 46: Not sure if we need a concrete example here. Maybe we can link to the forward example.

patriknw (commented):
looking good

docs/src/modules/java/pages/actions-publishing-subscribing.adoc
octonato: yeah, I think so. I will expand.

docs/src/modules/java/pages/value-entity.adoc
```diff
@@ -23,7 +23,7 @@ The following `counter_domain.proto` file defines our "Counter" Value Entity. Th
 [source,proto]
 .src/main/proto/value-entities/counter_domain.proto
 ----
-include::example$valueentity-counter/src/main/proto/value-entities/counter_domain.proto[]
+include::example$java-valueentity-counter/src/main/proto/value-entities/counter_domain.proto[]
```
octonato, line 26: I prefix the `valueentity-counter` with "java" in this PR. I'm planning to prefix all samples in another PR to accomodate for the upcoming scala sdk. But since I touched that sample here, it was better to prefix it now to avoid conflict.

octonato:
Tests failed because of #297. I had to disable the integration tests.

docs/src/modules/java/pages/actions-publishing-subscribing.adoc
```diff
@@ -0,0 +1,127 @@
...
+The class `CounterJournalToTopicAction` gets generated for us based on the proto file defined above.
+
+[source,protobuf,indent=0]
+.src/main/java/com/example/actions/CounterJournalToTopicAction.java
+----
+include::example$java-eventsourced-counter/src/main/java/com/example/actions/CounterJournalToTopicAction.java[tag=counter-topic]
+----
+<1> We convert the incoming domain event `CounterDomain.ValueIncreased` to the outgoing topic API `CounterTopicApi.Increased`.
+<2> We use the converted object to reply.
+
```
patriknw, line 42: would be good with a sentence that clarifies that it is the message in the `reply`  that will be published to the topic (when using eventing.out)

patriknw: ok, it's mentioned in the callouts above, but that is slightly wrong in saying returned value

> eventing.out annotation indicates that the returned value from this method should be published to the topic

patriknw: could just repeat it here "<2> We use the converted object as the reply, which will be published to the topic"

patriknw (approved):
LGTM, just a minor suggestion

docs/src/modules/java/pages/actions-publishing-subscribing.adoc
octonato: yep, I will adapt

octonato: Done in 2322358

docs/src/modules/java/nav.adoc
```diff
@@ -8,8 +8,7 @@
 *** xref:java:replicated-entity.adoc[Implementing Replicated Entities]
 *** xref:java:actions.adoc[Implementing Actions]
 *** xref:java:actions-as-controller.adoc[Actions as Controllers]
-*** xref:java:entity-eventing.adoc[Subscribing to a Journal]
-*** xref:java:topic-eventing.adoc[Publishing and Subscribing]
+*** xref:java:actions-publishing-subscribing.adoc[Publishing and Subscribing with Actions]
```
johanandren, line 11: No problems with that being a tad long in the left nav menu when rendered?

docs/src/modules/java/pages/actions-publishing-subscribing.adoc
```diff
@@ -0,0 +1,127 @@
...
+Another possible usage for Actions is to consume events and act upon. 
+
+For example, you may consume events from one entity or from a topic, transform to commands and send to an another entity or an external system. This is similar to the usage explained in xref:actions-as-controller.adoc[Actions as Controller], except that the Action is driven by the flow of incoming events instead of external user requests.
+
+For that purpose, it's enough to add the `(akkaserverless.method).eventing.in` and omit the `(akkaserverless.method).eventing.out`.
+
+== Accessing the Entity ID
+
+For many use cases, a subscriber to an event log will trigger other services and needs to pass the entity ID to the receiver. The events of an Event Sourced entity, by design, do not include the entity ID, but it is made available to the subscriber via the CloudEvent metadata field `subject`, for example via the `eventSubject` in the  link:{attachmentsdir}/api/com/akkaserverless/javasdk/action/ActionContext.html[`ActionContext`{tab-icon}, window="new"].
```
johanandren, line 99: Should we skip mentioning "cloud event" here since there is no more information about what that is here, and just say:
```suggestion
For many use cases, a subscriber to an event log will trigger other services and needs to pass the entity ID to the receiver. The events of an Event Sourced entity, by design, do not include the entity ID, but it is made available to the subscriber via the metadata field `subject`, accessible through `eventSubject` in the  link:{attachmentsdir}/api/com/akkaserverless/javasdk/action/ActionContext.html[`ActionContext`{tab-icon}, window="new"]. 
```

or include that partial somewhere here?

docs/src/modules/java/pages/actions-publishing-subscribing.adoc
```diff
@@ -0,0 +1,127 @@
...
+
+[source,protobuf,indent=0]
+.src/main/java/com/example/actions/CounterJournalToTopicAction.java
+----
+include::example$java-eventsourced-counter/src/main/java/com/example/actions/CounterJournalToTopicAction.java[tag=counter-topic-event-subject]
+----
+
+== Ignoring events
+
+When consuming events, each event must be matched by a Protobuf service method. In case your component is interested only in certain events, you may declare a method to receive all events that are not received by the other methods.
```
johanandren, line 111: "If all an event type is not handled the topic will be stuck and not progress" or some other explanation of what happens if this is missing?

docs/src/modules/java/nav.adoc
octonato: We get a break line, that's all. But happy to change if we find another option. 

Maybe only "Publishing and Subscribing" ?

octonato:
```suggestion
*** xref:java:actions-publishing-subscribing.adoc[Publishing and Subscribing]
```

docs/src/modules/java/pages/actions-publishing-subscribing.adoc
octonato:
```suggestion
When consuming events, each event must be matched by a Protobuf service method. In case your component is only interested in certain events, you may declare a method to receive all events that are not received by the other methods. If an event type is not handled, the Action will fail. Actions are designed to restart, but since the handler is missing, it will fail again. Therefore, it's important to define a such catch-all method in case you want to discard some events.
```

octonato: @johanandren, what do you think about the above phrasing?

johanandren (approved):

PR #377, opened 2021-09-14
feat: support protobuf scalar types for replicated entity codegen

### LINKED ISSUE #301: Support scalar types in codegen
Should we allow specifying domain types as scalar value types, rather than always being a message type? Probably most relevant to replicated entities, where something like string keys for maps may be common. This could be specified using the protobuf scalar value types, and then translated to an appropriate type in the language SDK.

So where it's currently:

```protobuf
option (akkaserverless.file).replicated_entity = {
  name: "SomeCounterMap"
  entity_type: "some-counter-map"
  replicated_counter_map: {
    key: "SomeKey"
  }
};

message SomeKey {
  string key = 1;
}
```

It could instead be:

```protobuf
option (akkaserverless.file).replicated_entity = {
  name: "SomeCounterMap"
  entity_type: "some-counter-map"
  replicated_counter_map: {
    key: "string"
  }
};
```

pvlugter:
Resolves #301. Support scalar types in the codegen annotation for replicated entities. Update some doc examples to use scalar value types instead of message types.

### DIFF
--- a/codegen/core/src/main/scala/com/lightbend/akkasls/codegen/ModelBuilder.scala
+++ b/codegen/core/src/main/scala/com/lightbend/akkasls/codegen/ModelBuilder.scala
@@ -89,7 +89,64 @@ object ModelBuilder {
   /**
    * Type argument for generic replicated data types with type parameters.
    */
-  case class TypeArgument(fqn: FullyQualifiedName)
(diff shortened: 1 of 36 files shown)


patriknw (approved):
LGTM

PR #386, opened 2021-09-14
Build structure, first step

RayRoestenburg:


RayRoestenburg:
There are some deps from java-sdk to core, Effect and other types, so first just moving everything to sdkJava, and get the new build structure in place (more sub projects), copy pasted quite a bit, which can be fixed over time.

RayRoestenburg:
Requesting review a little early, would like to get this in quickly, since the moves are quite large..

build.sbt
```diff
@@ -2,10 +2,46 @@ import Dependencies._
...
+  .aggregate(sdkCore, sdkJava, sdkScala, testkitJava, tck, codegenCore, codegenJava, codegenJavaCompilationTest)
 
-lazy val sdk = project
-  .in(file("sdk"))
+//FIXME duplicating most settings in sdkCore, sdkJava, sdkScala for now, cleanup later.
+lazy val sdkCore = project
+  .in(file("sdk/core"))
+  .enablePlugins(AkkaGrpcPlugin, BuildInfoPlugin, PublishSonatype)
+  .settings(
+    name := "akkaserverless-java-sdk-core",
```
raboof, line 12: maybe `akkaserverless-jvm-sdk` or `akkaserverless-runtime`?

build.sbt
```diff
@@ -2,10 +2,46 @@ import Dependencies._
...
+    crossPaths := false,
+    buildInfoKeys := Seq[BuildInfoKey](
```
raboof, line 14: do we need `BuildInfo` in all 3 artifacts?

build.sbt
```diff
@@ -2,10 +2,46 @@ import Dependencies._
...
+      val javaSourceDir = (Compile / javaSource).value.getAbsolutePath
+      (Compile / doc / sources).value.filter(_.getAbsolutePath.startsWith(javaSourceDir))
+    },
+    // javadoc (I think java 9 onwards) refuses to compile javadocs if it can't compile the entire source path.
+    // but since we have java files depending on Scala files, we need to include ourselves on the classpath.
+    Compile / doc / dependencyClasspath := (Compile / fullClasspath).value,
+    Compile / compile / javacOptions ++= Seq("--release", "8"),
+    Compile / scalacOptions ++= Seq("-release", "8"),
+    Compile / akkaGrpcGeneratedSources := Seq(AkkaGrpc.Server),
+    Compile / akkaGrpcGeneratedLanguages := Seq(AkkaGrpc.Scala), // FIXME should be Java, but here be dragons
```
raboof, line 33: do we need Akka gRPC in all projects/configurations?

raboof (approved):
good to get the initial structure in 'early' to avoid conflicts, but let's not wait too long with cleaning up further ;)

build.sbt
RayRoestenburg: I like the jvm-sdk suggestion

build.sbt
RayRoestenburg: I'm not sure, first step was to copy paste, and remove later. I can remove them from core. I guess they should be there for scala and java sdk.

build.sbt
RayRoestenburg: I'll remove this from core now, though I need to look deeper into where stuff is generated etc.

RayRoestenburg: (Since there is nothing in core yet, can always bring it back when needed)

raboof: it looks like this is the server that the sdk provides to the proxy, so perhaps this should later move along with the implementation to core

PR #387, opened 2021-09-14
Make all codegen warnings fatal

raboof:
Since in sbt we don't have a good way to signal them anyway


ennru (approved):
LGTM.

codegen/java-gen/src/main/scala/com/lightbend/akkasls/codegen/java/SourceGenerator.scala
```diff
@@ -70,11 +70,9 @@ object SourceGenerator {
       case service: ModelBuilder.EntityService =>
         model.entities.get(service.componentFullName) match {
           case None =>
-            // TODO perhaps we even want to make this an error, to really go all-in on codegen?
-            log.warning(
+            throw new IllegalStateException(
```
octonato, line 73: Nice! 

But surprising that we have only one warning.

PR #389, opened 2021-09-14
Document inter service calls

johanandren:
Refs https://github.com/lightbend/akkaserverless-docs/issues/903

Works in dev and soon everywhere (depending on infra fix in https://github.com/lightbend/akkaserverless/pull/5056 )


docs/src/modules/java/pages/call-another-as-service.adoc
```diff
@@ -0,0 +1,118 @@
...
+
+It is possible to use any gRPC client you want, but since Akka gRPC is already used by the SDK we recommend using it.
+
+Since the proto file of an Akka Serverless service contains annotations that cause the Java SDK code generation to generate services we need to start by removing those. Copy the API definition proto file from the other service into the `proto` directory but strip all `akkaserverless.service` option blocks as well as all other Akka Serverless annotations and the `import "akkaserverless/annotations.proto"` from it.
+
+The build needs to be updated to generate Akka gRPC client classes. In the existing configuration for the `protobuf-maven-plugin` in your `pom.xml` add the following:
+
+---
+// FIXME this block is a bit unwieldy, but har to understand where/what to add if not
+// FIXME should we _always_ include something instead, like a src/main/ex-proto and avoid clients for the service itself?
```
patriknw, line 18: I think we need clients for integration tests. I don't know why we have to be restrictive in generating clients. Doesn't hurt to have all of them?

Now we have `grpc-client` execution for `test-compile` in the pom.xml. We could have that for `compile` by default?

docs/src/modules/java/pages/call-another-as-service.adoc
```diff
@@ -0,0 +1,118 @@
+= Calling other Akka Serverless services
```
patriknw, line 1: I think we want to have this in the JavaScript and Java specific pages instead, i.e. in the sdk repo.

docs/src/modules/java/pages/call-another-as-service.adoc
```diff
@@ -0,0 +1,118 @@
...
+public class ProxyAction extends AbstractProxyAction {
+
+  private final CounterServiceClient counterClient;
+
+  public ProxyAction(ActorSystem system, ActionCreationContext creationContext) { // <1>
+    // Configure the client by code:
+    GrpcClientSettings settings =
+        GrpcClientSettings.connectToServiceAt("my-service", 80, system) // <2>
+          .withTls(false); // <3>
+    counterClient = CounterServiceClient.create(settings, system);
```
patriknw, line 75: This is the part I think we should provide some convenience for in the Java SDK. To also solve the client re-use problem we could have a `actionContext().getGrpcClient(CounterService.class, "my-service", 80)`.

Alternatively we could have corresponding `createGrpcClient` in `AkkaServerless` and it has to be passed in to the components in a constructor parameter. That DI style might be better for testing (and should be a familiar style for Java devs).

docs/src/modules/java/pages/call-another-as-service.adoc
patriknw: sorry, that is already the case

docs/src/modules/java/pages/call-another-as-service.adoc
octonato: I was think on something like this: 

For external services, user will need to register clients to `AkkaServerless`. We take care of setup and teardown. 

Clients for local components are registered automatically for the user. They don't need to do anything. Seamless experience.

Then, when it comes to use it, user would call: 

```scala
actionContext().clientFor(CounterService.class)
```
`CounterService` could be local or not. At that moment it doesn't matter anymore. 

User could choose to lookup via context or inject via constructor. If they want to inject via constructor, they will first need to register the client (if non-local) and then look it up before registering the component where they want to register it.

octonato: And ideally, users should not have access to the actor system.

docs/src/modules/java/pages/call-another-as-service.adoc
johanandren: Yeah, sure, we could, I guess it is about putting stuff in the prod classpath that will actually never be used, but might not matter _that_ much.

docs/src/modules/java/pages/call-another-as-service.adoc
johanandren: It would have to be something like Patrik suggested because the name the service is deployed as is not possible to deduce from the generated client class.

octonato: We can use `localhost:proxyPort` and avoid going out and back for local components.

johanandren: this is for inter-service though, so two separate deployments - two separate clusters.

johanandren: Updated with an api on Context much like what Patrik suggested, covering both internal Akka Serverless gRpc consumption and external - if the name is present in the `akka.grpc.client` config, that is used (external service) and if not, it is expected to be an internal service over the service mesh mathingy.

docs/src/modules/java/pages/call-another-as-service.adoc
johanandren: All samples and archetypes now updated to generate Akka gRPC clients

docs/src/modules/java/pages/call-another-service.adoc
```diff
@@ -0,0 +1,59 @@
+= Calling other services
+
+// FIXME this does not actually work (yet?) https://github.com/lightbend/akkaserverless-docs/issues/903#issuecomment-919174111
+
+== Akka Serverless services
+// FIXME possible from the same project only, or cross projects as well?
```
pvlugter, line 6: I think it's just in the same project / namespace.

docs/src/modules/java/pages/call-another-service.adoc
```diff
@@ -0,0 +1,59 @@
...
+<3> Instead we register a callback to execute once the increase call completes successfully, here we ignore the empty increase response and call `getCurrentCounter`, this makes the two calls _sequential_. (Note however that some other client calling the counter API could have changed the counter in between the calls)
+<4> Once we get a response for the `getCurrentCounter` we transform that into a `Response` - the response type of this action method.
+<5> We can reply with a `CompletionStage<Reply>` using `effects().asyncReply`. Once the `CompletionStage` completes, the client will get the response back.
+
+For additional documentation about Akka gRPC see the https://doc.akka.io/docs/akka-grpc/current/index.html[Akka gRPC documentation].
+
+== External gRPC services
+
+Calling an external service is done the same way as described above, with the difference that the name passed to `getGrpcClient` is present in the `application.conf` file of the project with details on how to connect to the external service. The for a service named `external-example-service` config is expected to be found under `akka.grpc.client.external-example-service`.
```
pvlugter, line 49:
```suggestion
Calling an external service is done the same way as described above, with the difference that the name passed to `getGrpcClient` is present in the `application.conf` file of the project with details on how to connect to the external service. For a service named `external-example-service` config is expected to be found under `akka.grpc.client.external-example-service`.
```

sdk/java-sdk/src/main/java/com/akkaserverless/javasdk/Context.java
```diff
@@ -16,8 +16,23 @@
...
+
 /** Root class of all contexts. */
 public interface Context {
   /** Get the service call factory for this stateful service. */
   ServiceCallFactory serviceCallFactory();
+
+  /**
+   * Get an Akka gRPC client for the given service name. The same client instance is shared across
+   * components in the application. The lifecycle of the client is managed by the SDK and it should
+   * not be stopped be user code.
```
pvlugter, line 29:
```suggestion
   * not be stopped by user code.
```

docs/src/modules/java/pages/call-another-as-service.adoc
octonato: So, maybe you didn't understand what I was suggesting. 

For inter-services calls, the user need to register a client programmatically or create on the fly based on some config (like in this PR). For that I agree, the user needs to have a name for it for the service discovery. 

For local components, instead of requiring a name to be passed, the user could simply call `clientFor(SomeService.class)` and hit the proxy directly (no service mesh). 

That said, the method `getGrpcClient(Class<T> clientClass, String service)`, could be simplify to `getGrpcClient(Class<T> clientClass)` (I personally prefer `clientFor` though). 

The point being, why bother with a service name for local components? I think we can have a solution where the serviceName and lookup is not required all the time.

For external services, the config key could be based on the FQCN. If we find a config for the class, we use it, the service name should be registered there.

johanandren: For local calls I think it shouldn't be a lookup with class like this at all but rather a class/object that is codegen-generated with clients for each of the components in the service itself. I think it is out of scope for this PR though.

johanandren: The FQCN of the class isn't necessarily the identity of a running external service, multiple servers could implement the same protobuf protocol, so I think it is better to have an explicit name than using the generated client class name, since you anyway have to add config block I don't think there is a lot of mental dev overhead there.

docs/src/modules/java/pages/call-another-service.adoc
jroper: Cross project communication is blocked by firewall rules.

johanandren: Thanks for verifying, I added that for cross project you have to consume the public API/routes of the service.

docs/src/modules/java/pages/call-another-service.adoc
```diff
@@ -0,0 +1,57 @@
...
+
+Calling other Akka Serverless services in the same project from an Action is done by invoking them over gRPC much like how an external client would. The service is however identified only by the name it has been deployed as, Akka Serverless takes care of routing requests to the service and keeping the data safe by encrypting the connection for us.
+
+In this sample we will make an action that does two sequential calls to the xref:value-entity.adoc[Value Entity Counter] service, deployed with the service name "counter".
+
+We start by adding the public API of the counter to the `src/main/proto` directory of our project.
+
+Since the proto file of an Akka Serverless service contains annotations that cause the Java SDK code generation to generate services, and we only want to consume the service, we need to start by removing the annotations.
+
+Copy the API definition proto file from the other service into the `proto` directory but strip all `akkaserverless.service` option blocks as well as all other Akka Serverless annotations and the `import "akkaserverless/annotations.proto"` from it.
```
patriknw, line 15: is it clear that "strip all" means "remove all"? otherwise we could say less fancy "remove all"

docs/src/modules/java/pages/call-another-service.adoc
```diff
@@ -0,0 +1,57 @@
...
+For additional documentation about Akka gRPC see the https://doc.akka.io/docs/akka-grpc/current/index.html[Akka gRPC documentation].
+
+== External gRPC services
+
+Calling an Akka Serverless service in another project, or an arbitrary external gRPC service is done the same way as described above, with the difference that the service name string passed to `getGrpcClient` is present in the `application.conf` file of the project with details on how to connect to the external service. For a service named `external-example-service` config is expected to be found under `akka.grpc.client.external-example-service`.
+
+See the https://doc.akka.io/docs/akka-grpc/current/client/configuration.html#by-configuration[Akka gRPC docs on consuming a service] for details on configuring the client for a service name.
+
+// FIXME
+// == External HTTP services
```
patriknw, line 52: we can leave this out, until such feature is requested

docs/src/modules/java/pages/call-another-service.adoc
```diff
@@ -0,0 +1,57 @@
...
+.src/main/java/com/example/DelegatingServiceAction.java
+----
+include::java:example$java-doc-snippets/src/main/java/com/example/DelegatingServiceAction.java[tag=delegating-action]
+----
+
+<1> We call `actionContext().getGrpcClient` with the client class of the API and the name that the service was deployed with
+<2> Calling a unary gRPC method returns the reply as a `java.util.concurrent.CompletionStage`. When execution continues through our method, the result will likely not yet have arrived, so we cannot look at it like a regular variable.
+<3> Instead we register a callback to execute once the increase call completes successfully, here we ignore the empty increase response and call `getCurrentCounter`, this makes the two calls _sequential_. (Note however that some other client calling the counter API could have changed the counter in between the calls)
+<4> Once we get a response for the `getCurrentCounter` we transform that into a `Response` - the response type of this action method.
+<5> We can reply with a `CompletionStage<Reply>` using `effects().asyncReply`. Once the `CompletionStage` completes, the client will get the response back.
```
patriknw, line 41: in the code it is using `Result` not `<Reply>`

samples/java-doc-snippets/src/main/java/com/example/DelegatingServiceAction.java
```diff
@@ -0,0 +1,41 @@
...
+  public Effect<DelegatingServiceApi.Result> addAndReturn(DelegatingServiceApi.Request request) {
+    CounterServiceClient counterClient = actionContext().getGrpcClient(CounterServiceClient.class, "counter"); // <1>
+
+    CounterApi.IncreaseValue increaseValue = CounterApi.IncreaseValue.newBuilder()
+        .setCounterId(request.getCounterId())
+        .setValue(1)
+        .build();
+    CompletionStage<Empty> increaseCompleted = counterClient.increase(increaseValue);  // <2>
+
+    CompletionStage<CounterApi.CurrentCounter> currentCounterValueAfter = increaseCompleted.thenCompose((__) -> // <3>
```
patriknw, line 29: that `(__)` looks a bit fancy (foreign)? `notUsed ->` instead?

sdk/java-sdk/src/main/scala/com/akkaserverless/javasdk/impl/GrpcClients.scala
```diff
@@ -0,0 +1,86 @@
...
+  private val clients = new ConcurrentHashMap[Key, AkkaGrpcClient]()
+
+  CoordinatedShutdown(system).addTask(CoordinatedShutdown.PhaseServiceStop, "stop-grpc-clients")(() =>
+    Future.traverse(clients.values().asScala)(_.close().asScala).map(_ => Done))
+
+  def getGrpcClient[T <: AkkaGrpcClient](clientClass: Class[T], service: String): T =
+    clients.computeIfAbsent(Key(clientClass, service), createClient(_)).asInstanceOf[T]
+
+  private def createClient(key: Key): AkkaGrpcClient = {
+    val settings = if (!system.settings.config.hasPath(s"""akka.grpc.client."${key.service}"""")) {
```
patriknw, line 63: clever way to differentiate between external and internal

sdk/java-sdk/src/main/scala/com/akkaserverless/javasdk/impl/GrpcClients.scala
```diff
@@ -0,0 +1,86 @@
...
+        // (TLS is handled for us by Akka Serverless infra)
+        .withTls(false)
+    } else {
+      log.debug("Creating gRPC client for external service [{}]", key.service)
+      // external service, defined in config
+      GrpcClientSettings.fromConfig(key.service)(system)
+    }
+
+    val create = key.clientClass.getMethod("create", classOf[GrpcClientSettings])
+    val client: AkkaGrpcClient = create.invoke(null, settings).asInstanceOf[AkkaGrpcClient]
```
patriknw, line 77: We have the client class as the clientClass and return value. It would have been nice if that could be the service interface instead. `CounterService` instead of `CounterServiceClient`. Because then we could more easily support test fakes for clients (in the future).

Since we anyway do a bit of reflection here, could we derive the client class from the service interface (by naming convention adding "Client"?

patriknw (commented):
looking good, a few thoughts...

sdk/java-sdk/src/main/scala/com/akkaserverless/javasdk/impl/GrpcClients.scala
johanandren: We could, but then we'd have no guarantee from the type system that what the user is passing is actually an Akka gRPC client. That's not an amazing type system guarantee though, so maybe that is ok.

samples/java-doc-snippets/src/main/java/com/example/DelegatingServiceAction.java
johanandren: I thought `__` was the canonical way to mark not used parameters in Java land, but sure.

patriknw: if that's the way it's done by others (not Scala ppl) I'm fine with that

johanandren: I can't find any reference to it on the line, so maybe I misremember (`_` did become a keyword in Java 9 though, so maybe I heard some JVM talk about future plans for that)

sdk/java-sdk/src/main/scala/com/akkaserverless/javasdk/impl/GrpcClients.scala
johanandren: Ok, updated to take the service interface instead.

patriknw: Thanks for that change. Good to not expose more than needed. I think the risk of passing in wrong class is low and would have immediate error feedback.

patriknw (approved):
LGTM

patriknw:
The commit message should be something like
```
feat(javasdk): Add factory of gRPC client for inter-service communication

* added getGrpcClient in context
* documentation
* always generate grpc client
```

Issue #379, opened 2021-09-14
sbt plugin for Scala codegen


raboof:
following the same approach as in Akka gRPC: bringing in sbt-protoc transitively and registering the codegen (for now a dummy) as a protoc generator:

* [x] sbt-protoc integration #390
* [x] pass proto definitions into 'model' for generation #390
* [x] separate protoc generators for both 'compile' and 'test' scope #390
* [x] create an sbt task for generating the code that the users will then maintain themselves. In the default project template we'll make a dependency between the 'compile' task and this task so the user won't have to worry about this at first, but can disable it when it gets in the way later.

PR #390, opened 2021-09-14
Populate sbt-plugin 'model', split compile/test

raboof:
Refs #379

### DIFF
--- a/codegen/scala-gen/src/main/scala/com/akkaserverless/codegen/scalasdk/DebugPrintlnLog.scala
+++ b/codegen/scala-gen/src/main/scala/com/akkaserverless/codegen/scalasdk/DebugPrintlnLog.scala
@@ -1,0 +1,26 @@
+/*
+ * Copyright 2021 Lightbend Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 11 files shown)


codegen/scala-gen/src/main/scala/com/akkaserverless/codegen/scalasdk/DebugPrintlnLog.scala
```diff
@@ -0,0 +1,26 @@
...
+
+package com.akkaserverless.codegen.scalasdk
+
+import com.lightbend.akkasls.codegen.Log
+
+case class DebugPrintlnLog(debugEnabled: Boolean) extends Log {
+  override def debug(message: String): Unit =
+    if (debugEnabled) println(s"[DEBUG] $message")
+  override def info(message: String): Unit =
+    if (debugEnabled) println(s"[INFO] $message")
```
octonato, line 25: Who need a log framework? 😄

octonato (approved):

codegen/scala-gen/src/main/scala/com/akkaserverless/codegen/scalasdk/DebugPrintlnLog.scala
raboof: Yeah... in Akka gRPC we have a hack to pass the sbt logger to the codegen (https://github.com/akka/akka-grpc/blob/e4abdfa60a5b1d3b7b7e816d05e859e16230db5b/sbt-plugin/src/main/scala/akka/grpc/sbt/AkkaGrpcPlugin.scala#L29), but perhaps we can do without here. We can pass fatal problems out, after all.

PR #392, opened 2021-09-14
doc: fix links to main docs and validate links in CI

pvlugter:
The restructure in the main docs broke some links. There's already a `validate-links` target, add this to CI.

For Cloudstate, we found the external link checking to be flaky, so only ran it manually. But the Cloudstate docs also had a different setup which allowed separate modules to use xrefs rather than external links, so these would be validated already, but here we're using external links to the main docs. We can see how it goes with running it regularly, and how often it needs to be rerun.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -121,7 +121,9 @@ jobs:
       - restore_deps_cache
       - run:
           name: Validate docs
-          command: make -C docs
+          command: |
+            make -C docs
+            make -C docs validate-links
       - save_deps_cache
 
   tests:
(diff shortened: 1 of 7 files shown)


.circleci/config.yml
```diff
@@ -121,7 +121,9 @@ jobs:
       - restore_deps_cache
       - run:
           name: Validate docs
-          command: make -C docs
+          command: |
+            make -C docs
+            make -C docs validate-links
```
raboof, line 126: do we need both?

raboof (approved):

octonato (approved):

PR #394, opened 2021-09-15
prefix all java samples

octonato:
This prefix all existing samples with `java-` in preparation for the upcoming Scala SDK as we will start to write samples for Scala as well. 

I left the valueentity-counter out because it's already done in #373


octonato:
It seems that two of the three replicated entity samples are not used in the docs. There is no link to shopping-cart nor to the counter sample. 

Should we remove them to reduce the number of samples to support?

docs/Makefile
```diff
@@ -46,21 +46,21 @@ apidocs:
...
-	rsync -a --exclude-from=.examplesignore ../samples/fibonacci-action/ "${managed_examples}/fibonacci-action/"
-	rsync -a --exclude-from=.examplesignore ../samples/doc-snippets/ "${managed_examples}/doc-snippets/"
-	rsync -a --exclude-from=.examplesignore ../samples/first-service/ "${managed_examples}/first-service/"
+	rsync -a --exclude-from=.examplesignore ../samples/java-fibonacci-action/ "${managed_examples}/java-fibonacci-action/"
+	rsync -a --exclude-from=.examplesignore ../samples/java-doc-snippets/ "${managed_examples}/java-doc-snippets/"
+	rsync -a --exclude-from=.examplesignore ../samples/java-first-service/ "${managed_examples}/java-first-service/"
 
 bundles:
-	bin/bundle.sh --zip "${managed_attachments}/customer-registry-quickstart.zip" ../samples/customer-registry-quickstart
+	bin/bundle.sh --zip "${managed_attachments}/java-customer-registry-quickstart.zip" ../samples/java-customer-registry-quickstart
```
patriknw, line 63: this will probably require an update to the download link from the quickstart adoc page

patriknw (approved):
LGTM

docs/Makefile
octonato: Good catch.  
Done in https://github.com/lightbend/akkaserverless-docs/pull/920

johanandren (approved):

pvlugter:
> It seems that two of the three replicated entity samples are not used in the docs. There is no link to shopping-cart nor to the counter sample.

I added `java-replicatedentity-shopping-cart` and `java-replicatedentity-examples` for the docs, and they should both be used there. The counter sample is redundant given the new examples for all the replicated data types, and I think that can be removed.

octonato:
@pvlugter, I may be wrong, but I didn't find any references to `replicatedentity-shopping-cart` in the docs.

octonato:
Oh, my bad. Checked one again and it's indeed used in `replicated-entity-crdt.adoc`. I will PR to remove the counter one.

PR #397, opened 2021-09-15
build: remove eventing sample

octonato:
The eventing sample is not being used in the docs anymore and it's very bloated. Time to say au revoir.


patriknw (approved):
LGTM

PR #398, opened 2021-09-15
feat!(scalasdk): WIP

RayRoestenburg:


sdk/scala-sdk/src/main/resources/reference.conf
```diff
@@ -0,0 +1,67 @@
+akkaserverless {
```
patriknw, line 1: This will not work, since we include the reference.conf from javasdk dependency, with the same config properties.
Good news is that it's not needed at all.

sdk/scala-sdk/src/main/scala/com/akkaserverless/scalasdk/Metadata.scala
```diff
@@ -0,0 +1,304 @@
...
+   * @param type
+   *   The type of the CloudEvent.
+   * @return
+   *   This metadata, represented as a CloudEvent with the specified fields.
+   */
+  def asCloudEvent(id: String, source: URI, `type`: String): CloudEvent
+}
+
+object Metadata {
+  def empty = new MetadataImpl(JMetadataImpl.Empty)
```
patriknw, line 256: `val` ?

sdk/scala-sdk/src/main/scala/com/akkaserverless/scalasdk/Metadata.scala
```diff
@@ -0,0 +1,304 @@
...
+  def apply() = empty
```
patriknw, line 257: I think `Metadata.empty` is enough.

sdk/scala-sdk/src/main/scala/com/akkaserverless/scalasdk/Metadata.scala
patriknw, line 256: always define explicit return types in public api: `empty: Metadata`

sdk/scala-sdk/src/main/resources/reference.conf
RayRoestenburg: Cool will remove

sdk/scala-sdk/src/main/scala/com/akkaserverless/scalasdk/Metadata.scala
RayRoestenburg: sure

sdk/scala-sdk/src/main/scala/com/akkaserverless/scalasdk/Metadata.scala
RayRoestenburg: ok

sdk/scala-sdk/src/main/scala/com/akkaserverless/scalasdk/Metadata.scala
RayRoestenburg: ah yes, I forget this every time!

patriknw (commented):
Looking good, let us know when it's in a good shape for first merge. We can add more and fixup incrementally, and that might also make it possible to split the work.

Before starting on all fronts we should probably spike ValueEntity all the way.

patriknw (approved):
LGTM, lets merge this and continue in additional PRs

patriknw:
Sorry to complain about such unimportant things as the commit message
```
feat!(scalasdk): WIP (#398)
```

It's not a breaking change, so it shouldn't have `!`.

The title should be more descriptive, such as
```
feat(scalasdk): First parts of public API for scalasdk, #381
```

Not a big deal at all, but let's improve.

PR #401, opened 2021-09-15
'unmanaged' codegen

raboof:
(draft because some TODO's left)

Not entirely happy with the approach: generating the code from the regular protoc generators, but then 'filtering them out' of the managed sources (that they get added to automatically) afterwards, and copying them over 'manually'. It does save us from duplicating part of the sbt-protoc implementation (in particular finding the correct `.proto` files and converting them into `com.google.protobuf.Descriptors.FileDescriptor` instances). Let's try it out, but if it leads to trouble we should definitely consider avoiding the `generators` and doing that step 'ourselves'.


pvlugter (approved):
The filter on the managed sources does seem a little awkward, but sounds good to start with this approach and see how it goes. And the separate task is good for disabling with `Compile / generateUnmanaged := ()`.

PR #396, opened 2021-09-15
fix: require java_outer_classname, not allow java_multiple_files, #239

patriknw:
* at least for now we don't support java_multiple_files=true and java_outer_classname is required
* better with fail fast with clear error message than weird compilation errors

Refs #239

PR #403, opened 2021-09-16
fix: codegen for outer class and multiple files variations

### LINKED ISSUE #239: Compilation problems with Actions without java_outer_classname
Action codegen generates invalid code if there isn't a `java_outer_classname` option. That's not the case of the other components. 

See https://github.com/lightbend/akkaserverless-java-sdk/pull/237/files#diff-e2cacc5b119d9b80e7513c32929fe8e3603b078893d30718ea82c7024482db8dR19

pvlugter:
Fixes #239 and follow-up to #396.

Code generation now takes into account what the generated outer class will be, with or without the `java_outer_classname` option, and whether generated classes will be wrapped or not based on `java_multiple_files`. Add some compilation tests — add all variations for actions, and a multiple files + default outer class for event sourced and value entities.

### DIFF
--- a/codegen/core/src/main/scala/com/lightbend/akkasls/codegen/Packaging.scala
+++ b/codegen/core/src/main/scala/com/lightbend/akkasls/codegen/Packaging.scala
@@ -28,10 +28,7 @@ case class FullyQualifiedName(protoName: String, name: String, parent: PackageNa
   lazy val fullQualifiedName = s"${parent.javaPackage}.$name"
   lazy val fullName = {
     if (parent.javaMultipleFiles) name
-    else
-      parent.javaOuterClassnameOption
-        .map(outer => s"$outer.$name")
-        .getOrElse(name)
+    else s"${parent.javaOuterClassname}.$name"
   }
 
   lazy val typeImport = s"${parent.javaPackage}.$fullName"
@@ -86,13 +83,29 @@ case class PackageNaming(
     pkg: String,
     goPackage: Option[String],
     javaPackageOption: Option[String],
-    javaOuterClassnameOption: Option[String],
+    javaOuterClassname: String,
     javaMultipleFiles: Boolean) {
   lazy val javaPackage: String = javaPackageOption.getOrElse(pkg)
-  lazy val javaOuterClassname: String = javaOuterClassnameOption.getOrElse(name)
 }
 
 object PackageNaming {
+  def apply(
+      protoFileName: String,
+      name: String,
+      pkg: String,
+      goPackage: Option[String],
+      javaPackageOption: Option[String],
+      javaOuterClassnameOption: Option[String],
+      javaMultipleFiles: Boolean): PackageNaming =
+    PackageNaming(
+      protoFileName,
+      name,
+      pkg,
+      goPackage,
+      javaPackageOption,
+      javaOuterClassnameOption.getOrElse(name),
+      javaMultipleFiles)
+
   def from(descriptor: Descriptors.FileDescriptor): PackageNaming = {
     val name =
       descriptor.getName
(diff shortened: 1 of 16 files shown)


codegen/core/src/main/scala/com/lightbend/akkasls/codegen/Packaging.scala
```diff
@@ -28,10 +28,7 @@ case class FullyQualifiedName(protoName: String, name: String, parent: PackageNa
   lazy val fullQualifiedName = s"${parent.javaPackage}.$name"
   lazy val fullName = {
     if (parent.javaMultipleFiles) name
-    else
-      parent.javaOuterClassnameOption
-        .map(outer => s"$outer.$name")
-        .getOrElse(name)
+    else s"${parent.javaOuterClassname}.$name"
```
pvlugter, line 31: First part of the fix, types will always be wrapped in the outer class if it's not multiple files. And use this `fullName` method in the generation to support the variations (rather than always adding the outer class).

codegen/core/src/main/scala/com/lightbend/akkasls/codegen/Packaging.scala
```diff
@@ -114,11 +127,20 @@ object PackageNaming {
...
         .map(_.toString())
 
+    val javaOuterClassname =
+      javaOuterClassnameOption.getOrElse {
+        val existingNames =
+          descriptor.getMessageTypes.asScala.map(_.getName) ++
+          descriptor.getEnumTypes.asScala.map(_.getName) ++
+          descriptor.getServices.asScala.map(_.getName)
+        if (existingNames.contains(name)) name + "OuterClass" else name
+      }
```
pvlugter, line 142: Second part of the fix, check whether `OuterClass` will be added to the name.

octonato: My memory is failing me. Is that the default behavior when generating java classes from protobuf? 

This is how I read the current PR: 

* if multiple-files = true, each type has it's one file. No wrapper, all is fine.
* if needs a wrapper, the wrapper defaults to file name.
* if name generated from the file name conflicts with service name, wrapper gets `OuterClass` appended to it.

Anyway, I'm fine with it. I'm just wondering if that's an existing convention.

codegen/java-gen/src/main/scala/com/lightbend/akkasls/codegen/java/EntityServiceSourceGenerator.scala
```diff
@@ -160,15 +160,14 @@ object EntityServiceSourceGenerator {
         "com.akkaserverless.javasdk.eventsourcedentity.EventSourcedEntity",
         "com.akkaserverless.javasdk.impl.eventsourcedentity.EventSourcedEntityHandler"))
 
-    val serviceApiOuterClass = service.fqn.parent.javaOuterClassname
-    val outerClassAndState = s"${entity.fqn.parent.javaOuterClassname}.${entity.state.fqn.name}"
```
octonato, line 164: We have been doing too much of string concatenation and repeating it. Good to push that to the internal model! 👍

octonato (approved):

codegen/core/src/main/scala/com/lightbend/akkasls/codegen/Packaging.scala
raboof: Yes, appending `OuterClass` when it is needed for disambiguation is a protoc-java convention, i.e. https://github.com/akka/akka-grpc/commit/8b39a4098f6a12bce2b909f4b94fb00424f63d85 and https://github.com/protocolbuffers/protobuf/blob/4e0a1119c0c5cdfe89b54b9d66cb76223f17861a/src/google/protobuf/compiler/java/java_name_resolver.cc#L157

octonato: thanks for clarifying

patriknw (commented):
Thanks

codegen/core/src/main/scala/com/lightbend/akkasls/codegen/Packaging.scala
pvlugter: Yes, should be aligned with the protobuf generation now. And we also still use the outer class for accessing the descriptor (whether or not it's multiple files enabled), so we need to have the right name for that.

PR #404, opened 2021-09-16
build: remove replicated-counter sample

octonato:


patriknw (approved):
LGTM, already covered by the other samples I guess

PR #405, opened 2021-09-16
fix: align Action / View name convention

octonato:
Refs: https://github.com/lightbend/akkaserverless-java-sdk/issues/238

Align the convention and add `Impl` if name contains `Action|View` (not only at the end)


octonato:
reverting to draft, I messed up with something here

patriknw (commented):
looking good, is it ready?

octonato:
Ready now, but surprisingly the registration order in in valueentity-customer-registry changed. 🤷

samples/java-valueentity-customer-registry/src/main/java/customer/Main.java
```diff
@@ -25,12 +25,11 @@ public static AkkaServerless createAkkaServerless() {
       CustomerValueEntity::new,
       // end::register[]
       CustomerSummaryByNameView::new,
-      CustomerByEmailView::new,
       CustomersResponseByNameView::new,
+      CustomerByEmailView::new,
```
octonato, line 29: So, here the names didn't change, but the ordering changed. Note that the `CustomerS` and `Customers` (plural). 

In any case, the ordering doesn't seen to be deterministic. 

CustomerByEmailView
CustomerActionImpl
CustomerByNameView

Alphabetical? Per component type? File system ordering?

patriknw: I created issue https://github.com/lightbend/akkaserverless-java-sdk/issues/407 for this.

samples/java-eventsourced-customer-registry/src/test/java/customer/domain/CustomerEntityTest.java
```diff
@@ -0,0 +1,65 @@
...
+import customer.api.CustomerApi;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.NoSuchElementException;
+import org.junit.Test;
+import scala.jdk.javaapi.CollectionConverters;
+
+import static org.junit.Assert.*;
+
+public class CustomerEntityTest {
```
octonato, line 20: I compiled all project locally and found some files that were not checked in

patriknw:
To be strict, the commit comment should be `fix!` since it's a breaking change. Not sure how strict we are yet with versioning of such changes. I think @retgits has hinted that we should deliver fast without too much worry of breaking user's code. It's just some cases of Action/View names that will be changed.

octonato:
I changed the commit message and fixed the build, we had a failure in the codegen. 
I will look into #407

patriknw (approved):
LGTM

PR #406, opened 2021-09-16
doc: add note about how to fix local linux development

jrudolph:
https://github.com/lightbend/akkaserverless-framework/issues/581 has more information about future improvements in this regard.

### DIFF
--- a/samples/java-customer-registry-quickstart/README.md
+++ b/samples/java-customer-registry-quickstart/README.md
@@ -49,6 +49,9 @@ To start the application locally, the `exec-maven-plugin` is used. Use the follo
 mvn compile exec:java
 ```
 
+> On Linux, the application will bind to localhost which will not be accessible from the docker containers depending on the
+> docker networking model. Use `HOST=0.0.0.0 mvn compile exec:java`, instead to make the user application bind to all interfaces.
+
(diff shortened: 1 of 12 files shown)


octonato (approved):
Looks good, but we need to add the same to the generated archetypes

https://github.com/lightbend/akkaserverless-java-sdk/blob/main/maven-java/akkaserverless-maven-archetype-value-entity/src/main/resources/archetype-resources/README.md

and 

https://github.com/lightbend/akkaserverless-java-sdk/blob/main/maven-java/akkaserverless-maven-archetype-event-sourced-entity/src/main/resources/archetype-resources/README.md

samples/java-customer-registry-quickstart/README.md
```diff
@@ -49,6 +49,9 @@ To start the application locally, the `exec-maven-plugin` is used. Use the follo
 mvn compile exec:java
 ```
 
+> On Linux, the application will bind to localhost which will not be accessible from the docker containers depending on the
+> docker networking model. Use `HOST=0.0.0.0 mvn compile exec:java`, instead to make the user application bind to all interfaces.
```
raboof, line 53: was this sufficient in your case? on my machine I also have to open up some firewall rules, but perhaps that's uncommon?

jrudolph: Yes, seems it was enough. What kind of firewall do you use or rules do you set?

raboof: > What kind of firewall do you use

I don't really know, the defaults from NixOS :). I can dump the rules if you want?

I ended up just doing `iptables -I INPUT -i the_interface -j ACCEPT` for the interface docker-compose was using.

johanandren:
Superseeded by #436

samples/java-customer-registry-quickstart/README.md
jrudolph: Ok, thanks. Maybe NixOS is just more locked down by default...

PR #414, opened 2021-09-17
Move Service and ComponentOptions to impl.

RayRoestenburg:


sdk/scala-sdk/src/main/scala/com/akkaserverless/scalasdk/ComponentOptions.scala
```diff
@@ -16,5 +16,5 @@
 
 package com.akkaserverless.scalasdk
 
-//FIXME see if needed
+//FIXME see if needed, maybe should be moved to impl
```
patriknw, line 19: if it is moved to impl we should (also) include `withForwardHeaders` and `forwardHeaders` in the more concrete options interfaces, such as EventSourcedEntityOptions

RayRoestenburg: yes will do

RayRoestenburg: These interfaces still extend ComponentOptions in impl, so should be good?

RayRoestenburg:
Ok with this @patriknw ?

sdk/java-sdk/src/main/scala/com/akkaserverless/javasdk/impl/ComponentOptions.scala
```diff
@@ -14,31 +14,25 @@
...
+trait ComponentOptions {
 
   /**
-   * @return the headers requested to be forwarded as metadata (cannot be mutated, use
-   *     withForwardHeaders)
+   * @return
+   *   the headers requested to be forwarded as metadata (cannot be mutated, use withForwardHeaders)
    */
-  Set<String> forwardHeaders();
+  def forwardHeaders(): Set[String]
```
patriknw, line 31: I would be more explicit with that `Set` type. I see that there is an import `java.util.Set` above but I'd rather have the full package here, or rename the import to JSet.

sdk/java-sdk/src/main/scala/com/akkaserverless/javasdk/impl/ComponentOptions.scala
```diff
@@ -14,31 +14,25 @@
...
-package com.akkaserverless.javasdk;
+package com.akkaserverless.javasdk.impl
 
-import com.akkaserverless.javasdk.eventsourcedentity.EventSourcedEntityOptions;
-import com.akkaserverless.javasdk.impl.eventsourcedentity.EventSourcedEntityOptionsImpl;
+import com.akkaserverless.javasdk.eventsourcedentity.EventSourcedEntityOptions
+import com.akkaserverless.javasdk.impl.eventsourcedentity.EventSourcedEntityOptionsImpl
 
 import java.util.Collections;
 import java.util.Set;
```
patriknw, line 23: some semicolon leftovers

sdk/java-sdk/src/main/scala/com/akkaserverless/javasdk/impl/ComponentOptions.scala
```diff
@@ -14,31 +14,25 @@
...
+   * commands. By default no headers except "X-Server-Timing" are forwarded.
    */
-  ComponentOptions withForwardHeaders(Set<String> headers);
-
-  /** Create a default components option */
-  static ComponentOptions defaults() {
-    return new EventSourcedEntityOptionsImpl(
-        0, PassivationStrategy.defaultTimeout(), Collections.emptySet());
-  }
+  def withForwardHeaders(headers: Set[String]): ComponentOptions
```
patriknw, line 37: ^ regarding `Set`

patriknw: This is public api and it's somewhat hidden (strange) to only have it via `impl`. We should add override in all that `extends ComponentOptions`, for example ActionOptions

RayRoestenburg: ok

patriknw (approved):
LGTM

johanandren (approved):

Issue #395, opened 2021-09-15
Missing API to specify options for a component


johanandren:
We have a whole set of `com.akkaserverless.javasdk.ComponentOptions` and subtypes to define options for entites but we are missing a way to hand those to the generated `Provider` when registering.

Also there is no `ActionOptions`.

patriknw:
I might missunderstand, but we have `withX` in the generated providers.

Example from docs:
```
akkaServerless.register(CustomerByNameViewProvider.of(CustomerByNameView::new)
      .withViewId("CustomerByNameV2"));
```

For ActionOptions we have a TODO but since there are nothing to configure yet we didn't add it.

johanandren:
Oh, maybe I was looking at an old generated provider.

Actions has the request headers forward whitelist that should be possible to configure. (That's actually what caused me to add this issue)

patriknw:
> Actions has the request headers forward whitelist that should be possible to configure.

good, then we have a reason for adding the ActionOptions

The `withViewId` was perhaps not a bad example of this because that is not options. We don't have ViewOptions.

In the generated EventSourcedEntityProvider we have `withOptions`, that can be used for defining for example withSnapshotEvery and withForwardHeaders.

Conclusion: make sure we have options for all components and that they can be defined via the provider.

PR #415, opened 2021-09-17
options for actions 

johanandren:
Refs #395


patriknw (approved):
LGTM, I think same is needed for views

Issue #381, opened 2021-09-14
Public API in Scala, corresponding to .java in `com.akkaserverless.javasdk`, share same `impl` as Java SDK


RayRoestenburg:

patriknw:
Done, see https://github.com/lightbend/akkaserverless-java-sdk/issues/372 for remaining tasks.

PR #417, opened 2021-09-17
feat(scalasdk): Public API for Views, #381

patriknw:
* also an attempt of using adapters to/from javasdk

Refs #381

### DIFF
--- a/sdk/scala-sdk/src/main/scala/com/akkaserverless/scalasdk/impl/view/ViewAdapter.scala
+++ b/sdk/scala-sdk/src/main/scala/com/akkaserverless/scalasdk/impl/view/ViewAdapter.scala
@@ -1,0 +1,98 @@
+/*
+ * Copyright 2021 Lightbend Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.akkaserverless.scalasdk.impl.view
+
+import java.util.Optional
+
+import scala.jdk.OptionConverters._
+
+import com.akkaserverless.javasdk
+import com.akkaserverless.javasdk.ServiceCallFactory
+import com.akkaserverless.scalasdk.Metadata
+import com.akkaserverless.scalasdk.impl.MetadataImpl
+import com.akkaserverless.scalasdk.view.UpdateContext
+import com.akkaserverless.scalasdk.view.View
+import com.akkaserverless.scalasdk.view.ViewCreationContext
+import com.akkaserverless.scalasdk.view.ViewProvider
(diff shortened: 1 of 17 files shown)


sdk/scala-sdk/src/main/scala/com/akkaserverless/scalasdk/impl/view/ViewAdapter.scala
```diff
@@ -0,0 +1,98 @@
...
+import com.akkaserverless.javasdk.ServiceCallFactory
+import com.akkaserverless.scalasdk.Metadata
+import com.akkaserverless.scalasdk.impl.MetadataImpl
+import com.akkaserverless.scalasdk.view.UpdateContext
+import com.akkaserverless.scalasdk.view.View
+import com.akkaserverless.scalasdk.view.ViewCreationContext
+import com.akkaserverless.scalasdk.view.ViewProvider
+import com.google.protobuf.Descriptors
+
+private[scalasdk] class Scala2JavaViewAdapter[S](scalasdkView: View[S]) extends javasdk.view.View[S] {
```
patriknw, line 33: This adapter approach is completely untested and I'm not sure it's the right approach.

patriknw: This is the "flow":

1. AkkaServerless.register(scalasdk.ViewProvider)
  - delegates to javasdk by wrapping the provider with Scala2JavaViewProviderAdapter
2. javasdk.ViewProvider.newHandler is invoked
  - adapter delegates to scaladsl new handler and adapts the ViewCreationContext
  - the action View implementation is created
3. For each incoming event
  - javasdk.ViewHandler._internalSetUpdateContext  is invoked
    - adapter delegates to scaladsl and adapts the UpdateContext
  - javasdk.ViewHandler.handleUpdate is invoked
    - adapter delegates to scaladsl
    - adapter converts the returned effect

johanandren: A bit of wrapping/unwrapping work but sounds reasonable.

johanandren (commented):
Looks ok to me

sdk/scala-sdk/src/main/scala/com/akkaserverless/scalasdk/view/View.scala
```diff
@@ -0,0 +1,99 @@
...
+   */
+  protected final def updateContext(): UpdateContext =
+    _updateContext.getOrElse(
+      throw new IllegalStateException("UpdateContext is only available when handling an update."))
+
+  /** INTERNAL API */
+  private[scalasdk] def _internalSetUpdateContext(context: Option[UpdateContext]): Unit =
+    _updateContext = context
+
+  protected final def effects(): View.UpdateEffect.Builder[S] =
```
octonato, line 89: shall we remove the `()` on the methods?

patriknw: yes, that might read be better

raboof (approved):
*Looks* good AFAICS so far ;)

octonato (approved):
I think there are a few more methods with `()`, but we can revisit it later on in an API polishing round.

patriknw:
There is also `updateContext`, but for that one I think it's good to have it as `updateContext()` since it changes for each call.

In the EffectBuilder we have `ignore()`. That could be argued, but I kind of like having the `()` in the effect builder even though it might be ok without for immutable builders like that.

PR #418, opened 2021-09-17
fix: only discuss Maven tooling

ennru:
I worked through a part of the Java SDK docs to only show the Maven tooling, as it is required nowadays.﻿


retgits (approved):
LGTM! Added one textual suggestion

PR #424, opened 2021-09-20
fix: syntax renamed to Format

octonato:
Syntax was a bad name that I picked back in the time. This PR change it to Format. 

Not urgent. I can rebase later when we finish the round of changes we are doing.


patriknw (approved):
LGTM

PR #425, opened 2021-09-20
feat(codegen-scala): Scaffold empty ValueEntity implementation

raboof:
This is very incomplete (actually referring to the ScalaPB-generated types
or to the SDK is not there yet), but I thought it might be good to PR 
already to avoid double work or conflicts on the lower-level changes.


codegen/scala-gen/src/main/scala/com/akkaserverless/codegen/scalasdk/impl/ValueEntitySourceGenerator.scala
```diff
@@ -0,0 +1,56 @@
...
+         |
+         |${imports.map(i => s"import $i").mkString("\n")}
+         |
+         |class ${entity.fqn.name} /* extends Abstract${entity.fqn.name} */ {
+         |  ${Syntax.indent(methods, 2)}
+         |}
+         |""".stripMargin)
+  }
+
+  private def lowerFirst(text: String): String =
```
patriknw, line 51: this exists in SourceGeneratorUtils

patriknw (approved):
LGTM as starting point. Otherwise I found it straightforward to copy the corresponding SourceGenerator and SourceGeneratorSuite from codegenJava and convert all of it to Scala in one go.

PR #426, opened 2021-09-20
feat: add SideEffect type placeholder for scalasdk

octonato:
not full impl, just so we can use this type when defining the new components Effects

### DIFF
--- a/sdk/scala-sdk/src/main/scala/com/akkaserverless/scalasdk/SideEffect.scala
+++ b/sdk/scala-sdk/src/main/scala/com/akkaserverless/scalasdk/SideEffect.scala
@@ -1,0 +1,51 @@
+/*
+ * Copyright 2021 Lightbend Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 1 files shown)


sdk/scala-sdk/src/main/scala/com/akkaserverless/scalasdk/SideEffect.scala
```diff
@@ -0,0 +1,51 @@
...
+   * Create a side effect of the given service call.
+   *
+   * @param serviceCall
+   *   The service call to effect.
+   * @param synchronous
+   *   Whether this effect should be executed synchronously.
+   * @return
+   *   The side effect.
+   */
+  def of(serviceCall: ServiceCall, synchronous: Boolean): SideEffect = ???
```
patriknw, line 32: Is there a reason for not naming it `apply`?

octonato: yes, me doing mechanical copy-and-paste. 

I will change

octonato: done

patriknw (approved):
LGTM

johanandren (approved):

PR #427, opened 2021-09-20
fix: invert adapter name, Java2Scala

octonato:


sdk/scala-sdk/src/main/scala/com/akkaserverless/scalasdk/impl/view/ViewAdapter.scala
```diff
@@ -81,7 +81,7 @@ private[scalasdk] class Java2ScalaViewCreationContextAdapter(javasdkContext: jav
     javasdkContext.getGrpcClient(clientClass, service)
 }
 
-private[scalasdk] class Scala2JavaUpdateContextAdapter(val javasdkContext: javasdk.view.UpdateContext)
+private[scalasdk] class Java2ScalaUpdateContextAdapter(val javasdkContext: javasdk.view.UpdateContext)
     extends UpdateContext {
```
patriknw, line 85: yup, that UpdateContext is Scala

patriknw (approved):
LGTM

PR #434, opened 2021-09-21
feat(codegen-scala): referencing scalapb-generated classes

raboof:
Not completely done yet (the reference to the state is still problematic), but that can be a follow-up PR

Might already be useful for review & merge just to get the input/output types, though


codegen/scala-gen/src/test/scala/com/akkaserverless/codegen/scalasdk/ValueEntitySourceGeneratorSuite.scala
```diff
@@ -16,23 +16,38 @@
...
+         |import com.example.service.MyState
+         |import com.example.service.SetValue
+         |import com.external.Empty
          |
          |class MyValueEntity /* extends AbstractMyValueEntity */ {
-         |  def set(currentState: Unit, command: Unit): Unit = ???
+         |  def set(currentState: Unit, command: SetValue): Empty = ???
          |
-         |  def get(currentState: Unit, command: Unit): Unit = ???
+         |  def get(currentState: Unit, command: GetValue): MyState = ???
```
patriknw, line 50: `currentState: Unit`? Is that still a missing piece?

patriknw (approved):
LGTM

codegen/scala-gen/src/test/scala/com/akkaserverless/codegen/scalasdk/ValueEntitySourceGeneratorSuite.scala
raboof: Correct! Referring to the input and output messages of a service now works, but there's still something going wrong with the state model, likely because it is a reference from a custom annotation. That's the next I'm going to pick up, but can be a separate PR I think

patriknw:
ok, merging this as first step

Issue #388, opened 2021-09-14
Instructions for local development don't work on Linux in all cases


jrudolph:
As observed here [in the forum](https://discuss.lightbend.com/t/java-kickstart-on-ubuntu-linux-16-04-lts-20-04-lts/8429( and also by myself the instructions given in the documentation and in the various example READMEs might not work on Linux.

The basic reason is that the proxy is started using `docker-compose` and the user-function is run on the host computer and binding to `localhost`. The host `localhost` interface is not necessarily accessible from a docker container running with the usual `bridge` networking.

Instead, the user-function should bind to all interfaces or the `docker0` interface in particular. You can achieve that by using

```bash
HOST=0.0.0.0 mvn compile exec:java
```

Is there a better way? If not, maybe that's what we should recommend everywhere.

raboof:
(some related discussion at https://github.com/lightbend/akkaserverless-framework/issues/581 )

PR #436, opened 2021-09-21
fix: simplify running locally

johanandren:
Refs #388
Refs lightbend/akkaserverless-framework#581

Changes:
 * Use `exec:exec` to run to be able to set `HOST` to `0.0.0.0` from `pom.xml` when running locally, in all samples and archetypes, to allow proxy in local docker connecting on all platform
 * Set `PUBSUB_EMULATOR_HOST` in `docker-compose.yml` to make pub sub emulator work on all platforms
 * Include additional `extra_host` needed for Linux in the shared `docker-compose.yml` and get rid of the Linux specific one

### DIFF
--- a/maven-java/akkaserverless-maven-archetype-event-sourced-entity/src/main/resources/archetype-resources/pom.xml
+++ b/maven-java/akkaserverless-maven-archetype-event-sourced-entity/src/main/resources/archetype-resources/pom.xml
@@ -183,12 +183,21 @@
         <executions>
           <execution>
             <goals>
-              <goal>java</goal>
+              <goal>exec</goal>
             </goals>
           </execution>
         </executions>
         <configuration>
-          <mainClass>${D}{mainClass}</mainClass>
+          <executable>java</executable>
+          <arguments>
+            <argument>-classpath</argument>
+            <classpath/>
+            <argument>${D}{mainClass}</argument>
+          </arguments>
+          <environmentVariables>
+            <!-- needed for the proxy to access the user function on all platforms -->
+            <HOST>0.0.0.0</HOST>
+          </environmentVariables>
         </configuration>
       </plugin>
 
--- a/samples/java-valueentity-shopping-cart/pom.xml
+++ b/samples/java-valueentity-shopping-cart/pom.xml
@@ -169,12 +169,21 @@
         <executions>
           <execution>
             <goals>
-              <goal>java</goal>
+              <goal>exec</goal>
             </goals>
           </execution>
         </executions>
         <configuration>
-          <mainClass>${mainClass}</mainClass>
+          <executable>java</executable>
+          <arguments>
+            <argument>-classpath</argument>
+            <classpath/>
+            <argument>${mainClass}</argument>
+          </arguments>
+          <environmentVariables>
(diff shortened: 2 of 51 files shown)


maven-java/akkaserverless-maven-archetype-event-sourced-entity/src/main/resources/archetype-resources/pom.xml
```diff
@@ -183,12 +183,21 @@
...
+          <executable>java</executable>
+          <arguments>
+            <argument>-classpath</argument>
+            <classpath/>
+            <argument>${D}{mainClass}</argument>
+          </arguments>
+          <environmentVariables>
+            <!-- needed for the proxy to access the user function on all platforms -->
+            <HOST>0.0.0.0</HOST>
+          </environmentVariables>
```
patriknw, line 200: 👍

patriknw (approved):
LGTM

samples/java-valueentity-shopping-cart/pom.xml
```diff
@@ -169,12 +169,21 @@
...
           <execution>
             <goals>
-              <goal>java</goal>
+              <goal>exec</goal>
             </goals>
           </execution>
         </executions>
         <configuration>
-          <mainClass>${mainClass}</mainClass>
+          <executable>java</executable>
```
jrudolph, line 177: This doesn't pick up the `JAVA_HOME` directory as `exec:java` does (because it runs it in the same JVM basically). So, that might be a bit of a pain to use if you have multiple JDKs installed and the `java` in the `PATH` does not correspond to `JAVA_HOME`. Or should we say that's an oddity of the environment?

We could keep using `exec:java` with a configuration like this:

```xml
      <plugin>
        <groupId>org.codehaus.mojo</groupId>
        <artifactId>exec-maven-plugin</artifactId>
        <version>3.0.0</version>
        <executions>
          <execution>
            <goals>
              <goal>java</goal>
            </goals>
          </execution>
        </executions>
        <configuration>
          <mainClass>${mainClass}</mainClass>
          <systemProperties>
            <systemProperty>
              <key>akkaserverless.user-function-interface</key>
              <value>0.0.0.0</value>
            </systemProperty>
          </systemProperties>
        </configuration>
      </plugin>
```

johanandren: Aha, I didn't think about sysprops being mutable so thought that wouldn't work.

But, wouldn't you either have some additional proxy tool `java` on the `PATH` if you have many installations, which in turn looks for `JAVA_HOME` or if you use jabba/sdkman whatnot select with that and get a java on `PATH`? (That's how it works out of the box with Mac at least).

Feels like a very cornery case env to me (but I might be wrong).

octonato (approved):

samples/java-valueentity-shopping-cart/pom.xml
johanandren: Let's go with this for now and we can re-think if it turns out it is a common problem.

jrudolph: :+1: yeah might be how I use my environment.

Issue #437, opened 2021-09-21
Harden against class naming conflicts in generated code


raboof:
Right now, in generated code, we often import class names via their full path, and then refer to them by their class name.

This will not work when here are classes with the same name in different packages (generated and/or library code).

Perhaps for the first versions this is not a big limitation, but eventually we might want to harden against this. A simplistic solution is to just refer to classes by their full path everywhere, though that gets rather ugly. Referring to full paths "only when needed" would be nicer but also a bit harder.

patriknw:
> A simplistic solution is to just refer to classes by their full path everywhere

That could be ok for the managed sourced, but not ok for unmanaged (one time skeleton) in my opinion.

patriknw:
For Java our recommendation should be to use `java_outer_classname`. Then I don't think this is problem? They become `CustomerApi.Address` and `CustomerDomain.Address` and should always be used with such full names.

Otherwise entity implementations will be a total mess, since they often (always) convert between api and domain types.

Is the convention to not define `java_outer_classname` for Scala? Otherwise we should recommend that for Scala too. We could then generate fully qualified named when java_outer_classname is not defined, too support that also but at the price of some uglier code.

raboof:
> Is the convention to not define `java_outer_classname` for Scala? Otherwise we should recommend that for Scala too. We could then generate fully qualified named when java_outer_classname is not defined, too support that also but at the price of some uglier code.

I don't think ScalaPB has such an option at all (https://scalapb.github.io/docs/customizations)

> For Java our recommendation should be to use `java_outer_classname`. Then I don't think this is problem? They become `CustomerApi.Address` and `CustomerDomain.Address` and should always be used with such full names.

I agree for Java it's not a problem when using `java_outer_classname`. For Scala for this common case we can probably use a package import, so something like `def getCart(currentState: Cart, command: shoppingcart_api.GetShoppingCart): shoppingcart_api.Cart`?

patriknw:
Correct me if I'm wrong but I think we have this solved?
so nice with
```
import com.example.shoppingcart.api

api.LineItem
```

raboof:
It works for this 'common case', but there's still a lot of other possible naming conflicts.

Hardening against those is perhaps less important/urgent.

PR #438, opened 2021-09-21
feat(codegen-scala): correctly determine ValueEntity state package

raboof:
The same probably needs to be applied to some of the other
components, and there's #437


codegen/scala-gen/src/test/scala/com/akkaserverless/codegen/scalasdk/ValueEntitySourceGeneratorSuite.scala
```diff
@@ -45,9 +36,9 @@ class ValueEntitySourceGeneratorSuite extends munit.FunSuite {
          |import com.external.Empty
          |
          |class MyValueEntity /* extends AbstractMyValueEntity */ {
-         |  def set(currentState: Unit, command: SetValue): Empty = ???
+         |  def set(currentState: com.example.service.domain.MyState, command: SetValue): Empty = ???
```
patriknw, line 39: ok, I will add another comment to the ticket

patriknw (approved):
LGTM, just some minor style thing

octonato (approved):
LGTM, but also agree on changing the short name `e`

Ok to merge after changing it.

PR #327, opened 2021-09-09
fix: readme links to documentation

ennru:

PR #443, opened 2021-09-21
feat(codegen-scala): use package imports

raboof:
To make code nicer to read without introducing naming conflicts for the
common case where domain and API contain the same names.

Adding this structure could hopefully also help with detecting and avoiding
conflicts later (#327).

### DIFF
--- a/codegen/core/src/main/scala/com/lightbend/akkasls/codegen/SourceGeneratorUtils.scala
+++ b/codegen/core/src/main/scala/com/lightbend/akkasls/codegen/SourceGeneratorUtils.scala
@@ -94,12 +94,24 @@ object SourceGeneratorUtils {
   def packageAsPath(packageName: String): Path =
     Paths.get(packageName.replace(".", "/"))
 
+  class Imports(val currentPackage: String, val imports: Seq[String], val semi: Boolean) {
+    def contains(imp: String) = imports.contains(imp)
+
+    override def toString = {
+      val suffix = if (semi) ";" else ""
+      imports
+        .map(pkg => s"import $pkg$suffix")
+        .mkString("\n")
+    }
+
+  }
+
   def generateImports(
       types: Iterable[FullyQualifiedName],
       packageName: String,
       otherImports: Seq[String],
       packageImports: Seq[String] = Seq.empty,
-      semi: Boolean = true): String = {
+      semi: Boolean = true): Imports = {
     val messageTypeImports = types
       .filterNot { typ =>
         typ.parent.javaPackage == packageName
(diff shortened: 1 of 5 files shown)


codegen/core/src/main/scala/com/lightbend/akkasls/codegen/SourceGeneratorUtils.scala
```diff
@@ -80,7 +80,8 @@ object SourceGeneratorUtils {
   def typeImport(fullyQualifiedName: FullyQualifiedName): String = {
     val name =
       if (fullyQualifiedName.parent.javaMultipleFiles) fullyQualifiedName.name
-      else fullyQualifiedName.parent.javaOuterClassname
+      else if (fullyQualifiedName.parent.javaOuterClassname.nonEmpty) fullyQualifiedName.parent.javaOuterClassname
+      else fullyQualifiedName.name
```
patriknw, line 84: what will happen if `java_outer_classname` is defined when generating Scala? You said that ScalaPB don't have that option so maybe things will be totally wrong if that is accidentally defined?

Ignore and fallback to right defaults for Scala or fail fast if `java_outer_classname` is defined for Scala?

raboof: This code is shared, so we don't "know" if we're generating Scala here.

We can make sure in the Scala `FullyQualifiedNameExtractor` that we never populate `javaOuterClassname` there, but beyond that I'm not sure there's an easy way to add more safeguards. One thing we could do is remove `PackageNaming.from(Descriptor)` and make sure this is only done in `FullyQualifiedNameExtractor`, but I haven't dared to pull that thread yet ;).

(the ideal solution IMO would be to remove `javaOuterClassname` from `FullyQualifiedName` entirely and interpret it inside the Java `FullyQualifiedNameExtractor` already, but that might not be possible/easy)

codegen/core/src/main/scala/com/lightbend/akkasls/codegen/SourceGeneratorUtils.scala
```diff
@@ -93,43 +94,56 @@ object SourceGeneratorUtils {
   def packageAsPath(packageName: String): Path =
     Paths.get(packageName.replace(".", "/"))
 
+  class Imports(val currentPackage: String, val imports: Seq[String], val semi: Boolean) {
+    def contains(imp: String) = imports.contains(imp)
+
+    override def toString = {
+      val suffix = if (semi) ";" else ""
```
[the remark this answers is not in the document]
patriknw: Let's take this cleanup in a separate PR.

patriknw (approved):
LGTM

PR #449, opened 2021-09-22
chore(scalasdk): instructions for working with samples

patriknw:


raboof (approved):
(we should probably also document how to work with scripted tests, but that's for a different PR)

PR #456, opened 2021-09-22
build: move proto test files to common place

octonato:
This is in preparation for full generation tests for the Scala SDK. 

We have a bunch of files used by `codegenJavaCompilationTest` that exercise many different options. 

As we progress with the codegen for Scala, we want the Scala codegen to run on those same files.


raboof:
> This is in preparation for full generation tests for the Scala SDK.

Don't we already have full generation tests, the scripted tests of the sbt plugin?

octonato:
> Don't we already have full generation tests, the scripted tests of the sbt plugin?

My sentence is not clear. Yes, we do have full generation. What I mean, we should be able to have exhaustive generations to be able to catch corner cases. Like usage of outer java class, multi-files, import from other packages, etc. That's why we have this list of files.

build.sbt
```diff
@@ -171,6 +171,7 @@ lazy val codegenJavaCompilationTest = project
   .settings(
     (publish / skip) := true,
     name := "akkaserverless-codegen-java-compilation-tests",
+    Compile / PB.protoSources += baseDirectory.value / ".." / ".." / "proto-test-files",
```
patriknw, line 174: Are we going to do something similar like this in the scripted test to include them? Have you tried that it works?

patriknw (commented):
LGTM, with a question about how to use them...

patriknw (approved):

PR #460, opened 2021-09-22
fix(sbt-plugin): make sure akkaserverless classes are generated

raboof:
This way we don't get the class files from a dependencies, but we get
the proto files and generate the class files ourselves: this way we have
more control over how they are generated.

### DIFF
--- a/sbt-plugin/src/main/scala/com/akkaserverless/sbt/AkkaserverlessPlugin.scala
+++ b/sbt-plugin/src/main/scala/com/akkaserverless/sbt/AkkaserverlessPlugin.scala
@@ -43,7 +43,7 @@ object AkkaserverlessPlugin extends AutoPlugin {
 
   override def projectSettings: Seq[sbt.Setting[_]] = Seq(
     libraryDependencies ++= Seq(
-      "com.akkaserverless" % "akkaserverless-sdk-protocol" % "0.7.0-beta.19" % "protobuf",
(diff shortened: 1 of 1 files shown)


sbt-plugin/src/main/scala/com/akkaserverless/sbt/AkkaserverlessPlugin.scala
```diff
@@ -43,7 +43,7 @@ object AkkaserverlessPlugin extends AutoPlugin {
 
   override def projectSettings: Seq[sbt.Setting[_]] = Seq(
     libraryDependencies ++= Seq(
-      "com.akkaserverless" % "akkaserverless-sdk-protocol" % "0.7.0-beta.19" % "protobuf",
+      "com.akkaserverless" % "akkaserverless-sdk-protocol" % "0.7.0-beta.19" % "protobuf-src",
```
patriknw, line 46: so this will solve the AnnotationsProto issue?

patriknw (approved):
LGTM

sbt-plugin/src/main/scala/com/akkaserverless/sbt/AkkaserverlessPlugin.scala
raboof: Yes, I'm fairly confident it will (though there might be further problems lurking there, around flat/non-flat packages and naming clashes)

octonato (approved):

PR #462, opened 2021-09-22
feat(sbt-plugin): automatically depend on Akka gRPC

raboof:
Will be useful later when we change the ServiceCall infrastructure to
use Akka gRPC as well

### DIFF
--- a/sbt-plugin/src/main/scala/com/akkaserverless/sbt/AkkaserverlessPlugin.scala
+++ b/sbt-plugin/src/main/scala/com/akkaserverless/sbt/AkkaserverlessPlugin.scala
@@ -16,10 +16,11 @@
 
 package com.akkaserverless.sbt
 
+import akka.grpc.sbt.AkkaGrpcPlugin
+
 import java.nio.file.Path
 import java.nio.file.Paths
 import java.nio.file.Files
-
 import com.akkaserverless.codegen.scalasdk.{ gen, genTests, genUnmanaged, AkkaserverlessGenerator }
 import sbt.{ Compile, _ }
 import sbt.Keys._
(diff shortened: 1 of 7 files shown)


octonato (approved):

patriknw:
> Will be useful later when we change the ServiceCall infrastructure to
use Akka gRPC as well

That is in fact not for later. It's already needed for the `getGrcpClient` that Johan added recently.

sbt-plugin/src/main/scala/com/akkaserverless/sbt/AkkaserverlessPlugin.scala
```diff
@@ -28,8 +29,8 @@ import sbtprotoc.ProtocPlugin.autoImport.PB
 import scalapb.GeneratorOption
 
 object AkkaserverlessPlugin extends AutoPlugin {
-  override def trigger = allRequirements
-  override def requires = ProtocPlugin
+  override def trigger = noTrigger
+  override def requires = ProtocPlugin && AkkaGrpcPlugin
```
patriknw, line 33: could you also add the `akkaserverless-scala-sdk` to libraryDependencies while you are at it? (we have a FIXME in the samples for that)

raboof: I thought that should already happen automatically because of https://github.com/lightbend/akkaserverless-java-sdk/blob/main/codegen/scala-gen/src/main/scala/com/akkaserverless/codegen/scalasdk/AkkaserverlessGenerator.scala#L58 - will verify

raboof: > will verify

Yes, seems to work, added #465

patriknw (approved):
LGTM

PR #465, opened 2021-09-22
chore(scala-samples): remove sdk dependency

raboof:
that's added automatically through
https://github.com/lightbend/akkaserverless-java-sdk/blob/main/codegen/scala-gen/src/main/scala/com/akkaserverless/codegen/scalasdk/AkkaserverlessGenerator.scala#L58


patriknw (approved):
ok, then it must have something else that got me on the wrong track here

johanandren (approved):

Issue #464, opened 2021-09-22
Action handler with streamed out does not compile


johanandren:
Needs an additional cast or something:

```proto
rpc NextNumbers(Number) returns (stream Number) {}
```

gives:
```java
  @Override
  public Source<Action.Effect<?>, NotUsed> handleStreamedOut(String commandName, MessageEnvelope<Object> message) {
    switch (commandName) {
      case "NextNumbers":
        return action()
                 .nextNumbers((FibonacciApi.Number) message.payload());
      default:
        throw new ActionHandler.HandlerNotFound(commandName);
    }
  }
```

Which fails to compile with:
```
incompatible types: akka.stream.javadsl.Source<com.akkaserverless.javasdk.action.Action.Effect<com.example.fibonacci.FibonacciApi.Number>,akka.NotUsed> cannot be converted to akka.stream.javadsl.Source<com.akkaserverless.javasdk.action.Action.Effect<?>,akka.NotUsed>
```

PR #467, opened 2021-09-22
fix: Streaming out handlers did not compile 

johanandren:
Refs #464

### DIFF
--- a/codegen/java-gen/src/main/scala/com/lightbend/akkasls/codegen/java/ActionServiceSourceGenerator.scala
+++ b/codegen/java-gen/src/main/scala/com/lightbend/akkasls/codegen/java/ActionServiceSourceGenerator.scala
@@ -278,6 +278,7 @@ object ActionServiceSourceGenerator {
         |  }
         |
         |  @Override
+        |  @SuppressWarnings("unchecked")
         |  public Source<Action.Effect<?>, NotUsed> handleStreamedOut(String commandName, MessageEnvelope<Object> message) {
         |    switch (commandName) {
         |      ${Format.indent(streamOutCases, 6)}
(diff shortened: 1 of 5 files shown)


octonato (approved):
LGTM

So, our protos are not that exhaustive (yet).

codegen/java-gen/src/main/scala/com/lightbend/akkasls/codegen/java/ActionServiceSourceGenerator.scala
```diff
@@ -241,7 +241,7 @@ object ActionServiceSourceGenerator {
       val inputTypeFullName = cmd.inputType.fullName
 
       s"""|case "$methodName":
-          |  return action()
+          |  return (Source<Action.Effect<?>, NotUsed>)(Object) action()
```
patriknw, line 244: my Java parser is in for repair; does it first cast to Object followed by another cast to Source?

johanandren: Yes, since the Java doesn't understand variance/Java `Source` is invariant. I don't think there is any other way to do it. We _know_ it is safe though.

patriknw: I trust you ™️ 
Shall we add some supress warning annotation?

johanandren: Yes, we should 👍

octonato (approved):

PR #469, opened 2021-09-22
feat(scalasdk): codegen for ValueEntityHandler.

RayRoestenburg:


RayRoestenburg:
It will fail scripted, since it needs the actual entity implementation (If I've translated this correctly of course).

codegen/core/src/main/scala/com/lightbend/akkasls/codegen/ModelBuilder.scala
```diff
@@ -77,7 +77,7 @@ object ModelBuilder {
       state: State)
       extends Entity(fqn, entityType) {
     val abstractEntityName = "Abstract" + fqn.name
-
+    val handlerName = fqn.name + "Handler"
```
raboof, line 80: (I wonder if it would be nicer to keep them in `fqn` form here, but that's not for this PR)

raboof (approved):
some conflicts, probably because the squash-and-merge of #471? so probably a matter of rebasing

codegen/scala-gen/src/main/scala/com/akkaserverless/codegen/scalasdk/impl/ValueEntitySourceGenerator.scala
```diff
@@ -86,6 +87,54 @@ object ValueEntitySourceGenerator {
...
+      generateImports(
+        Seq(valueEntity.state.fqn) ++
+        service.commands.map(_.inputType) ++
+        service.commands.map(_.outputType),
+        valueEntity.fqn.parent.scalaPackage,
+        otherImports = Seq(
+          "com.akkaserverless.scalasdk.valueentity.CommandContext",
+          "com.akkaserverless.scalasdk.valueentity.ValueEntity",
+          "com.akkaserverless.scalasdk.impl.valueentity.ValueEntityHandler",
+          "com.akkaserverless.scalasdk.impl.valueentity.ValueEntityHandler.CommandHandlerNotFound"),
```
patriknw, line 104: This might be wrong. In View and Actions we used the corresponding javasdk exception instead of defining a new in scalasdk. It's not something seen by the user. In that we we don't need to catch-convert-throw.

patriknw (commented):
looking good, but one thing could be wrong...

PR #471, opened 2021-09-23
feat(codegen-scala): make value entity impl extend abstract

raboof:


RayRoestenburg (approved):
LGTM

codegen/scala-gen/src/test/scala/com/akkaserverless/codegen/scalasdk/ValueEntitySourceGeneratorSuite.scala
```diff
@@ -37,13 +37,24 @@ class ValueEntitySourceGeneratorSuite extends munit.FunSuite {
...
+         |/** A value entity. */
+         |class MyValueEntity(val entityId: String) extends AbstractMyValueEntity {
+         |  override def set(currentState: MyState, command: service.SetValue): ValueEntity.Effect[Empty] =
+         |    effects.error("The command handler for `Set` is not implemented, yet");
          |
-         |  def get(currentState: MyState, command: service.GetValue): service.MyState = ???
+         |  override def get(currentState: MyState, command: service.GetValue): ValueEntity.Effect[service.MyState] =
+         |    effects.error("The command handler for `Get` is not implemented, yet");
+         |
+         |  override def emptyState =
```
patriknw, line 52: I'd prefer to have emptyState first, before the methods.

Also include explicit return type.

codegen/scala-gen/src/test/scala/com/akkaserverless/codegen/scalasdk/ValueEntitySourceGeneratorSuite.scala
```diff
@@ -37,13 +37,24 @@ class ValueEntitySourceGeneratorSuite extends munit.FunSuite {
...
+         |    throw new UnsupportedOperationException("Not implemented yet, replace with your empty entity state")
+         |}
+         |object MyValueEntity {
+         |  def apply(context: Unit /* ValueEntityContext */): MyValueEntity =
```
patriknw, line 56: yes, it should have ValueEntityContext parameter

patriknw: I don't think there is much value in generating the companion object. We can use `new` as the default from Main (that is how it's generated currently). If user would like to change constructor parameter the companion will just make it two places to change.

If they really want a companion they can add it themselves.

patriknw (commented):
LGTM

PR #472, opened 2021-09-23
fix: remove warning on additionalDescriptors

octonato:


sdk/scala-sdk/src/main/scala/com/akkaserverless/scalasdk/view/ViewProvider.scala
```diff
@@ -30,5 +30,5 @@ trait ViewProvider[S, V <: View[S]] {
 
   def newHandler(context: ViewCreationContext): ViewHandler[S, V]
 
-  def additionalDescriptors(): immutable.Seq[Descriptors.FileDescriptor]
```
octonato, line 33: the generated code doesn't use `()` and we get a warning on compilation. Since it's Scala, we can remove the it. This aligns with `serviceDescriptor` a few lines above.

raboof (approved):

patriknw (approved):
LGTM, thanks

PR #473, opened 2021-09-23
chore(scalasdk): some more proto directory structure align

johanandren:
Not super important but good to keep it consistent across all proto directories in the sdk, makes it more likely we get it right  everywhere in the future (or until we add linting)


octonato (approved):
sure, good to align

I'm removing the shopping cart from the sbt scripted test because I'm bringing all the other files there. Anyway, I will fix the conflicts later.

patriknw (approved):
LGTM

PR #474, opened 2021-09-23
feat(codegen-scala): value entity provider

raboof:
Let's finish this up after merging #469 so we can actually hook up the handler

### DIFF
--- a/codegen/scala-gen/src/main/scala/com/akkaserverless/codegen/scalasdk/impl/ValueEntitySourceGenerator.scala
+++ b/codegen/scala-gen/src/main/scala/com/akkaserverless/codegen/scalasdk/impl/ValueEntitySourceGenerator.scala
@@ -134,9 +134,8 @@ object ValueEntitySourceGenerator {
     val packageName = entity.fqn.parent.scalaPackage
     val className = entity.fqn.name + "Provider"
 
-    val imports = generateImports(
+    implicit val imports: Imports = generateImports(
       Seq(entity.state.fqn) ++ service.commands.map(_.inputType) ++ service.commands.map(_.outputType),
-      // TODO descriptor imports
       packageName,
       Seq(
         "com.akkaserverless.scalasdk.valueentity.ValueEntityContext",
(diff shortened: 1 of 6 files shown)


codegen/scala-gen/src/test/scala/com/akkaserverless/codegen/scalasdk/ValueEntitySourceGeneratorSuite.scala
```diff
@@ -77,4 +77,36 @@ class ValueEntitySourceGeneratorSuite extends munit.FunSuite {
           |""".stripMargin)
   }
 
+  test("it can generate a provider") {
+    val file =
+      provider(testData.valueEntity(domainParent), testData.simpleEntityService(apiParent))
+    assertNoDiff(
+      file.content,
```
patriknw, line 84: ok, then we have a cleanup task to align all other generation to return File instead of String
(consistency police here)

codegen/scala-gen/src/main/scala/com/akkaserverless/codegen/scalasdk/impl/ValueEntitySourceGenerator.scala
```diff
@@ -86,6 +81,48 @@ object ValueEntitySourceGenerator {
...
+         |  override final val serviceDescriptor: Descriptors.ServiceDescriptor =
+         |    ${typeName(service.descriptorObject)}.javaDescriptor.findServiceByName("${service.fqn.protoName}")
+         |
+         |  override final val entityType = "${entity.entityType}"
+         |
+         |  override final def newHandler(context: ValueEntityContext) = ???
+         |
+         |  override final val additionalDescriptors =
+         |    ${typeName(service.descriptorObject)}.javaDescriptor :: Nil
+         |}
```
patriknw, line 122: should also have `def withOptions`

codegen/scala-gen/src/test/scala/com/akkaserverless/codegen/scalasdk/ValueEntitySourceGeneratorSuite.scala
raboof: I'd like that ;) (and yeah, we have a bunch of consistency things to clean up ;) ). Will PR.

patriknw (approved):
LGTM

PR #470, opened 2021-09-23
fix: action provider compilation

octonato:

PR #475, opened 2021-09-23
fix: move proto test files under scripted

octonato:
On top of #470. 

This brings the proto test file under the scripted tests. I needed to move the files there and then let the java compilation test point to the scripted proto dir. 

The reason for that is that the whole project needs to be copied to /tmp when running and we need the files to be copied as well. That said, the scripted tests is now the host for those files.


sbt-plugin/src/sbt-test/sbt-akkaserverless/basic/test
```diff
@@ -1,13 +1,73 @@
 # TODO replace with just ``> compile` when issue #454 has been fixed
+> protocSources
 > protocGenerate
 > generateUnmanaged
 > compile
-$ exists src/main/scala/foo/bar/Baz.scala
```
octonato, line 5: btw, we are still generating this file. I will remove it later.

sbt-plugin/src/sbt-test/sbt-akkaserverless/basic/test
```diff
@@ -1,13 +1,73 @@
...
+$ exists target/scala-2.13/src_managed/main/akkaserverless/com/example/valueentity/domain/some_value_entity_domain/AbstractSomeValueEntity.scala
+
+# com/example/valueentity/user_api.proto
+# com/example/valueentity/domain/user_domain.proto
+$ exists src/main/scala/com/example/valueentity/domain/user_domain/User.scala
+$ exists target/scala-2.13/src_managed/main/akkaserverless/com/example/valueentity/domain/user_domain/AbstractUser.scala
 
 > Test/compile
-$ exists target/scala-2.13/src_managed/test/akkaserverless/foo/bar/BazSpec.scala
-$ absent target/scala-2.13/src_managed/test/akkaserverless/foo/bar/Baz.scala
```
octonato, line 13: I didn't find it useful to keep testing this. Very soon we will have the real stuff here.

sbt-plugin/src/sbt-test/sbt-akkaserverless/basic/build.sbt
```diff
@@ -2,7 +2,11 @@ import com.akkaserverless.sbt.AkkaserverlessPlugin.autoImport.generateUnmanaged
 
 scalaVersion := "2.13.6"
 
+<<<<<<< Updated upstream
```
patriknw, line 5: unresolved merge conflict

octonato: Oops

codegen/scala-gen/src/main/scala/com/akkaserverless/codegen/scalasdk/impl/ActionServiceSourceGenerator.scala
```diff
@@ -298,13 +298,13 @@ object ActionServiceSourceGenerator {
...
         |  override final def serviceDescriptor: Descriptors.ServiceDescriptor =
-        |    ${service.fqn.parent.name}Proto.javaDescriptor.findServiceByName("${service.fqn.protoName}")
+        |    ${typeName(service.descriptorObject)}.javaDescriptor.findServiceByName("${service.fqn.protoName}")
         |
         |  override final def newHandler(context: ActionCreationContext): ${service.handlerName} =
         |    new ${service.handlerName}(actionFactory(context))
         |
         |  override final def additionalDescriptors: immutable.Seq[Descriptors.FileDescriptor] =
-        |    ${service.fqn.parent.name}Proto.javaDescriptor ::
+        |    ${typeName(service.descriptorObject)}.javaDescriptor ::
```
raboof, line 307: This change looks good, but the test trips over it. Likely because the test still uses a 'Java-style' model, I'll PR making that easier.

raboof (approved):
I liked having a 'realistic' example in the scripted tests, but I guess we can re-introduce that when we want ;)

codegen/scala-gen/src/main/scala/com/akkaserverless/codegen/scalasdk/impl/ActionServiceSourceGenerator.scala
raboof: (PR'ed #481, though not sure anymore it's the same problem)

patriknw (approved):
LGTM, an additional scripted setup we could use is to setup something only for CI and have a job copying proto files from the samples and verify that they compile. Compared to the samples themselves it would validate that the generated unmanaged code compiles.

patriknw:
> We will need to make sure that the samples protos don't conflict with each other.

We could setup CI job that run copy and run them one by one.

> But this is already happening, no? We do run the tests for the samples. They compile and run.

Not fully testing with an empty project so that we verify that the generated unmanaged sources are correct. The unmanaged sources are in git and implemented by us.

An easier approach to my idea would be to not use scripted, but just have a CI job that removes `src/main/` and runs `sbt compile`. That approach also works for `mvn`. I'll create a ticket.

patriknw:
Created issue https://github.com/lightbend/akkaserverless-java-sdk/issues/486 for that additional CI idea.

Issue #463, opened 2021-09-22
Streaming in not possible to use for anything


johanandren:
Any scenario I can think of requires the user to **run** the incoming `Source` of messages and then make an async reply out of that, but there is currently no access to an `ActorSystem` or a `Materializer` needed to do that. 

Streaming out and bidi streaming is possible since those gives a chance to hand off the stream blueprint for the SDK to materialize/run.

johanandren:
Perhaps we should make the `Materializer` available through `ComponentContext`?

patriknw:
> Perhaps we should make the Materializer available through ComponentContext?

I can't come up with any better alternative. It becomes rather low level Akka Streams stuff but it's already Streams.

pvlugter:
I think the TCK implementation for streaming in just creates its own actor system currently. Making the materializer available does seem like the best option.

PR #476, opened 2021-09-23
fix: access to materializer for stream in actions 

johanandren:
Refs #463


testkit-java/src/main/scala/com/akkaserverless/javasdk/testkit/impl/TestKitActorSystem.scala
```diff
@@ -0,0 +1,28 @@
...
+
+import akka.actor.ActorSystem
+
+/**
+ * INTERNAL API
+ */
+private[akkaserverless] object TestKitActorSystem {
+  // Only used to materialize streams (expected to be short lived) in user logic, same system re-used across tests is fine
+  // FIXME leaks one actor system per test run though, because no callback to terminate, that's not great
+  lazy val system = ActorSystem(s"testkit-actor-system-${System.currentTimeMillis()}")
```
johanandren, line 27: The only other option I could think of is to let the concrete testkit classes lazy-create and own the actor system, problem is the testkit also does not have a lifecycle, so would require adding that further complicating things.

Not allowing unit tests to materializing the incoming stream does not feel like an option.
Open for any good ideas.

octonato: I don't see how we could do that differently. Since we generate the test skeleton, we can add a teardown method and stop the testkit with a good chunk of comments explaining the reasons. 

If users touch that method then yes, they will be leaking.

johanandren: Decided to go back on this for now and revisit supporting it in the testkit separately as that is a bigger task involving changes to how the testkit is interacted with from the tests (both for in progress Scala testkits and the Java ones).

patriknw: yes, let's take that later

patriknw (approved):
LGTM

PR #477, opened 2021-09-23
fix: favor scala function syntax

octonato:
Pure aesthetics. Favor Scala function syntax so when devs navigate to withComponents method they see Scala functions as they are used to.


raboof (approved):

patriknw (approved):
LGTM

Issue #372, opened 2021-09-13
Scala SDK


patriknw:
Task breakdown:

- [ ] Build project structure in this repo
- [ ] Public API in Scala, corresponding to .java in `com.akkaserverless.javasdk`, share same `impl` as Java SDK
- [ ] sbt plugin for codegen
- [ ] Codegen of Scala, correspdonding to `codegenJava`
- [ ] .g8 template, corresponding to the maven archetype
- [ ] Samples and documentation, use Java/Scala tabs

PR #478, opened 2021-09-23
feat(scalasdk): api for ValueEntity TestKit

patriknw:
I have started on the codegen too, but PRed this first in case someone want to start with integration tests in parallel (in same testkit project)

Refs #372


raboof (approved):
The changes look good, surprising there isn't more to it, but I guess that's all in the generated code ;)

PR #479, opened 2021-09-23
fix: make sure unmanaged generated files are detected on first compilation attempt

### LINKED ISSUE #435: Unmanaged classes sometimes need a second 'compile'
When opening a new project and running `compile`, it seems any missing 'unmanaged' classes are correctly generated into `src/main/scala`, but not actually picked up for that `compile` step. Running `compile` again fixes it.

Perhaps they are created 'too late' to be picked up when sbt checks which files have changed and have to be (re)compiled?

jrudolph:
There's no happens-before relationship between mentioned tasks in a task
definition, so we need to make sure to include generated files
explicitly.

Fixes #435

### DIFF
--- a/samples/scala-value-entity-customer-registry/build.sbt
+++ b/samples/scala-value-entity-customer-registry/build.sbt
@@ -1,5 +1,3 @@
-import com.akkaserverless.sbt.AkkaserverlessPlugin.autoImport.generateUnmanaged
-
 name := "customer-registry"
 
 organization := "com.akkaseverless.samples"
@@ -32,11 +30,3 @@ Test / logBuffered := false
 
 run / fork := false
 Global / cancelable := false // ctrl-c
-
-Compile / compile := {
-  // Make sure 'generateUnmanaged' is executed on each compile, to generate scaffolding code for
-  // newly-introduced concepts.
-  // After initial generation they are to be maintained manually and will not be overwritten.
-  (Compile / generateUnmanaged).value
-  (Compile / compile).value
-}
--- a/sbt-plugin/src/main/scala/com/akkaserverless/sbt/AkkaserverlessPlugin.scala
+++ b/sbt-plugin/src/main/scala/com/akkaserverless/sbt/AkkaserverlessPlugin.scala
@@ -33,7 +33,7 @@ object AkkaserverlessPlugin extends AutoPlugin {
   override def requires = ProtocPlugin && AkkaGrpcPlugin
 
   trait Keys { _: autoImport.type =>
-    val generateUnmanaged = taskKey[Unit](
+    val generateUnmanaged = taskKey[Seq[File]](
       "Generate \"unmanaged\" akkaserverless scaffolding code based on the available .proto definitions.\n" +
       "These are the source files that are placed in the source tree, and after initial generation should typically be maintained by the user.\n" +
       "Files that already exist they are not re-generated.")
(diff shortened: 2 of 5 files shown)


samples/scala-value-entity-customer-registry/build.sbt
```diff
@@ -29,11 +27,3 @@ Test / logBuffered := false
 
 run / fork := false
 Global / cancelable := false // ctrl-c
-
-Compile / compile := {
-  // Make sure 'generateUnmanaged' is executed on each compile, to generate scaffolding code for
-  // newly-introduced concepts.
-  // After initial generation they are to be maintained manually and will not be overwritten.
-  (Compile / generateUnmanaged).value
-  (Compile / compile).value
```
jrudolph, line 38: These two task are not run in any defined order.

patriknw (approved):
LGTM!

sbt-plugin/src/main/scala/com/akkaserverless/sbt/AkkaserverlessPlugin.scala
```diff
@@ -59,28 +59,33 @@ object AkkaserverlessPlugin extends AutoPlugin {
...
       // Then copy over any new generated unmanaged sources
       copyIfNotExist(
         Paths.get((Compile / temporaryUnmanagedDirectory).value.toURI),
         Paths.get((Compile / sourceDirectory).value.toURI).resolve("scala"))
     },
     Compile / managedSources :=
-      (Compile / managedSources).value.filter(s => !isIn(s, (Compile / temporaryUnmanagedDirectory).value)))
+      (Compile / managedSources).value.filter(s => !isIn(s, (Compile / temporaryUnmanagedDirectory).value)),
+    Compile / unmanagedSources :=
+      (Compile / generateUnmanaged).value ++ (Compile / unmanagedSources).value)
```
raboof, line 71: Nice!

(this makes the regeneration non-optional, but that's probably fine)

raboof (approved):

sbt-plugin/src/main/scala/com/akkaserverless/sbt/AkkaserverlessPlugin.scala
jrudolph: You mean, when all files are generated, you could skip that step? Probably true, but maybe no problem in practice.

raboof: Yeah - earlier it came up that it is sometimes annoying when the code generation for user-managed  classes runs automatically - the case that came up was when the user removes the skeleton integration test (this was feedback on the java plugin). On the other hand it's nice when it's immediately generated for the first compile. For that reason we initially put the dependency in the user `build.sbt`, so they have it by default, but can easily disable it.

I now think it's fine to put it in the plugin though, not worth it to expose users to this detail for such a niche concern. We could always revisit when we get feedback.

jrudolph: :+1:

pvlugter: Yeah, can wait to see if this comes up with users. Can still disable by overriding the setting if needed: `Compile / generateUnmanaged := Seq.empty`.

PR #483, opened 2021-09-23
feat: implementation of scala fibonacci sample

octonato:
Not working yet hence draft. 
I tried to run service locally and it's not responding. So, probably some issue when converting?

PR #484, opened 2021-09-23
fix: remove dummy foo/bar/baz

octonato:
I removed it initially in #483, but sending on a different PR since #483 is still broken.


raboof (approved):
Good to remove now that we have 'actual' generated classes. Looks like there's still a test that needs updating?

octonato:
> Looks like there's still a test that needs updating?

Yes, that was the scripted tests. Solved now that #475

PR #485, opened 2021-09-23
chore(scalasdk): codegenScala sbt project

patriknw:
* similar to codegenJava
* annoying that they have different naming conventions

tab completion:
```
sbt:akkaserverless-java-sdk> codegen
codegenCore/                  codegenJava/                  codegenJavaCompilationTest/   codegenScala/
```

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -136,7 +136,7 @@ jobs:
       - restore_deps_cache
       - run:
           name: Run tests
-          command: sbt test codegenCore/publishLocal akkaserverless-codegen-scala/publishLocal sdkCore/publishLocal sdkJava/publishLocal sdkScala/publishLocal scripted
+          command: sbt test codegenCore/publishLocal codegenScala/publishLocal sdkCore/publishLocal sdkJava/publishLocal sdkScala/publishLocal scripted
       - save_deps_cache
 
   integration-tests:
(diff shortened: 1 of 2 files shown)


raboof (approved):
Good to align.

I think I prefer having the sbt project id the same as the artifact name, so I'd maybe rather change things the 'other way around' - but this is OK

patriknw:
> I think I prefer having the sbt project id the same as the artifact name

Personally I agree, but I thought I'd just follow the style that existed

johanandren:
+1 on following artifact names (but I think it might interfere with how @RayRoestenburg s dev env is set up?)

RayRoestenburg:
I can do something temporarily, or fix sbt-sublime, which is so old, I should probably rip off the band-aid :-) so do what is best.

patriknw:
Could we merge this as a first sanity and do a bigger build refactoring later if we feel that is important?

raboof:
(ah I didn't realize the package names were interfering with sbt-sublime)

RayRoestenburg:
They don't, it is actually the name of the project variables, (It does not support identifiers in backticks) so this is not an issue.

RayRoestenburg:
@raboof this https://github.com/lightbend/akkaserverless-java-sdk/pull/408 was what caused an issue for sbt-sublime, FYI.

PR #489, opened 2021-09-24
fix(codegen-java): honour java_package option

raboof:
Some places used the `pkg` rather than the `javaPackage` when generating
Java code. Renamed `pkg` to `protoPackage` to avoid this mistake in the
future.


patriknw (approved):
LGTM

johanandren (approved):

PR #491, opened 2021-09-24
fix(sbt-plugin): rely on Akka gRPC-generated model classes

raboof:
Don't generate the Akka gRPC server for now, as its handlers clash with Akkaserverless-generated handlers


patriknw (approved):
LGTM

PR #492, opened 2021-09-24
fix(scalasdk): fix passivation strategy.

RayRoestenburg:


sdk/scala-sdk/src/main/scala/com/akkaserverless/scalasdk/impl/valueentity/ValueEntityAdapter.scala
```diff
@@ -84,8 +84,13 @@ private[scalasdk] class JavaValueEntityOptionsAdapter(scalasdkValueEntityOptions
...
 
-  def passivationStrategy(): javasdk.PassivationStrategy =
-    new JavaPassivationStrategyAdapter(scalasdkValueEntityOptions.passivationStrategy)
+  def passivationStrategy(): javasdk.PassivationStrategy = {
+    scalasdkValueEntityOptions.passivationStrategy match {
+      case com.akkaserverless.scalasdk.impl.Timeout(Some(duration)) =>
+        javasdk.PassivationStrategy.timeout(duration.toJava)
+      case com.akkaserverless.scalasdk.impl.Timeout(None) => javasdk.PassivationStrategy.defaultTimeout()
+    }
+  }
```
patriknw, line 94: at the top of the file, remove
```
import com.akkaserverless.javasdk.impl.Timeout
```
so that we always reference javasdk classes more explicitly here, like: `javasdk.impl.Timeout`

patriknw (approved):
LGTM, after one small cleanup

Issue #490, opened 2021-09-24
Async failure is not turned into error protocol response


johanandren:
Completing a call (action) with a `CompletionStage`/`effects().async` where the future fails should be turned into a protocol response error message just like throwing from the command handler and not fail the grpc stream

johanandren:
Possibly something unclear/not quite right about a streaming response stream failing as well, should that be turned into an error response message rather than fail the stream (if that is in fact what is happening currently).

johanandren:
There is also no logging for these kinds of errors as far as I can see, would be nice if the SDK logged an error with exception details, at least when the user handler throws?

PR #493, opened 2021-09-24
fix: Turn thrown exception from user Action code into failure responses

johanandren:
Refs #490

### DIFF
--- a/sdk/java-sdk/src/main/scala/com/akkaserverless/javasdk/impl/action/ActionsImpl.scala
+++ b/sdk/java-sdk/src/main/scala/com/akkaserverless/javasdk/impl/action/ActionsImpl.scala
@@ -38,6 +38,7 @@ import scala.collection.immutable
 import scala.concurrent.Future
 import scala.jdk.CollectionConverters.SeqHasAsJava
 import com.akkaserverless.javasdk.impl.ActionFactory
+import org.slf4j.LoggerFactory
 
 import scala.util.control.NonFatal
 
@@ -59,6 +60,7 @@ final class ActionService(
 final class ActionsImpl(_system: ActorSystem, services: Map[String, ActionService], rootContext: Context)
     extends Actions {
 
+  private val log = LoggerFactory.getLogger(classOf[Action])
   import _system.dispatcher
   implicit val system: ActorSystem = _system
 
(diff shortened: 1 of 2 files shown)


sdk/java-sdk/src/main/scala/com/akkaserverless/javasdk/impl/action/ActionsImpl.scala
```diff
@@ -83,10 +84,14 @@ final class ActionsImpl(_system: ActorSystem, services: Map[String, ActionServic
...
-          val withSurroundingSideEffects = effect.addSideEffects(sideEffects.asJava)
-          effectToResponse(withSurroundingSideEffects, anySupport)
-        }
+        futureEffect
+          .flatMap { effect =>
+            val withSurroundingSideEffects = effect.addSideEffects(sideEffects.asJava)
+            effectToResponse(withSurroundingSideEffects, anySupport)
+          }
+          .recover { case NonFatal(ex) =>
+            ActionResponse(ActionResponse.Response.Failure(Failure(description = ex.getMessage)))
```
patriknw, line 93: Will this become the error handling behavior as the entities? For them I can see that exceptions are turned into EntityException...

johanandren: In the entities it is always a stream of events/commands => responses, and that has a recover in the "far end" catching any failure happening inside of the stream: 
https://github.com/lightbend/akkaserverless-java-sdk/blob/main/sdk/java-sdk/src/main/scala/com/akkaserverless/javasdk/impl/eventsourcedentity/EventSourcedEntitiesImpl.scala#L132

So this is aligning the various action calls with how the entities behave.

johanandren: I think we should add error logging for all these just like for entities though.

patriknw (approved):
LGTM, but I think you wanted to add some more logging

sdk/java-sdk/src/main/scala/com/akkaserverless/javasdk/impl/action/ActionsImpl.scala
johanandren: Ah, one thing that is different wrt is that for entities the exception is hidden and the returned message is "internal error" unless it is specifically an `EntityException`, so that's the solution there for not "leaking" impl details in the error. Maybe I should do that here as well?

johanandren:
Another thought of a possible improvement (for all error logging) would be to pick up the Logger for the user component class when reporting those errors, so that they can more easily be filtered etc, now they will come from `AaaBbbbImpl`.

PR #498, opened 2021-09-24
feat(sbt-plugin): add generateUnmanagedTest

patriknw:
Refs #372


raboof (approved):
scripted test looks like it needs an invocation of `publishLocal` for the testkit, otherwise looks OK to me!

PR #503, opened 2021-09-26
Asserting different effects

franciscolopezsancho:
- [x] creating ad-hoc testkit to test Actions. 
Created a Testkit in `java-eventsourced-counter` for `CounterJournalToTopicAction`
Created one test for it. Test passing. 
- [x] create codegen.
Created Testkit thorught codegen. Test passing.
Added loading user-function.desc from sample. To generate exact `BuilderModel.Model`
Added codegen for generating examples on `src/test`
- [x] add docs.


samples/java-eventsourced-counter/src/main/java/com/example/domain/CounterJournalToTopicActionTestKit.java
```diff
@@ -0,0 +1,73 @@
...
+import com.akkaserverless.javasdk.testkit.impl.StubActionContext;
+import com.example.actions.CounterTopicApi;
+import java.util.Optional;
+
+/**
+ * TestKit for unit testing Counter
+ */
+public final class CounterJournalToTopicActionTestKit {
+
+  private CounterJournalToTopicAction action;
```
johanandren, line 24: Should be a new instance for every call, so don't create it upfront and keep in a field, but use the factory in the generated command methods.

samples/java-eventsourced-counter/src/main/java/com/example/domain/CounterJournalToTopicActionTestKit.java
```diff
@@ -0,0 +1,73 @@
...
+  private List<Object> effects = new ArrayList<Object>();
```
johanandren, line 25: As the actions commands are completely independent and stateless I don't think it makes sense to keep track of them like this, just return the effect to the user for asserting on each call and that's it.

testkit-java/src/main/java/com/akkaserverless/javasdk/testkit/ActionResult.java
```diff
@@ -0,0 +1,59 @@
...
+  // boolean didEmitEffect();??? TODO review if useful
+
+  /**
+   * Look at the next effect and verify that it is of type E or fail if not or if there is no next
+   * effect. If successful this consumes the effect, so that the next call to this method looks at
+   * the next effect from here.
+   *
+   * @return The next effect if it is of type E, for additional assertions.
+   */
+  <E> E getEffectOfType(Class<E> expectedClass);
```
johanandren, line 58: Skip this general effect extractor since there is no public API for users to do anything to inspect the values of those effect interfaces. 

Instead, for each possible effect, provide a pair of `bool isThatEffect()` and a method to inspect that returning something useful according to the type of effect (the `R` reply if it is a reply, the error description for `effects().error(desciption)` for example), see the other testkit Result types for examples of that.

samples/java-eventsourced-counter/src/main/java/com/example/domain/CounterJournalToTopicActionTestKit.java
franciscolopezsancho: done.

samples/java-eventsourced-counter/src/main/java/com/example/domain/CounterJournalToTopicActionTestKit.java
franciscolopezsancho: done

testkit-java/src/main/java/com/akkaserverless/javasdk/testkit/ActionResult.java
franciscolopezsancho: I skipped your suggestion to return R and added the message from the reply instead. I could add its metadata as well but can't think of any specific use case. 
I kept `getEffectOfType` method but for private use only. 

LMKWYT

testkit-java/src/main/java/com/akkaserverless/javasdk/testkit/ActionResult.java
```diff
@@ -0,0 +1,51 @@
...
+import com.akkaserverless.javasdk.ServiceCall;
+import java.util.concurrent.CompletableFuture;
+import com.akkaserverless.javasdk.action.Action;
+
+public interface ActionResult<T> {
+
+  /** @return true if the call had an effect with a reply, false if not */
+  boolean isReply();
+
+  T getReplyMessage();
```
johanandren, line 31: This is exactly what I meant, make sure it the method names aligns with the same in the other Result types though.

testkit-java/src/main/java/com/akkaserverless/javasdk/testkit/ActionResult.java
```diff
@@ -0,0 +1,51 @@
...
+  /** @return true if the call was forwarded, false if not */
+  boolean isForward();
+
+  ServiceCall getForwardServiceCall();
+
+  // TODO rewrite name. Async doesn't seem very descriptive. Are the rest sync?
+  /** @return true if the call was async, false if not */
+  boolean isAsync();
+
+  CompletableFuture<Action.Effect<T>> getAsyncEffect();
```
johanandren, line 42: This one is tricky, as we are back to no user API for this one, I wonder if it perhaps could be `CompletionStage<ActionResult<T>>` , that way a user can block/wait for it in the test and _then_ inspect that result.

Another option could perhaps be to block in here for async effects to complete, in the normal calls if it is async, but then there is the question about what timeout to use, so maybe better to hand the waiting off to the user/test logic.

Note that the return type should be `CompletionStage`, the `CompletableFuture` type is the one a user can complete (like `Promise`).

testkit-java/src/main/scala/com/akkaserverless/javasdk/testkit/impl/StubActionContext.scala
```diff
@@ -0,0 +1,41 @@
...
+import com.akkaserverless.javasdk.valueentity.ValueEntityContext
+import com.akkaserverless.javasdk.action.ActionCreationContext
+import akka.stream.Materializer
+
+/** Context for action calls. */
+final class StubActionContext extends ActionContext {
+
+  override def metadata() = Metadata.EMPTY
+
+  override def eventSubject() = Optional.of("testkitkat")
```
johanandren, line 32: `test-subject-id` or something like that

testkit-java/src/main/scala/com/akkaserverless/javasdk/testkit/impl/StubActionContext.scala
```diff
@@ -0,0 +1,41 @@
...
+import java.util.Optional;
+import com.akkaserverless.javasdk.ServiceCallFactory
+import com.akkaserverless.javasdk.valueentity.ValueEntityContext
+import com.akkaserverless.javasdk.action.ActionCreationContext
+import akka.stream.Materializer
+
+/** Context for action calls. */
+final class StubActionContext extends ActionContext {
+
+  override def metadata() = Metadata.EMPTY
```
johanandren, line 30: Could be better to throw saying it is not supported yet.

samples/java-eventsourced-counter/src/test/java/com/example/domain/CounterJournalToTopicActionTest.java
```diff
@@ -0,0 +1,25 @@
...
+public class CounterJournalToTopicActionTest {
+
+  @Test
+  public void increaseTest() {
+    CounterJournalToTopicActionTestKit testKit = CounterJournalToTopicActionTestKit.of(CounterJournalToTopicAction::new);
+    int valueToincrease = 1;
+    ActionResult<CounterTopicApi.Increased> result = testKit.increase(CounterDomain.ValueIncreased.newBuilder().setValue(valueToincrease).build());
+    assertTrue(result.isReply());
+    CounterTopicApi.Increased replyMessage = result.getReplyMessage();
+    assertEquals(replyMessage.getValue(),valueToincrease);
```
patriknw, line 22: for `assertEquals` the expect value is supposed to be the first parameter

patriknw (commented):
looking good

testkit-java/src/main/scala/com/akkaserverless/javasdk/testkit/impl/StubActionContext.scala
franciscolopezsancho: done now

testkit-java/src/main/scala/com/akkaserverless/javasdk/testkit/impl/StubActionContext.scala
franciscolopezsancho: right!

samples/java-eventsourced-counter/src/test/java/com/example/domain/CounterJournalToTopicActionTest.java
franciscolopezsancho: done

codegen/java-gen/src/test/scala/com/lightbend/akkasls/codegen/java/ActionTestKitGeneratorSuite.scala
```diff
@@ -0,0 +1,165 @@
...
+    val className = "CounterJournalToTopicAction"
+    val service: ModelBuilder.ActionService = {
+      model.get.services("com.example.actions.CounterJournalToTopic").asInstanceOf[ModelBuilder.ActionService]
+
+    }
+    val entity: ModelBuilder.EventSourcedEntity = {
+      model.get.entities("com.example.domain.Counter").asInstanceOf[ModelBuilder.EventSourcedEntity]
+    }
+
+    //This prints the contents of the model, services or entity. Necessary to
```
franciscolopezsancho, line 52: This is a new approach to create the initial `BuilderModel.Model` by reading the `user-function.desc` file from one of the samples. Then we can inspect its content by printing it with `munit.Assertions.munitPrint(model)`.

codegen/java-gen/src/test/scala/com/lightbend/akkasls/codegen/java/ActionTestKitGeneratorSuite.scala
```diff
@@ -0,0 +1,165 @@
...
+  }
+
+  test("it can generate an specific Test stub for the entity") {}
+
+  /**
+   * This ModelBuilder.EventSourcedEntity is equivalent to the entity in
+   * test/resources/testkit/shoppingcart_domain.proto
+   */
+  def generateModel(): Try[ModelBuilder.Model] = {
+
```
franciscolopezsancho, line 130: This is a new approach to create the initial BuilderModel.Model by reading the user-function.desc file from one of the samples. Then we can inspect its content by printing it with munit.Assertions.munitPrint(model).

codegen/java-gen/src/main/scala/com/lightbend/akkasls/codegen/java/ActionTestKitGenerator.scala
```diff
@@ -0,0 +1,127 @@
...
+        "Empty"
+      } else {
+        command.outputType.fullName
+      }
+
+    val domainOuterClass = entity.fqn.parent.name
+
+    service.commands
+      .map { command =>
+        s"""|public ActionResult<${selectOutput(command)}> ${lowerFirst(command.name)}(${domainOuterClass}.${command.inputType.protoName} event) {
```
johanandren, line 118: it's not an `event` here, use either `command` or the actual name of the input parameter type (with `lowerFirst`).

codegen/java-gen/src/test/scala/com/lightbend/akkasls/codegen/java/ActionTestKitGeneratorSuite.scala
johanandren: Ok, was it added because that is easier than to build up a realistic model manually like the TestData?

samples/java-eventsourced-counter/src/main/java/com/example/actions/CounterJournalToTopicActionTestKit.java
```diff
@@ -1,24 +1,25 @@
-package com.example.domain;
+/* This code is managed by Akka Serverless tooling.
+ * It will be re-generated to reflect any changes to your protobuf definitions.
+ * DO NOT EDIT
+ */
+
+package com.example.actions;
```
johanandren, line 6: Shouldn't be checked in now anymore that we have the codegen?

testkit-java/src/main/java/com/akkaserverless/javasdk/testkit/ActionResult.java
```diff
@@ -20,15 +20,15 @@
...
+import java.util.concurrent.CompletionStage;
 import com.akkaserverless.javasdk.action.Action;
 
 public interface ActionResult<T> {
 
   /** @return true if the call had an effect with a reply, false if not */
   boolean isReply();
 
-  T getReplyMessage();
+  T getReplyMsg();
```
johanandren, line 31: I think it should be `getReply()` like in the value entity and es entity `Result` types.

testkit-java/src/main/java/com/akkaserverless/javasdk/testkit/ActionResult.java
```diff
@@ -39,7 +39,9 @@
   /** @return true if the call was async, false if not */
   boolean isAsync();
 
-  CompletableFuture<Action.Effect<T>> getAsyncEffect();
+  CompletionStage<Action.Effect<T>> getAsyncEffect();
+
+  ActionResult<T> createResult(Action.Effect<T> effect);
```
johanandren, line 44: Will not be needed with the above fix.

testkit-java/src/main/java/com/akkaserverless/javasdk/testkit/ActionResult.java
```diff
@@ -39,7 +39,9 @@
   /** @return true if the call was async, false if not */
   boolean isAsync();
 
-  CompletableFuture<Action.Effect<T>> getAsyncEffect();
+  CompletionStage<Action.Effect<T>> getAsyncEffect();
```
johanandren, line 42: Should be `CompletionStage<ActionResult<T>>`

johanandren (commented):
Good progress!

samples/java-eventsourced-counter/src/main/java/com/example/actions/CounterJournalToTopicActionTestKit.java
franciscolopezsancho: It was added because of that, yes. Easier and less error-prone. 

You're right I can remove it now. I guess I was sentimentally attached to it :))

testkit-java/src/main/java/com/akkaserverless/javasdk/testkit/ActionResult.java
franciscolopezsancho: ok

testkit-java/src/main/java/com/akkaserverless/javasdk/testkit/ActionResult.java
franciscolopezsancho: right. done, now

testkit-java/src/main/java/com/akkaserverless/javasdk/testkit/ActionResult.java
franciscolopezsancho: removed

codegen/java-gen/src/main/scala/com/lightbend/akkasls/codegen/java/ActionTestKitGenerator.scala
franciscolopezsancho: ok, done. 
Question why is then in the past tense?

codegen/java-gen/src/test/scala/com/lightbend/akkasls/codegen/java/ActionTestKitGeneratorSuite.scala
franciscolopezsancho: yep. I find it easier and less prone to error.

codegen/java-gen/src/main/scala/com/lightbend/akkasls/codegen/java/ActionTestKitGenerator.scala
```diff
@@ -0,0 +1,127 @@
...
+        "com.akkaserverless.javasdk.testkit.impl.ActionResultImpl",
+        "com.akkaserverless.javasdk.impl.action.ActionEffectImpl",
+        "com.example.actions.CounterJournalToTopicAction",
+        "com.example.actions.CounterTopicApi",
+        "com.akkaserverless.javasdk.testkit.impl.StubActionCreationContext",
+        "com.akkaserverless.javasdk.testkit.impl.StubActionContext"))
+
+    val testKitClassName = s"${className}TestKit"
+
+    println(packageName)
```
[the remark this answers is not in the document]
franciscolopezsancho: done

testkit-java/src/main/java/com/akkaserverless/javasdk/testkit/ActionResult.java
franciscolopezsancho: done

testkit-java/src/main/java/com/akkaserverless/javasdk/testkit/ActionResult.java
franciscolopezsancho: done

codegen/java-gen/src/main/scala/com/lightbend/akkasls/codegen/java/ActionTestKitGenerator.scala
johanandren: What is in past tense?

franciscolopezsancho: for instance  the input in `public ActionResult<com.example.actions.CounterTopicApi.Increased> increase(com.example.domain.CounterDomain.ValueIncreased valueIncreased)`

codegen/java-gen/src/test/scala/com/lightbend/akkasls/codegen/java/ActionTestKitGeneratorSuite.scala
```diff
@@ -34,26 +34,25 @@ import scala.util.Try
...
-    val entity: ModelBuilder.EventSourcedEntity = {
-      model.get.entities("com.example.domain.Counter").asInstanceOf[ModelBuilder.EventSourcedEntity]
     }
+    val className = service.className
 
-    //This prints the contents of the model, services or entity. Necessary to
+    //This prints the contents of the model, services or entity. It's necessary to
     // see how the user-function.desc (see generateModel below) maps to ModelBuilder objects.
-    // println(munit.Assertions.munitPrint(entity))
+    log.debug(munit.Assertions.munitPrint(service))
```
johanandren, line 53: I think using println temporarily to inspect when changing something is fine, instead of adding test logging config etc. Once the test passes it's not super interesting with that model printed?

testkit-java/src/main/java/com/akkaserverless/javasdk/testkit/ActionResult.java
```diff
@@ -39,9 +39,7 @@
   /** @return true if the call was async, false if not */
   boolean isAsync();
 
-  CompletionStage<Action.Effect<T>> getAsyncEffect();
-
-  ActionResult<T> createResult(Action.Effect<T> effect);
+  CompletionStage<? extends ActionResult<T>> getAsyncEffect();
```
johanandren, line 42: Hmm. I don't think a wildcard is needed here and that it could be just `<ActionResult<T>>`.

testkit-java/src/main/scala/com/akkaserverless/javasdk/testkit/impl/ActionResultImpl.scala
```diff
@@ -48,13 +51,11 @@ final class ActionResultImpl[T](effect: Action.Effect[T]) extends ActionResult[T
   /** @return true if the call was async, false if not */
   def isAsync(): Boolean = effect.isInstanceOf[ActionEffectImpl.AsyncEffect[T]]
 
-  def getAsyncEffect(): CompletionStage[Action.Effect[T]] = {
+  def getAsyncEffect(): CompletionStage[ActionResultImpl[T]] = {
```
johanandren, line 54: No need to be specific here, with `CompletionStage<ActionResult<T>>` above this should be `CompletionStage[ActionResult[T]]` as well.

johanandren (commented):
It's shaping up nicely!

codegen/java-gen/src/test/scala/com/lightbend/akkasls/codegen/java/ActionTestKitGeneratorSuite.scala
franciscolopezsancho: Ok. I wasn't even sure myself. Adding the log file and the logger might be a bit too much 👍

testkit-java/src/main/java/com/akkaserverless/javasdk/testkit/ActionResult.java
franciscolopezsancho: I tried and failed :) I found it was necessary as I got the error that it was invariant.

franciscolopezsancho: But now that I read your next comment. Let me double check

franciscolopezsancho: Double checked. I need this as Java is invariant on Generics by default. Would you agree?

testkit-java/src/main/scala/com/akkaserverless/javasdk/testkit/impl/ActionResultImpl.scala
franciscolopezsancho: You're right. I need the covariant version, though `  def getAsyncEffect(): CompletionStage[_ <: ActionResult[T]]`.

testkit-java/src/main/java/com/akkaserverless/javasdk/testkit/ActionResult.java
johanandren: I think you can just cast on the impl side if the compiler is complaining, we know it is safe.

testkit-java/src/main/scala/com/akkaserverless/javasdk/testkit/impl/ActionResultImpl.scala
johanandren: Make the return type invariant and make sure the returned value aligns (we know it does), a type ascription inside map (to `ActionResult[T]`), an explicit type param to `toJava[ActionResult[T]]` or casting the entire before returning result should do the trick.

franciscolopezsancho: That sounds better than my approach. Nice!

testkit-java/src/main/java/com/akkaserverless/javasdk/testkit/ActionResult.java
franciscolopezsancho: I'll give it a go

franciscolopezsancho: done

testkit-java/src/main/scala/com/akkaserverless/javasdk/testkit/impl/ActionResultImpl.scala
franciscolopezsancho: done

codegen/java-gen/src/test/scala/com/lightbend/akkasls/codegen/java/ActionTestKitGeneratorSuite.scala
```diff
@@ -0,0 +1,218 @@
...
+
+  test("it can generate an specific Test stub for the entity") {
+
+    val sourceCode = ActionTestKitGenerator.generateTestSourceCode(service)
+
+    val expected =
+      """/* This code was generated by Akka Serverless tooling.
+        | * As long as this file exists it will not be re-generated.
+        | * You are free to make changes to this file.
+        | */
```
johanandren, line 128: Note that we did some changes wrt the generated banner last week in #550

sdk/java-sdk/src/main/java/com/akkaserverless/javasdk/action/ActionContext.java
```diff
@@ -35,7 +35,7 @@
    *
    * @return The call level metadata.
    */
-  Metadata metadata();
+  Metadata metadata(); // TODO this is not necessary
```
johanandren, line 38: It has better docs than the MetadataContext one though :)

codegen/java-gen/src/test/scala/com/lightbend/akkasls/codegen/java/ActionTestKitGeneratorSuite.scala
franciscolopezsancho: Thanks!!

sdk/java-sdk/src/main/java/com/akkaserverless/javasdk/action/ActionContext.java
franciscolopezsancho: hehe 
Shall I add in there `FIX: remove and move docs from this method class to MetadataContext`? for the moment?

johanandren: Sure

franciscolopezsancho: ok

codegen/java-gen/src/main/scala/com/lightbend/akkasls/codegen/java/ActionTestKitGenerator.scala
```diff
@@ -0,0 +1,199 @@
...
+  }
+
+  def filterJavaOuterClassname(javaOuterClassnameOption: Option[String]): String = {
+    javaOuterClassnameOption match {
+      case None => ""
+      case Some(value) =>
+        if (value == "EmptyProto" || value == "AnyProto")
+          "" //TODO review this. Might not be needed if proto are 'well' defined.
+        // original .proto file on test might be badly defined
+        // e.g. `rpc Ignore(google.protobuf.Any) returns (google.protobuf.Empty)` might have to be `rpc Ignore(com.google.protobuf.Any) returns (com.google.protobuf.Empty)`
```
johanandren, line 194: This is something that I think works in the other generators, so why is it special here, or is it broken there as well?

samples/java-eventsourced-counter/src/test/java/com/example/actions/CounterJournalToTopicActionTest.java
```diff
@@ -0,0 +1,43 @@
...
+    // ActionResult<com.example.actions.CounterTopicApi.Decreased> result = testKit.decrease(com.example.domain.CounterDomain.ValueDecreased.newBuilder()...build());
+  }
+
+  @Test
+  public void ignoreTest() {
+    CounterJournalToTopicActionTestKit testKit = CounterJournalToTopicActionTestKit.of(CounterJournalToTopicAction::new);
+    // ActionResult<com.google.protobuf.Empty> result = testKit.ignore(com.google.protobuf.Any.newBuilder()...build());
+  }
+
+}
```
octonato, line 43: we can maybe include that last `}` for completeness. As such the code show on screen is compilable.
```suggestion
// tag::class[]
}
// end::class[]
```

docs/src/modules/java/pages/actions.adoc
```diff
@@ -130,8 +130,43 @@ Scala::
...
+== Testing the Action 
+
+
+=== Unit tests
+
+The following snippet shows how the `CounterJournalToTopicActionTestKit` is used to test the `CounterJournalToTopicAction` implementation.
+
+AkkaServerless provides the `CounterJournalToTopicActionTestKit` that allows us to call the methods of `CounterJournalToTopicAction`. For each `Action` AkkaServerless generates an specific test kit for it, with the name `[ActionName]TestKit`. This test kit hold an instance of its `Action`. Each call we execute over to the test kit returns an `ActionResult` that holds the effect produced by the underlying `CounterJournalToTopicAction`. 
+
+Apart from the test kit, to facilitate the creation of unit tests, AkkaServerless generates test classes, ready to use, based on the Action defined in the .proto files. This is shown in the snippet below.
```
octonato, line 148:
```suggestion
Apart from the test kit AkkaServerless generates test classes based on the Action defined in the .proto files. This is shown in the snippet below. 
```

We can have a more straightforward text if we remove the `, something in the middle,`. In that case here, it's clear that we generate it to facilitate the creation of tests and that the generated tests is ready for use.

codegen/java-gen/src/test/scala/com/lightbend/akkasls/codegen/java/ActionTestKitGeneratorSuite.scala
```diff
@@ -122,13 +120,12 @@ class ActionTestKitGeneratorSuite extends munit.FunSuite {
...
 
     val expected =
-      """/* This code was generated by Akka Serverless tooling.
-        | * As long as this file exists it will not be re-generated.
-        | * You are free to make changes to this file.
-        | */
+      """// This class was initially generated based on the .proto definition by Akka Serverless tooling.
+        |//
+        |// As long as this file exists it will not be overwritten: you can maintain it yourself,
+        |// or delete it so it is regenerated as needed.
```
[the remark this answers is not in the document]
octonato: yeah, it was. 

@franciscolopezsancho, the reason is that IntelliJ collapse the first comment in the file and we want to have it visible to the users. Check the other generators.

docs/src/modules/java/pages/actions.adoc
```diff
@@ -130,8 +130,43 @@ Scala::
...
+----
+include::example$java-eventsourced-counter/src/test/java/com/example/actions/CounterJournalToTopicActionTest.java[tag=class] 
+----
+<1> creating the test kit to test `increase` Action's method.
+<2> calling `increase` method with some value.
+<3> retrieving the `Reply` from `increase` method.
+
+*ActionResult*
+
+Calling an entity method through the TestKit gives us back an link:{attachmentsdir}/testkit/com/akkaserverless/javasdk/testkit/ActionResult.html[`ActionResult` {tab-icon}, window="new"]. This class has methods that we can use to assert our expectations, such as:
```
octonato, line 163:
```suggestion
Calling an action method through the TestKit gives us back an link:{attachmentsdir}/testkit/com/akkaserverless/javasdk/testkit/ActionResult.html[`ActionResult` {tab-icon}, window="new"]. This class has methods that we can use to assert our expectations, such as:
```

docs/src/modules/java/pages/actions.adoc
```diff
@@ -130,8 +130,43 @@ Scala::
...
+
+* `getReply` is the response of the method called in the test
+* `getForwardServiceCall` returns the `ServiceCall` to which is fowarded the command the `Action` receives.
```
octonato, line 166: this two sentences are not so clear to me, what about...

```suggestion
* `getReply` returns the value passed to the `Effect.reply`, if any.
* `getForwardServiceCall` returns the `ServiceCall` to use to forward the call, if any.
```

octonato (commented):
Looking good. 

Added a few comments...

codegen/java-gen/src/main/scala/com/lightbend/akkasls/codegen/java/ActionTestKitGenerator.scala
franciscolopezsancho: Is there no more. You're totally right is wasn't needed

codegen/java-gen/src/test/scala/com/lightbend/akkasls/codegen/java/ActionTestKitGeneratorSuite.scala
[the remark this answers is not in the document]
franciscolopezsancho: right. One sec.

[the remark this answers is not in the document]
franciscolopezsancho: done

docs/src/modules/java/pages/actions.adoc
```diff
@@ -130,8 +130,43 @@ Scala::
...
+*ActionResult*
+
+Calling an action method through the TestKit gives us back an link:{attachmentsdir}/testkit/com/akkaserverless/javasdk/testkit/ActionResult.html[`ActionResult` {tab-icon}, window="new"]. This class has methods that we can use to assert our expectations, such as:
+
+* `getReply` returns the value passed to the `Effect.reply`, if any.
+* `getForwardServiceCall` returns the `ServiceCall` to use to forward the call, if any.
+
+
+
+*CounterJournalToTopicActionTestKit*
```
johanandren, line 170: I don't think you need to mention the generated test kit again, it's introduced in the text above.

johanandren (commented):
Closing up on being done this one. 👍

docs/src/modules/java/pages/actions.adoc
franciscolopezsancho: less is more

docs/src/modules/java/pages/actions.adoc
```diff
@@ -143,7 +143,7 @@ When more components are added the `AkkaServerlessFactory` is regenerated, and y
 
 The following snippet shows how the `CounterJournalToTopicActionTestKit` is used to test the `CounterJournalToTopicAction` implementation.
 
-AkkaServerless provides the `CounterJournalToTopicActionTestKit` that allows us to call the methods of `CounterJournalToTopicAction`. For each `Action` AkkaServerless generates an specific test kit for it, with the name `[ActionName]TestKit`. This test kit hold an instance of its `Action`. Each call we execute over to the test kit returns an `ActionResult` that holds the effect produced by the underlying `CounterJournalToTopicAction`. 
+AkkaServerless generates the `CounterJournalToTopicActionTestKit` that allows us to call the methods of `CounterJournalToTopicAction`. For each `Action` Akka Serverless generates a specific test kit for it, with the name `[ActionName]TestKit`. This test kit hold an instance of its `Action`. Each call we execute over to the test kit returns an `ActionResult` that holds the effect produced by the underlying `CounterJournalToTopicAction`.
```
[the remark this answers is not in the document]
johanandren: And one more in the paragraph below

Issue #486, opened 2021-09-23
CI testing of unmanaged sources in samples


patriknw:
Unless I have missed something we are not testing that the generated unmanaged sources for the samples compile correctly. The unmanaged sources are in git and implemented by us.

We could have a CI job that removes src/main/ in the samples and runs sbt/mvn compile.

octonato:
Ok, I see what you mean. Makes sense. 

We can do it after we run the samples, we can have another round that remove all the implementations and do the test-compile

patriknw:
This was fixed.

PR #505, opened 2021-09-27
chore: Verify unmanaged sources compile

johanandren:
References #486

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -86,6 +86,14 @@ commands:
             echo "Running maven-java with SDK version: '$SDK_VERSION'"
             mvn versions:set -DnewVersion=$SDK_VERSION
             mvn verify install
+
+  build-sbt-scala:
+    description: Build sbt plugin and scala artifacts
+    steps:
+      - run:
+          name: publish local
+          command: sbt publishLocal
+
 jobs:
   checks:
     docker:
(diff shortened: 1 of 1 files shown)


johanandren:
Not the best task to pick up when CircleCI is having an outage :)

.circleci/config.yml
```diff
@@ -86,6 +86,14 @@ commands:
             echo "Running maven-java with SDK version: '$SDK_VERSION'"
             mvn versions:set -DnewVersion=$SDK_VERSION
             mvn verify install
+
+  build-sbt-scala:
+    description: Build sbt plugin and scala artifacts
+    steps:
+      - run:
+          name: publish local
+          command: sbt publishLocal
```
patriknw, line 95: I think it might be better to `publishLocal` before above maven publish, because `mvn version:set` modifies some files and the version number will not become the same here.

patriknw (commented):
Looking good with one proposal..

.circleci/config.yml
johanandren: Ah, that's probably causing the problem I mentioned that I didn't quite understand yesterday. thx

.circleci/config.yml
```diff
@@ -73,19 +73,20 @@ commands:
...
-  build-maven-java:
-    description: Build maven-java
+  publish-local:
+    description: Build and publish artifacts and plugins locally for other tests to use
     steps:
       - run:
-          name: publish m2
-          command: sbt publishM2
+          name: publish m2 and local
+          command: sbt publishM2 publishLocal
```
patriknw, line 81: good

patriknw (approved):
LGTM

PR #507, opened 2021-09-27
Scala value entity counter sample

johanandren:
Java value entity counter sample ported and value entity doc page updated with tabs for each language.


samples/scala-valueentity-counter/build.sbt
```diff
@@ -0,0 +1,48 @@
...
+val AkkaServerlessSdkVersion = System.getProperty("akkaserverless-sdk.version", "0.7.2")
+libraryDependencies ++= Seq(
+  logback,
+  logbackJson,
+  logbackJackson,
+  "org.scalatest" %% "scalatest" % "3.2.7" % Test,
+  // FIXME include testkit dependency via the AkkaserverlessPlugin
+  "com.akkaserverless" %% "akkaserverless-scala-sdk" % AkkaServerlessSdkVersion,
+  "com.akkaserverless" %% "akkaserverless-scala-sdk-testkit" % AkkaServerlessSdkVersion % Test)
+
```
patriknw, line 48: dependencies should be updated according to #508
only 
```
libraryDependencies ++= Seq("org.scalatest" %% "scalatest" % "3.2.7" % Test)
```

samples/scala-valueentity-counter/src/main/proto/com/example/actions/double-counter.proto
```diff
@@ -0,0 +1,33 @@
...
+//
+// tag::double-action-proto[]
+syntax = "proto3";
+package com.example.actions;
+
+import "akkaserverless/annotations.proto"; 
+import "com/example/counter_api.proto"; // <1>
+import "google/protobuf/empty.proto";
+
+option java_outer_classname = "DoubleCounterApi";
```
patriknw, line 23: `java_outer_classname` can be removed from Scala samples. It will not be used, but could be confusing.

samples/scala-valueentity-counter/src/main/scala/com/example/Main.scala
```diff
@@ -0,0 +1,51 @@
...
+    AkkaServerlessFactory.withComponents(
+      new Counter(_),
+      new DoubleCounterAction(_));
+    // end::registration[]
+    */
+
+    /* the comment hack bellow is needed to only show the Counter::new
+    // tag::registration-value-entity[]
+    AkkaServerlessFactory.withComponents(
+      new Counter(_));
```
patriknw, line 38: some semi leftovers here

samples/scala-valueentity-counter/src/main/scala/com/example/actions/DoubleCounterAction.scala
```diff
@@ -0,0 +1,47 @@
...
+import com.akkaserverless.scalasdk.SideEffect
+import com.akkaserverless.scalasdk.action.Action
+import com.akkaserverless.scalasdk.action.ActionCreationContext
+import com.example.IncreaseValue
+import com.google.protobuf.empty.Empty
+
+// tag::controller-forward[]
+// tag::controller-side-effect[]
+/** An action. */
+class DoubleCounterAction(creationContext: ActionCreationContext) extends AbstractDoubleCounterAction {
```
patriknw, line 16: @RayRoestenburg note that some of the samples for actions docs is already included here

samples/scala-valueentity-counter/src/test/scala/com/example/domain/CounterSpec.scala
```diff
@@ -0,0 +1,57 @@
...
+    extends AnyWordSpec
+    with Matchers {
+
+  "Counter" must {
+
+    "handle command Increase" in {
+      val testKit = CounterTestKit(new Counter(_))
+
+      val result1 = testKit.increase(IncreaseValue(value = 1))
+      result1.getReply should === (Empty.defaultInstance)
```
patriknw, line 23: We should decide what assert matchers to use. In Akka we used `===` for a while because we wanted to have the compile time type check, but I'm not sure that is working here (without adding more stuff)?

I'd suggest we use the more simplistic
```
result1.getReply should shouldBe Empty.defaultInstance
```

samples/scala-valueentity-counter/src/test/scala/com/example/domain/CounterSpec.scala
```diff
@@ -0,0 +1,57 @@
...
+      testKit.getState.value should ===(1)
+
+      val result2 = testKit.decrease(DecreaseValue(value = 1))
+      result2.getReply should ===(Empty.defaultInstance)
+      testKit.getState.value should ===(0)
+    }
+
+    "handle command Reset" in {
+      val testKit = CounterTestKit(new Counter(_))
+      // val result = testKit.increase(example.IncreaseValue(...))
```
patriknw, line 46: remove, or is there a reason to keep?

samples/scala-valueentity-counter/src/test/scala/com/example/domain/CounterSpec.scala
```diff
@@ -0,0 +1,57 @@
...
+      // one more time
+      val result2 = testKit.increase(IncreaseValue(value = 1))
+      result2.getReply should === (Empty.defaultInstance)
+      testKit.getState.value should === (2)
+    }
+    // end::sample-unit-test[]
+
+    "handle command Decrease" in {
+      val testKit = CounterTestKit(new Counter(_))
+      // val result = testKit.increase(example.IncreaseValue(...))
```
patriknw, line 34: remove, or is there a reason to keep?

docs/dev/antora.yml
```diff
@@ -9,7 +9,7 @@ content:
 
 ui:
   bundle:
-    url: https://lightbend.github.io/antora-supplemental-ui-lightbend-theme/default_ui/ui-bundle-1023155878.zip
```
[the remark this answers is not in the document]
patriknw: not sure, but compare with Javascript https://github.com/lightbend/akkaserverless-javascript-sdk/pull/195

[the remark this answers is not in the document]
johanandren: Tried building the java docs, and that preview doesn't show the tabs either (and also didn't get an updated theme version). Maybe there was no new zip with the tabs created, or maybe it needs more infra to actually show tabs. Not super important just a minor annoyance.

samples/scala-valueentity-counter/src/test/scala/com/example/domain/CounterSpec.scala
johanandren: The test passes with any further imports etc. but I agree, the tripple equals is anyway a bit funky.

One thing I thought about: should we make the testkit more Scala:y with a `.reply` instead of `.getReply` and `.state` instead of `.getState` when we anyway have specific types?

patriknw: > The test passes with any further imports etc. but I agree, the tripple equals is anyway a bit funky.

I mean, does it find type errors at compile time? This shouldn't compile if it should have any added value
```
result1.getReply should === ("foo")
```

One thing I thought about: should we make the testkit more Scala:y

That would be better for the Result, but for the generated testkit we have `getState()` and that might be somewhat confusing as `state` or `state()` since it's changing. We could name that `currentState()` if we prefer that over `getState()`?

johanandren: I do like `currentState()` better.

johanandren: Changed, `getReply()` => `reply`, `getState()` => `currentState()`

samples/scala-valueentity-counter/build.sbt
```diff
@@ -40,14 +40,9 @@ val logback = "ch.qos.logback" % "logback-classic" % LogbackVersion
 val logbackJson = "ch.qos.logback.contrib" % "logback-json-classic" % LogbackContribVersion
 val logbackJackson = "ch.qos.logback.contrib" % "logback-jackson" % LogbackContribVersion
 
-// FIXME include these dependencies via the AkkaserverlessPlugin
-val AkkaServerlessSdkVersion = System.getProperty("akkaserverless-sdk.version", "0.7.2")
 libraryDependencies ++= Seq(
   logback,
   logbackJson,
   logbackJackson,
```
patriknw, line 46: Still some logback dependencies here that shouldn't be needed. They should be included via `akkaserverless-scala-sdk` unless we changed that.

docs/dev/antora.yml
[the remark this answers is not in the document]
patriknw: @pvlugter or @ennru Do you know what needs to be done to enable tabs in the javasdk docs? I assume we already have that working in javascript?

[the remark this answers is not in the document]
johanandren: To be clear: I'm expecting that it works when it is pulled into the main docs site but not when building and previewing the respective sdks (didn't try and verify though).

patriknw (approved):
LGTM, let's merge this first in the PR queue

docs/dev/antora.yml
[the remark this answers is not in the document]
patriknw: I can confirm that it's looking good when building from akkaserverless-docs using 
```
antora --stacktrace local.yml && cp get.sh build/site/
```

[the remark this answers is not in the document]
pvlugter: Yes, would be good to have the actual theme here in the SDK docs. For JS docs, I just viewed the main docs locally. Haven't looked at it properly yet, but akkaserverless-docs seems to have a bunch of things locally where it would be better to have everything in a published bundle.

PR #508, opened 2021-09-27
feat(scalasdk): add standard dependencies to project via sbt plugin.

RayRoestenburg:


raboof (approved):
Seems reasonable to me

samples/scala-value-entity-customer-registry/src/test/scala/customer/domain/CustomerValueEntitySpec.scala
```diff
@@ -0,0 +1,53 @@
+/* This code was generated by Akka Serverless tooling.
```
johanandren, line 1: This looks unrelated though?

johanandren (approved):

samples/scala-value-entity-customer-registry/src/test/scala/customer/domain/CustomerValueEntitySpec.scala
RayRoestenburg: yeah that was me running a test, I'll remove it from the PR.

RayRoestenburg:
I added some settings to exclude the standard added dependencies, maybe overkill.

samples/scala-value-entity-customer-registry/src/test/scala/customer/domain/CustomerValueEntitySpec.scala
RayRoestenburg: removed.

johanandren:
Can't removing them be done by a user anyway without additional settings here? `libraryDependencies := libraryDedendencies.value.filter()`

sbt-plugin/src/main/scala/com/akkaserverless/sbt/AkkaserverlessPlugin.scala
```diff
@@ -41,14 +48,39 @@ object AkkaserverlessPlugin extends AutoPlugin {
...
+    val excludeScalaTestDependencies = settingKey[Boolean]("Exclude scalatest dependencies")
   }
   object autoImport extends Keys
   import autoImport._
 
+  val LogbackVersion = "1.2.3"
+  val LogbackContribVersion = "0.1.5"
+  val logback = "ch.qos.logback" % "logback-classic" % LogbackVersion
+  val logbackJson = "ch.qos.logback.contrib" % "logback-json-classic" % LogbackContribVersion
+  val logbackJackson = "ch.qos.logback.contrib" % "logback-jackson" % LogbackContribVersion
```
patriknw, line 61: I don't think the Logback dependencies should be included by the plugin. Instead they should be included as ordinary transitive dependencies via the sdk library. I thought I had already fixed that in https://github.com/lightbend/akkaserverless-java-sdk/pull/496 ?

sbt-plugin/src/main/scala/com/akkaserverless/sbt/AkkaserverlessPlugin.scala
```diff
@@ -41,14 +48,39 @@ object AkkaserverlessPlugin extends AutoPlugin {
...
+  val ScalaTestVersion = "3.2.7"
```
patriknw, line 62: I'm leaning towards not including ScalaTest dependency because we all know there are opinions about test frameworks and by having that as an explicit dependency in the example build it's easier to switch. We only depend on it in the unmanaged (one time) generated test class so users can replace that code with their own test framework code.

We have a similar approach for Java where users can switch between Junit4 and 5 or whatnot.

patriknw (commented):
not sure this is the right approach...

sbt-plugin/src/main/scala/com/akkaserverless/sbt/AkkaserverlessPlugin.scala
RayRoestenburg: Ah, did not know that.. yes is fine.

sbt-plugin/src/main/scala/com/akkaserverless/sbt/AkkaserverlessPlugin.scala
RayRoestenburg: ok

samples/scala-value-entity-customer-registry/build.sbt
```diff
@@ -28,20 +28,3 @@ Test / logBuffered := false
...
-val logbackJson = "ch.qos.logback.contrib" % "logback-json-classic" % LogbackContribVersion
-val logbackJackson = "ch.qos.logback.contrib" % "logback-jackson" % LogbackContribVersion
-
-// FIXME include these dependencies via the AkkaserverlessPlugin
-val AkkaServerlessSdkVersion = System.getProperty("akkaserverless-sdk.version", "0.7.2")
-libraryDependencies ++= Seq(
-  logback,
-  logbackJson,
-  logbackJackson,
-  "org.scalatest" %% "scalatest" % "3.2.7" % Test,
```
patriknw, line 44: ScalaTest must remain here, if you agree with https://github.com/lightbend/akkaserverless-java-sdk/pull/508#discussion_r717260609

sbt-plugin/src/sbt-test/sbt-akkaserverless/basic/build.sbt
```diff
@@ -1,9 +1,3 @@
 scalaVersion := "2.13.6"
 
 enablePlugins(AkkaserverlessPlugin)
-
-val AkkaServerlessSdkVersion = System.getProperty("plugin.version", "0.7.2")
-libraryDependencies ++= Seq(
-  "org.scalatest" %% "scalatest" % "3.2.7" % Test,
```
patriknw, line 7: ScalaTest must remain here, if you agree with https://github.com/lightbend/akkaserverless-java-sdk/pull/508#discussion_r717260609

samples/scala-value-entity-customer-registry/build.sbt
```diff
@@ -28,20 +28,3 @@ Test / logBuffered := false
...
-  // FIXME include testkit dependency via the AkkaserverlessPlugin
-  "com.akkaserverless" %% "akkaserverless-scala-sdk-testkit" % AkkaServerlessSdkVersion % Test)
-
```
patriknw, line 47: don't we have another Scala sample? Fibonacci? Is that up to date?

sbt-plugin/src/main/scala/com/akkaserverless/sbt/AkkaserverlessPlugin.scala
RayRoestenburg: removed these.

sbt-plugin/src/main/scala/com/akkaserverless/sbt/AkkaserverlessPlugin.scala
RayRoestenburg: removed.

samples/scala-value-entity-customer-registry/build.sbt
RayRoestenburg: ah good catch

sbt-plugin/src/main/scala/com/akkaserverless/sbt/AkkaserverlessPlugin.scala
RayRoestenburg: (yeah this is why I added the settings to remove the deps but yeah that was just an idea)

PR #512, opened 2021-09-28
docs: Scala in views docs

patriknw:
* still rather many things left as TODO because they depend on the event sourced sample

References https://github.com/lightbend/akkaserverless-java-sdk/issues/372

### DIFF
--- a/docs/src/modules/java/pages/views.adoc
+++ b/docs/src/modules/java/pages/views.adoc
@@ -1,5 +1,5 @@
-= Implementing Views in Java
-
+= Implementing Views in Java or Scala
+:page-supergroup-java-scala: Language
 include::ROOT:partial$include.adoc[]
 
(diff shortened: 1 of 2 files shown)


docs/src/modules/java/pages/views.adoc
```diff
@@ -29,24 +29,48 @@ Consider an example of a Customer Registry service with a `customer` Value Entit
...
+[.tabset]
+Java::
++
 [source,proto,indent=0]
 .src/main/proto/customer/domain/customer_domain.proto
 ----
 include::example$java-valueentity-customer-registry/src/main/proto/customer/domain/customer_domain.proto[tags=declarations;domain]
 ----
 
+[.tabset]
```
johanandren, line 41: I think this is not right (i did the same in the value entity docs) and that it should be a single `[.tabset]` for the two language tabs rather.

johanandren (approved):
LGTM so far except the duplicated tabsets-tags

docs/src/modules/java/pages/views.adoc
patriknw: thanks, I'll change that

patriknw:
I need to rebase this after https://github.com/lightbend/akkaserverless-java-sdk/pull/515 because the name change of the sample.

PR #515, opened 2021-09-28
chore(scalasdk): rename scala-value-entity-customer-registry

RayRoestenburg:
to scala-valueentity-customer-registry (consistency with other project directories).


raboof (approved):
Hmm, I guess that seems reasonable - would that mean we'd also want to rename the maven archetypes?

RayRoestenburg:
@raboof yeah, sounds like a good cleanup.

samples/scala-valueentity-customer-registry/scala-value-entity-customer-registry.sublime-project
```diff
@@ -0,0 +1,7 @@
+{
```
patriknw, line 1: could you remove all sublime files and add to gitignore

patriknw (commented):
good to rename, some sublime files...

samples/scala-valueentity-customer-registry/scala-value-entity-customer-registry.sublime-project
RayRoestenburg: ah crap, will do.

patriknw (approved):
LGTM

PR #519, opened 2021-09-28
chore: align command names

octonato:
Pure aesthetics, but nice to users...

Aligns the command names in the abstract and implementation classes. 

The 'issue' manifests as following: 

The abstract class will have methods defined as: 

`abstract public Effect<Empty> increase(CounterDomain.CounterState currentState, CounterApi.IncreaseValue increaseValue)`

The impl class will have: 

`public Effect<Empty> increase(CounterDomain.CounterState currentState, CounterApi.IncreaseValue command)`

Note the difference in the command name: `increaseValue` vs. `command`.

When we add a new method to proto file, the abstract class will get a new method with an argument with the name of the command. 

In IntelliJ, user gets an error in impl class and let the IDE generate the method for them. The new generated method will have an argument like `increaseValue` (following the abstract method name). Now, the impl class will have a mix of commands names as `command` and as `someCommandName`.

### DIFF
--- a/codegen/java-gen/src/main/scala/com/lightbend/akkasls/codegen/java/ReplicatedEntitySourceGenerator.scala
+++ b/codegen/java-gen/src/main/scala/com/lightbend/akkasls/codegen/java/ReplicatedEntitySourceGenerator.scala
@@ -55,7 +55,8 @@ object ReplicatedEntitySourceGenerator {
       val inputType = cmd.inputType.fullName
       val outputType = cmd.outputType.fullName
       s"""|@Override
-          |public Effect<$outputType> ${lowerFirst(methodName)}($parameterizedDataType currentData, $inputType command) {
+          |public Effect<$outputType> ${lowerFirst(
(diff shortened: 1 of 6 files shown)


johanandren (approved):

patriknw (approved):
LGTM

patriknw:
on the topic of conventional commits; `fix` should be used when fixing real bugs, this is more of a `chore`

octonato:
> on the topic of conventional commits; `fix` should be used when fixing real bugs, this is more of a `chore`

correct, I will fix it. In any case, it's broken so I will do a force push soon

octonato:
This is fixed now. I was too enthusiastic and changed a argument on the handler generator and of course, that was wrong. 

Anyway, fixed and commit is now labeled with `chore`

patriknw:
something failing here @octonato

octonato:
> something failing here @octonato

It was the renaming of a scala sample. Not clear why CI rebasing didn't fix it. 

Anyway, fixed now.

RayRoestenburg (approved):

PR #522, opened 2021-09-28
chore: favor usage of FQN for proto component

octonato:
When I was preparing the value entity video, I found myself having to explain the syntax used to point to the component and came to the conclusion that this is an unnecessary complexity / distraction that we are imposing to the users. 

I still think that the annotation as we have today can be simplified to avoid this references, so maybe the change in this PR will become obsolete. Still, I would like to submit it to collect ideas.


maven-java/akkaserverless-maven-archetype-event-sourced-entity/src/main/resources/archetype-resources/src/main/proto/__packageInPathFormat__/counter_api.proto
```diff
@@ -34,7 +34,7 @@ message CurrentCounter {
 service CounterService {
     option (akkaserverless.service) = {
         type : SERVICE_TYPE_ENTITY
-        component : ".domain.Counter"
+        component : ${package}".domain.Counter"
```
octonato, line 37: this syntax is wrong. It needs to be inside the quotes. 
I will fix it.

patriknw (approved):
LGTM

PR #527, opened 2021-09-29
chore: configure release drafter

patriknw:


.github/release-drafter.yml
```diff
@@ -1,24 +1,31 @@
...
+name-template: 'Akka Serverless Java and Scala SDKs $NEXT_PATCH_VERSION'
 tag-template: 'v$NEXT_PATCH_VERSION'
 categories:
   - title: '🚀 Features'
     labels:
       - 'Feature'
+  - title: '🐛 Bug Fixes'
+    labels:
+      - 'bug'
+      - 'fix'
```
ennru, line 11: I've been looking at this again today, as well.
Having "Feature" and "Bug fixes" as separate categories might not be too helpful. The PRs will be duplicated in the listing for e.g. the "Java SDK" category. Within those categories the prefix will give some useful grouping. 
So labelling by file location might be enough for decent release notes?

.github/workflows/release-drafter.yml
```diff
@@ -1,20 +0,0 @@
-name: Release Drafter
```
ennru, line 1: The action must be kept.

.github/release-drafter.yml
patriknw: That is true and I agree. Would have been nice to have 

* Java SDK Features
* Java SDK Bug Fixes
* Scala SDK Features
* Scala SDK Bug Fixes
* Other
  - all chore and other misc stuff goes here and are not duplicated into the above categeries 

Then it might be better to only use convention commit title for the labeling and not the file locations.

patriknw: Then we also have a good reason to use of the conventional commit comments. They will drive what the release notes look like. Touched files is too blunt in my opinion. A change is often most important for some area, such as a feature in javasdk, but also touches some other things for small cleanup or consistency. The person writing that commit knows better best.

patriknw:
I simplified this. Only Java and Scala categories that are assigned from the commit title. Also kept Documentation, but that might be redundant.

.github/release-drafter.yml
```diff
@@ -28,15 +24,12 @@ template: |
   Special thanks to the following contributors who helped with this release: $CONTRIBUTORS
 
 autolabeler:
-  - label: 'Maven'
-    files:
-      - 'maven-java/*'
+  - label: 'java-sdk'
+    title:
+      - '/.*\(.*java.*\):/i'
```
ennru, line 29: I believe using the file locations for labelling works well (that what Alpakka does). The title regex can complement that.
```suggestion
  - label: 'java-sdk'
    title:
      - '/.*\(.*java.*\):/i'
   files:
      - 'sdk/java-sdk/*'
```

patriknw: > I believe using the file locations for labelling works well (that what Alpakka does).

It works, but it will create duplicates that might not be desired. Let's say I add a feature for Scala SDK, but cleanup a typo in the java-sdk, then that feature will show up in both categories. Let's try to use conventional commits and give the author of the commit the power to decide.

ennru: That's just a matter of removing the label in the UI.

ennru (approved):
Let's start with this.

Issue #523, opened 2021-09-29
Run some pr validation tasks in parallel


patriknw:
With the many samples it's starting to take quite some time. We should be able to run `samples-tests` tasks in parallel to some degree.

johanandren:
Also, docs validation taking 10 minutes, and blocking all the other testing must be possible to improve

PR #537, opened 2021-09-30
Speed up ci PR validation

johanandren:
References #523

Starting point ~35 minutes

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -603,6 +603,8 @@ workflows:
   CI:
     jobs:
       - checks
+      - publish-local
+
       - validate-docs:
           requires:
             - checks
@@ -618,9 +620,6 @@ workflows:
           requires: 
             - checks
 
-      - publish-local:
-          requires:
-            - checks
 
       - maven-java-tests:
           requires:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -86,6 +86,22 @@ commands:
             echo "Running maven-java with SDK version: '$SDK_VERSION'"
             mvn versions:set -DnewVersion=$SDK_VERSION
             mvn verify install
+      - run: mkdir -p workspace
+      - run: cp -r ~/.m2 workspace/
+      - run: cp -r ~/.ivy2 workspace/
+      # copy published files to workspace for other jobs to pick up rather than build their own
+      - persist_to_workspace:
+          root: workspace
+          paths:
+            - .m2
+            - .ivy2
+  copy-from-workspace:
+    description: Copy locally published artifacts from workspace
+    steps:
+      - attach_workspace:
+          at: workspace
+      - run: rsync -r workspace/.m2 .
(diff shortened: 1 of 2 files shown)


johanandren:
~26 minutes by not depending on the doc generation/validation

johanandren:
~14 minutes (== check + slowest sample test run in sample-replicated-entity-examples)

johanandren:
Down to ~12 minutes which essentially is 2 minutes code check and 10 minutes doc build.

Some curious things though: the set SDK version command takes ~1 minute in all jobs even though it only starts sbt to print a version. One thing we could do is to write it to a file in the workspace from the publish task, and read that rather than start sbt in all the sample builds. Probably not worth it as long as the doc build takes so long time.

johanandren:
BTW this is a good view looking at how I parallelized the build: https://app.circleci.com/pipelines/github/lightbend/akkaserverless-java-sdk/2000/workflows/95044ee6-2025-4a68-b010-43ff2db1ab91

octonato (approved):
Nice!!

pvlugter:
Nice improvement by parallelising 👍🏼 

> slowest sample test run in sample-replicated-entity-examples

It has a separate integration test (with proxy docker container) for each replicated data type. But probably not worthwhile breaking it up if the docs are longer.

> 10 minutes doc build

The docs build does the external link validation now. That could be a separate job to parallelise this.

> Some curious things though: the set SDK version command takes ~1 minute in all jobs even though it only starts sbt to print a version. One thing we could do is to write it to a file in the workspace from the publish task, and read that rather than start sbt in all the sample builds.

Could also run the appropriate `git describe` command directly — but would need to be aligned with whatever sbt-dynver is doing. There's a scripted version in the JavaScript SDK:

https://github.com/lightbend/akkaserverless-javascript-sdk/blob/458b48f745c5f0587f7f30ab057eb1644b8b1376/sdk/bin/version.sh#L22-L39

.circleci/config.yml
```diff
@@ -78,14 +78,30 @@ commands:
     steps:
       - run:
           name: publish m2 and local
-          command: sbt publishM2 publishLocal
+          command: sbt publishM2 publishLocal -Ddisable.apidocs=true
```
patriknw, line 81: That's a good one. Doc generation is incredible slow. That is probably what takes most time of the doc Make.

johanandren:
Extracting the link validation didn't make a difference but maybe worth it anyway to get a smaller/quicker job to re-run when some link fails temporarily.

patriknw (approved):
LGTM, that should make a big difference

pvlugter:
> Extracting the link validation didn't make a difference but maybe worth it anyway to get a smaller/quicker job to re-run when some link fails temporarily.

Ah, that doesn't as long as I thought it might. Makes sense to have separate as it could be flaky.

pvlugter (approved):

PR #542, opened 2021-10-01
fix(scalasdk): support JDK 8 for generated Java classes

pvlugter:
Scala SDK has protobuf generated Java classes, so needs the release 8 option for javac as well.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -89,6 +89,7 @@ lazy val sdkScala = project
       "scalaVersion" -> scalaVersion.value),
     buildInfoPackage := "com.akkaserverless.scalasdk",
     Compile / scalacOptions ++= Seq("-release", "8"),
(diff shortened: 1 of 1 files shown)


build.sbt
```diff
@@ -89,6 +89,7 @@ lazy val sdkScala = project
       "scalaVersion" -> scalaVersion.value),
     buildInfoPackage := "com.akkaserverless.scalasdk",
     Compile / scalacOptions ++= Seq("-release", "8"),
+    Compile / javacOptions ++= Seq("--release", "8"),
```
patriknw, line 92: we should test at least a few of the samples in CI with JDK 8

patriknw: or just drop support for 8: https://github.com/lightbend/akkaserverless/discussions/5194

patriknw (approved):
LGTM

PR #506, opened 2021-09-27
chore: better quickstart deploy instruction 

johanandren:
References #211

PR #543, opened 2021-10-01
chore: update all the deploy instructions for java

pvlugter:
Follow-up to apply the changes from #506 to the archetypes themselves and all the java samples.


raboof (approved):
I wonder if we should have a script or checklist for these kinds of tasks

johanandren (approved):

Issue #549, opened 2021-10-01
'unmanaged code'-comment is easy to miss


raboof:
For code that we generate once and then has to be maintained by the user, we produce a comment explaining this at the top of this file. However, this is easy to miss, especially in IntelliJ which collapses it by default:

![comment](https://user-images.githubusercontent.com/131856/135600816-730c7154-7778-46ce-96de-428233ca96ec.png)

This is confusing for the first user experience, as most users will start by renaming their service. I think we should move it to the comment above the class rather than at the top of the file.

patriknw:
Sounds good to move it to a more visible place

PR #550, opened 2021-10-01
feat(codegen): more visible 'unmanaged' comment

raboof:
Refs #549

I'd like a review on this before doing all the other
classes as well ;)


codegen/java-gen/src/main/scala/com/lightbend/akkasls/codegen/java/ActionServiceSourceGenerator.scala
```diff
@@ -137,13 +137,13 @@ object ActionServiceSourceGenerator {
...
-    s"""|$unmanagedComment
-        |
-        |package $packageName;
+    s"""|package $packageName;
         |
         |$imports
         |
-        |/** An action. */
+        |/**
+        | * An action.$unmanagedComment
```
patriknw, line 145: I think that comment is more a developer description than something that belongs in api doc section. Could we have it as a separate comment (as it was before, or with `//` style) right after the imports but not inside the api doc?

raboof: works for me, updated

johanandren (approved):

octonato (approved):

patriknw (approved):
LGTM when all of them have been updated

pvlugter:
👍🏼

PR #555, opened 2021-10-04
bump: update commons-compress (security vulnerabilities)

pvlugter:
We don't expect the vulnerabilities to be exposed, but this should resolves the FOSSA checks in framework.

Refs https://github.com/lightbend/akkaserverless-framework/issues/869

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -52,7 +52,6 @@ object Dependencies {
   val jacksonParameterNames = "com.fasterxml.jackson.module" % "jackson-module-parameter-names" % JacksonVersion
   val jacksonScala = "com.fasterxml.jackson.module" %% "jackson-module-scala" % JacksonVersion
 
-  val testcontainers = "org.testcontainers" % "testcontainers" % TestContainersVersion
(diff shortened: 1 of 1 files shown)


johanandren (approved):

project/Dependencies.scala
```diff
@@ -52,7 +52,6 @@ object Dependencies {
   val jacksonParameterNames = "com.fasterxml.jackson.module" % "jackson-module-parameter-names" % JacksonVersion
   val jacksonScala = "com.fasterxml.jackson.module" %% "jackson-module-scala" % JacksonVersion
 
-  val testcontainers = "org.testcontainers" % "testcontainers" % TestContainersVersion
```
raboof, line 55: I liked `testcontainers` over `testContainers` as it's written without a - in the artifact name, but :+1:

raboof (approved):

project/Dependencies.scala
pvlugter: Yes, the official name is `Testcontainers` as one word. I removed the dependency that wasn't being used. Didn't think of switching to this one.

Issue #504, opened 2021-09-27
Error details in responses


johanandren:
Currently for entities (and soon for actions #493) we include the exception message if a user method throws, that could be leaking internal details and is also not something you'd expect the client/user of the API to be able to handle anyway.

Should we rather only return a nondescriptive "Unexpected error" and expect that users will debug through logging rather than their clients telling them what was wrong?

Also, if we do that, should we attach a unique id (say uuid) to the error message and include in the log for correlating?

patriknw:
Interesting, I had a similar suspicion when we worked on the Java SDK and I'm pretty sure I tried it and was happy to see that it didn't include the internal details in the public error message. Maybe I was testing it wrong (or there are several cases, or something changed)?

I absolutely agree that we should not expose internals in returned response.

The idea of a correlation id is also nice.

johanandren:
It's a bit involved so I think I may have parsed it wrong, but it looks as though an unexpected exception is turned into an EntityException with toString of the exception as part of the `message` (https://github.com/lightbend/akkaserverless-java-sdk/blob/main/sdk/java-sdk/src/main/scala/com/akkaserverless/javasdk/impl/eventsourcedentity/EventSourcedEntitiesImpl.scala#L191) and then later when recovering that failure in the stream to produce a failure response it is included in the failure (https://github.com/lightbend/akkaserverless-java-sdk/blob/main/sdk/java-sdk/src/main/scala/com/akkaserverless/javasdk/impl/eventsourcedentity/EventSourcedEntitiesImpl.scala#L132 and https://github.com/lightbend/akkaserverless-java-sdk/blob/main/sdk/java-sdk/src/main/scala/com/akkaserverless/javasdk/impl/EntityExceptions.scala#L83)

Maybe they are silenced by the proxy rather than passed through though?

pvlugter:
I also thought this was already preventing unexpected exception messages to go to the client, but it does look like it includes them still. Good to not expose them. And automatically adding correlation ids for unexpected failures is a good idea.

PR #557, opened 2021-10-04
feat: better error handling

johanandren:
References #504


sdk/java-sdk/src/main/scala/com/akkaserverless/javasdk/impl/ErrorHandling.scala
```diff
@@ -0,0 +1,38 @@
...
+
+  def withCorrelationId[T](block: String => T): T = {
+    val correlationId = UUID.randomUUID().toString
+    MDC.put("correlationID", correlationId)
+    try {
+      block(correlationId)
+    } finally {
+      MDC.remove("correlationID")
+    }
+  }
```
johanandren, line 36: Turns out you don't really see mdc values in the console as I was hoping. I wonder if it is worth doing like this and see if we can get that added, or just put the id in the log entry message string which will work right away?

project/Dependencies.scala
```diff
@@ -80,6 +80,7 @@ object Dependencies {
     akkaslsProxyProtocol % "protobuf-src",
     akkaslsSdkProtocol % "compile;protobuf-src",
     akkaDependency("akka-testkit") % Test,
+    akkaDependency("akka-actor-testkit-typed") % Test,
```
johanandren, line 83: Needed since we use the SLF4J logger directly and no way to inspect those log entries with classic testkit.

johanandren:
For some reason the samples are trying to use the wrong snapshot version now, fixing in #558

sdk/java-sdk/src/main/scala/com/akkaserverless/javasdk/impl/ErrorHandling.scala
patriknw: but MDC values will show up in production logs (otherwise we have a problem) and isn't it for that this is most useful?

patriknw: on the other hand, if we don't log it as mdc on proxy side we could just make it the same here, i.e. include in log message instead

(that also might make it easier for having more similar logging from JS)

sdk/java-sdk/src/main/scala/com/akkaserverless/javasdk/impl/action/ActionsImpl.scala
```diff
@@ -123,10 +146,6 @@ final class ActionsImpl(_system: ActorSystem, services: Map[String, ActionServic
         throw new RuntimeException(s"Unknown metadata implementation: ${other.getClass}, cannot send")
     }
 
-  private def toProtocol(ex: Throwable): ActionResponse = {
-    ActionResponse(ActionResponse.Response.Failure(Failure(0, ex.getMessage)))
```
patriknw, line 127: ok, so this `ex.getMessage` was the leak (via proxy) out to the client?

patriknw (commented):
looking good

sdk/java-sdk/src/main/scala/com/akkaserverless/javasdk/impl/action/ActionsImpl.scala
johanandren: Yeah, but there are more places for the other component types, I just started out with actions as a first step

sdk/java-sdk/src/main/scala/com/akkaserverless/javasdk/impl/ErrorHandling.scala
johanandren: I wonder if I could use MDC but use the layout to include in the log message as well for now (until log viewer/production logs can show it)

johanandren:
A decision that is perhaps questionable: discerning the "protocol errors" in case the proxy misbehaves from arbitrary exceptions seemed like a bit of extra work, so I just went with logging and returning an `Unexpected error` for those as well. 

Not sure I understand the use case where A. that would happen and B. forwarding that to the proxy as a response is useful, as long as error details are logged in the user function, but maybe I'm missing something?

johanandren:
Not sure it's worth doing the last one with using the logger for the concrete class, even if it would be nice. Action is kinda hacky and for the entities it will require some restructuring or capturing things in exceptions.

sdk/java-sdk/src/main/scala/com/akkaserverless/javasdk/impl/action/ActionsImpl.scala
```diff
@@ -57,10 +74,28 @@ final class ActionService(
   override final val componentType = Actions.name
 }
 
-final class ActionsImpl(_system: ActorSystem, services: Map[String, ActionService], rootContext: Context)
+private[javasdk] object ActionsImpl {
+  private[action] val log = LoggerFactory.getLogger(classOf[Action])
```
octonato, line 78: I think we can remove this logger here and initialize it in the `def log` above. 
It's kind of confusing. 

btw, there is a direct reference to it on line 308, we should use `service.log`

sdk/java-sdk/src/main/scala/com/akkaserverless/javasdk/impl/action/ActionsImpl.scala
```diff
@@ -258,17 +296,18 @@ final class ActionsImpl(_system: ActorSystem, services: Map[String, ActionServic
...
+                    handleUnexpectedException(service, call, ex)
                   }
               } catch {
                 case NonFatal(ex) =>
-                  // command handler threw
-                  log.error(s"Failure during handling of command ${call.serviceName}.${call.name}", ex)
-                  Source.single(toProtocol(ex))
+                  // command handler threw an "unexpected" error
+                  ErrorHandling.withCorrelationId { correlationId =>
+                    log.error(s"Failure during handling of command ${call.serviceName}.${call.name}", ex)
```
octonato, line 308:
```suggestion
                    service.log.error(s"Failure during handling of command ${call.serviceName}.${call.name}", ex)
```

sdk/java-sdk/src/main/scala/com/akkaserverless/javasdk/impl/eventsourcedentity/EventSourcedEntitiesImpl.scala
```diff
@@ -121,15 +122,18 @@ final class EventSourcedEntitiesImpl(
...
         case (Seq(EventSourcedStreamIn(other, _)), _) =>
           throw ProtocolException(
             s"Expected init message for Event Sourced Entity, but received [${other.getClass.getName}]")
       }
       .recover { case error =>
-        log.error(error, failureMessage(error))
-        EventSourcedStreamOut(OutFailure(failure(error)))
+        // only "unexpected" exceptions should end up here
+        ErrorHandling.withCorrelationId { correlationId =>
+          log.error(error, failureMessageForLog(error))
```
octonato, line 134: I'm wondering if we should do the same trick as for actions for all kind of entities. We will need to init the handler outside the flow with minor penalty when we don't use it (case 2 and 3 in the pattern matching), but I think that case 2 and 3 are rather rare. So might be an option?

octonato (commented):
Looking good. 

I have a comment/suggest about init handlers for entities and usage of same logger in all places.

sdk/java-sdk/src/main/scala/com/akkaserverless/javasdk/impl/action/ActionsImpl.scala
johanandren: We also use it for things not related to a specific service below, so it can't be in a specific service.

sdk/java-sdk/src/main/scala/com/akkaserverless/javasdk/impl/eventsourcedentity/EventSourcedEntitiesImpl.scala
johanandren: We don't know which entity to init unless we have the init message with the service name in it, so no service to tie log to here.

sdk/java-sdk/src/main/scala/com/akkaserverless/javasdk/impl/action/ActionsImpl.scala
johanandren: Should perhaps be `classOf[ActionsImpl]` though

octonato: ok and it makes sense to use ActionsImpl.

sdk/java-sdk/src/main/scala/com/akkaserverless/javasdk/impl/eventsourcedentity/EventSourcedEntitiesImpl.scala
octonato: true, I missed that.

octonato (commented):
realised that we have covered Actions, ES, VE and Replicated entities, but no Views.

sdk/java-sdk/src/main/scala/com/akkaserverless/javasdk/impl/ErrorHandling.scala
johanandren: Verified that the MDC value is visible through the akkasls logs command, but with the correlation in the log entry through the formatter you also see it already through the console log ui (otoh you only see the first line of the log message so it's anyway not super useful there)

sdk/java-sdk/src/main/java/com/akkaserverless/javasdk/logging/LogbackJsonLayout.java
```diff
@@ -35,5 +38,23 @@ public LogbackJsonLayout() {
...
+            this.includeMessage,
+            map.get(MESSAGE_ATTR_NAME) + " [" + correlationID + "]",
+            map);
+      }
+      if (this.includeFormattedMessage) {
+        add(
+            FORMATTED_MESSAGE_ATTR_NAME,
+            this.includeFormattedMessage,
+            map.get(FORMATTED_MESSAGE_ATTR_NAME) + " [" + correlationID + "]",
+            map);
```
patriknw, line 56: wonder if we are going over board and making this more magic than necessary?

We always log these with `ErrorHandling.withCorrelationId { correlationId =>` right? Can't we just include the correlationId in the log message explicitly. Maybe even skip the idea of including it in MDC altogether?

johanandren: Yeah, definitely questionable, I think MDC is the nice thing, so I'd rather cut it out of the log entries then, the only place that is not visible is the console ui (as it is right now).

johanandren: Reverted that now, so only in MDC

johanandren:
This is ready for final review

PR #558, opened 2021-10-04
fix: samples using wrong snapshot version

johanandren:
I didn't figure out the reason, but something we merged touched the files somewhere between publishing locally and other jobs trying to determine SDK version and picking that up from the circle ci workspace, so that the SDK version published and the one the samples tried to use were not the same.

This change writes the version number, that was published locally, to a file in the workspace and jobs that wants to use the built artifact read that file instead of trying to use sbt to figure out what the current snapshot version is.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -222,10 +222,10 @@ jobs:
       - checkout-and-merge-to-main
       - restore_deps_cache
       - setup_sbt
-      - set-sdk-version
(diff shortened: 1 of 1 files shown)


johanandren:
Bonus: cuts a minute of work starting sbt up in each sample test to figure out version down to "0s"

patriknw (approved):
LGTM

Issue #551, opened 2021-10-01
autolabeler not working


patriknw:
It's supposed to add labels to the PRs. https://github.com/lightbend/akkaserverless-java-sdk/blob/main/.github/release-drafter.yml#L26

We have some closed PRs with labels but seems like those labels were manually added?

ennru:
Yes, I added labels manually to a few PRs prior to that release as the PRs existed before we merged the release drafter updates. Created #566 now to see if file-based labeling kicks in.

ennru:
"Auto labeler won't work with forks as the token cannot work across repo permissions, and that would be a problem with GitHub permissions." 
https://github.com/release-drafter/release-drafter/issues/785#issuecomment-778700859

ennru:
If we want auto-labelling by file only, we should consider https://github.com/mithro/autolabeler which we use in Alpakka.

PR #566, opened 2021-10-05
chore: try file-based labeling

ennru:
Refs #551﻿

### DIFF
--- a/.github/release-drafter.yml
+++ b/.github/release-drafter.yml
@@ -27,9 +27,15 @@ autolabeler:
   - label: 'java-sdk'
     title:
       - '/.*\(.*java.*\):/i'
+    files:
+      - '/sdk/java-sdk/**'
   - label: 'scala-sdk'
     title:
       - '/.*\(.*scala.*\):/i'
+    files:
+      - '/sdk/scala-sdk/**'
   - label: 'Documentation'
     title:
       - '/doc.*:/i'
+    files:
+      - '/docs/**'


octonato (approved):
Seems that this should work better. 
We may want to also tag changed in testkits, maven/sbt. But we can add later once we confirm this will work as expected.

.github/release-drafter.yml
```diff
@@ -27,9 +27,15 @@ autolabeler:
   - label: 'java-sdk'
     title:
       - '/.*\(.*java.*\):/i'
+    files:
+      - '/sdk/java-sdk/**'
   - label: 'scala-sdk'
     title:
       - '/.*\(.*scala.*\):/i'
```
patriknw, line 34: can we remove the title sections now?

patriknw:
> We may want to also tag changed in testkits, maven/sbt

I don't think we should have too many categories because each PR that touches many files will show up duplicated in all matching sections. I think users will see testkits and build tools as part of the sdk.

patriknw:
You can try by touching some matching file, but I still think there is something missing to make this work. We had file based matching previously but I couldn't see any labels that weren't manually added.

.github/release-drafter.yml
ennru: Let's see if it kicks in with the file globs. Something else might be broken.

octonato:
> I don't think we should have too many categories because each PR that touches many files will show up duplicated in all matching sections. I think users will see testkits and build tools as part of the sdk.

@patriknw, I was not suggesting more categories, but also labeling maven changes with `java-sdk` and sbt change with `scala-sdk`.

patriknw:
That makes total sense, agreed.

PR #567, opened 2021-10-05
chore: make sure to always return the full set of generated files

jrudolph:
That's what `sourceGenerators` in sbt expects. Otherwise, those files will
be missing from the set of generated files and will not be included in
compilation. This will lead to error during incremental recompilation

(I noticed that compilation of java-gen-compilation-tests only worked the first time).


pvlugter (approved):
👍🏼 LGTM

pvlugter:
There's this FIXME that could be removed now:

https://github.com/lightbend/akkaserverless-java-sdk/blob/679efb228dc305eb9900a15542c9ab2ea5501099/project/ReflectiveCodeGen.scala#L19

octonato (approved):

raboof (approved):
Aah, we have to tell sbt it is 'still a generated file' even if we generated it before, not in 'this' generation run. Great to know!

jrudolph:
> Aah, we have to tell sbt it is 'still a generated file' even if we generated it before, not in 'this' generation run. Great to know!

Yep, it seems managed source files are not automatically enumerated to be included as sources.

PR #575, opened 2021-10-07
chore: remove obsolete build job

octonato:


octonato:
> I think that is fine for the samples, they all compile in their own job so to speak. Doing it in the checks would delay all other tasks since everything depends on that so I think we should avoid making that take longer time. 

I was talking about a full compilation cycle, ie: `sbt test:compile`. We could do it if we can guarantee that here is no double compilation. But I'm fine to let it happen when we run the tests.

The samples are compiled in their own job and that's ok. Note that the previous job here was doing: `sbt compile`, not `sbt samples/compile`. We don't even have a `samples` parent module anymore.

Issue #572, opened 2021-10-06
Action testkit for Scala 


johanandren:
Scala side of #336

PR #583, opened 2021-10-07
feat: Action testkit for Scala SDK

johanandren:
Including some fixes on the Java Action testkit side

References #572


johanandren:
I think I'll look into adding support for streaming calls separately.

PR #599, opened 2021-10-11
chore: generate sources for tck tests for Java SDK

jrudolph:
In this PR, actions, value entities, and event-sourced entities TCK tests are converted to use auto-generated code.

For now, I imported the TCK protos to be more easily editable. When all TCK tests are converted they should be backported to the proxy and then again be imported from there.


jrudolph:
Yes, I fixed the remaining issues.

Issue #590, opened 2021-10-08
Support for streaming calls in action testkit


johanandren:
Currently the Java and Scala Action testkits only allow testing unary actions

johanandren:
I see that @franciscolopezsancho prepared something almost working here, which I disabled when it didn't go all the way.

PR #602, opened 2021-10-11
fix: (re)-introduce action testkit support for streaming commands 

johanandren:
References #590


johanandren:
Only tests for unary yet, perhaps if we add some action samples for streaming that would be a good place for such unit tests.

PR #611, opened 2021-10-12
feat: implement Scala TCK checks

jrudolph:
Some early code generation attempts, currently fails because of missing generated imports when multiple entities share their domain.


jrudolph:
Seems I need to fix https://github.com/jrudolph/akkaserverless-java-sdk/blob/b6868c575e838ea460ab08b3ef01aac6c469fcfc/codegen/core/src/main/scala/com/lightbend/akkasls/codegen/ModelBuilder.scala#L452-L453 to fix the imports.

jrudolph:
The setup now seems to work fine. So far, I have still only implemented the action tests all other ones are not yet implemented and are ignored in `RunTck`. Since we want to use only one set of protos between Scala and Java (and later port that back to the actual TCK and use it for all languages), I had to revert a bit from #628 and replace it with manually generated code for now in the Java side of things.

I'll create a new umbrella ticket (or add to the existing Java TCK one) for actually implementing the Scala TCK tests.

patriknw (approved):
LGTM

Issue #576, opened 2021-10-07
Make Scala ReplicatedData implementations more idiomatic


octonato:
> Also wondering if we want to have the method names and extra methods that are idiomatic for Scala data structures. So `contains` rather than `containsKey` on maps, additional `keys` method, maybe `apply` method for Set contains, and so on.

_Originally posted by @pvlugter in https://github.com/lightbend/akkaserverless-java-sdk/pull/541#pullrequestreview-773431327_

PR #613, opened 2021-10-12
feat(scalasdk): replicated entities examples for Scala

octonato:
This is the implementation of the scala-replicatedentities-examples. 

I realised that it will be better to implement it after #576 so the examples show more idiomatic scala code. Therefore, I'm keeping it on draft for now.


pvlugter:
> I noticed that we don't have an example for ReplicatedVote, also not in the Java API. I can add it on follow-up PR and in both languages.

I left out documenting Vote and opened: https://github.com/lightbend/akkaserverless-framework/issues/859

Issue #577, opened 2021-10-07
Let Scala SDK wrap Java Impl to minimize conversions on collections


octonato:
> On the double conversions (Scala -> Java -> Scala) going from the wrapper to the Java API and then to the internal API in Scala: maybe the Scala wrappers should take the internal impl versions of the data structures, instead of the Java interfaces, and have access to some internal methods that accept Scala collections directly. So that there's no extra conversion from Scala. 

_Originally posted by @pvlugter in https://github.com/lightbend/akkaserverless-java-sdk/pull/541#pullrequestreview-773431327_

PR #627, opened 2021-10-14
chore: wrap java replicated types impl and avoid extra conversions

### LINKED ISSUE #576: Make Scala ReplicatedData implementations more idiomatic

octonato:
Fixes #576 and #577


octonato:
> No need to make the methods private[akkaserverlesss] when the entire class is already.

True, I will remove them

PR #630, opened 2021-10-14
chore: auto-generate imports

raboof:
Experiment: this way we don't have to list the imports separately, which seems much easier to maintain.


octonato:
This is super cool. I never looked how interpolation works and I'm amazed.

PR #679, opened 2021-11-03
fix(proxy): update proxy version.

RayRoestenburg:


jroper:
Good, this failed. It's supposed to. The version can't just be bumped without implementing the 0.8 protocol support.

Issue #683, opened 2021-11-09
Implement more Scala TCK tests


jrudolph:
Most tests are still disabled in `RunTck`.

 * [x] action #611 
 * [ ] value entity @jrudolph 
 * [ ] event-sourced entity
 * [ ] view
 * [ ] eventing
 * [ ] replicated entity
 * [ ] local persistence eventing

Some of those might have to be implemented together with their codegen Java parts because they should share the same proto files, #see 600

Issue #673, opened 2021-11-01
Deprecate deferred calls from entities


johanandren:
In the bigger discussion we said we don't think entities should forward anything anywhere, (but maybe still allow side effects, even if they don't seem very useful?).

They can already not do async work and include it in their effect - only actions has the `effects.asyncReply` etc. (even if we can not in general stop arbitrary fire-forget async calls triggered from a command handler).

I left this out of the components-PR #628 since some of the TCK tests depend on being able to forward, if we deprecate we should start by removing those tests from the TCK suite.

> I'm surprised by this. I understood that we will treat all components as leafs (except Actions). Did I miss the discussion where we changed our mind about it? 
> If that's the reason, could we mark them all as deprecated? I mean the methods in the Effect API in leaf components receiving `DeferredCall` and the `components()` method here.

_Originally posted by @octonato in https://github.com/lightbend/akkaserverless-java-sdk/pull/628#discussion_r739252192_

PR #684, opened 2021-11-09
feat: Scala TCK value entity implementation

jrudolph:
This one currently fails the configuration tests. Is there anything needed to make that work or could that be a bug in the Scala SDK?

Refs #683

### DIFF
--- a/tck/scala-tck/src/main/scala/akkaserverless/tck/model/TckService.scala
+++ b/tck/scala-tck/src/main/scala/akkaserverless/tck/model/TckService.scala
@@ -1,0 +1,33 @@
+/*
+ * Copyright 2021 Lightbend Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package akkaserverless.tck.model
+
+import scala.concurrent.duration._
+
+import com.akkaserverless.scalasdk.PassivationStrategy
+import com.akkaserverless.scalasdk.valueentity.ValueEntityOptions
+import com.akkaserverless.tck.model.valueentity.{ ValueEntityConfiguredEntity, ValueEntityConfiguredEntityProvider }
+
+/** Create the AkkaServerless instance with some required configuration changes */
+object TckService {
+  def createService() =
+    Main
+      .createAkkaServerless()
+      .register(
+        ValueEntityConfiguredEntityProvider(new ValueEntityConfiguredEntity(_))
+          .withOptions(ValueEntityOptions.defaults.withPassivationStrategy(PassivationStrategy.timeout(100.millis))))
+}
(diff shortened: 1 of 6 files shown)


jrudolph:
> This one currently fails the configuration tests. Is there anything needed to make that work or could that be a bug in the Scala SDK?

Found it, the Java TCK configures it in the startup code...

jrudolph:
> Found it, the Java TCK configures it in the startup code...

Is there a way to get that functionality without having to copy in the factory and adapt its code?

tck/scala-tck/src/main/scala/akkaserverless/tck/model/TckService.scala
```diff
@@ -0,0 +1,33 @@
...
+import com.akkaserverless.scalasdk.PassivationStrategy
+import com.akkaserverless.scalasdk.valueentity.ValueEntityOptions
+import com.akkaserverless.tck.model.valueentity.{ ValueEntityConfiguredEntity, ValueEntityConfiguredEntityProvider }
+
+/** Create the AkkaServerless instance with some required configuration changes */
+object TckService {
+  def createService() =
+    Main
+      .createAkkaServerless()
+      .register(
```
jrudolph, line 30: We just re-register the component that needs configuration. This allows us to regenerate Main as is when entities are added or removed.

tck/scala-tck/src/main/scala/com/akkaserverless/scalasdk/Access.scala
```diff
@@ -0,0 +1,39 @@
...
+import com.akkaserverless.javasdk.impl
+import com.akkaserverless.scalasdk.impl.InternalContext
+import com.akkaserverless.scalasdk.impl.valueentity.ScalaValueEntityContextAdapter
+import com.akkaserverless.scalasdk.valueentity.ValueEntityContext
+
+object Access {
+  // FIXME: Access to components has been restricted in Scala SDK, see https://github.com/lightbend/akkaserverless-java-sdk/issues/673
+  // On the Scala side there's no way to make calls right now. The TCK, however, currently requires those calls, so we need this hack.
+  // Ultimately, these tests are probably removed from the SDK at some point, in which case we can remove this method again.
+  def components(context: ValueEntityContext): Components =
```
jrudolph, line 29: Quick hack to allow to access other components from more places.

johanandren: Hmm, the intent was that they should now be available in all entities as well as actions but maybe  I missed opening that up in the codegen for Scala and only did it in Java?

The ticket was about discussing maybe limiting that part in the future rather (but maybe not).

jrudolph: #673 sounds as if you had only added `components` support for the Java side in #628  because the TCK required it.

jrudolph: But anyway if you say the discrepancy is an oversight, let's create a ticket to fix that?

johanandren: https://github.com/lightbend/akkaserverless-java-sdk/pull/689

johanandren: Or a PR :)

tck/scala-tck/src/main/scala/akkaserverless/tck/model/TckService.scala
johanandren: Neat!

tck/scala-tck/src/main/scala/com/akkaserverless/scalasdk/Access.scala
jrudolph: Even better :) Will update after that one is merged.

johanandren (commented):
Looks good!

johanandren (approved):

Issue #682, opened 2021-11-09
Scala EventSourced customer-registry deploy problems


johanandren:
Seeing
```
Error: failed to start container "user-function": Error response from daemon: OCI runtime create failed: container_linux.go:380: starting container process caused: exec: "/opt/docker/bin/customer-registry": stat /opt/docker/bin/customer-registry: no such file or directory: unknown
```

While the scala-valuerentity-customer-registry build.sbt and generated docker stuff under `target/` etc looks pretty much identical but successfully deploys. Could perhaps be something with the env rather?

johanandren:
Generated docker file says `ENTRYPOINT ["/opt/docker/bin/customer-registry"]` but staged image contents contain `bin/main` rather. Where does that expected entry point name come from?

PR #687, opened 2021-11-09
chore: Fix scala eventsourced customer registry docker build

johanandren:
Refs #682

Two main classes lead to entry points not matching the auto-generated one so it could not actually be deployed.

Also, align the two scala customer registry projects a little bit

### DIFF
--- a/samples/scala-eventsourced-customer-registry/build.sbt
+++ b/samples/scala-eventsourced-customer-registry/build.sbt
@@ -1,6 +1,6 @@
-name := "eventsourced-customer-registry"
+name := "customer-registry"
 
-organization := "com.akkaseverless.samples"
+organization := "com.akkaserverless.samples"
 organizationHomepage := Some(url("https://akkaserverless.com"))
 licenses := Seq(("CC0", url("https://creativecommons.org/publicdomain/zero/1.0")))
 
@@ -10,6 +10,8 @@ enablePlugins(AkkaserverlessPlugin, JavaAppPackaging, DockerPlugin)
 dockerBaseImage := "docker.io/library/adoptopenjdk:11-jre-hotspot"
 dockerUsername := sys.props.get("docker.username")
 dockerRepository := sys.props.get("docker.registry")
+// two Main files in this project changes entry point
+dockerEntrypoint := Seq("bin/main")
 ThisBuild / dynverSeparator := "-"
 
 Compile / scalacOptions ++= Seq(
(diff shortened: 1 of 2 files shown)


samples/scala-valueentity-customer-registry/build.sbt
```diff
@@ -26,6 +26,11 @@ Test / parallelExecution := false
 Test / testOptions += Tests.Argument("-oDF")
 Test / logBuffered := false
 
+Compile / run := {
+  // needed for the proxy to access the user function on all platforms
+  sys.props += "akkaserverless.user-function-interface" -> "0.0.0.0"
+  (Compile / run).evaluated
+}
```
raboof, line 33: ha, nice trick

samples/scala-eventsourced-customer-registry/build.sbt
```diff
@@ -1,6 +1,6 @@
-name := "eventsourced-customer-registry"
+name := "customer-registry"
```
raboof, line 1: why not eventsourced? OK by me though.

samples/scala-eventsourced-customer-registry/build.sbt
```diff
@@ -1,6 +1,6 @@
...
 
-organization := "com.akkaseverless.samples"
+organization := "com.akkaserverless.samples"
```
raboof, line 3: :+1:

samples/scala-eventsourced-customer-registry/build.sbt
```diff
@@ -10,6 +10,8 @@ enablePlugins(AkkaserverlessPlugin, JavaAppPackaging, DockerPlugin)
 dockerBaseImage := "docker.io/library/adoptopenjdk:11-jre-hotspot"
 dockerUsername := sys.props.get("docker.username")
 dockerRepository := sys.props.get("docker.registry")
+// two Main files in this project changes entry point
+dockerEntrypoint := Seq("bin/main")
```
raboof, line 14: does that second Main really belong in this sample? OK I guess :)

raboof (approved):

samples/scala-eventsourced-customer-registry/build.sbt
johanandren: Alternative would be to put `valueentity` in front of the name in the other project, I'm fine with either.

samples/scala-eventsourced-customer-registry/build.sbt
johanandren: I think it is used for the docs somewhere, we have a "snippets" project now but then we'd need to copy some components to show registration with properties there as well.

pvlugter (approved):

Issue #571, opened 2021-10-06
Convenient way to only run unit tests


patriknw:
We have integration tests in test scope and it would be nice to have a way to run all unit tests without these.

We could name them `UnitSpec` and `IntegrationSpec` so that `testOnly **UnitSpec` would work?

An alternative is to use ScalaTest tags for the integration tests.

_Originally posted by @patriknw in https://github.com/lightbend/akkaserverless-java-sdk/pull/565#r722377573_

patriknw:
@raboof I have a new idea. For the integration tests can have a check if the proxy is running on expected port and mark the test as pending (with a warning log) if it's not. If you agree I can add this.

We have similar in some Akka tests:
```
  "Resolver" must {
    if (!dockerAvailable()) {
      system.log.error("Test not run as docker is not available")
      pending
    } else {
      system.log.info("Docker available. Running DNS tests")
    }
```

patriknw:
ah, 😭

PR #693, opened 2021-11-09
feat: convinient way to run only unit tests

franciscolopezsancho:
By default would only run unit tests. If `onlyUnit` is set to false then would also run integration tests.

References #571


sbt-plugin/src/main/scala/com/akkaserverless/sbt/AkkaserverlessPlugin.scala
```diff
@@ -48,6 +48,8 @@ object AkkaserverlessPlugin extends AutoPlugin {
       "These are the source files that are placed in the source tree, and after initial generation should typically be maintained by the user.\n" +
       "Files that already exist they are not re-generated.")
     val temporaryUnmanagedDirectory = settingKey[File]("Directory to generate 'unmanaged' sources into")
+    val onlyUnit = settingKey[Boolean]("Filters out integration tests. By default: true")
```
johanandren, line 51: Shouldn't it be through a "configuration" scope rather than a setting key, so we'd have `Test` which is the regular ones and `IntegrationTest` (already in sbt) which is the integration ones?

```scala
override def projectConfigurations = Seq(IntegrationTest)

override def projectSettings = 
  ... ++ inConfig(IntegrationTest)(only it tests)
  ++  inConfig(Test)(only unit tests)
```

patriknw:
@octonato here is the unit test vs integration test PR, see also the background discussion in #571

sbt-plugin/src/main/scala/com/akkaserverless/sbt/AkkaserverlessPlugin.scala
patriknw: wouldn't that introduce all the problems that we have seen with sbt `it:test` ?

patriknw, line 51: I think the default should be `false`, and that we also read the setting from a System.property so that it can be used like this:
```
sbt -DonlyUnitTest=true
```

raboof: FWIW I saved my WIP trying to put the integration tests into the `IntegrationTest` scope in https://github.com/lightbend/akkaserverless-java-sdk/compare/main...raboof:wip-integration-tests-into-IntegrationTest-configuration?expand=1 - but ran into various problems and abandoned that approach in favour of just adding them as regular tests. /cc @octonato

franciscolopezsancho: yep, sounds good.

franciscolopezsancho: OK, I talked with @raboof and it seems the main problem was trying to add `it:test` through `Defaults.itSettings`. This wouldn't allow further configuring the project.

johanandren: Yeah, sorry, I had forgotten what I had heard previously about problems with that.

Still a bit annoying to have our own way for integration tests when sbt has some concept for it.

raboof: > Still a bit annoying to have our own way for integration tests when sbt has some concept for it.

yeah, I agree - perhaps we can revisit it some day, but I reached my sbt threshold last time I tried ;)

sbt-plugin/src/main/scala/com/akkaserverless/sbt/AkkaserverlessPlugin.scala
```diff
@@ -104,7 +106,18 @@ object AkkaserverlessPlugin extends AutoPlugin {
     Test / managedSources :=
       (Test / managedSources).value.filter(s => !isIn(s, (Test / temporaryUnmanagedDirectory).value)),
     Test / unmanagedSources :=
-      (Test / generateUnmanaged).value ++ (Test / unmanagedSources).value)
+      (Test / generateUnmanaged).value ++ (Test / unmanagedSources).value,
+    onlyUnitTest := {
+      sys.props.get("onlyUnitTest") match {
+        case Some("true") => true
+        case _            => false
+      }
```
patriknw, line 114: that could be simplified to 
```
onlyUnitTest := Boolean.getBoolean("onlyUnitTest")
```

(yeah, that is plain old Java)

patriknw (commented):
looking good

mention it in the docs where we explain how to write unit tests

sbt-plugin/src/main/scala/com/akkaserverless/sbt/AkkaserverlessPlugin.scala
franciscolopezsancho: :) niiice! 

About the docs, will do. Actually, I was about to add it but realized there's no unique place where there's a reference to this, and also that it doesn't appear on "Implementing Event Sourced Entities" nor in "Implementing Actions". I'll add it accordingly.

franciscolopezsancho:
I don't see that in [config.yml](https://github.com/lightbend/akkaserverless-java-sdk/blob/main/.circleci/config.yml) that `sys.env.get("CIRCLECI").contains("true")` can be met as per [here](https://github.com/lightbend/akkaserverless-java-sdk/blob/9161d96402973b7ec65b93d5449189760d890e41/sbt-plugin/src/sbt-test/sbt-akkaserverless/eventsourcedentity/build.sbt#L9). Hence Integration test should run. Do we want the integration test running in CI?

docs/src/modules/java/pages/actions.adoc
```diff
@@ -187,4 +187,6 @@ Calling an action method through the test kit gives us back an link:{attachments
 +
 * `reply` returns the reply message passed to `effects.reply()` or throws an exception failing the test, if the effect returned was not a reply.
 * `errorDescription` returns the error description when `effects().error()` was returned to signal an error.
-* `forwardedTo` returns details about what message was forwarded and where the call was forwarded (since it is a unit test the forward is not actually executed).
\ No newline at end of file
+* `forwardedTo` returns details about what message was forwarded and where the call was forwarded (since it is a unit test the forward is not actually executed).
++
+By default the integration and unit test are both invoked by `sbt test`. To only run unit tests run `sbt -DonlyUnitTest test` or set up that paremeter to `true` in the sbt console by `set onlyUnitTest := true` and then run `sbt test`
```
franciscolopezsancho, line 192: I'm not sure this applies to Actions. They only have unit tests, right?

docs/src/modules/java/pages/event-sourced-entities.adoc
```diff
@@ -303,6 +303,9 @@ include::example$scala-eventsourced-shopping-cart/src/test/scala/com/example/sho
 <7> retrieving the second event generated from the first call to `addItem`. There is no such event as our implementation only generates one event when addItem it's called.
 <8> retrieving the response from the call to `addItem`.
 <9> retrieving the state of the entity after all the calls to `addItem`.
++
+By default the integration and unit test are both invoked by `sbt test`. To invoke only the unit tests run `sbt -DonlyUnitTest test` or set up that paremeter to `true` in the sbt console by `set onlyUnitTest := true` and then run `sbt test`
+
```
patriknw, line 308: This shouldn't be shown when Java has been selected as language. It can be shown for Scala only in this way:
```
[.group-scala]#By default the integration and unit test are both invoked by `sbt test`. To invoke only the unit tests run `sbt -DonlyUnitTest test` or set up that paremeter to `true` in the sbt console by `set onlyUnitTest := true` and then run `sbt test`#
```

patriknw: if you set it in the sbt console you will not run `sbt test`. You will run `test` in the console.

I think it would be better to say "sbt session" rather than sbt console.

docs/src/modules/java/pages/quickstart-template.adoc
```diff
@@ -203,10 +203,13 @@ Scala::
 +
 [source,bash]
 ----
-sbt compile Test/compile
+sbt -DonlyUnitTest test
```
patriknw, line 206: There is also a wrong `compile` at line 192

docs/src/modules/java/pages/quickstart-template.adoc
```diff
@@ -203,10 +203,13 @@ Scala::
...
 ----
+By default the integration and unit tests are both triggered with `sbt test`. To invoke only the unit tests run `sbt -DonlyUnitTest test` or set up that value to `true` in the sbt console by `set onlyUnitTest := true`. Then run `sbt test`
```
patriknw, line 208: is this only shown when the Scala tab is selected?

docs/src/modules/java/pages/actions.adoc
patriknw: I think we should mention this here anyway. At some point we should have integration tests for actions, if we don't have it already. And, there are probably other integration tests in the project that you don't want to run when developing the unit tests of actions.

docs/src/modules/java/pages/event-sourced-entities.adoc
franciscolopezsancho: done

docs/src/modules/java/pages/quickstart-template.adoc
franciscolopezsancho: I added [.group-scala]#

docs/src/modules/java/pages/event-sourced-entities.adoc
patriknw: at the end I think it would be better with
```
and then run `test`
```
because you are already expected to be inside the sbt session if you have set the setting there

sbt-plugin/src/main/scala/com/akkaserverless/sbt/AkkaserverlessPlugin.scala
```diff
@@ -116,7 +118,13 @@ object AkkaserverlessPlugin extends AutoPlugin {
     Test / managedSources :=
       (Test / managedSources).value.filter(s => !isIn(s, (Test / temporaryUnmanagedDirectory).value)),
     Test / unmanagedSources :=
-      (Test / generateUnmanaged).value ++ (Test / unmanagedSources).value)
+      (Test / generateUnmanaged).value ++ (Test / unmanagedSources).value,
+    onlyUnitTest := JBoolean.getBoolean("onlyUnitTest"),
```
patriknw, line 122: this doesn't work with how it's documented

`-DonlyUnitTest` will add the system property but the value will be `""`

it would have to be `-DonlyUnitTest=true`, but maybe better to change this again so that both `""` and `true` is supported

docs/src/modules/java/pages/actions.adoc
```diff
@@ -187,4 +187,6 @@ Calling an action method through the test kit gives us back an link:{attachments
 +
 * `reply` returns the reply message passed to `effects.reply()` or throws an exception failing the test, if the effect returned was not a reply.
 * `errorDescription` returns the error description when `effects().error()` was returned to signal an error.
-* `forwardedTo` returns details about what message was forwarded and where the call was forwarded (since it is a unit test the forward is not actually executed).
\ No newline at end of file
+* `forwardedTo` returns details about what message was forwarded and where the call was forwarded (since it is a unit test the forward is not actually executed).
++
+[.group-scala]#By default the integration and unit test are both invoked by `sbt test`. To only run unit tests run `sbt -DonlyUnitTest test` or set up that value to `true` in the sbt session by `set onlyUnitTest := true` and then run `sbt test`
```
patriknw, line 192: It must also end with a `#` for the group

docs/src/modules/java/pages/quickstart-template.adoc
```diff
@@ -227,6 +230,15 @@ The Maven failsafe plugin runs the integration tests when the `it` profile is en
...
 
+[.group-scala]
+By default the integration and unit test are both invoked by `sbt test`.
+
+[.group-scala]
+[source,bash]
+----
+sbt test
+----
+
```
patriknw, line 241: why don't we use `[.tabset]` here? `[.tabset]` is used just a few lines up for similar sbt and mvn commands

docs/src/modules/java/pages/actions.adoc
franciscolopezsancho: Is there a way to create the docs exactly as in the public docs? with tabs included?  `make dev-html` shows all the tabs at once.

sbt-plugin/src/main/scala/com/akkaserverless/sbt/AkkaserverlessPlugin.scala
franciscolopezsancho: Ups. 
OK, I'll make `-DonlyUnitTest` behave as `-DonlyUnitTest=true` and update the docs accordingly.

jrudolph (commented):
I know this approach was suggested in the ticket but wouldn't it make sense to stick to a less custom approach?

E.g. we could [tag](https://www.scalatest.org/user_guide/using_the_runner#filtering) integration tests as such and then use `testOnly -- -lit` in sbt to exclude those tests if that's what needed?

raboof:
> I know this approach was suggested in the ticket but wouldn't it make sense to stick to a less custom approach?

I agree it would be nice to have something less custom if we can pull it off ;)

> E.g. we could [tag](https://www.scalatest.org/user_guide/using_the_runner#filtering) integration tests as such and then use `testOnly -- -lit` in sbt to exclude those tests if that's what needed?

Unfortunately I tried that, but because AFAIK tags must work on the individual test level and not on the test class level, it will execute the `beforeAll`/`afterAll` etc even when all actual tests are excluded, which kind of defeats the point in this case... (I'm *fairly* sure I tried this, though it looks like I didn't share that, sorry about that)

docs/src/modules/java/pages/quickstart-template.adoc
franciscolopezsancho: I think because the content of the tabs alternate after the `[.tabset]` but I guess I could group them and use only the `[.tabset]`. I'm just don't feel very confident touching anything tab-related as I can't see the result until published.

docs/src/modules/java/pages/actions.adoc
patriknw: that can be done from the akkaserverless-docs repo with the local.yml, see https://github.com/lightbend/akkaserverless-docs/blob/master/README-antora.md

docs/src/modules/java/pages/event-sourced-entities.adoc
```diff
@@ -303,6 +303,9 @@ include::example$scala-eventsourced-shopping-cart/src/test/scala/com/example/sho
 <7> retrieving the second event generated from the first call to `addItem`. There is no such event as our implementation only generates one event when addItem it's called.
 <8> retrieving the response from the call to `addItem`.
 <9> retrieving the state of the entity after all the calls to `addItem`.
++
+[.group-scala]#By default the integration and unit test are both invoked by `sbt test`. To invoke only the unit tests run `sbt -DonlyUnitTest test` or set up that value to `true` in the sbt session by `set onlyUnitTest := true` and then run `sbt test`
```
patriknw, line 307: also missing `#` at the end

franciscolopezsancho: yep, I realized later. :) thanks

franciscolopezsancho:
Tested on local (tabs included)

patriknw (approved):
LGTM

Issue #656, opened 2021-10-18
Use the auto-import approach in codegen


patriknw:
See https://github.com/lightbend/akkaserverless-java-sdk/pull/630

would also be nice for Java

PR #699, opened 2021-11-11
Convert Java View code generation to new style

jrudolph:
Refs #656

Part of my View problems in the TCK are again related to naming problems, so I started fixing this here.

### DIFF
--- a/codegen/core/src/main/scala/com/lightbend/akkasls/codegen/SourceGeneratorUtils.scala
+++ b/codegen/core/src/main/scala/com/lightbend/akkasls/codegen/SourceGeneratorUtils.scala
@@ -18,16 +18,15 @@ package com.lightbend.akkasls.codegen
 
 import java.nio.file.Path
 import java.nio.file.Paths
-
 import scala.annotation.tailrec
 import scala.collection.immutable
-
 import com.lightbend.akkasls.codegen.ModelBuilder.Command
 import com.lightbend.akkasls.codegen.ModelBuilder.MessageTypeArgument
 import com.lightbend.akkasls.codegen.ModelBuilder.ScalarType
 import com.lightbend.akkasls.codegen.ModelBuilder.ScalarTypeArgument
 import com.lightbend.akkasls.codegen.ModelBuilder.State
 import com.lightbend.akkasls.codegen.ModelBuilder.TypeArgument
+import com.lightbend.akkasls.codegen.SourceGeneratorUtils.CodeElement.FQNElement
 
 object SourceGeneratorUtils {
   val managedComment =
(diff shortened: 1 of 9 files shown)


codegen/java-gen/src/test/scala/com/lightbend/akkasls/codegen/java/ViewServiceSourceGeneratorSuite.scala
```diff
@@ -208,15 +207,11 @@ class ViewServiceSourceGeneratorSuite extends munit.FunSuite {
       generatedSrc,
       """package com.example.service;
        |
-       |import com.akkaserverless.javasdk.impl.view.UpdateHandlerNotFound;
-       |import com.akkaserverless.javasdk.impl.view.ViewRouter;
-       |import com.akkaserverless.javasdk.view.View;
+       |import com.akkaserverless.javasdk.view.ViewProvider;
```
jrudolph, line 210: Imports are currently all over the place and we should sort them ultimately, but that will require touching all tests.

codegen/core/src/main/scala/com/lightbend/akkasls/codegen/SourceGeneratorUtils.scala
```diff
@@ -181,38 +180,61 @@ object SourceGeneratorUtils {
...
-      case name: FullyQualifiedName => Seq(name)
-      case block: CodeBlock         => block.fqns
-      case seq: Seq[_]              => new CodeBlock(seq).fqns
-      case _                        => Seq.empty
+      case FQNElement(name) => Seq(name)
+      case block: CodeBlock => block.fqns
+      case _: StringElement => Seq.empty
     }
+
+    def write(imports: Imports, typeName: (FullyQualifiedName, Imports) => String): String =
```
raboof, line 213: I guess this could even be

```suggestion
    def write(typeName: FullyQualifiedName => String): String =
```

but perhaps not worth the change

codegen/core/src/main/scala/com/lightbend/akkasls/codegen/SourceGeneratorUtils.scala
```diff
@@ -181,38 +180,61 @@ object SourceGeneratorUtils {
...
     code.foreach(validateType)
 
     private def validateType(seq: Any): Unit = seq match {
-      case _: String             => ()
-      case _: FullyQualifiedName => ()
-      case block: CodeBlock      => block.code.foreach(validateType)
-      case s: Seq[_]             => s.foreach(validateType)
-      case other =>
-        throw new IllegalArgumentException(s"Unexpected value of type [${other.getClass}] in block: [$other]")
+      case _: StringElement => ()
```
raboof, line 201: I think `validateType` can be removed completely now that we convert to `CodeElement` earlier

codegen/java-gen/src/main/scala/com/lightbend/akkasls/codegen/java/ViewServiceSourceGenerator.scala
```diff
@@ -75,213 +76,186 @@ object ViewServiceSourceGenerator {
...
-        |  }
-        |
-        |}
-        |""".stripMargin
+    JavaGeneratorUtils.generate(
+      packageName,
+      c"""|$managedComment
+          |
+          |/** A view handler */
+          |public class ${view.routerName} extends $ViewRouter<${view.state.fqn}, ${view.impl}> {
```
raboof, line 97: Maybe even change the `view.routerName` string to a `view.router` FQN everywhere - but not important for this PR.

codegen/java-gen/src/main/scala/com/lightbend/akkasls/codegen/java/ViewServiceSourceGenerator.scala
```diff
@@ -75,213 +76,186 @@ object ViewServiceSourceGenerator {
...
-        "com.akkaserverless.javasdk.impl.view.UpdateHandlerNotFound",
-        "com.akkaserverless.javasdk.impl.view.ViewRouter",
-        "com.akkaserverless.javasdk.view.ViewProvider",
-        "com.akkaserverless.javasdk.view.ViewCreationContext",
-        "com.akkaserverless.javasdk.view.View",
-        "com.akkaserverless.javasdk.view.ViewOptions",
-        view.classNameQualified,
-        "com.google.protobuf.Descriptors",
-        "com.google.protobuf.EmptyProto",
-        "java.util.function.Function"))
```
raboof, line 145: Satisfying that this is no longer needed :)

codegen/java-gen/src/main/scala/com/lightbend/akkasls/codegen/java/ViewServiceSourceGenerator.scala
```diff
@@ -75,213 +76,186 @@ object ViewServiceSourceGenerator {
...
+          |  private final String viewId;
+          |  private final $ViewOptions options;
+          |
+          |  /** Factory method of ${view.impl} */
+          |  public static ${view.providerName} of(
+          |      $Function<$ViewCreationContext, ${view.impl}> viewFactory) {
+          |    return new ${view.providerName}(viewFactory, "${view.viewId}", $ViewOptions.defaults());
+          |  }
+          |
+          |  private ${view.providerName}(
```
raboof, line 141: Same for `view.providerName`->`view.provider`? It shouldn't lead to different output since this is the file where `view.provider` is defined, but I think it'd read a bit nicer and be 'good style' to prefer FQN's over strings where possible.

project/ReflectiveCodeGen.scala
```diff
@@ -6,9 +6,10 @@ import ProtocPlugin.autoImport.PB
 import akka.grpc.sbt.AkkaGrpcPlugin
 import akka.grpc.sbt.AkkaGrpcPlugin.autoImport._
 import protocbridge.{ Artifact, SandboxedJvmGenerator }
-import sbt.internal.inc.classpath.ClasspathUtilities
+import sbt.internal.inc.classpath.{ ClasspathUtil, ClasspathUtilities }
```
raboof, line 9: :thinking:

codegen/core/src/main/scala/com/lightbend/akkasls/codegen/SourceGeneratorUtils.scala
jrudolph: :+1: It's worth and simple to do :)

codegen/java-gen/src/main/scala/com/lightbend/akkasls/codegen/java/ViewServiceSourceGenerator.scala
jrudolph: In general it makes sense, but we cannot use FQN directly for name declarations directly because it might start to qualify them in spaces like this where we don't want it.

project/ReflectiveCodeGen.scala
jrudolph: Fixed and split out to https://github.com/lightbend/akkaserverless-java-sdk/pull/707.

codegen/java-gen/src/main/scala/com/lightbend/akkasls/codegen/java/ViewServiceSourceGenerator.scala
raboof: I'm pretty sure it won't ever qualify names that are in the 'current package', but that's indeed a requirement we'd be tied to then.

raboof (approved):

codegen/java-gen/src/main/scala/com/lightbend/akkasls/codegen/java/ViewServiceSourceGenerator.scala
jrudolph: Ah, good point. I guess it would be hard to generate code in a foreign package in the first place. So, I agree the change would be nice but let's fix it later.

Issue #509, opened 2021-09-28
user-function.desc not on classpath in Scala projects


johanandren:
Local proxy reports:
```
{"timestamp":"2021-09-28T09:22:26.677Z","thread":"akkaserverless-akka.actor.default-dispatcher-2","logger":"com.akkaserverless.javasdk.impl.DiscoveryImpl","message":"Source info descriptor [protobuf/descriptor-sets/user-function.desc] not found on classpath. Reporting descriptor errors against source locations will be disabled. To fix this, ensure that the following configuration applied to the protobuf maven plugin: \n\n<writeDescriptorSet>true</writeDescriptorSet>\n<includeSourceInfoInDescriptorSet>true</includeSourceInfoInDescriptorSet>\n<descriptorSetFileName>user-function.desc</descriptorSetFileName>\n\nand also that the generated resources directory is included in the classpath:\n\n  <build>\n    <resources>\n      <resource>\n        <directory>${project.build.directory}/generated-resources</directory>\n      </resource>\n    </resources>\n    ...\n","context":"default","severity":"WARN"}
```

patriknw:
We have this enabled in the akkaserverless-framework sbt build and should include something similar via the sbt plugin (or application build.sbt). Look at the PR when James added it https://github.com/lightbend/akkaserverless-framework/pull/323

reid-spencer:
I've noticed this too.  Can we just put James' code into the sbt plugin?

franciscolopezsancho:
After some tests, I think this issue has already been fixed for two reasons:
- I can't reproduce, no matter how many times I run it, the user-function.desc gets into the classpath and the message from the Local proxy doesn't appear, as stated in the header of this issue. 
- [This line](https://github.com/lightbend/akkaserverless-java-sdk/blob/e7eaf196f9f0be9c2d1d1d12f67e3521e4e07ce7/sbt-plugin/src/main/scala/com/akkaserverless/sbt/AkkaserverlessPlugin.scala#L121) introduced in this [PR](https://github.com/lightbend/akkaserverless-java-sdk/pull/708/files#diff-57082be4488fc43ec5980cc20ded3f4c32532e932dfba9fc2a538e4bb1f7fa61L111) makes the difference. After compiling the .proto sources the resource `user-function.desc` is generated by the build. One can reproduce the current issue by removing that line.

franciscolopezsancho:
Shall we close it @johanandren?

johanandren:
[...]

PR #701, opened 2021-11-13
fix: Include service descriptor file in scala images 

johanandren:
References #509

Manually verified that a deployed service built with this that has invalid views shows the right error message including the view query definition from the service descriptor 👍


pvlugter (approved):
👍🏼

raboof (approved):

sbt-plugin/src/main/scala/com/akkaserverless/sbt/AkkaserverlessPlugin.scala
```diff
@@ -79,7 +90,7 @@ object AkkaserverlessPlugin extends AutoPlugin {
         akkaGrpcCodeGeneratorSettings.value :+ AkkaserverlessGenerator.enableDebug) -> (Test / sourceManaged).value,
     Compile / generateUnmanaged := {
       Files.createDirectories(Paths.get((Compile / temporaryUnmanagedDirectory).value.toURI))
-      // Make sure generation has happened
+      // Make sure generation has happened and that the protobuf descriptor file is created
```
jrudolph, line 93: That doesn't work as intended, because there's no guarantee that the descriptor has been generated when `managedResources` are copied to the classpath.

jrudolph: Will fix this.

PR #675, opened 2021-11-01
docs: Composition examples

johanandren:
Added scala shopping cart value entity project

A few samples replaced + added
 * new sample for forwarding using the shopping cart (slightly more realistic but still not great, keeping the samples to only be about the shopping cart though) 
 * async call + mapped response for id generation
 * composing a sequence of calls
 * counter-example to show case the lack of transaction/boundary for an action/composition

Also fixed integration test support for the testkit (dynamic port for proxy needed to come from test containers)

PR #706, opened 2021-11-15
fix protobuf descriptor introspection tests

octonato:
The protobuf introspections was using some old descriptor files. 

While working on the new codegen annotation, I need to update those files and the tests got broken. 

Therefore, this PR does the following:

1. update the descriptors to use the latest files in the shopping cart samples
2. adds a view to the event sourced shopping cart (it was previously using the eventing sample, but this one is gone)
3. re-implemented the integration test for the event sourced sample (we were using an old version of it.)

The plan is to modify the descriptors to use the new annotation, but we still need to support the old format, so I will duplicate the tests and run it against the new and the old format. And this PR is a preparation for it.


samples/java-eventsourced-shopping-cart/src/it/java/com/example/shoppingcart/domain/ShoppingCartIntegrationTest.java
```diff
@@ -13,7 +13,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package com.example.shoppingcart.api;
```
octonato, line 16: This is surprising. The test was defined in package `api`, but it was in directory `com/example/shoppingcart`. 

Noow moved to `com.example.shoppingcart.domain` which is the place where it gets generated.

johanandren: https://github.com/lightbend/akkaserverless-java-sdk/issues/678 I thought I fixed by moving in #675 though, strange.

samples/java-eventsourced-shopping-cart/src/main/java/com/example/shoppingcart/view/ShoppingCartViewServiceImpl.java
```diff
@@ -0,0 +1,35 @@
...
+
+import com.akkaserverless.javasdk.view.ViewContext;
+import com.example.shoppingcart.domain.ShoppingCartDomain;
+
+// This class was initially generated based on the .proto definition by Akka Serverless tooling.
+//
+// As long as this file exists it will not be overwritten: you can maintain it yourself,
+// or delete it so it is regenerated as needed.
+
+public class ShoppingCartViewServiceImpl extends AbstractShoppingCartViewServiceView {
```
johanandren, line 11: If there is nothing but the generated stub then maybe don't check it in and let it regenerate on build?

johanandren (commented):
Would be nice if those descs weren't checked in but generated by compiling the respective protos instead so they keep up to date.

samples/java-eventsourced-shopping-cart/src/main/java/com/example/shoppingcart/view/ShoppingCartViewServiceImpl.java
octonato: yep, good point

octonato:
> Would be nice if those descs weren't checked in but generated by compiling the respective protos instead so they keep up to date.

The problem with that is that we will need to checkout the framework code as well since we depend on the proto annotation defined there. We could eventually take them from some published artifact.

johanandren (approved):

PR #708, opened 2021-11-16
fix: make sure user-function.desc has actually been copied for Scala SDK projects

jrudolph:
Follow up to #701.

If it worked before then only by chance, because there was a race condition between `user-function.desc` being generated and being copied to the `target/scala_2.13/classes` folder.

### DIFF
--- a/sbt-plugin/src/sbt-test/sbt-akkaserverless/compile-only/test
+++ b/sbt-plugin/src/sbt-test/sbt-akkaserverless/compile-only/test
@@ -1,4 +1,4 @@
-> compile
+> products
 
 $ exists src/main/scala/com/example/Main.scala
 $ exists target/scala-2.13/src_managed/main/com/example/AkkaServerlessFactory.scala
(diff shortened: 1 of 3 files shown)


sbt-plugin/src/sbt-test/sbt-akkaserverless/compile-only/test
```diff
@@ -149,5 +149,8 @@ $ exists target/scala-2.13/src_managed/main/com/example/replicated/set/domain/So
 $ exists src/test/scala/com/example/valueentity/domain/SomeValueEntitySpec.scala
 $ exists src/test/scala/com/example/valueentity/SomeValueEntityServiceIntegrationSpec.scala
 
+# Descriptor set must be included in classpath
+$ must-mirror target/scala-2.13/classes/protobuf/descriptor-sets/user-function.desc target/scala-2.13/resource_managed/main/protobuf/descriptor-sets/user-function.desc
```
jrudolph, line 153: Without the fix it fails here and for some reason in that state the target file is a directory and not a file...

raboof (approved):

PR #716, opened 2021-11-17
Generate services and entities from new codegen annotation

octonato:
Depends on https://github.com/lightbend/akkaserverless-framework/pull/970 
References https://github.com/lightbend/akkaserverless-framework/issues/898


codegen/core/src/main/protobuf/akkaserverless/codegen/annotations.proto
```diff
@@ -0,0 +1,155 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// All extension points for Akka Serverless
+
+syntax = "proto3";
+
+package akkaserverless;
+
+option java_package = "com.akkaserverless.codegen";
```
octonato, line 21: This is now bringing all codegen annotation to the SDK project. 

I few things to note: 

* File to import is `akkaserverless/codege/annotations.proto`. Stays consistent with other annotations files.
* Proto package is still `akkaserverless`. this allow us to use `option(akkaserverless.codegen)`. We stay on the same namespace
* Java package must be `com.akkaserverless.codegen`, other we get two `com.akkaserverless.Annotations` files. Now we have  `com.akkaserverless.Annotations` (from framework) and `com.akkaserverless.codegen.Annotations` from the SDK
* All definitions are suffixed with `Def` to avoid conflicts with proto message in `akkaserverless`

codegen/core/src/test/resources/test-files/descriptor-sets/generate-from-samples.sh
```diff
@@ -32,15 +34,29 @@ echo "Generated event-sourced-shoppingcart.desc"
...
+
+protoc --include_imports \
+--proto_path="$1/protocols/sdk/src/main/protobuf" \
+--proto_path="$2/codegen/core/src/main/protobuf" \
+--proto_path="$2/samples/java-valueentity-shopping-cart/src/main/proto" \
+--descriptor_set_out=action-shoppingcart.desc \
+"$2/samples/java-valueentity-shopping-cart/src/main/proto/com/example/shoppingcart/shoppingcart_controller_api.proto" \
+"$2/samples/java-valueentity-shopping-cart/src/main/proto/com/example/shoppingcart/shoppingcart_api.proto"
+
+echo "Generated action-shoppingcart.desc"
```
octonato, line 54: We don't have introspection test for Actions. Also not on the previous format. I will add them. Next thing on my list.

codegen/core/src/test/scala/com/lightbend/akkasls/codegen/ModelBuilderSuite.scala
```diff
@@ -30,7 +30,39 @@ import scala.collection.mutable
...
+    def descriptorSetPath: String
+  }
+  object CodegenAnnotationConfig extends Config {
+    override val label: String = "CodegenAnnotation"
+    override val descriptorSetPath: String = "descriptor-sets"
+
+    def registry: ExtensionRegistry = {
+      val reg = ExtensionRegistry.newInstance()
+      reg.add(com.akkaserverless.codegen.Annotations.codegen)
+      reg.add(com.akkaserverless.Annotations.method)
```
octonato, line 46: I initially thought that `method` was only used on the proxy side and didn't copied it here. 

It's surely used there because that's what tell the proxy how to create the views and eventing. Not clear why we need this in the build tests.

octonato (commented):
I will be doing some more refinements and adding tests for Actions, but this is ready for a first review.

samples/java-eventsourced-shopping-cart/src/main/proto/com/example/shoppingcart/shoppingcart_api.proto
```diff
@@ -37,9 +38,16 @@ message Cart { // <6>
...
+  option (akkaserverless.codegen) = {
+    event_sourced_entity: {
+        name: "ShoppingCart" 
+        package_name: "com.example.shoppingcart.domain"
+        entity_type: "shopping-cart" 
+        state: "Cart"
+        events: [
+          "ItemAdded",
+          "ItemRemoved" ]
+    }
```
johanandren, line 50: This looks great.

samples/java-eventsourced-shopping-cart/src/main/proto/com/example/shoppingcart/view/shopping_cart_view_model.proto
```diff
@@ -19,11 +19,12 @@ package com.example.shoppingcart.view;
 option java_outer_classname = "ShoppingCartViewModel";
 
 import "akkaserverless/annotations.proto";
+import "akkaserverless/codegen/annotations.proto";
 import "com/example/shoppingcart/domain/shoppingcart_domain.proto";
 
 service ShoppingCartViewService {
-  option (akkaserverless.service) = {
-    type : SERVICE_TYPE_VIEW
+  option (akkaserverless.codegen) = {
```
johanandren, line 26: Nice that it is explicit that it is for codegen now, I was honestly not 100% sure before if any parts were read and handed to the proxy.

johanandren (commented):
Looks very promising from a first quick review.

octonato:
Build is broken because artifact is not including the new proto file (I think, didn't look into detail)

Another thing that came to my mind, the docs will break because I changed the files in the shopping cart samples. Instead, we should duplicate those files in the codegen for the introspection tests. We should not re-use files that are used in the docs. 

I will fix this later today.

build.sbt
```diff
@@ -22,6 +23,12 @@ def common: Seq[Setting[_]] =
     Compile / javacOptions ++= Seq("-encoding", "UTF-8", "--release", "11"),
     Compile / scalacOptions ++= Seq("-encoding", "UTF-8", "-release", "11"))
 
+lazy val codegenProtoAnnotations = project 
+  .in(file("codegen/proto-annotations"))
+  .enablePlugins(PublicProtocolProject)
+  .settings(common)
+  .settings(name := "akkaserverless-codegen-proto-annotations")
```
octonato, line 30: If we want to keep the annotations in the SDK, we need it in an apart sbt module like this because we need it in two palaces. 

We need the codegenCore to compile it and run the introspections tests and we need it to be available for the use code. This is done by adding the annotations as a dependency on the SDK artifacts.

build.sbt
```diff
@@ -74,6 +82,7 @@ lazy val sdkJava = project
     // We need to generate the java files for things like entity_key.proto so that downstream libraries can use them
     // without needing to generate them themselves
     Compile / PB.targets += PB.gens.java -> crossTarget.value / "akka-grpc" / "main",
+    Compile / PB.protoSources ++= (codegenProtoAnnotations / Compile / PB.protoSources).value,
```
octonato, line 85: Note that there are two things here. 

First, `codegenProtoAnnotations` is added as a dependency. We want users to get the proto file in the classpath when the pick the SDK jar so they can use them in their proto files. 
That could be an explicit dependency, but I think it should be included as a transitive deps of the SDK. The reason is that if an user update their project to the new SDK version, they will get the annotations and are good to go. Otherwise they would need to update their pom.xml. I think we should avoid it.

The SDK ships with the generated classes from the akkaserverless protocol, so here I'm doing the same. I let the SDK generated the new codegen proto classes and that then get included in the SDK jar.

build.sbt
```diff
@@ -203,7 +213,8 @@ lazy val codegenCore =
     .settings(
       name := "akkaserverless-codegen-core",
       testFrameworks += new TestFramework("munit.Framework"),
-      Test / fork := false)
+      Test / fork := false,
+      Compile / PB.protoSources ++= (codegenProtoAnnotations / Compile / PB.protoSources).value)
```
octonato, line 217: here we only need the proto annotations when compiling. We don't need to add it as a transitive dependency. Users will get it from the SDK already.

codegen/core/src/test/scala/com/lightbend/akkasls/codegen/TestData.scala
```diff
@@ -226,7 +226,6 @@ class TestData(val packageNamingTemplate: PackageNaming) {
     valueEntity(domainProto(suffix), suffix)
   def valueEntity(parent: PackageNaming, suffix: String = ""): ModelBuilder.ValueEntity =
     ModelBuilder.ValueEntity(
-      parent.protoPackage + s".MyValueEntity$suffix",
```
octonato, line 229: Clean-up: this is not being used.

codegen/scala-gen/src/test/scala/com/akkaserverless/codegen/scalasdk/ClassNameSuite.scala
```diff
@@ -23,7 +23,7 @@ class ClassNameSuite extends munit.FunSuite {
     val d = DescriptorSet.descriptors(
       classOf[ClassNameSuite].getResourceAsStream("/test-files/descriptor-sets/value-shoppingcart.desc"))
     val descriptors = d.right.get.right.get.toList
-    assertEquals(descriptors.size, 11)
+    assertEquals(descriptors.size, 12)
```
octonato, line 26: that's normal, we have one more proto file.

project/ProtocolProject.scala
```diff
@@ -0,0 +1,59 @@
+import sbt._
+import sbt.Keys._
+import sbt.internal.util.ManagedLogger
+import sbtprotoc.ProtocPlugin
+import scala.sys.process._
+
+/**
+ * Settings for protocol projects.
+ */
+object ProtocolProject extends AutoPlugin {
```
octonato, line 10: Picked from the framework project. After we release a new SDK we can also copy the plugin part that checks for breaking changes. I will ticketify this so we don't forget.

samples/java-eventsourced-shopping-cart/src/main/proto/com/example/shoppingcart/domain/shoppingcart_domain.proto
```diff
@@ -5,20 +5,7 @@
...
-// tag::event_sourced_entity[]
-// Describes how this domain relates to an event sourced entity
-option (akkaserverless.file).event_sourced_entity = { // <4>
-  name: "ShoppingCart" // <5>
-  entity_type: "shopping-cart" // <6>
-  state: "Cart" // <7>
-  events: ["ItemAdded", "ItemRemoved"] // <8>
-};
-// end::event_sourced_entity[]
+option java_outer_classname = "ShoppingCartDomain"; // <2>
```
octonato, line 8: oh, I still need to clean up this. That will make the docs to break. 

Actually I naïvely tried to fixed this today by swapping the descriptor-sets folders in d17d48cb8028c205db0c321d3feca055021b7534. Silly, the problem in the docs are the files, not the descriptors. 🤦

build.sbt
jrudolph: > The reason is that if an user update their project to the new SDK version, they will get the annotations and are good to go. Otherwise they would need to update their pom.xml. I think we should avoid it.

I agree, no extra dependency should be needed.

johanandren (approved):

build.sbt
```diff
@@ -22,6 +23,12 @@ def common: Seq[Setting[_]] =
     Compile / javacOptions ++= Seq("-encoding", "UTF-8", "--release", "11"),
     Compile / scalacOptions ++= Seq("-encoding", "UTF-8", "-release", "11"))
 
+lazy val codegenProtoAnnotations = project
+  .in(file("codegen/proto-annotations"))
+  .enablePlugins(PublicProtocolProject)
+  .settings(common)
+  .settings(name := "akkaserverless-codegen-proto-annotations")
```
patriknw, line 30: will these be part of the java-sdk or is this only temporary?

codegen/core/src/test/resources/test-files/codegen-annotation/protos/event-sourced-entity/com/example/shoppingcart/shoppingcart_api.proto
```diff
@@ -0,0 +1,84 @@
...
+
+message Cart { // <6>
+  repeated LineItem items = 1;
+}
+
+service ShoppingCartService { // <7>
+  option (akkaserverless.codegen) = {
+    event_sourced_entity: {
+        name: "ShoppingCart" 
+        package_name: "com.example.shoppingcart.domain"
```
patriknw, line 56: I guess `package` is a reserved word?

patriknw (commented):
looking good

codegen/core/src/test/resources/test-files/codegen-annotation/protos/event-sourced-entity/com/example/shoppingcart/shoppingcart_api.proto
octonato: yes, it's reserved. 

I've been struggling to name this property. 

It's a proto package when we use it to localize the state and events messages, but it's the java package in which we will generate the ShoppingCart entity. 

Ideally, we should have two different fields, but that feels too verbose. We will need to document it well. In any case, we do need to have a codegen section in the docs explaining the codegen.

build.sbt
octonato: We have been discussing it. Johan made the point that since it's a codegen concern, we could keep it in the codegen itself and make it part of the SDK. There are pros and cons about it and I don't think the decision is settled. 

We are keeping it here for now to move fast and have more time to think about it.

codegen/core/src/test/resources/test-files/codegen-annotation/generate-from-samples.sh
```diff
@@ -0,0 +1,76 @@
+#!/bin/bash
+set -euo pipefail
+IFS=$'\n\t'
+
+# Re-generates sample proto file descriptors from the source
+# Requires https://github.com/lightbend/akkaserverless-java-sdk to be cloned locally, and its path provided
+# Usage: ./generate-from-samples.sh /path/to/akkaserverless-framework /path/to/akkaserverless-java-sdk
```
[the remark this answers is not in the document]
octonato: Not sure if that's important.

If we use one that doesn't align, we will produce description files that are in principle not what we want. The introspection tests will pass, but the sample and compilation test won't. And if they do, then we are also good.

Anyway, I think we have enough tests to cover it.

codegen/core/src/test/resources/test-files/codegen-annotation/protos/event-sourced-entity/com/example/shoppingcart/shoppingcart_api.proto
jrudolph: > It's a proto package when we use it to localize the state and events messages, but it's the java package in which we will generate the ShoppingCart entity.

jrudolph: > It's a proto package when we use it to localize the state and events messages, but it's the java package in which we will generate the ShoppingCart entity.

That sounds weird. Why should it apply to resolving state and events? Let's try to keep it such that state and event resolution strictly uses proto scope like other types in a proto file as well. Where those live in the Java program should be a different concern.

The question is then if we still need the package annotation at all for generation as well? Couldn't we just remove that setting and generate the entity in the same package that has been declared in the `java_package_name` option for the file? Which kind of usages would that break?

octonato: The problem is that this is an annotation on the service itself. The service is part of the external API. 

For example, the file above has proto package `com.example.shoppingcart`. The Java class for this service will be generated in `com.example.shoppingcart. ShoppingCartApi.ShoppingCartService` by akka grpc. But this is not really visible to the user, unless they build a client.

And then we have the entity to be generated. It's true that what we generate is a mix of a service and domain. It will have methods that will be derived from the service definition, like `addItem` and `removeItem`. Incoming traffic will be forwarded to those methods. In that sense, they are kind of service methods or external API. 

We could then say, the package for this entity is: `com.example.shoppingcart` and the fully qualified name will be `com.example.shoppingcart.ShoppingCart`.
From there we could say: if you want it in another package, prefix `ShoppingCart` with whatever you want, like `domain.ShoppingCart`. Then it becomes `com.example.shoppingcart.domain.ShoppingCart`.

Same for the state and events. If the user define then on the same package (ie: `com.example.shoppingcart`) all good. If not, they need to prefix them. 

The drawback is that entity, state and events will have a common root package with the service. Not that this is a big deal.

octonato: @jrudolph, I'm not sure I understand this sentence

> Let's try to keep it such that state and event resolution strictly uses proto scope like other types in a proto file as well. 

Can you clarify? 

For the record, both state and events are pure strings and we can't reference then as proto Messages. It's not that we can import a proto file and reference them.

octonato: hmm, now I'm in doubt about this: 
> It's not that we can import a proto file and reference them. 

I hope I'm wrong about it. Maybe it's possible to write: 

```scala
    event_sourced_entity: {
        name: "ShoppingCart" 
        package_name: "com.example.shoppingcart.domain"
        entity_type: "shopping-cart" 
        state: Cart
        events: [
          ItemAdded,
          ItemRemoved ]
    }
  };
```

jrudolph: > Can you clarify?

It was about how we resolve state and event, not how to declare them in the proto. It doesn't seem possible to refer to message types in annotation, so they have to be strings. In our resolution code, we can still use the same logic as proto does itself, i.e if the message is defined in the same file, you can refer to it simply by just the name, but if it is defined in another proto of another package, you can also use the fully-qualified (proto) name to reference those (but not using the Java packages).

> The problem is that this is an annotation on the service itself. The service is part of the external API.

Ah, right, so we cannot move the entity definition easily to another file (as before) to change the target package using the per-file option.

Since we now want to bundle service and entity more closely together, do we really need to provide different packages for service and entity after all? Functionally, we need the distinction between `api` and `domain` only to tell messages apart that should be named the same between `api` and `domain` but the API service never turns up in user code anyway.

octonato: Yeah, I think it's ok to generate the entity in the package defined by the current file. 

What about the following...

```scala
package com.example.shoppingcart; 

service ShoppingCartService {
	      event_sourced_entity: {
	          name: "ShoppingCart" 
	          entity_type: "shopping-cart" 
	          state: "Cart"
	          events: [
	            "ItemAdded",
	            "ItemRemoved" ]
	      }
	    };
...
}
```

Entity generated in `com.example.shoppingcart`. 
State and Events must be define in proto package `com.example.shoppingcart`.

Then, if they are elsewhere...

```scala
package com.example.shoppingcart; 

service ShoppingCartService {
	      event_sourced_entity: {
	          name: "ShoppingCart" 
	          entity_type: "shopping-cart" 
	          state: ".domain.Cart"
	          events: [
	            ".domain.ItemAdded",
	            ".domain.ItemRemoved" ]
	      }
	    };
...
}
```

Meaning the must be in `com.example.shoppingcart.domain`. 

Would that make more sense for you?

octonato: of course, once could also reference state and events by using a full package, eg `foo.bar.Cart`. In which case, the domain message `Cart` is not sharing any package with the service.

jrudolph: So, summing up, with this change, "Entity" and "Service" are not siblings any more in the proto file but one unit. Wouldn't it make sense to try to treat those more as a unit from the point of the user in the first place? So, that there's only a single thing which is (conceptually) "the entity" which provides an API through its (implicit) service.

Couldn't we even remove the `name` then by using a naming convention, e.g. by just appending `Entity` to the service name or similar (even better if we could just use the service name but that probably doesn't work because of the existing akka-grpc code gen)?

```
service ShoppingCart {
    event_sourced_entity: {
        entity_type: "shopping-cart" 
        state: "Cart"
        events: [
          "ItemAdded",
          "ItemRemoved" ]
    }
  };

  rpc AddItem (AddLineItem) returns (google.protobuf.Empty) {
    option (google.api.http) = {
      post: "/cart/{cart_id}/items/add"
      body: "*"
    };
  }

  rpc RemoveItem (RemoveLineItem) returns (google.protobuf.Empty) {
    option (google.api.http).post = "/cart/{cart_id}/items/{product_id}/remove";
  }

  rpc GetCart (GetShoppingCart) returns (Cart) {
    option (google.api.http) = {
      get: "/carts/{cart_id}"
      additional_bindings: {
          get: "/carts/{cart_id}/items"
          response_body: "items"
      } };
  }
}
```

octonato: I think we tried with name conventions before we unsatisfying results. 

We had things like `ShoppingCartService` and `ShoppingCartServiceEntity` which only make things even more convoluted. And the reason is that we needed to sufix it because of conflicts with the akka-grpc generated service.

At the end the user get exposed to the akka-grpc service as well. That happens in the integration tests for instance. So, we should not reuse the name nor suffix it with Entity, IMO. 

Also, if I'm coding a ShoppingCart, I want a class classed ShoppingCart, not ShoppingCartEntity. Entity is a technicality that has nothing to do with the domain I'm modeling. But that's maybe just me. 😄 

In addition to that, I think the new annotations should generate exactly the same code as before. So when updating, I can simply change my protos without touching any of my classes.

jrudolph: I'm fine with an optional `name` annotation but we should try remove required options if possible to minimize choice a user has to make before getting started.

> Also, if I'm coding a ShoppingCart, I want a class classed ShoppingCart, not ShoppingCartEntity. Entity is a technicality that has nothing to do with the domain I'm modeling. But that's maybe just me.

I agree, but as we do in other places we can use whatever logic we want to deduce a suitable default name (e.g. if service ends with "Service" just remove it, if it doesn't end with service, add `Entity` etc.).

octonato: ok, let me rework this a bit. 

I will start by removing the package name. 

We can then experiment with re-using service name (with convention to append entity) and make the name an optional thing. 

Let's see where we get.

octonato: About the naming convention, I think you are right. We are using the Service name for Actions and Views. 

We will simplify the game if we follow the same convention for entities. Not that the convention is obvious at first, but once the user learn it, they can look to any proto definition and understand what will be generated. 

Btw, we could also add a option `name` field to actions and views to close the circle.

octonato: Wow, we had it already

```
message ActionDef {
  // Optional name for the view - if not defined, will follow the name of the service
  string name = 1;
}
```

Not sure if we were using this.

octonato: @jrudolph, the package name is removed as discussed here (see 1568f0fe)

I will work on the change to make the name optional and use the service name instead, but will send it in another PR on top of this as I see it as a change in the current behavior. 

I would like to avoid bloating this PR even further. There is already a lot in here.

johanandren (approved):

PR #717, opened 2021-11-18
chore(codegen): more explicitly order imports

raboof:


codegen/core/src/main/scala/com/lightbend/akkasls/codegen/Imports.scala
```diff
@@ -30,6 +30,12 @@ class Imports(val currentPackage: String, val _imports: Seq[String]) {
       .filter(_._2.size > 1)
       .keySet
 
+  def ordered: List[Seq[String]] = {
+    val (stdlib, other) =
+      imports.partition(i => i.startsWith("java.") || i.startsWith("javax.") || i.startsWith("scala."))
```
patriknw, line 35: would we like akkaserverless and akka as another group?

codegen/java-gen/src/test/scala/com/lightbend/akkasls/codegen/java/MainSourceGeneratorSuite.scala
```diff
@@ -182,6 +183,7 @@ class MainSourceGeneratorSuite extends munit.FunSuite {
         |import com.example.service.view.MyServiceViewImpl;
         |import com.example.service.view.MyServiceViewProvider;
         |import com.example.service.view.ServiceOuterClass;
+        |
         |import java.util.function.Function;
```
patriknw, line 187: shall we have the std lib imports at the top?

raboof: I agree that seems nicer, but I tried to follow the IntelliJ default conventions which seem to put them at the bottom...

octonato: I'm not sure that this has any added value for us (and for the user). 

A simple alphabetical sort should be enough for generated code. 

And for the generated code that can be edited by the user, we should not care because their IDE will sort it according to their preferences.

jrudolph (approved):
LGTM

codegen/java-gen/src/test/scala/com/lightbend/akkasls/codegen/java/MainSourceGeneratorSuite.scala
jrudolph: I agree the exact ordering is not super important but having that grouping is quite nice.

Issue #715, opened 2021-11-17
Document and improve topic message type handling


johanandren:
Comes out of the #691 investigation

We have no mention at all in the documentation on how events coming from topics are handled with regard to the payload type.

Since the switch from annotation driven method resolution we also do not support all "raw" message types that may come from a topic into Akka Serverless.

Some detective work results:
 * The content type of messages coming in from a topic differs a bit between the two current supported brokers parsing events into "CloudEvents" which always has a content type:
   * Google Pub/Sub has a `Content-Type` attribute in their protocol
   * Current Kafka integration looks for a message attribute called `ce-datacontenttype`
   * if an event from either broker does not have that present the type is `application/octet-stream`
 * The CloudEvent content-type is then used to turn the event into a protobuf Any to be handed to the user function
   * for `application/json` or `application/something+json` it is encoded as Any with the type_url `json.akkaserverless.com/` + cloud event "type" field (empty if not present)
   * for `text/plain; charset=utf-8` it is encoded as `p.akkaserverless.com/string`
   * for `application/octet-stream` it is encoded as `p.akkaserverless.com/bytes`

We should document how these message types are passed on and or handled by default, per supported broker in the docs.

The old reflection based Java SDK had support for making a parameter or return type have String, protobuf ByteString or a Jackson Annotated type as parameter and the serialization in the SDK handling our internal representations of those types. 

We currently have JSON covered by accepting proto `Any` and provide an explicit conversion. For ByteStrings and raw Strings this feels a bit more clunky but maybe we should make similar utilities for those.

raboof:
> Current Kafka integration looks for a message attribute called ce-datacontenttype

Looks like that's consistent with the 'binary content mode' of the kafka-cloudevent binding spec, https://github.com/cloudevents/spec/blob/v1.0.1/kafka-protocol-binding.md#3231-property-names

Issue #691, opened 2021-11-09
ClassCast when receiving raw bytes from topic


johanandren:
This could potentially be a bug in the proxy rather than the SDK as well:

If an event coming from a topic does not have any content type it is passed as raw `ByteString`, from looking at the logic in the proxy I'd expect the bytes to wrapped in a proto `Any` with the type_url `p.akkaserverless.com/bytes` but @samcschneider reported getting this error:

```
java.lang.ClassCastException: class com.google.protobuf.ByteString$LiteralByteString cannot be cast to class com.google.protobuf.any.Any (com.google.protobuf.ByteString$LiteralByteString and com.google.protobuf.any.Any are in unnamed module of loader 'app')
  com.lightbend.domain.SchedulerTopicSubscriptionActionRouter.handleUnary(SchedulerTopicSubscriptionActionRouter.scala:22)\n\tat com.akkaserverless.scalasdk.impl.action.JavaActionRouterAdapter.handleUnary(ActionAdapters.scala:77)
  com.akkaserverless.javasdk.impl.action.ActionRouter.$anonfun$handleUnary$1(ActionRouter.scala:47)
  com.akkaserverless.javasdk.impl.action.ActionRouter.callWithContext(ActionRouter.scala:165)
  com.akkaserverless.javasdk.impl.action.ActionRouter.handleUnary(ActionRouter.scala:46)
  com.akkaserverless.javasdk.impl.action.ActionsImpl.handleUnary(ActionsImpl.scala:181)
  com.akkaserverless.protocol.action.ActionsHandler$.$anonfun$partial$2(ActionsHandler.scala:115)
  ...
```

johanandren:
The problem is that the leftovers of the old reflection centered serialization/deserialization support currently makes it impossible to accept events from topics that Akka Serverless interpreted as bytes or text. See #715 for more details.

johanandren:
Tried just dropping the primitive support since API request, response and event messages should never be primitives, but we rely on being able to use primitives in the CRDTs, so we'd need two different paths for deserialization for that vs service messages at least.

Found the following pre-existing wrapper types `com.google.protobuf.BytesValue`, `com.google.protobuf.StringValue` that could perhaps be used in the proto descriptor and that we could serialize/deserialize into our own `Any`-representation. I'll investigate that track covering bytes, strings and JSON instead of the current `Any` scheme.

PR #719, opened 2021-11-18
fix: Support for byte string and text messages from topics 

johanandren:
Refs #691 and #715 

Currently a service receiving messages from a topic can handle JSON through the somewhat clunky pass-through protobuf Any scheme but while the proxy can pick up raw bytes/non-proto payloads and string messages from topics, there is no current way to accept those into a Java/Scala SDK app.

This PR adds that by changes to AnySupport for regular service messages (and less importantly, also entity state/snapshot) deserialization of messages from the proxy now turns:

*  `p.akkaserverless.com/string` into `com.google.protobuf.StringValue` 
*  `p.akkaserverless.com/bytes` into `com.google.protobuf.BytesValue`

Or the corresponding ScalaPB `com.google.protobuf.wrapper` case class types for Scala.

The idea is that a user proto descriptor could use those built in proto wrapper types in their schema and codegen will generate the right wrapper types in the Java/Scala code like this: 

```protobuf
import "akkaserverless/annotations.proto";
import "google/protobuf/wrappers.proto";

service DelegatingService {
  option (akkaserverless.service) = {
    type : SERVICE_TYPE_ACTION
  };

  rpc ConsumeRawByteMessages(google.protobuf.BytesValue) returns (google.protobuf.Empty) {
    option (akkaserverless.method).eventing.in = {
      topic:  "bytes-topic"
    };
  };

  rpc ConsumeTextMessages(google.protobuf.StringValue) returns (google.protobuf.Empty) {
    option (akkaserverless.method).eventing.in = {
      topic:  "text-topic"
    };
  };
}
```

It should also allow publishing raw bytes or strings to a topic by using the same types on the output side.

I have also made it fail fast on primitive types which the old SDK could encode/decode as parameters and return value but there currently is no reasonable type safe way to represent in the Java SDK (or the proto descriptor)

If we agree on this I'll continue on codegen and docs (and eventually reach the "other side" documenting ingestion and emitting of data with explanation of types.

I did on purpose not change how the JSON payloads work, even though tempting to turn them into StringValue, since the type_url in those can contain a discriminator that makes it possible to chose among many possible types to deserialize into, and with the wrapper we'd loose that information.


johanandren:
Started on docs directly in this PR but hit a problem I had not foreseen. The proxy resolves which subscriber message an event should go to for `p.akkaserverless.com/` and `json.akkaserverless.com/` based on the expectation that they accept proto Any. To allow `StringValue` and `ByteValue` the proxy eventing infra will need some changes as well.

johanandren:
Pausing this while waiting for lightbend/akkaserverless-framework#973 reaching prod (since these API/doc changes won't work for users until then)

johanandren:
Note to self when continuing: the topic docs should cross link to the set-up-broker section, seems like it only links to google pub sub docs now.

johanandren:
This now _almost_ works. Consuming topics with bytes, strings, json works fine but publishing needs another round of changes to the proxy before working quite like what I have described in the docs.

The problem that needs to be solved is that we use the `reply` both for reply to a client and the value to emit to a topic if annotated to. There are two options that I can think of:
1.  the actually returned protobuf message must match the service descriptor return type and eventing needs to transform the specific wrapper types to the right kind of content-type/message to publish
2. what is published is what we returned and the proxy must transform the response to the client to match the proto descriptor

I think 1. makes most sense, so I'll do another round of change where `BytesValue` is turned into raw bytes with content type `application/octet-stream` and `StringValue` is turned into `text/plain;encoding=utf-8` messages.

There is one additional caveat for JSON where I can't really come up with anything that makes sense: we currently get a user function error that `json.akkaserverless.com/com.example.MyJsonMessage` is not the `type.googleapis.com/google.protobuf.Any` that the proxy expects because of the descriptor. This also means that a gRPC client triggering publishing of JSON to a topic is in for a surprising answer. We can silence the warning in this case but there isn't really anything sensible that we can return to the user.

johanandren:
Fix for the last step in: https://github.com/lightbend/akkaserverless-framework/pull/1007

johanandren:
Fix is now live (proxy 0.8.3) so this will now work as described in this PR.

docs/src/modules/java/partials/pubsub-note.adoc
```diff
@@ -1,6 +1,4 @@
 [NOTE]
 ====
-To use Google Cloud Pub/Sub in for your Akka Serverless project, please follow the instructing at https://developer.lightbend.com/docs/akka-serverless/projects/message-brokers.html[how to configure message brokers].
-
-It is your responsibility to create the topics in Google Cloud Pub/Sub before configuring publishers or subscribers.
+In the protobuf descriptors topic names are referenced but no additional details about how to connect to the topics. When deploying the application there must be a broker configuration with credentials and details on how connect in the Akka Serverless project. For details about configuring a broker see https://developer.lightbend.com/docs/akka-serverless/projects/message-brokers.html[Configure message brokers]
```
octonato, line 3:
```suggestion
In the Protobuf descriptors topic names are referenced but no additional details about how to connect to the topics. When deploying the application there must be a broker configuration with credentials and details on how connect in the Akka Serverless project. For details about configuring a broker see https://developer.lightbend.com/docs/akka-serverless/projects/message-brokers.html[Configure message brokers]
```

Now, I don't know how consistent we have been in using the word `Protobuf`. There are probably other places in text that we use lowercase.

octonato: Or maybe we should say, "In the gRPC descriptors..." ?

sdk/java-sdk/src/main/scala/com/akkaserverless/javasdk/impl/AnySupport.scala
```diff
@@ -355,6 +355,20 @@ class AnySupport(
       case javaPbAny: JavaPbAny   => ScalaPbAny.fromJavaProto(javaPbAny)
       case scalaPbAny: ScalaPbAny => scalaPbAny
 
+      // these are all generated message so needs to go before GeneratedMessage,
+      // but we encode them in side any just like regular message, we just need to get the type_url right
```
octonato, line 359:
```suggestion
      // but we encode them inside Any just like regular message, we just need to get the type_url right
```

octonato (approved):
LGTM, left a few typo fixes though

docs/src/modules/java/pages/serialization.adoc
```diff
@@ -7,40 +7,23 @@ You do not need to handle serialization for messages. Akka Serverless functions
...
+If the content type starts with `application/protobuf`, `application/x-protobuf` or `application/vnd.google.protobuf` the payload is expected to also have a `ce-type` header or attribute identifying the concrete protobuf message type. Such messages will be decoded into the described message type before handed to a topic subscriber method, which must accept that specific message type.
 
-[cols="a,a"]
-|===
-| Protobuf type | Java type
+If the publishing service is also an Akka Serverless service, this is handled transparently for you as shown in xref:actions-publishing-subscribing.adoc[].
 
-| string
-| java.lang.String
+For messages consumed from or published to topics to interact with external services it can be a requirement to use another format than `protobuf`.
```
pvlugter, line 18: Just a suggested tweaking:

```suggestion
For messages that are consumed from or published to topics when interacting with external services, it can be a requirement to use a format other than `protobuf`. Other supported message formats include JSON, text, or raw bytes.
```

docs/src/modules/java/pages/serialization.adoc
```diff
@@ -102,4 +85,104 @@ include::example$scala-doc-snippets/src/main/scala/com/example/json/MyServiceAct
...
+If an action has a return type of `StringValue` and publishes to a topic, the events published to the topic will have content-type `text/plain; charset=utf-8`.
+
+== Bytes
+
+If the content type is `application/octet-stream`, no content type is present, or the type is unknown to Akka Serverless the message is treated as a binary message. The topic subscriber method must accept the `google.protobuf.BytesValue` message.
+
+[.tabset]
+Java::
++
+A `proto` definition of an Action that consumes String messages can look like this:
```
pvlugter, line 126:
```suggestion
A `proto` definition of an Action that consumes binary messages with raw bytes can look like this:
```

docs/src/modules/java/pages/serialization.adoc
```diff
@@ -102,4 +85,104 @@ include::example$scala-doc-snippets/src/main/scala/com/example/json/MyServiceAct
...
+.src/main/proto/com/example/topics_action.proto
+----
+include::example$java-doc-snippets/src/main/proto/com/example/topics_action.proto[tag=bytes]
+----
+<1> `google.protobuf.BytesValue` requires the import `google/protobuf/wrappers.proto`.
+<2> When consuming raw bytes messages from a topic the input type must be `google.protobuf.BytesValue`.
+
+Scala::
++
+A `proto` definition of an Action that consumes String messages can look like this:
```
pvlugter, line 138:
```suggestion
A `proto` definition of an Action that consumes binary messages with raw bytes can look like this:
```

docs/src/modules/java/pages/serialization.adoc
```diff
@@ -102,4 +85,104 @@ include::example$scala-doc-snippets/src/main/scala/com/example/json/MyServiceAct
...
+- `ce-type` = fully qualified protobuf message name (eg. `shopping.cart.api.TopicOperation`)
+
+(The `ce-` prefixed attributes are part of the CloudEvents specification.)
+
+The Protobuf rpc declaration uses the expected Protobuf message type and specifies the topic to subscribe to. You'll normally want to share the exact Protobuf message declaration with the sending system.
+
+[.tabset]
+Java::
++
+A `proto` definition of an Action that consumes cloud event messages can look like this:
```
pvlugter, line 170: Should it be `CloudEvent` rather than `cloud event`? (and references below)

pvlugter:
```suggestion
A `proto` definition of an Action that consumes CloudEvent messages can look like this:
```

docs/src/modules/java/pages/serialization.adoc
```diff
@@ -102,4 +85,104 @@ include::example$scala-doc-snippets/src/main/scala/com/example/json/MyServiceAct
...
++
+[source,proto,indent=0]
+.src/main/proto/com/example/topics_action.proto
+----
+include::example$java-doc-snippets/src/main/proto/com/example/topics_action.proto[tag=cloudEvent]
+----
+<1> When consuming a cloud event containing a protobuf message the handler request must have the message type specified in the metadata
```
pvlugter, line 177:
```suggestion
<1> When consuming a CloudEvent containing a Protobuf message the handler request must have the message type specified in the metadata
```

docs/src/modules/java/pages/serialization.adoc
```diff
@@ -102,4 +85,104 @@ include::example$scala-doc-snippets/src/main/scala/com/example/json/MyServiceAct
...
+
+Scala::
++
+A `proto` definition of an Action that consumes cloud event messages can look like this:
```
pvlugter, line 181:
```suggestion
A `proto` definition of an Action that consumes CloudEvent messages can look like this:
```

docs/src/modules/java/pages/serialization.adoc
```diff
@@ -102,4 +85,104 @@ include::example$scala-doc-snippets/src/main/scala/com/example/json/MyServiceAct
...
++
+[source,proto,indent=0]
+.src/main/proto/com/example/json/json_api.proto
+----
+include::example$scala-doc-snippets/src/main/proto/com/example/topics_action.proto[tag=cloudEvent]
+----
+<1> When consuming a cloud event containing a protobuf message the handler request must have the message type specified in the metadata
```
pvlugter, line 188:
```suggestion
<1> When consuming a CloudEvent containing a Protobuf message the handler request must have the message type specified in the metadata
```

docs/src/modules/java/partials/pubsub-note.adoc
```diff
@@ -1,6 +1,4 @@
 [NOTE]
 ====
-To use Google Cloud Pub/Sub in for your Akka Serverless project, please follow the instructing at https://developer.lightbend.com/docs/akka-serverless/projects/message-brokers.html[how to configure message brokers].
-
-It is your responsibility to create the topics in Google Cloud Pub/Sub before configuring publishers or subscribers.
+In the Protobuf descriptors topic names are referenced but no additional details about how to connect to the topics. When deploying the application there must be a broker configuration with credentials and details on how connect in the Akka Serverless project. For details about configuring a broker see https://developer.lightbend.com/docs/akka-serverless/projects/message-brokers.html[Configure message brokers]
```
pvlugter, line 3: Suggested tweak:

```suggestion
In the Protobuf descriptors, only topic names are referenced and no additional details about how to connect to the topics are needed. When deploying the application there must be a broker configuration in the Akka Serverless project, with credentials and details on how connect to the broker. For details about configuring a broker see https://developer.lightbend.com/docs/akka-serverless/projects/message-brokers.html[Configure message brokers]
```

pvlugter (approved):
Nice. LGTM. Added some small suggestions for the docs, which can be used or not.

docs/src/modules/java/pages/serialization.adoc
```diff
@@ -7,40 +7,23 @@ You do not need to handle serialization for messages. Akka Serverless functions
 
 The gRPC services are also exposed as HTTP endpoints with JSON messages. See xref:proto.adoc#_transcoding_http[Transcoding HTTP].
 
-For messages consumed from or published to topics it can be needed to use another format than `protobuf`. Akka Serverless provides a utility to serialize and deserialize JSON messages.
+== When consuming Messages from Topics
 
-== Primitive types
+When a message arrives from a topic, Akka Serverless detects the message payload type based on the `Content-Type` or `ce-datacontenttype` header or attribute of the message. If there is no such metadata, the content is seen as raw bytes.
```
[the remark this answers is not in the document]
johanandren: Pretty sure your native english is better than my "eu speak". 👍

Issue #722, opened 2021-11-19
addSideEffect throws exception when trying to access the context


franciscolopezsancho:
Adding to the scala-eventsourced-conter the following call implementation
```
  override def increaseWithSideEffect(
      currentState: CounterState,
      increaseValue: example.IncreaseValue): EventSourcedEntity.Effect[Empty] =
    if (increaseValue.value < 0) {
      effects.error("Value must be a zero or a positive number")
    } else {
      effects
        .emitEvent(ValueIncreased(increaseValue.value))
        .thenAddSideEffect(_ =>
          SideEffect(components.counter.increase(example.IncreaseValue(context.entityId, increaseValue.value * 2))))
        .thenReply(_ => Empty())
    }
```
On `sbt run` when calling like this ```grpcurl -plaintext  -d '{"counter_id": "1", "value": 1}' localhost:9000 com.example.CounterService/IncreaseWithSideEffect```
raises the error below
```
[info]   java.lang.IllegalStateException: CommandContext is only available when handling a command.
[info]   at com.akkaserverless.scalasdk.eventsourcedentity.EventSourcedEntity.commandContext(EventSourcedEntity.scala:214)
[info]   at com.example.domain.AbstractCounter.components(AbstractCounter.scala:15)
[info]   at com.example.domain.Counter.$anonfun$increaseWithSideEffect$1(Counter.scala:40)
[info]   at com.akkaserverless.scalasdk.impl.eventsourcedentity.EventSourcedEntityEffectImpl.$anonfun$thenAddSideEffect$1(EventSourcedEntityEffectImpl.scala:84)
[info]   at com.akkaserverless.javasdk.impl.eventsourcedentity.EventSourcedEntityEffectImpl.$anonfun$secondaryEffect$1(EventSourcedEntityEffectImpl.scala:62)
```

PR #723, opened 2021-11-19
fix: trying to access to context on a side effect

franciscolopezsancho:
References #722


sdk/java-sdk/src/main/scala/com/akkaserverless/javasdk/impl/eventsourcedentity/EventSourcedEntityRouter.scala
```diff
@@ -131,9 +131,22 @@ abstract class EventSourcedEntityRouter[S, E <: EventSourcedEntity[S]](protected
         val snapshot =
           if (shouldSnapshot) Option(endState)
           else None
-        CommandResult(events.toVector, commandEffect.secondaryEffect(endState), snapshot, currentSequence)
+
+        try {
+          context.asInstanceOf[CommandContextImpl].checkActive()
```
franciscolopezsancho, line 136: As this is the `EventSourcedEntityRouter` we know it has been created with an instance of `CommandContextImpl`. Not too sure though if this is the right approach.

franciscolopezsancho:
Even though the integration test passes and the application also works when `sbt run` it raises the same exception (#722) when running an unit test. I understand that the test kit is not leveraging `EventSourceEntityImpl.runEntity` which is what is fixed here.

johanandren:
The testkit is a little bit tricky because we expect that the secondary effect can be evaluated lazily in the `EventSourcedResultImpl` where we do not have access to the actual entity instance, but because of this issue I think we'll have to do that eagerly in the generated event sourced testkit, so that we can surround it with setting and clearing out the context, and then pass the evaluated effects to EventSourcedResultImpl.

sdk/java-sdk/src/main/scala/com/akkaserverless/javasdk/impl/eventsourcedentity/EventSourcedEntityRouter.scala
johanandren: We shouldn't check active though but rather toggle it to active, no such method right now (but only add that if needed and you see it fail without this `checkActive()` call, I think we may be fine without it)

sdk/java-sdk/src/main/scala/com/akkaserverless/javasdk/impl/eventsourcedentity/EventSourcedEntityRouter.scala
```diff
@@ -131,9 +131,22 @@ abstract class EventSourcedEntityRouter[S, E <: EventSourcedEntity[S]](protected
...
+          entity._internalSetCommandContext(Optional.of(context))
```
johanandren, line 137:
```suggestion
          // side effect callbacks may want to access context or components which is valid
          entity._internalSetCommandContext(Optional.of(context))
```

sdk/java-sdk/src/main/scala/com/akkaserverless/javasdk/impl/eventsourcedentity/EventSourcedEntitiesImpl.scala
```diff
@@ -252,20 +251,21 @@ final class EventSourcedEntitiesImpl(
...
-
   private class EventSourcedEntityContextImpl(override final val entityId: String)
       extends AbstractContext(system)
       with EventSourcedEntityContext
   private final class EventContextImpl(entityId: String, override val sequenceNumber: Long)
       extends EventSourcedEntityContextImpl(entityId)
       with EventContext
 }
+
+class CommandContextImpl(
```
johanandren, line 262: `private[eventsourcedentity]` and `final` (if it still needs to be out here)

sdk/java-sdk/src/main/scala/com/akkaserverless/javasdk/impl/eventsourcedentity/EventSourcedEntityRouter.scala
```diff
@@ -131,9 +131,23 @@ abstract class EventSourcedEntityRouter[S, E <: EventSourcedEntity[S]](protected
...
+          // side effect callbacks may want to access context or components which is valid
+          entity._internalSetCommandContext(Optional.of(context))
+          CommandResult(events.toVector, commandEffect.secondaryEffect(endState), snapshot, currentSequence)
+        } finally {
+          entity._internalSetCommandContext(Optional.empty())
+        }
       case NoPrimaryEffect =>
-        CommandResult(Vector.empty, commandEffect.secondaryEffect(_stateOrEmpty()), None, context.sequenceNumber())
+        try {
+          context.asInstanceOf[CommandContextImpl].checkActive()
```
johanandren, line 145: Same here
```suggestion
```

sdk/java-sdk/src/main/scala/com/akkaserverless/javasdk/impl/eventsourcedentity/EventSourcedEntitiesImpl.scala
franciscolopezsancho: not anymore, I'll put it back

franciscolopezsancho:
continuing in #737

PR #725, opened 2021-11-22
Only apply CustomerCreated once, on an empty state

sebastian-alfers:
If we get a CustomerCreated and the state is already initialized (non-empty), do not apply twice.


samples/scala-eventsourced-customer-registry/src/main/scala/customer/view/CustomerByNameView.scala
```diff
@@ -18,7 +18,7 @@ class CustomerByNameView(context: ViewContext) extends AbstractCustomerByNameVie
 
   override def processCustomerCreated(
     state: Customer, customerCreated: CustomerCreated): UpdateEffect[Customer] = // <3>
-    if (state == emptyState) effects.ignore() // already  created
+    if (state != emptyState) effects.ignore() // already  created
```
pvlugter, line 21: may as well remove the extra space in the comment too

```suggestion
    if (state != emptyState) effects.ignore() // already created
```

pvlugter (approved):
Good catch

johanandren (approved):

Issue #608, opened 2021-10-12
AkkaServerlessFactory might end up in empty package


jrudolph:
The package name for the AkkaServerlessFactory is determined by the `longestCommonPrefix` which might be empty for a set of protos. In that case we should either fail or actually put it on the root level.

It would also be nice if the package could be customized (sbt configuration?).

patriknw:
I assume this would be similar problem for Java

johanandren:
Having heterogenous packages in proto descriptors will now break a project since the component generation incorrectly handles empty package. I think we could fix that, but there will be problems with the Components type in the default package for Java since the actual components, in other packages, refer to it and there is no way to import root package classes in Java as far as I know.

In addition to allow for specifying this, we should fail when the root package is empty with an explanation what to do about it.

PR #730, opened 2021-11-23
fix: configurable root package

johanandren:
And better warning when not configured and guessed root package is empty
(because of no common prefix in proto packages)

References #608


johanandren:
Java is different and already has a `mainClass` configuration for the maven plugin which we use for deciding root package, so may not need any changes.

samples/scala-doc-snippets/src/main/proto/com/example/my_action.proto
```diff
@@ -0,0 +1,24 @@
...
+      topic:  "bytes-topic"
+    };
+  };
+
+  rpc ConsumeTextMessages(google.protobuf.StringValue) returns (google.protobuf.Empty) {
+    option (akkaserverless.method).eventing.in = {
+      topic:  "text-topic"
+    };
+  };
+}
```
johanandren, line 24: Oh, this one snuck in here but doesn't belong

codegen/core/src/main/scala/com/lightbend/akkasls/codegen/SourceGeneratorUtils.scala
```diff
@@ -48,7 +50,16 @@ object SourceGeneratorUtils {
...
       .toSet
     if (packages.isEmpty) throw new IllegalArgumentException("Nothing to generate!")
-    longestCommonPrefix(packages.head, packages.tail)
+    longestCommonPrefix(packages.head, packages.tail) match {
+      case Nil =>
+        val packageNames = packages.map(_.mkString("."))
+        throw new RuntimeException(
+          s"No common package prefix found for protobuf packages [${packageNames.mkString(", ")}]. " +
+          "Auto-selection of root package for shared classes not possible, please configure explicit root package for " +
+          "Akka Serverless code generation in your build.") with NoStackTrace
```
patriknw, line 59: It would have been nice to include a link to documentation that describes how to define that in the build.

patriknw (commented):
I think we need some more instructions of how to define that build setting

codegen/core/src/main/scala/com/lightbend/akkasls/codegen/SourceGeneratorUtils.scala
johanandren: Since this only happens with sbt, wdyt about just explicitly mentioning the sbt setting here?

I expect that it will be a corner case that most users won't end up with, so my thinking is that you only need to know about it if you need it - which will mean you end up with this error.

octonato (approved):
LGTM

codegen/core/src/main/scala/com/lightbend/akkasls/codegen/SourceGeneratorUtils.scala
johanandren: Amended error message, let me know what you think, but already started release so not urgent, this will go in next release.

codegen/core/src/main/scala/com/lightbend/akkasls/codegen/SourceGeneratorUtils.scala
```diff
@@ -48,7 +50,17 @@ object SourceGeneratorUtils {
...
     if (packages.isEmpty) throw new IllegalArgumentException("Nothing to generate!")
-    longestCommonPrefix(packages.head, packages.tail)
+    longestCommonPrefix(packages.head, packages.tail) match {
+      case Nil =>
+        val packageNames = packages.map(_.mkString("."))
+        throw new RuntimeException(
+          s"No common package prefix found for protobuf packages [${packageNames.mkString(", ")}]. " +
+          "Auto-selection of root package for shared generated classes not possible, please configure explicit root package for " +
+          "Akka Serverless code generation in your sbt build using the Akka Serverless sbt plugin setting 'rootPackage', " +
+          """for example: 'rootPackage := Some("com.example")'""") with NoStackTrace
```
jrudolph, line 60: :+1: for providing a helpful message here.

jrudolph (approved):
LGTM

codegen/core/src/main/scala/com/lightbend/akkasls/codegen/SourceGeneratorUtils.scala
patriknw: lgtm, thanks!

PR #735, opened 2021-11-25
feat: allow custom component names

octonato:
Ready for first review.


codegen/core/src/test/scala/com/lightbend/akkasls/codegen/ModelBuilderSuite.scala
```diff
@@ -345,77 +804,289 @@ abstract class ModelBuilderSuite(config: ModelBuilderSuite.Config) extends munit
...
+            transformName = false)))
     }.get
   }
+}
 
-  test(s"deriving java package from proto options (${config.label})") {
-    val protoFileName = "name.proto"
-    val name = "Name"
-    val pkg = "com.example"
+class ModelBuilderWithServiceAnnotationSuite extends ModelBuilderSuite(ModelBuilderSuite.ServiceAnnotationConfig) {
```
octonato, line 854: I wish I could have kept the three tests below under `ModelBuilderSuite` and not duplicate them, but the are slightly different from the new codegen annotation one because of the choice of descriptors.

octonato (commented):
I added a few comments here and there to explain the changes. Most files were added to because of the introspection test. The main changes are in the `ModelBuilder` and `ModelBuilderSuite`.

One important thing to note is that the entities will use a descriptor that is derived from the service descriptor and not from the domain descriptor anymore (when using the new annotation).

Previously we use the domain descriptor because it defined also the other domain messages, but the entity itself is not a proto message. It's entirely generated by us and the user can define the entity to be in any package, not necessary the same package as the domain messages. 

So instead of trying to find out where is the domain descriptor (may even be absent), the entity 'descriptor' is now a derivation from the service descriptor.

codegen/core/src/test/resources/test-files/codegen-annotation/protos/action-named/com/example/shoppingcart/shoppingcart_controller_api.proto
```diff
@@ -31,11 +28,12 @@ message NewCartCreated {
 
 service ShoppingCartAction {
   option (akkaserverless.codegen) = {
-    action: {}
+    action: {
+      name: ".controllers.ShoppingCartController"
```
johanandren, line 32: I think I missed some discussion, or maybe this is just for the tests, but, the syntax `.nested.packages.Type` is about the Java/Scala type, so that is appended to `com.example.shoppingcart` in this case (confusing how it relates to the java outer class name, I guess it is a bail out from that)? 

Shouldn't we rather use fully qualified names for these?

johanandren (approved):
LGTM but a question about the relative path name in descriptors.

codegen/core/src/test/resources/test-files/codegen-annotation/protos/action-named/com/example/shoppingcart/shoppingcart_controller_api.proto
octonato: The outer class doesn't come into play here. All components that the codegen generates are without an outer class. 

I have mixed feeling about this syntax. We were using it already for the component field before and at some point I change the templates to use a fully-qualified package. 

With the new annotation, we need to point out where we want the component to be generated but also where to find the other domain proto messages (state and events). So now, I'm more in favor of it because it will reduce some verbose. 

The resolution logic goes as follow: 

* protoPackage: com.example, name: foo.bar.Baz => class is generated as `foo.bar.Baz`
* protoPackage: com.example, name: .bar.Baz => class is generated as `com.example.bar.Baz`
* protoPackage: com.example, name: Baz => class is generated as `com.example.Baz`

The same resolution logic will be used to find out the descriptors where are the state and events messages are defined. 

We will need to document it and decide how we want the templates. I don't mind to use the short notation. 

One thing I had in mind is to have a codegen page explaining everything about the codegen. Then we could add a link to in the template to point to the documentation page. So if users were in doubt about it, they could check the docs.

PR #740, opened 2021-12-01
feat: JWT claim metadata

jroper:
This provides convenient access to JWT claim metadata.

### DIFF
--- a/sdk/java-sdk/src/main/java/com/akkaserverless/javasdk/JwtClaims.java
+++ b/sdk/java-sdk/src/main/java/com/akkaserverless/javasdk/JwtClaims.java
@@ -1,0 +1,417 @@
+/*
+ * Copyright 2021 Lightbend Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.akkaserverless.javasdk;
+
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.JsonNode;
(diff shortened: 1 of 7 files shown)


sdk/java-sdk/src/main/java/com/akkaserverless/javasdk/JwtClaims.java
```diff
@@ -0,0 +1,418 @@
...
+   * @param name The name of the claim.
+   * @return The object claim, if present. Returns empty if the claim is not an object or can't be
+   *     parsed as an object.
+   */
+  default Optional<JsonNode> getObject(String name) {
+    return getString(name)
+        .flatMap(
+            value -> {
+              try {
+                return Optional.of(new ObjectMapper().readTree(value));
```
johanandren, line 256: Not sure it matters when it is only reading the tree, but, perhaps there some value to re-using the already existing, configured object mapper in `com.akkaserverless.javasdk.JsonSupport` for this rather than a new one for every call?

sdk/java-sdk/src/test/scala/com/akkaserverless/javasdk/impl/MetadataImplSpec.scala
```diff
@@ -0,0 +1,48 @@
...
+    "support getting the subject JWT claim" in {
+      metadata("_akkasls-jwt-claim-sub" -> "some-subject").jwtClaims.subject().toScala.value shouldBe "some-subject"
+    }
+
+    "support getting the expiriation JWT claim" in {
+      metadata("_akkasls-jwt-claim-exp" -> "12345").jwtClaims.expirationTime().toScala.value shouldBe Instant
+        .ofEpochSecond(12345)
+    }
+  }
+
```
johanandren, line 41: Would be good with some coverage of the additional parsing claim methods here.

johanandren (commented):
Two small things, looks good except for those.

sdk/java-sdk/src/main/java/com/akkaserverless/javasdk/JwtClaims.java
jroper: Agreed that there is value to doing that, will change.

sdk/java-sdk/src/test/scala/com/akkaserverless/javasdk/impl/MetadataImplSpec.scala
jroper: I've added more tests, particularly for the more complex parsers.

johanandren (approved):
LGTM!

PR #746, opened 2021-12-07
chore: move quickstart pages to sdk

octonato:
Moving here some pages that are in the https://github.com/lightbend/akkaserverless-docs/, but should be here. They import files from the samples and therefore must be include in the build here. Otherwise we risk to break them easily. 

The same will need to be done on the javascript sdk. 

I will also PR the docs to remove the pages there. 

There is one thing though, the pages in the `akkaserverless-docs` have a `include::ROOT:partial$include.adoc[]` and it's not clear to me why we have this file. 
In any case, it doesn't seem to make a difference. I could build the full documentation locally and didn't notice a difference. This file is probably added already in all pages and therefore this include is probably obsolete. 

It will be good to have someone with some more experience with Antora to clarify this.

### DIFF
--- a/docs/src/modules/java/pages/quickstart/cr-value-entity-java.adoc
+++ b/docs/src/modules/java/pages/quickstart/cr-value-entity-java.adoc
@@ -1,0 +1,231 @@
+= Quickstart: Customer Registry in Java
+
+include::java:partial$attributes.adoc[]
+
+Learn how to create a customer registry in Java, package it into a container, and run it on Akka Serverless.
+
+== Before you begin
+
+* If you're new to Akka Serverless, {console}[create an account, window="console"] so you can try out Akka Serverless for free.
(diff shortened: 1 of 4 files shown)


octonato:
Oh, I see now that it's much more complicated than what I thought. I ran it locally, but without validating the docs. Those pages have links to documents that only exist in akkaserverless-docs which explain why they are there instead of in the SDK.

pvlugter:
When the SDK docs were moved here from the main docs, all xrefs needed to be converted into external links. We can do the same here. These links should be validated, so we'll find out here if the main docs change, although possible to break the links in the main docs for a while.

For Cloudstate we had a better setup, where all modules (including the main docs) were built together and could be cross-referenced.

octonato:
Superseded by #756

PR #752, opened 2021-12-13
fix: deserializing EventSourced entity events

raboof:


jrudolph:
Will be easy to create a test case for this once #753 is merged.

raboof:
> Will be easy to create a test case for this once #753 is merged.

Yes, actually I was surprised our unit test data didn't catch this since I think this should be the typical scenario...

jrudolph:
I just checked the TCK which didn't fail because it uses the same type for events and state.

jrudolph:
> Yes, actually I was surprised our unit test data didn't catch this since I think this should be the typical scenario...

The unusual part is that state and events are defined in different files.

raboof:
> The unusual part is that state and events are defined in different files.

Aaah good eyes :+1:

johanandren (approved):

Issue #761, opened 2021-12-14
Named actions breaks ComponentsImpl compilation


octonato:
The new annotation allow us to define a custom a name for an action. 

The ComponentsImpl generation picks the Action component, while it expect the Akka gRPC service.

I was adding this to a PR when I detect this issue. I will push the PR as a reproducer.

```scala 
[error] /Users/renato/GitHub/serverless/jvm-sdk/name-action/codegen/scala-gen-compilation-tests/target/scala-2.13/src_managed/main/com/example/ComponentsImpl.scala:151:80: type mismatch;
[error]  found   : com.akkaserverless.scalasdk.action.Action.Effect[com.example.action6.Response]
[error]  required: scala.concurrent.Future[com.example.action6.Response]
[error]        () => getGrpcClient(classOf[_root_.com.example.action6.MyAction]).method(command)
[error]                                                                                ^
[error] one error found
```

PR #762, opened 2021-12-14
chore: extra compile test for named action

octonato:
This is a reproducer for #761. It's broken, so I'm keeping as draft. 

This is not urgent to fix, IMO.


octonato:
Obsolete, there is a fix for this in #774

PR #753, opened 2021-12-13
chore: add round-trip tests for code-generation (Java)

jrudolph:
See the new README.md for a description.

PR #764, opened 2021-12-14
chore: extra compile tests

octonato:
This adds some extra compile time tests to exercise some new feature:

* entities (ES, Value and Replicate) can be unnamed
* views and actions can get an explicit name

However, action with explicit name is broken, see https://github.com/lightbend/akkaserverless-java-sdk/pull/762

There are some other functionality that we could also add extra tests, for instance domain classes lookup. But we have already enough test touching this indirectly. For instances, all samples use fully-qualified names and all compile time tests use relative lookups, eg: `.domain.Foo`.


octonato:
I added one more test that exercise the case where the domain is define in the same package as the service and we want the entity to be generate in the same package as well. 

For example...

```proto
    value_entity: {
      name: "User"
      entity_type: "user"
      state: "UserState"
    }
  };
```

octonato:
This can be reviewed and eventually merged, but I plan to rework it to use what Johannes added in #753.

octonato:
Obsolete, I'm moving this to ExampleSuite tests

PR #770, opened 2021-12-15
fix: remove board step from release template

raboof:
We have no 'Shipped' column anymore, tickets are
moved out of 'To be shipped' after the demo


ennru (approved):
Thanks.

PR #771, opened 2021-12-20
docs: Paging with OFFSET and LIMIT

johanandren:
Covers https://github.com/lightbend/akkaserverless-framework/pull/1043 and https://github.com/lightbend/akkaserverless-framework/pull/1049 so no need to merge until those are released and deployed.

I think the query language overview should move in the main docs and not in the SDK since it will be the same across SDKs.

### DIFF
--- a/docs/src/modules/java/pages/views.adoc
+++ b/docs/src/modules/java/pages/views.adoc
@@ -488,8 +488,6 @@ SELECT * FROM customers OFFSET :start_from LIMIT :max_customers
 
 Note: Using numeric offsets can lead to missing or duplicated entries in the result if entries are added to or removed from the view between requests for the pages.
 
-`OFFSET` and `LIMIT` cannot be used in queries containing `OR`.
-
 ==== Token based offset ====
 
 The count based offset requires that you keep track of how far you got by adding the page size to the offset for each query.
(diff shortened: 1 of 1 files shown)


docs/src/modules/java/pages/views.adoc
```diff
@@ -486,7 +486,9 @@ Or come from fields in the request message:
 SELECT * FROM customers OFFSET :start_from LIMIT :max_customers
 ----
 
-Note: `OFFSET` and `LIMIT` cannot be used in queries containing `OR`.
+Note: Using numeric offsets can lead to missing or duplicated entries in the result if entries are added to or removed from the view between requests for the pages. The token based paging handles this better.
```
raboof, line 489: Perhaps we should just describe in what cases entries are duplicated/missing without pointing to token-based paging? token-based paging is not *much* better, after all...

johanandren: Lets' remove the token based reference, but not sure I agree we should spend doc-space detailing exact scenarios.

johanandren:
Can't be merged until we transition to the new database, or needs the token based paging separated since only index based paging works with spanner.

johanandren:
Now only contains the index based paging

ennru (approved):
LGTM.

Issue #766, opened 2021-12-14
fail: flaky CustomerByEmailIntegrationTest in java-customer-registry-views


octonato:
We have seen this failing at least twice today. 

https://app.circleci.com/pipelines/github/lightbend/akkaserverless-java-sdk/2774/workflows/80ca377b-9f7a-4a67-b03d-44f90b1c5244/jobs/25803

johanandren:
https://app.circleci.com/pipelines/github/lightbend/kalix-jvm-sdk/3717/workflows/1952129a-e5a7-4131-9a52-f034d0462d7c/jobs/48436

efgpinto:
https://app.circleci.com/pipelines/github/lightbend/kalix-jvm-sdk/3731/workflows/6147fab4-1cff-4d24-ab4a-9b1c9ba81c7e/jobs/48787

PR #772, opened 2021-12-20
fix: Failing integration test #766

johanandren:
References #766

Turns out that awaiter library does not quite like Akka testkit await out of the box but will fail on exception unless explicitly asked to ignore exceptions.


octonato (approved):
LGTM

PR #775, opened 2021-12-21
chore: script to compile ExampleSuite tests

octonato:
I discovered that the existing tests for the ExampleSuite doesn't really compiled. 

So, I'm adding this script to help us compile then locally and fix. For the moment, I'm also running this script in CI. Not yet clear how we can best do this. Sbt task? 

Anyway, I just want to move fast with this for now.


octonato:
The script I added in this PR is not good. It should have failed the build, but we got a green one with the following in the logs:

```java
Build failed [/home/circleci/project/codegen/java-gen/src/test/resources/tests/action-service/with-action-in-name/proto]
-----------------------------------------------------------------
CircleCI received exit code 0
```

codegen/java-gen/src/test/resources/tests/action-service/simple-new-style/proto/example-action.proto
```diff
@@ -27,9 +27,9 @@ service MyService {
   };
 
   rpc simpleMethod(MyRequest) returns (org.external.Empty) {}
-  rpc streamedOutputMethod(MyRequest) returns (stream google.protobuf.Empty) {}
-  rpc streamedInputMethod(stream MyRequest) returns (google.protobuf.Empty) {}
-  rpc fullStreamedMethod(stream MyRequest) returns (stream google.protobuf.Empty) {}
```
octonato, line 32: This is causing a compilation error. Akka gRPC will generate a service two serializers using the same name. 
I reported it here: 
https://github.com/akka/akka-grpc/issues/1526

.circleci/config.yml
```diff
@@ -226,6 +226,21 @@ jobs:
       - publish-local
       - save_deps_cache
 
+  compile-example-suite-java: 
+    machine: true
+    description: "Check that all samples are included in CI-verification"
```
johanandren, line 231:
```suggestion
    description: "Compile Test Projects from Example Suite (Java)"
```

codegen/java-gen/src/test/resources/tests/action-service/simple-old-style/generated-managed/org/example/AkkaServerlessFactory.java
```diff
@@ -0,0 +1,23 @@
...
+
+public final class AkkaServerlessFactory {
+
+  public static AkkaServerless withComponents(
+      Function<ActionCreationContext, MyServiceAction> createMyServiceAction) {
+    AkkaServerless akkaServerless = new AkkaServerless();
+    return akkaServerless
+      .register(MyServiceActionProvider.of(createMyServiceAction));
+  }
+}
```
johanandren, line 23: These managed/generated ones, should they really be checked in, vs regenerated on build?

johanandren (approved):
LGTM but I wonder about those generated classes and if they should really be checked in.

codegen/java-gen/src/test/resources/tests/action-service/simple-old-style/generated-managed/org/example/AkkaServerlessFactory.java
octonato: That's the goal. It allow us to do first check that what we generate is indeed what we want (now we can even compile them) and they help us to see how a change in the codegen will affect the user. 

Now, we have different tests styles in the codegen and we should start to clean it up. We probably have a few that are becoming irrelevant, but I will write an issue about it later.

PR #776, opened 2021-12-22
chore: ExampleSuite tests for ES entities (new codegen)

octonato:
LIke #774, but for event sourced entities. 

Also split in three commits to make it easy to review

The first commit fix a bug related to import of events. The previous code was expecting that the events were defined in the same proto as the State, but the new annotation allow us to split them. Actually, each event can technically be on a different file. I don't think this is will affect the Scala generator because there we use the code block interpolator. 

My fix here will also become obsolete once we start to use the interpolator for java, but this would be too much work to do now, so I opt for a quick fix. 

The second commit at the protos for events sourced tests services

The third commit can be ignored, it contains the generate code.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -43,3 +43,4 @@ tags
 target
 dependency-reduced-pom.xml
 *.versionsBackup
+regenerate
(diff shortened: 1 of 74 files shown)


johanandren (approved):
LGTM

PR #780, opened 2021-12-23
chore: ExampleSuite tests for replicated entities (new codegen)

octonato:
Also with two commits: 

* protos
* generated code

Only running it for multi-map since it uses two extra domain classes (key and value)
There is also tests for scalar KV.

### DIFF
--- a/codegen/java-gen/src/test/resources/tests/replicated-entity/multimap-absolute-packages/proto/multi_map_api.proto
+++ b/codegen/java-gen/src/test/resources/tests/replicated-entity/multimap-absolute-packages/proto/multi_map_api.proto
@@ -1,0 +1,51 @@
+// Copyright 2021 Lightbend Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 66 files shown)


codegen/java-gen/src/test/resources/tests/replicated-entity/multimap-absolute-packages/proto/multi_map_api.proto
```diff
@@ -0,0 +1,51 @@
...
+}
+
+service MultiMapService {
+  option (akkaserverless.codegen) = {
+    replicated_entity: {
+      name: "com.example.replicated.multimap.domain.SomeMultiMap"
+      entity_type: "some-multi-map"
+      replicated_multi_map: {
+        key: "com.example.replicated.multimap.domain.SomeKey"
+        value: "com.example.replicated.multimap.domain.SomeValue"
```
johanandren, line 45: No version with primitive key or value?

johanandren (approved):

codegen/java-gen/src/test/resources/tests/replicated-entity/multimap-absolute-packages/proto/multi_map_api.proto
octonato: There is one, here: 
https://github.com/lightbend/akkaserverless-java-sdk/pull/780/files#diff-f95343914fb9bf1aaa9d2d6687ed719df48773ea5780b68f9a098b3e2809e8c2R38-R48

Issue #587, opened 2021-10-08
Support for testing 'forward'


raboof:
Part of #372

For both ValueEntity and EventSourcedEntity

johanandren:
Covered by `ValueEntityResult.forwardedTo` and `EventSourcedResult.forwardedTo`

PR #784, opened 2022-01-04
feat: Support side effects and forward assertion in testkits 

johanandren:
References #322 and #587

With some additional cleanup and deduplication of code in the testkit result classes


octonato (approved):
LGTM

PR #785, opened 2022-01-04
chore: move closing ']]#'

octonato:
I'm not sure about this `]]#`, but it seems that it's used to escape the `$`. I couldn't find docs about it. 

In any case, I think we should only enclose the lines that seems to need it. Having it in a comment line make it less visible for us when maintaining this code.

### DIFF
--- a/maven-java/akkaserverless-maven-archetype-event-sourced-entity/src/main/resources/archetype-resources/docker-compose.yml
+++ b/maven-java/akkaserverless-maven-archetype-event-sourced-entity/src/main/resources/archetype-resources/docker-compose.yml
@@ -9,12 +9,12 @@ services:
       - "host.docker.internal:host-gateway"
     environment:
 #[[      USER_FUNCTION_HOST: ${USER_FUNCTION_HOST:-host.docker.internal}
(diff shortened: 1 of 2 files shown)


raboof:
> I'm not sure about this `]]#`, but it seems that it's used to escape the `$`. I couldn't find docs about it.

Maven archetypes use velocity for the templating, the corresponding docs are in https://velocity.apache.org/engine/1.7/user-guide.html#literals . I agree it's not so obvious, we could add a comment there?
 
> In any case, I think we should only enclose the lines that seems to need it. Having it in a comment line make it less visible for us when maintaining this code.

:+1:

raboof (approved):

johanandren (approved):

octonato:
> I agree it's not so obvious, we could add a comment there? 

That would be good, but this is a yml file and we need to preserve indentation. That's why we have this weird formatting

```yml
    environment:
#[[      USER_FUNCTION_HOST: ${USER_FUNCTION_HOST:-host.docker.internal}
      USER_FUNCTION_PORT: ${USER_FUNCTION_PORT:-8080} ]]#
```

Adding a comment will require some extra gymnastic I think, so I prefer to leave it as such.

PR #783, opened 2022-01-04
feat: placeholder configuration for JWT dev secrets

jroper:
This adds commented out placeholders for configuring JWT dev secrets as implemented in https://github.com/lightbend/akkaserverless-framework/pull/1055.

PR #787, opened 2022-01-06
chore: wrap whole docker-compose with Velocity escaping

octonato:
References #783

### DIFF
--- a/maven-java/akkaserverless-maven-archetype-event-sourced-entity/src/main/resources/archetype-resources/docker-compose.yml
+++ b/maven-java/akkaserverless-maven-archetype-event-sourced-entity/src/main/resources/archetype-resources/docker-compose.yml
@@ -1,3 +1,4 @@
+#[[
 version: "3"
 services:
   akka-serverless-proxy:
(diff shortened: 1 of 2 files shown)


raboof (approved):
actually I wonder if it wouldn't be nice to use the velocity templating to replace the akkaserverless-proxy version, but until then no objection :+1:

octonato:
> actually I wonder if it wouldn't be nice to use the velocity templating to replace the akkaserverless-proxy version, but until then no objection

I don't think it has much value to ask user which version of the proxy they want to use. At the end, they should use what is deployed by us.

raboof:
> > actually I wonder if it wouldn't be nice to use the velocity templating to replace the akkaserverless-proxy version, but until then no objection
> 
> I don't think it has much value to ask user which version of the proxy they want to use. At the end, they should use what is deployed by us.

I agree we shouldn't ask the user, but it would be convenient if we didn't have to update this version in so many places when doing a proxy release - I figured the templating might allow us to 'inherit' it from the build somehow. Perhaps that's not so easy though - anyway not important.

octonato:
We have some scripts to that update the versions for the sdk. They are not in good shape. We have two of them and they overlap. We could review them and include the dockers as well. We will still need to run them after a release though.

PR #788, opened 2022-01-07
chore: update proxy version to 0.8.6

octonato:


raboof (approved):

raboof:
prod is still on 0.8.4, should we wait for 0.8.6 to hit before updating the sdk?

octonato:
> the proxy on prod is still on 0.8.4, should we wait for 0.8.6 to hit before updating the sdk?

Good question. 

The framework is released and the sdk as well. Maven users will get the new versions already. Sbt users don't unless we merge https://github.com/lightbend/akkaserverless-value-entity.g8/pull/16. 

For the codegen annotation is ok, as it doesn't depend on the proxy. For the new JWT functionality we do depend on the proxy, but they are not yet documented and also not announced (see https://github.com/lightbend/akkaserverless-docs/pull/1133). 

So, in principle we don't need to wait for the proxy to be deployed, but it won't hurt if we wait.

raboof:
> So, in principle we don't need to wait for the proxy to be deployed

this is only true if all changes to the proxy between 0.8.5 and 0.8.6 are forward-compatible, which we don't require, right? Have you checked they are?

octonato:
So, for clarity. 

The SDK was released yesterday and the main goal for it was to get the documentation published. The new codegen was already released, but it was a hidden feature. 

Actually, I could have wait with the release, but I wanted to publish the docs so the main docs PR that changed the quickstart pages could build and be tested in https://docs.akkaserverless.dev/. 

In this PR we update the framework version so, after merging, the users will be able to use the new JWT annotations.

octonato:
> this is only true if all changes to the proxy between 0.8.5 and 0.8.6 are forward-compatible, which we don't require, right? Have you checked they are?

Well, I think that's the case for the JWT features. If you get the new SDK (next release) and start to use the new JWT features, it will break when deploying, but the feature is hidden. 

I'm not aware of anything else that could cause an issue. Do we have a process to track this kind of situation?

raboof:
> The SDK was released yesterday

It was released with sdk 0.8.5, not 0.8.6, though, right?

> after merging, the users will be able to use the new JWT annotations.

They could, but it wouldn't work, since the deployed proxy doesn't support them yet, right?

> Do we have a process to track this kind of situation?

The process is to not update the SDK before the corresponding proxy made it to prod (https://github.com/lightbend/akkaserverless-framework/blob/master/.github/release-issue-template.md#wait-for-the-release-to-make-it-to-prod), unless you've made sure there's no problems ;). We don't have a process to make sure there's no problems, AFAiK.

octonato:
> It was released with sdk 0.8.5, not 0.8.6, though, right?

correct, and that include the codegen annotation and some of the JWT features. In 0.8.6, we have more JWT features

> They could, but it wouldn't work, since the deployed proxy doesn't support them yet, right?

Correct, but since it's 'hidden' I think the risk is minimal. We could even argue that if we don't officially announce it (release notes / documentation), then it's a non-feature.

> The process is to not update the SDK before the corresponding proxy made it to prod

Yep, seems the safest to do.

PR #793, opened 2022-01-10
chore: setup Scala ExampleSuite

jrudolph:
For now just copying test cases from the Java side (but maybe we can move the proto files for test cases to codegenCore to share a common set between Java and Scala?


octonato:
> but maybe we can move the proto files for test cases to codegenCore to share a common set between Java and Scala?

Yes!

octonato:
Let's merge this and later move the protos.

PR #795, opened 2022-01-10
[WIP] fancy way of doing compilation tests for all defined ExampleSuite examples

jrudolph:
Here's a fancy way of doing compilation testing for the generated test code of the example suite test files.

I created a new sbt build to avoid that the main build is overloaded with all those test builds. We would compile those files only from CI.

@octonato this could be an alternative to the script you mentioned before.


octonato (commented):
Yeah, I find this better than the script I wrote. I mentioned, in my original PR, the idea of creating something in sbt to compile those files, but I think this solution is simple enough.

Let's got for it!

octonato:
@jrudolph, I'm closing this one since #799 supersedes it.

PR #796, opened 2022-01-11
feat: allow POJO domain classes

octonato:
Still draft for ready for a first review round. 

This add the possibility to define the domain classes as POJOs, need more tests I think and there is some bits that I find hacky, but good as a first stab, IMO.


.circleci/config.yml
```diff
@@ -526,6 +526,29 @@ jobs:
             rm -rf src/main/java src/test/java src/it/java
             mvn -Dakkaserverless-sdk.version=$SDK_VERSION test-compile
 
+
+  sample-java-eventsourced-counter-pojo:
```
octonato, line 530: I duplicated the counter samples because I wanted to have at least something to run integration tests and prove that the serialization is working as expected. 

Needed duplication because domain proto (that is deleted in this sample) is used in the docs.

codegen/core/src/main/scala/com/lightbend/akkasls/codegen/ModelBuilder.scala
```diff
@@ -484,11 +488,32 @@ object ModelBuilder {
...
+        PackageNaming(
+          protoFileName = "",
+          name = className,
+          protoPackage = "",
+          javaPackageOption = Some(packageName),
+          javaOuterClassnameOption = None,
+          javaMultipleFiles = true),
+        None)
+    fqn.copy(descriptorObject = Some(fqn))
+  }
```
octonato, line 515: This is the hacky part. We need a `FullyQualifiedName` that is build from a simple package and class name. But our  `FullyQualifiedName` is very proto oriented, hence the `protoName = ""` etc.

codegen/core/src/main/scala/com/lightbend/akkasls/codegen/Packaging.scala
```diff
@@ -32,6 +32,9 @@ case class FullyQualifiedName(
     parent: PackageNaming,
     descriptorObject: Option[FullyQualifiedName]) {
 
+  // FIXME: quick workaround for now. We may want to model this in a better way.
+  val isProtoMessage: Boolean = protoName.nonEmpty
```
octonato, line 36: yeah, let make it more hacky. :-)

We can find a better solution for that, but I wanted to have a confirmation that the POJO idea can fly first.

codegen/java-gen/src/main/scala/com/lightbend/akkasls/codegen/java/ActionServiceSourceGenerator.scala
```diff
@@ -293,10 +293,12 @@ object ActionServiceSourceGenerator {
...
 
+    val relevantDescriptors =
+      collectRelevantTypes(service.commandTypes, service.fqn)
+        .collect { case fqn if fqn.isProtoMessage => s"${fqn.parent.javaOuterClassname}.getDescriptor()" }
+
     val descriptors =
-      (collectRelevantTypes(service.commandTypes, service.fqn)
-        .map(d =>
-          s"${d.parent.javaOuterClassname}.getDescriptor()") :+ s"${service.fqn.parent.javaOuterClassname}.getDescriptor()").distinct.sorted
+      (relevantDescriptors :+ s"${service.fqn.parent.javaOuterClassname}.getDescriptor()").distinct.sorted
```
octonato, line 301: relevantDescriptors are used to generate the `additionalDescriptors` in the provide, but we don't want the POJOs in this list.

sdk/java-sdk/src/main/scala/com/akkaserverless/javasdk/impl/AnySupport.scala
```diff
@@ -389,8 +389,13 @@ class AnySupport(
...
 
       case other =>
-        throw SerializationException(
-          s"Don't know how to serialize object of type ${other.getClass}. Try passing a protobuf, using a primitive type, or using a type annotated with @Jsonable.")
+        try {
+          ScalaPbAny.fromJavaProto(JsonSupport.encodeJson(value))
+        } catch {
+          case _: IllegalArgumentException =>
+            throw SerializationException(
+              s"Don't know how to serialize object of type ${value.getClass}. Try passing a protobuf, using a primitive type, or using a type annotated with @Jsonable.")
```
octonato, line 397: `IllegalArgumentException` is thrown from `encodeJson` if it can't serialize it, but maybe we want to catch all exceptions here.

octonato (commented):
~The first three commits (616a17ad728ed45e69f39eff6bd786f9fada74aa, 0b599fcee062217ea0c870281e8b07e76801fe14 and a605fb643ea6c1d953cc762e142c5a04abd0e889) are the most relevant. ~

Updated, see -> https://github.com/lightbend/akkaserverless-java-sdk/pull/796#issuecomment-1010853785

octonato:
I did a force push to re-organise the commits. The first commit 2f9267f contains the feature and the rest is about the tests. That should make the review easier. 

The commit for the samples shows how it look like when modeling using POJOs. The business logic can then move from the component (which is more like a service) to the domain class itself.

samples/java-eventsourced-counter-pojo/src/main/java/org/example/domain/CounterState.java
```diff
@@ -0,0 +1,29 @@
...
+  public final int value;
+
+  @JsonCreator
+  public CounterState(@JsonProperty("value") int value) {
+    this.value = value;
+  }
+  
+  public static CounterState defaultInstance() {
+    return new CounterState(0);
+  }
```
johanandren, line 17: This is just mimicing the generated proto Javas, maybe just instantiate with 0 in the entity instead of this?

And also, if we really want to appeal to Java devs, perhaps the class should be mutable? I _think_ that should be safe since there is a single writer (but maybe we do future juggling in the  component impls so it isnt)

samples/java-valueentity-counter-pojo/src/test/java/com/example/domain/CounterTest.java
```diff
@@ -0,0 +1,74 @@
...
+  @Test
+  public void increaseCounterTest() {
+    CounterTestKit testKit = CounterTestKit.of(Counter::new);
+
+    CounterApi.IncreaseValue increaseValueCommand = CounterApi.IncreaseValue.newBuilder()
+        .setValue(1)
+        .build();
+    ValueEntityResult<Empty> result1 = testKit.increase(increaseValueCommand);
+    assertEquals(Empty.getDefaultInstance(), result1.getReply());
+    assertEquals(1, testKit.getState().value);
```
johanandren, line 26: The testkit just works out of the box? Neat.

sdk/java-sdk/src/main/scala/com/akkaserverless/javasdk/impl/AnySupport.scala
```diff
@@ -389,8 +389,13 @@ class AnySupport(
         ScalaPbAny(BytesPrimitive.fullName, primitiveToBytes(BytesPrimitive, byteString))
 
       case other =>
-        throw SerializationException(
-          s"Don't know how to serialize object of type ${other.getClass}. Try passing a protobuf, using a primitive type, or using a type annotated with @Jsonable.")
+        try {
+          ScalaPbAny.fromJavaProto(JsonSupport.encodeJson(value))
```
johanandren, line 393: Is there a risk that this accidentally json encodes something a user didn't mean to, given that it is the same code path for any response/forwarded/side effect message from any component? 

Should we narrow that down with some additional parameter/flag for AnySupport when serializing and only flag that for entity state?

sdk/java-sdk/src/main/scala/com/akkaserverless/javasdk/impl/AnySupport.scala
```diff
@@ -389,8 +389,13 @@ class AnySupport(
...
+        } catch {
+          case _: IllegalArgumentException =>
+            throw SerializationException(
+              s"Don't know how to serialize object of type ${value.getClass}. Try passing a protobuf, using a primitive type, or using a type annotated with @Jsonable.")
+        }
```
johanandren, line 398: I'm not seeing deserialization anywhere in this PR, how is json from the proxy turned into the pojo instances when starting the entity (and replaying events)?

octonato: I don't know! 🤷 
This still need to be verified. 

Seems to work or the code path is not being exercised. Most probably the latter. 
The integration test start from scratch and the entity stays in memory during the test so there is not replay or load of any state. That's probably why it works. 

I'm aware of this and will investigate this further.

samples/java-valueentity-counter-pojo/src/test/java/com/example/domain/CounterTest.java
octonato: Yep! They are just java types.

samples/java-eventsourced-counter-pojo/src/main/java/org/example/domain/CounterState.java
octonato: For the moment the samples are intended for us only. It will take some time before we consider that the POJO style is the one we want to promote the most. We can review this later. 

I personally think it's ok to promote immutable classes. I've seen immutable classes showing up in Java blog posts. Not saying that this is the most common idiom in Java world, but it's not as exotic as it used to be.

sdk/java-sdk/src/main/scala/com/akkaserverless/javasdk/impl/AnySupport.scala
octonato: Indeed, I think we should drive it somehow. 

Preferably not with an extra annotation (like the old `@Jsonable`), but let the registration passes something down the road. 

One thing I had in mind is to introduce a `Serializer` that is then registered by the provider. By default it would be protobuf, for POJOs we switch to Jackson. The user could choose something else as well. I'm think on Scala devs that would want something else, like play-json or circe. 

Anyway, this need to mature.

octonato:
I put this back to draft as I think that is far to be ready for merging. That's clear. :-)

samples/java-eventsourced-counter-pojo/src/main/java/org/example/domain/PojoCounterProvider.java
```diff
@@ -0,0 +1,86 @@
...
+  }
+
+
+  @Override
+  public Serializer serializer() {
+    return new CompositeSerializer(new JsonSerializer(CounterState.class))
+            .add(new JsonSerializer(ValueIncreased.class))
+            .add(new JsonSerializer(ValueDecreased.class))
+            .add(new JsonSerializer(ValueReset.class));
+  }
```
octonato, line 77: This is what it would look like when using POJOs. 

Users could overwrite this if they want and add some extra serialization/deserfialization logic. I'm thinking on Scala developers that may choose not to use Jackson but a more idiomatic Scala json library.

sdk/java-sdk/src/main/java/com/akkaserverless/javasdk/valueentity/ValueEntityProvider.java
```diff
@@ -35,4 +36,8 @@
   ValueEntityRouter<S, E> newRouter(ValueEntityContext context);
 
   Descriptors.FileDescriptor[] additionalDescriptors();
+
+  default Serializer serializer() {
+    return Serializer.noopSerializer();
+  }
```
octonato, line 42: The Provider defines the Serializer to use. This is then passed to `AnySupport` and used before anything. 

`noopSeriazlier` has no effect and therefore `AnySupport` will just use protobuf.

octonato: As a side idea, `AnySupport` could also be defined as a `Serializer`, but that's not in scope.

sdk/java-sdk/src/main/scala/com/akkaserverless/javasdk/impl/Serializer.scala
```diff
@@ -0,0 +1,90 @@
...
+
+    override def canDeserialize(any: JavaPbAny): Boolean = false
+
+    override def deserialize(any: JavaPbAny): Any =
+      throw new IllegalArgumentException("NoopSerializer can not deserialize types")
+
+  }
+
+}
+trait Serializer {
```
octonato, line 41: I originally called it a Codec because the JsonSupport and AnySupport has methods like `encode` / `decode`, but after looking at it once more I thought that it's more clear to users if we just call it a `Serializer`. 

Users are not expected to implement `Serializers` though, but they can if they want. Actually, they could use any format now, not only Json. 

Which makes me think that we should default to cbor instead. 

We could expand the annotation and let the user chose some pre-defined serialization formats, `cbor`, `json` or `custom`.  Where `custom` means, we don't generate anything and the user will need to write their serializers and configure the provider accordingly.

sdk/java-sdk/src/main/scala/com/akkaserverless/javasdk/impl/Serializer.scala
```diff
@@ -0,0 +1,90 @@
...
+  def canSerialize(any: Any): Boolean
+  def serialize(any: Any): JavaPbAny
+
+  def canDeserialize(any: JavaPbAny): Boolean
+  def deserialize(any: JavaPbAny): Any
```
octonato, line 46: Could have define to types, but then we will have to provide `Ser` and `Deser` while there isn't a case where only one should be provided.

sdk/java-sdk/src/main/scala/com/akkaserverless/javasdk/impl/Serializer.scala
```diff
@@ -0,0 +1,90 @@
...
+    any.getTypeUrl.contains(cls.getName)
+
+  override def deserialize(any: JavaPbAny): Any =
+    JsonSupport.decodeJson(cls, any)
+
+}
+
+object CompositeSerializer {
+  def add(serializer: Serializer): CompositeSerializer = new CompositeSerializer(Seq(serializer))
+}
```
octonato, line 66: This is odd. I can't call this method from Java while I can call `Serializer.noopSerializer()`. 

Maybe because there is a `class CompositeSerializer` while `Serializer` has only an `object` and a `trait`?

octonato (commented):
Added some comments to explain some points and eventually trigger some extra thoughts.

jroper:
I like the pluggable `Serializer` approach in `AnySupport`. Looks good to me.

octonato:
Ok, one more round. I'm marking it now as ready for review. 

I reorganise the commits to facilitate the review. 

* commit (b1b8895) contains only the feature logic
* commit (43bd1ec) adds two new samples using POJO domains
* commit (9bd56ca) is the massive regenerated of the ExampleSuite tests

Still missing a the Scala counterpart, but I will add it on a follow-up PR to avoid make this one too big. 

I'm also working on an improvement of the intermediate AST for FullyQualifiedName (see the hack I introduced to identity POJO types). This will be gone in follow-up PR. I don't want to add even more code here.

sdk/java-sdk/src/main/scala/com/akkaserverless/javasdk/impl/Serializer.scala
```diff
@@ -0,0 +1,95 @@
...
+
+  def add(serializer: Serializer): CompositeSerializer =
+    new CompositeSerializer(serializers :+ serializer)
+
+  override def canSerialize(any: Any): Boolean = serializers.exists(_.canSerialize(any))
+
+  override def serialize(any: Any): JavaPbAny =
+    serializers
+      .find(_.canSerialize(any))
+      .map(_.serialize(any))
```
patriknw, line 85: find+map = collectFirst

sdk/java-sdk/src/main/scala/com/akkaserverless/javasdk/impl/Serializer.scala
```diff
@@ -0,0 +1,95 @@
...
+      .getOrElse(throw new IllegalArgumentException(s"No serializer found for $any"))
+
+  override def canDeserialize(any: JavaPbAny): Boolean = serializers.exists(_.canDeserialize(any))
+
+  override def deserialize(any: JavaPbAny): Any =
+    serializers
+      .find(_.canDeserialize(any))
+      .map(_.deserialize(any))
+      .getOrElse(throw new IllegalArgumentException(s"No deserializer found for $any"))
```
patriknw, line 94: don't `toString` of unknown objects. can be large or could contain sensitive data that shouldn't be included in logs

patriknw (commented):
first commit looking mostly good

samples/java-eventsourced-counter-pojo/src/main/java/org/example/domain/CounterState.java
```diff
@@ -0,0 +1,29 @@
...
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+public class CounterState {
+
+  public final int value;
+
+  @JsonCreator
+  public CounterState(@JsonProperty("value") int value) {
```
patriknw, line 11: JsonProperty shouldn't be needed. Maybe you haven't added the `-parameters` Java compiler option?

samples/java-eventsourced-counter-pojo/src/main/java/org/example/domain/CounterState.java
```diff
@@ -0,0 +1,29 @@
...
+  }
+
+  public CounterState increase ( int value){
+    if (value <= 0) throw new IllegalArgumentException("decrease value must be a positive number");
+    return new CounterState(this.value + value);
+  }
+
+  public CounterState decrease ( int value){
+    if (value >= 0) throw new IllegalArgumentException("decrease value must be a negative number");
+    return new CounterState(this.value + value);
```
patriknw, line 26: `- value` ?

samples/java-eventsourced-counter-pojo/src/main/java/org/example/domain/ValueDecreased.java
```diff
@@ -0,0 +1,17 @@
+package org.example.domain;
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+public class ValueDecreased {
+  public final int value;
+
+  @JsonCreator
+  public ValueDecreased(@JsonProperty("value") int value) {
```
patriknw, line 10: ^

samples/java-eventsourced-counter-pojo/src/main/java/org/example/domain/ValueIncreased.java
```diff
@@ -0,0 +1,17 @@
+package org.example.domain;
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+public class ValueIncreased {
+  public final int value;
+
+  @JsonCreator
+  public ValueIncreased(@JsonProperty("value") int value) {
```
patriknw, line 10: ^

samples/java-eventsourced-counter-pojo/src/main/java/org/example/domain/ValueReset.java
```diff
@@ -0,0 +1,13 @@
+package org.example.domain;
+
+import com.fasterxml.jackson.annotation.JsonTypeName;
+
+@JsonTypeName("reset")
+public class ValueReset {
+
+  private static final ValueReset instance =  new ValueReset();
```
patriknw, line 8: singleton instances are better as `enum` in Java, or is there a problem with json serialization of that?

samples/java-eventsourced-counter-pojo/src/main/java/org/example/domain/ValueReset.java
```diff
@@ -0,0 +1,13 @@
+package org.example.domain;
+
+import com.fasterxml.jackson.annotation.JsonTypeName;
+
+@JsonTypeName("reset")
```
patriknw, line 5: is this needed?

samples/java-valueentity-counter-pojo/src/main/proto/com/example/counter_api.proto
```diff
@@ -0,0 +1,50 @@
...
+}
+
+service CounterService { // <7>
+  option (akkaserverless.codegen) = { // <8>
+    value_entity: { // <9>
+      name: "com.example.domain.Counter" // <10>
+      entity_type: "counter" // <11>
+
+      // NOTE: domain classed are in org.example because 
+      // the CI tests remove the generated classes in src/java/main/com
```
patriknw, line 41: That would be somewhat confusing if we use this sample in docs or as user example. Can we tweak CI?

patriknw (commented):
commit two

samples/java-eventsourced-counter-pojo/src/main/java/org/example/domain/CounterState.java
octonato: `+ value` is correct. The passed value must be a negative value so `this.value + (- value)`. Either that, or the other way around.

sdk/java-sdk/src/main/scala/com/akkaserverless/javasdk/impl/Serializer.scala
octonato: sure

samples/java-eventsourced-counter-pojo/src/main/java/org/example/domain/ValueReset.java
octonato: I didn't with enums. I will give it a try.

samples/java-eventsourced-counter-pojo/src/main/java/org/example/domain/ValueReset.java
octonato: Yes, it's needed if we use an object. Since there is now data, the payload is serialized to `{}`. When it comes back we need to know how to deserialize it back, so we need a type hint. 

But maybe not needed if we use enums.

samples/java-valueentity-counter-pojo/src/main/proto/com/example/counter_api.proto
octonato: Yep, we can change it. We can remove the Main and the Entity explicitly instead of deleting the parent dir.

samples/java-eventsourced-counter-pojo/src/main/java/org/example/domain/CounterState.java
octonato: I will try this out.

samples/java-eventsourced-counter-pojo/src/main/java/org/example/domain/CounterState.java
patriknw: strange, then I would have expected increase(-5) instead of having both increase and decrease

sdk/java-sdk/src/main/scala/com/akkaserverless/javasdk/impl/Serializer.scala
johanandren: I don't think we should open up for generic serialization, or support multiple formats, instead we should make sure that the format we _do_ support (`json`) works seamless throughout Akka Serverless, for example using the `json.akkaserverless.com` type url for the Any's and then special-handling that where needed in the proxy and SDKs.

samples/java-eventsourced-counter-pojo/src/main/java/org/example/domain/CounterState.java
octonato: Yeah, there are a few different ways to implement it. I just followed what we have done in the original sample.

samples/java-eventsourced-counter-pojo/src/main/java/org/example/domain/CounterState.java
octonato: This is removed now. But I couldn't remove the `@JsonCreator` one, only `@JsonProperty`.

sdk/java-sdk/src/main/scala/com/akkaserverless/javasdk/impl/Serializer.scala
octonato: Agree that a custom serialization may not be needed and even complicated things. That was not my intention to add it right now anyway. 

I do think that cbor can be interesting eventually since we do that in akka as well, but again, nothing that we should care about now.

samples/java-eventsourced-counter-pojo/src/main/java/org/example/domain/CounterState.java
octonato: removed the `defaultinstance()` and using default constructor now.

codegen/core/src/main/scala/com/lightbend/akkasls/codegen/ModelBuilder.scala
octonato: I'm working on another PR to fix this. I will send it on top of this one.

samples/java-eventsourced-counter-pojo/src/main/java/org/example/domain/CounterState.java
octonato: So, it seems that it's needed when the constructor has only one argument which is the case in all the example here. TIL

patriknw: Yes, single arg constructor requires JsonCreator

codegen/java-gen/src/test/resources/tests/replicated-entity/multimap-unnamed-new-style/generated-managed/com/example/replicated/multimap/MultiMapServiceEntityProvider.java
```diff
@@ -70,4 +71,9 @@ public final MultiMapServiceEntityRouter newRouter(ReplicatedEntityContext conte
       SomeMultiMapDomain.getDescriptor()
     };
   }
+  
+  @Override
+  public Serializer serializer() { 
+    return Serializer.noopSerializer();
+  }
```
johanandren, line 78: Seems noisy that the default empty/no serializer must be in each generated class, can't there be a default impl be in the base Provider class/trait instead? (And only generate override when needed)

codegen/java-gen/src/main/scala/com/lightbend/akkasls/codegen/java/ActionServiceSourceGenerator.scala
johanandren: Where would the POJO come from for an Action, it doesn't have state?

codegen/core/src/main/scala/com/lightbend/akkasls/codegen/ModelBuilder.scala
```diff
@@ -484,11 +488,32 @@ object ModelBuilder {
...
     else {
-      val descriptor = lookupDescriptor(revolvedPackage, resolvedName, additionalDescriptors)
-      MessageTypeArgument(resolveFullyQualifiedMessageType(resolvedName, descriptor, additionalDescriptors))
+      lookupDomainDescriptor(revolvedPackage, resolvedName, additionalDescriptors)
+        .map { descriptor =>
+          MessageTypeArgument(resolveFullyQualifiedMessageType(resolvedName, descriptor, additionalDescriptors))
+        }
+        .getOrElse {
+          MessageTypeArgument(buildFullyQualifiedNameForPojos(revolvedPackage, resolvedName))
+        }
```
johanandren, line 497: So if I read this right, if a user misspells the stringly typed state annotation so it does not match a message, it will silently continue and expect it to be a POJO, wouldn't it be better with something more explicit in the annotations so that we can know the intent of the user and fail accordingly instead?

sdk/java-sdk/src/main/scala/com/akkaserverless/javasdk/impl/Serializer.scala
```diff
@@ -0,0 +1,97 @@
...
+
+final class JsonSerializer(cls: Class[_]) extends Serializer {
+  override def canSerialize(any: Any): Boolean =
+    cls == any.getClass
+
+  override def serialize(any: Any): JavaPbAny =
+    JsonSupport.encodeJson(any)
+
+  override def canDeserialize(any: JavaPbAny): Boolean =
+    any.getTypeUrl.contains(cls.getName)
```
johanandren, line 62: Should be explicitly `json.akkaserverless.com/` or `json.akkaserverless.com/${cls.getName}` depending on how strict we want to be. 

If it doesn't have our custom type prefix calling `decodeJson` will later fail.

sdk/java-sdk/src/main/scala/com/akkaserverless/javasdk/impl/AnySupport.scala
```diff
@@ -195,7 +196,8 @@ class AnySupport(
     descriptors: Array[Descriptors.FileDescriptor],
     classLoader: ClassLoader,
     typeUrlPrefix: String = AnySupport.DefaultTypeUrlPrefix,
-    prefer: AnySupport.Prefer = AnySupport.Prefer.Java) {
+    prefer: AnySupport.Prefer = AnySupport.Prefer.Java,
+    serializer: Serializer) {
```
johanandren, line 200: Maybe `customSerializer` or something conveying that it is not usually used?

sdk/java-sdk/src/main/scala/com/akkaserverless/javasdk/impl/AnySupport.scala
```diff
@@ -379,7 +382,7 @@ class AnySupport(
 
       case null =>
         throw SerializationException(
-          s"Don't know how to serialize object of type null. Try passing a protobuf, using a primitive type, or using a type annotated with @Jsonable.")
+          s"Don't know how to serialize object of type null. Try passing a protobuf, using a primitive type, or defining a codec for this type.")
```
johanandren, line 385: Except that users cant "define a codec" since it is internal API, so not a great error message? Also this is if the _value_ was null, would fail regardless, so just a NPE could suffice?

sdk/java-sdk/src/main/scala/com/akkaserverless/javasdk/impl/AnySupport.scala
```diff
@@ -390,7 +393,7 @@ class AnySupport(
 
       case other =>
         throw SerializationException(
-          s"Don't know how to serialize object of type ${other.getClass}. Try passing a protobuf, using a primitive type, or using a type annotated with @Jsonable.")
+          s"Don't know how to serialize object of type ${other.getClass}. Try passing a protobuf, using a primitive type, or defining a codec for this type.")
```
johanandren, line 396: And same here, no public API for defining "codec", maybe should just say `"Don't know how to serialize object of type ${other.getClass}."` without the recommendation instead?

sdk/java-sdk/src/main/scala/com/akkaserverless/javasdk/impl/AnySupport.scala
```diff
@@ -421,7 +424,11 @@ class AnySupport(
    */
   def decodeMessage(any: ScalaPbAny): Any = {
     val typeUrl = any.typeUrl
-    if (typeUrl.equals(BytesPrimitive.fullName)) {
+    val javaPbAny = ScalaPbAny.toJavaProto(any)
+
+    if (serializer.canDeserialize(javaPbAny)) {
+      serializer.deserialize(javaPbAny)
```
johanandren, line 430: This `canDeserialize` call is now applied once for every incoming request, event even though the feature is only for entity state (unless I missed some logic somewhere).

I think it would be good if it was more explicitly toggled at call site, something like `decodeMessage(any, supportJson = true)`.

johanandren (commented):
I understand that we want the capability of using POJOs but I really dislike this way of achieving it and think that it is bad DX.

 * SDK users will anyway have to learn protobuf to define the protocol, just that now they will have to keep in mind that there is one part that works a bit different, with limitations and consequences that they will have to keep in mind (subscribing to events and updates, views, backoffice inspection)
 * JSON as a storage format for the state means we will have to implement version migration tooling, because state will be long lived and users will want to change their POJOs

I would much more have wanted to see some interop with the descriptors where we would transform between a user defined POJO and the storage format defined in protobuf (either through reflection or some custom Jackson protobuf backend). Something that could also allow POJOs for request/response giving a much more consistent user experience.

If we are sure this is what we want to do the implementation is ok (with some feedback added), but I'd wish that we take a step back and reconsider if this is a change to the Java SDK we really want.

codegen/java-gen/src/test/resources/tests/replicated-entity/multimap-unnamed-new-style/generated-managed/com/example/replicated/multimap/MultiMapServiceEntityProvider.java
octonato: Agree. I started with a default impl before I had the generators, but then moved for not clear reason (that I remember). I will bring it back.

sdk/java-sdk/src/main/scala/com/akkaserverless/javasdk/impl/Serializer.scala
octonato: Yes, indeed better. I will do it.

codegen/core/src/main/scala/com/lightbend/akkasls/codegen/ModelBuilder.scala
octonato: Yes, I think so. We can add something, but I prefer to add it later. I don't want now to go the full cycle of changing annotations in the framework. 

For the moment, this PR is more important to get the POJO basics out. We can then fine tune it on other iterations.

sdk/java-sdk/src/main/scala/com/akkaserverless/javasdk/impl/AnySupport.scala
octonato: In other places I called it `additionalSerializer`. I think I missed to rename here.

sdk/java-sdk/src/main/scala/com/akkaserverless/javasdk/impl/AnySupport.scala
octonato: Will fix.

codegen/core/src/main/scala/com/lightbend/akkasls/codegen/ModelBuilder.scala
johanandren: Could be done without any changes to the annotations, with some in-string prefix perhaps? `state: "java:MyState"` or `state: "pojo:MyState"`. That way it would also nicely fail if you tried to just move it to some other SDK/codegen.

octonato: hmm, that's sound interesting

I will experiment with it, but in another PR. This in one is getting too big.

octonato:
@johanandren, I'm ok to not merge this for now and keep it as a feature branch. 
We merge other PRs here and we go until we get it in the shape we want. 

I will apply your suggestions and squash the commit. And the we can consider that the squashed commit is good and don't need to be reviewed again.

octonato:
@johanandren, this is squashed now. I didn't apply your suggestion because it will impact #818 and #821. 

I will send a new PR on top of #821 with your change requests. 

That said, this PR is now the feature branch.

octonato:
This is back to the fridge, so marking it as draft. 

However, there quite some other improvements here (eg: ClassMessageType) that I don't want to miss, so I will bring it to another PR.

Issue #781, opened 2021-12-23
Testing strategies (take 2)


octonato:
We have discussed it previously in #56 and since them the state of testing in the SDK improved a lot. 

Time has passed and we have new kinds of tests and maybe some of then have become obsolete. 

This is a list of different types of tests we have. 

### Codegen Output

In this tests we use some test models (ModelBuilder) to directly test the output of the generator. 
It has a very fast turnaround when changing anything in the emitted sources.

**Problem**
There is no guarantee that the emitted code is compilable or not. 
Those are heavy to write tests and therefore we don't exercise all the possible combinations.

### Compilation Test 

We added those tests exactly to solve the shortcomings of the "Codegen Output Tests". 
In this tests we take a bunch of proto files in one single project and we compile everything. The same files are used for Java and Scala and once again by the sbt-plugin tests (for Scala only).

Those are easy to write tests. We only need to think how the proto should look like, make sure that there is no naming conflicts and we are ready to go.

**Problem**
We verify that the code compiles, but **not how the files look like**. Did the component is named correctly? Do we have unused imports? The ExampleSuite tests solved that problem.

### AST test - ModelBuilderSuite

For this tests we have a few files in `codegen/core/src/test/resources/test-files` that are used to generate protobuf .desc files. 
The .desc files are feeded to the ModelBuilder and we verify that the produced AST is what we expect. 

**Problem**
They are hard to maintain and have very low value. I don't care much how the AST looks like, the only thing I care is that given an proto file as input, I can generate code that is compilable and do what I expect. How we build an AST to get there is irrelevant, IMO.

### ExampleSuite - (java only)

New style of tests. Here we take proto files and generate code from them. We can then review the generated code and commit it.
On next run, we generate the same files again and if there are differences the test fails. That allow us to change the codegen and immediately see if the generated code is impact. 

Since #775, they are also compiled during tests, but needs an special shell script and requires publishing maven archetype locally which make them a little heavy when we are hacking the codegen. 

### Samples

In the samples we have full project with everything, including integration tests. 

## Suggestions

I think we should drop the AST tests. Too much work, low value. 

The "**Codegen Output**" could be replaced by ExampleSuite. We should read real proto files instead of a handcrafted test model and emit code that can be compared to an expect files. I would do this per generator, for instance, pick an Action proto file, generate the ActionProvider and verify. We don't need to have the `regenerate` trick since we will be testing each file in isolation. It's easier to just write the file and edit it until the test passes. 

[...]

PR #805, opened 2022-01-17
chore: removes redundant introspection tests

octonato:
This removes the introspection tests. These tests compares the structure of the intermediate AST generated by the ModelBuilder. They are annoying to maintain and have the ExampleSuite tests are doing a better job in validating the generated code.

References #781


octonato:
This PR is now fixed. I removed even more tests. 

I don't think they are adding extra protection because ExampleSuite are now basically covering those cases as well. 

What we could eventually do is to have a variation of ExampleSuite that takes proto files (not pre-compiled descriptors) and returns AST for us to inspect. That could be useful eventually. We can re-evaluate it later.

johanandren (approved):

PR #807, opened 2022-01-18
chore: abstract away ExampleSuite compilation project

octonato:


build.sbt
```diff
@@ -332,51 +332,22 @@ lazy val codegenScalaCompilationTest = project
...
-          .enablePlugins(ProtocPlugin, AkkaGrpcPlugin)
-      }
-    }
-  } else {
-    ExampleSuiteCompilationProject.empty
+  ExampleSuiteCompilationProject.compilationProject("java", "codegen/java-gen/src/test/resources/tests") {
+    testProject =>
+      testProject
+        .dependsOn(sdkJava % "compile", sdkJavaTestKit % "test")
+        .settings(akkaGrpcGeneratedLanguages := Seq(AkkaGrpc.Java))
```
octonato, line 339: to build a compilation project, we need three things: 

* the lang so we can use it as label to name / prefix things
* the path to the tests - (I didn't want to do concatenation on this to make it clear in the build.sbt file which path we are using
* a function `Project => Project` to configure the project with some specifics

project/ExampleSuiteCompilationProject.scala
```diff
@@ -1,12 +1,58 @@
...
+
+  def compilationProject(lang: String, pathToTests: String)(configureFunc: Project => Project) = {
     new CompositeProject {
-      def componentProjects = Seq.empty[Project]
+
+      lazy val root =
+        Project(id = s"codegen${lang.capitalize}CompilationExampleSuite", base = file(pathToTests))
+          .aggregate(innerProjects.map(p => p: ProjectReference): _*)
+
+      val innerProjects =
```
octonato, line 19: hmm, this should be lazy. I will fix it.

project/ExampleSuiteCompilationProject.scala
```diff
@@ -1,12 +1,58 @@
...
+              .disablePlugins(HeaderPlugin, ScalafmtPlugin, JavaFormatterPlugin)
+              .settings(
+                Compile / unmanagedSourceDirectories ++= Seq("generated-managed", "generated-unmanaged").map(
+                  baseDirectory.value / _),
+                Test / unmanagedSourceDirectories ++= Seq("generated-test-managed", "generated-test-unmanaged").map(
+                  baseDirectory.value / _),
+                Compile / PB.protoSources += baseDirectory.value / "proto")
+              .enablePlugins(ProtocPlugin, AkkaGrpcPlugin)
+          }
+          .map(configureFunc) // apply specific settings
```
octonato, line 32: two `.map` to make it more obvious we were applying this function. Anyway, just cosmetics.

octonato (commented):
@jrudolph, as discussed, here is a small PR to abstract away the building of a compilation project.

WDYT?

johanandren (approved):
Only quick-skimmed but LGTM as far as I can see.

jrudolph (commented):
LGTM

PR #808, opened 2022-01-19
chore: trigger ExampleSuite regeneration via -D property

octonato:
References #781 

With this PR we may be able to delete the codegen String comparison tests. 
It makes it a little easier to use the ExampleSuite + compilation as the base of our test strategy. 
By passing some flags to sbt we can force the regeneration of some files and then compile them. 

There is a little bit more of gymnastic is required when changing the codegen though. 

The updated README files contain some explanation on how we can use it to have a better dev roundtrip.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -217,6 +217,7 @@ lazy val codegenJava =
     .dependsOn(codegenCore % "compile->compile;test->test")
     .enablePlugins(PublishSonatype)
     .settings(
+      Test / fork := false, // needed to pass -D properties to ExampleSuite
       // to provide access to protoc to tests
       Test / buildInfoPackage := "com.lightbend.akkasls.codegen.java",
       Test / buildInfoKeys := Seq(
@@ -299,6 +300,7 @@ lazy val codegenScala =
     .settings(
       name := "akkaserverless-codegen-scala",
       scalaVersion := Dependencies.ScalaVersionForSbtPlugin,
+      Test / fork := false, // needed to pass -D properties to ExampleSuite
       buildInfoKeys := Seq[BuildInfoKey](
         name,
         organization,
--- a/codegen/core/src/main/scala/com/lightbend/akkasls/codegen/File.scala
+++ b/codegen/core/src/main/scala/com/lightbend/akkasls/codegen/File.scala
@@ -27,6 +27,7 @@ case class GeneratedFiles(
     managedTestFiles: Seq[File],
     unmanagedTestFiles: Seq[File],
     integrationTestFiles: Seq[File]) {
+
   def write(
       managedSourceDirectory: Path,
       unmanagedSourceDirectory: Path,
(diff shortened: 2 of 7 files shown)


build.sbt
```diff
@@ -217,6 +217,7 @@ lazy val codegenJava =
     .dependsOn(codegenCore % "compile->compile;test->test")
     .enablePlugins(PublishSonatype)
     .settings(
+      Test / fork := false, // needed to pass -D properties to ExampleSuite
```
octonato, line 220: if we find it's worth, we can keep forking and pass the -D properties down the road. I didn't find it was worth though, but happy to change my mind.

codegen/core/src/main/scala/com/lightbend/akkasls/codegen/File.scala
```diff
@@ -39,6 +40,18 @@ case class GeneratedFiles(
...
+      managedSourceDirectory: Path,
+      unmanagedSourceDirectory: Path,
+      managedTestSourceDirectory: Path,
+      unmanagedTestSourceDirectory: Path,
+      unmanagedIntegrationTestSourceDirectory: Path): Iterable[Path] =
+    managedFiles.map(_.writeToDirectory(managedSourceDirectory, onlyIfMissing = false)) ++
+    unmanagedFiles.map(_.writeToDirectory(unmanagedSourceDirectory, onlyIfMissing = false)) ++
+    managedTestFiles.map(_.writeToDirectory(managedTestSourceDirectory, onlyIfMissing = false)) ++
+    unmanagedTestFiles.map(_.writeToDirectory(unmanagedTestSourceDirectory, onlyIfMissing = false)) ++
+    integrationTestFiles.map(_.writeToDirectory(unmanagedIntegrationTestSourceDirectory, onlyIfMissing = false))
```
octonato, line 53: It turns out that we were not regenerating the `unmanaged` files. See above in method `write` (GitHub collapsed). 

This method is only used for testing though. We could eventually move it or make it `private [akkaserverless]`? In any case, codegen is not expected to be used directly by users so no big deal IMO.

octonato (commented):
Left some comments.

codegen/core/src/main/scala/com/lightbend/akkasls/codegen/File.scala
octonato: We should probably mark that whole class as internal API. We never know.

octonato: We can do it in follow-up PR. I think all the codegen API should be for internal usage only. 

They don't show up as a dependency in the project. In principle they are a dependency for the plugins. Still, they show up in maven central as an artifact. 

At that point in time (and probably later as well), we want to have full freedom to change the codegen code the way we find necessary without having to keep any sort of backward compatibility.

johanandren (approved):

PR #813, opened 2022-01-21
Docs for token based paging in views

johanandren:
Can't be merged/published until we switch db backend, since token based paging is not supported for spanner.

### DIFF
--- a/docs/src/modules/java/pages/views.adoc
+++ b/docs/src/modules/java/pages/views.adoc
@@ -656,11 +656,15 @@ SELECT * AS customers, COUNT(*) AS total, has_more() AS more FROM customers LIMI
 
 === Paging
 
-Reading through a query result one "page" at a time rather than returning the entire result at once is possible with `OFFSET` and `LIMIT`.
+Reading through a query result one "page" at a time rather than returning the entire result at once is possible in two ways, with a count based offset or a token based offset.
+
+In both cases `OFFSET` and `LIMIT` are used.
 
 `OFFSET` specifies at which offset in the result to start
 
 `LIMIT` specifies a maximum number of results to return
+
+==== Count based offset ====
 
 The values can either be static, defined up front in the query:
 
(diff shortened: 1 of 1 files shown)


docs/src/modules/java/pages/views.adoc
```diff
@@ -653,3 +653,76 @@ To get the total number of results that will be returned over all pages, use `CO
 ----
 SELECT * AS customers, COUNT(*) AS total, has_more() AS more FROM customers LIMIT 10
 ----
+
+=== Paging
+
+Reading through a query result one "page" at a time rather than returning the entire result at once is possible in two ways, with a count based offset or a token based offset.
```
ennru, line 659:
```suggestion
Splitting a query result into one "page" at a time rather than returning the entire result at once is possible in two ways:
* with a count based offset or 
* a token based offset.
```

docs/src/modules/java/pages/views.adoc
```diff
@@ -653,3 +653,76 @@ To get the total number of results that will be returned over all pages, use `CO
...
+SELECT * FROM customers OFFSET :start_from LIMIT :max_customers
+----
+
+Note: Using numeric offsets can lead to missing or duplicated entries in the result if entries are added to or removed from the view between requests for the pages.
+
+==== Token based offset ====
+
+The count based offset requires that you keep track of how far you got by adding the page size to the offset for each query.
+
+An alternative is to use a string token emitted by Akka Serverless identifying how far into the result set the paging has reached using the functions `next_page_token()` and `page_token_offset()`.
```
ennru, line 688:
```suggestion
An alternative to this is to use a string token emitted by Kalix identifying how far into the result set the paging has reached using the functions `next_page_token()` and `page_token_offset()`.
```

ennru (approved):
Some suggestions. LGTM.

ennru (changes requested):
Looking at views.adoc I notice the whole "Paging" section is duplicated.

PR #799, opened 2022-01-12
chore: dynamic project for ExampleSuite compilation

octonato:
Inspired by #795, but able to load the example suite projects dynamically using flags:

```
sbt -Dexample.suite.scala.enabled -Dexample.suite.java.enabled
```

I found this quite useful when working on #796. As such as could generate the code and compile without leaving sbt or switching consoles.

PR #817, opened 2022-01-21
chore: removes obsolte generation suites

octonato:
Removes all tests that were comparing strings emitted by the generator. They became redundant after we merged #799 and #808.

To progress with #796, I plan to change part of the AST emitted by the model builder. Removing these tests first will speed it up because we won't need to refactor the TestData and all the GeneratorSuites because of changes in the AST. 

References #781


codegen/core/src/test/scala/com/lightbend/akkasls/codegen/TestData.scala
```diff
@@ -86,156 +86,11 @@ object TestData {
...
-  }
-
-  def serviceProto(suffix: String = ""): PackageNaming =
-    packageNamingTemplate.copy(
-      "my_service.proto",
-      s"MyService$suffix",
-      "com.example.service",
-      javaOuterClassnameOption = packageNamingTemplate.javaOuterClassnameOption.map(_ => s"ServiceOuterClass$suffix"))
-
   def domainProto(suffix: String = ""): PackageNaming =
```
octonato, line 89: This is the only method that is still being used in the tests.

octonato:
One extra bit of information, the generator suites for Replicated Entity covered quite almost (if not all) replicated data types. 

We don't have the equivalent in the ExampleSuite, but we do have in the compile only tests (`sbt-script`, `java-gen-compilation-tests` and `scala-gen-compilation-tests`). So we are covered.

johanandren (approved):
LGTM

jrudolph (commented):
LGTM, thanks for cleaning up!

PR #818, opened 2022-01-24
chore: rework message types

octonato:
Draft because on top of #796.


codegen/core/src/main/scala/com/lightbend/akkasls/codegen/ModelBuilder.scala
```diff
@@ -147,19 +146,19 @@ object ModelBuilder {
...
-      else MessageTypeArgument(FullyQualifiedName(name, name, proto, fqn))
+      else MessageTypeArgument(ProtoMessageType(name, name, proto, messageTypeOpt))
     }
   }
 
   /**
    * Type argument for Protobuf message types.
    */
-  case class MessageTypeArgument(fqn: FullyQualifiedName) extends TypeArgument
+  case class MessageTypeArgument(messageType: MessageType) extends TypeArgument
```
octonato, line 161: `MessageTypeArgument` was taking a `FullyQualifiedName`, but now it takes `MessageType`. Kind of fits like a glove. :-)

octonato (commented):
Most of the changes is about creating the new hierarchy and rename the values to reflect the new type name. 

Whenever possible, we should prefer `MessageType` over `ProtoMessageType`, but this is working in progress. I will be adding more commit here, but buy small steps.

octonato (commented):
I'm not so happy with my choice for `PojoMessageType`

Strictly speaking, any Java / Scala class can be of this type. So for instance, the components that we generate are not proto message, but that simplified type. 

So, we have trait `MessageType` and `ProtoMessageType` extending `MessageType`. 

Then the `PojoMessageType`, but I'm looking for a new name for this. 
I have the following in mind: 

* SimpleMessageType 
* PlainMessageType
* FlatMessageType

That message type is only a package and a type name. Nothing else. 

Any preference or alternative name?

octonato:
This is ready for review. The goal here is to, when ready, to merge it onto #796 (feature branch). 

I'm looking for feedback on alternative names for `PojoMessageType` (see my comment above).

octonato:
I've been observing quite some CI failures due to network errors. 

The build here failed with 

```bash
76205aac4d5a: Pull complete 
[ERROR] DOCKER> Unable to pull 'docker.io/library/adoptopenjdk:11-jre-hotspot' from registry 'docker.io' : error pulling image configuration: ....
```

but I've seen other kind of errors.

codegen/core/src/main/scala/com/lightbend/akkasls/codegen/MessageType.scala
```diff
@@ -20,20 +20,58 @@ import scala.jdk.CollectionConverters._
...
+}
+
 /**
- * A fully qualified name that can be resolved in any target language
- *
- * @param descriptorObject
- *   code representation of the descriptor that defined this type, if any
+ * A message type representing a simple Java/Scala type
+ */
+object PojoMessageType {
```
johanandren, line 56: `ClassMessageType`, `NativeClassMessageType`, `NativeMessageType`?

octonato: `ClassMessageType` sounds good. Generic enough and unbiased.

octonato:
Build is broken because of changes in main. A rebase is need to fix it.

patriknw (approved):
LGTM

PR #819, opened 2022-01-25
Update instructions for samples

### LINKED ISSUE #750: Wrong instructions in readme for mvn exec:java
I think we have switched to `exec:exec`
```
[ERROR] Failed to execute goal org.codehaus.mojo:exec-maven-plugin:3.0.0:java (default-cli) on project eventsourced-customer-registry: Unable to parse configuration of mojo org.codehaus.mojo:exec-maven-plugin:3.0.0:java for parameter arguments: Cannot store value into array: arraycopy: element type mismatch: can not cast one of the elements of java.lang.Object[] to the type of the destination array, java.lang.String
```

raboof:
Fixes #750

### DIFF
--- a/samples/java-eventsourced-customer-registry/README.md
+++ b/samples/java-eventsourced-customer-registry/README.md
@@ -7,18 +7,12 @@ To run the example locally:
 * Start the example:
   * publish relevant projects for use from Maven
     ```
-    sbt publishM2
-    ```
-  * build the latest Maven plugin and archetypes
-    ```
-    cd maven-java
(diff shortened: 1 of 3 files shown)


johanandren (approved):
LGTM but was it only those three?

raboof:
> LGTM but was it only those three?

Those were the only occurances of exec:java... it looks like most of the other README's have the generic archetype-generated README which already recommended exec:exec

octonato (approved):
LGTM

PR #821, opened 2022-01-25
feat: add support for Scala domain classes

octonato:
Draft because on top of #818.


codegen/scala-gen/src/main/scala/com/akkaserverless/codegen/scalasdk/impl/ScalaGeneratorUtils.scala
```diff
@@ -81,6 +93,45 @@ object ScalaGeneratorUtils {
...
+        else {
+          var sep = c""
+          val separatedTypes = types.foldLeft(c"") { (acc, el) =>
+            val cur = c"$acc$sep$el"
+            sep = c", "
+            cur
+          }
+          c"[$separatedTypes]"
+        }
+    }
```
octonato, line 133: This is a duplication for `dataType` and `parameterizeDataType` that works with `CodeBlock` instead of String and implicit imports. 

The replicated entity generator is the only Scala generator that is not using codeblocks. Probably because of the lack of those methods. We can use it as a migration path to introduce CodeBlock here as well. 

Last but not least I added this dirty hack at the end that is basically a `mkString` for CodeBlocks. Can certainly be improved.

octonato (commented):
Breaks when compiling replicated entities. To be continued...

patriknw:
Isn't there still rather many open comments from https://github.com/lightbend/akkaserverless-java-sdk/pull/796? Wouldn't it be better to iterate and complete on the Java implementation before also bringing in Scala? So much code that must be touched for each change.

Also, I would like to understand what open questions we have around the whole feature. Do we have that discussion in some ticket?

octonato:
I'm closing since quite broken because of merges and because we parked the efforts. I will keep the branch though.

PR #826, opened 2022-01-27
doc: Sample for streaming view updates

johanandren:
References https://github.com/lightbend/akkaserverless-framework/pull/1068

Won't be ready for merge/publish until proxy 0.8.8 is published/available _and_ we have switched backing db:s

### DIFF
--- a/docs/src/modules/java/pages/views.adoc
+++ b/docs/src/modules/java/pages/views.adoc
@@ -408,6 +408,38 @@ include::example$scala-eventsourced-customer-registry/src/main/scala/customer/Ma
 
 The View definitions are stored and validated when a new version is deployed. There will be an error message if the changes are not compatible.
 
+== Streaming view updates
+
+A query can provide a near real time stream of results for the query, emitting new entries matching the query as they are added or updated in
+the view.
+
+This is done by adding the option `stream_updates` to a query method with a `stream` reply:
+
+[.tabset]
+Java::
++
+[source,proto,indent=0]
+----
+include::example$java-eventsourced-customer-registry/src/main/proto/customer/view/customer_view.proto[tag=stream-updates]
+----
+<1> A regular query, in this case showing all customers from a specific city
+<2> The `stream_updates` option set to `true`
+
+Scala::
++
+[source,proto,indent=0]
+----
+include::example$scala-eventsourced-customer-registry/src/main/proto/customer/view/customer_view.proto[tag=stream-updates]
+----
+<1> A regular query, in this case showing all customers from a specific city
+<2> The `stream_updates` option set to `true`
+
+This will first list the complete result for the query and then keep the response stream open, emitting new or updated
+entries matching the query as they are added to the view. The stream does not complete until the client closes it.
+
+Note: This is not intended as transport for service to service propagation of updates and does not guarantee delivery, for such use cases you
+should instead publish events to a topic, see xref:actions-publishing-subscribing.adoc[Publishing and Subscribing with Actions]
+
 [#query]
 == Query syntax reference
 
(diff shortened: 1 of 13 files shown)


docs/src/modules/java/pages/views.adoc
```diff
@@ -408,6 +408,38 @@ include::example$scala-eventsourced-customer-registry/src/main/scala/customer/Ma
...
+<2> The `stream_updates` option set to `true`
+
+Scala::
++
+[source,proto,indent=0]
+----
+include::example$scala-eventsourced-customer-registry/src/main/proto/customer/view/customer_view.proto[tag=stream-updates]
+----
+<1> A regular query, in this case showing all customers from a specific city
+<2> The `stream_updates` option set to `true`
```
patriknw, line 435: shall we also highlight the stream return type?

docs/src/modules/java/pages/views.adoc
```diff
@@ -408,6 +408,38 @@ include::example$scala-eventsourced-customer-registry/src/main/scala/customer/Ma
...
+
+This will first list the complete result for the query and then keep the response stream open, emitting new or updated
+entries matching the query as they are added to the view. The stream does not complete until the client closes it.
+
+Note: This is not intended as transport for service to service propagation of updates and does not guarantee delivery, for such use cases you
+should instead publish events to a topic, see xref:actions-publishing-subscribing.adoc[Publishing and Subscribing with Actions]
```
franciscolopezsancho, line 441: I found this NOTE to be very insightful. I wonder I fully understand it, though. 
- `service to service propagation` refers only to AkkaServerless services, right?  
- `for such use cases you should instead publish` is because of the delivery guarantees?

johanandren: Any service, Akka serverless or implemented with some other tech, should use message broker rather than this, for delivery guarantees and decoupling (a broker will still have the messages when the sending service is offline, and can still take published messages when the consumer is offline)

docs/src/modules/java/pages/views.adoc
```diff
@@ -408,6 +408,39 @@ include::example$scala-eventsourced-customer-registry/src/main/scala/customer/Ma
...
+
+[.tabset]
+Java::
++
+[source,proto,indent=0]
+----
+include::example$java-eventsourced-customer-registry/src/main/proto/customer/view/customer_view.proto[tag=stream-updates]
+----
+<1> A regular query, in this case showing all customers from a specific city
+<2> The `stream_updates` option set to `true`
```
ennru, line 426: As below:
```suggestion
<1> Return type must be a `stream`
<2> A regular query, in this case showing all customers from a specific city
<3> The `stream_updates` option set to `true`
```

samples/java-eventsourced-customer-registry/src/main/java/customer/view/CustomerByCityStreamingView.java
```diff
@@ -0,0 +1,78 @@
+package customer.view;
+
+import com.akkaserverless.javasdk.view.View;
+import com.akkaserverless.javasdk.view.ViewContext;
+import com.google.protobuf.Any;
+import customer.api.CustomerApi;
+import customer.domain.CustomerDomain;
+
+// This class was initially generated based on the .proto definition by Akka Serverless tooling.
```
ennru, line 9: Remove, or say Kalix.

samples/java-eventsourced-customer-registry/src/main/proto/customer/view/customer_view.proto
```diff
@@ -133,3 +134,71 @@ service CustomerByNameFromTopic {
...
 }
 // end::service-topic[]
+
+
+// tag::stream-updates[]
+message ByCityRequest {
+  string city = 1;
+}
+service CustomerByCityStreaming {
+  option (akkaserverless.codegen) = {
```
ennru, line 144: The annotations are not Kalix.

ennru (changes requested):

johanandren:
Thanks for the fixes @ennru

johanandren:
Actual deploy error failing this: "Error reported from Kalix system: KLX-00114 All View update methods must have the same eventing.in property.\n\nTo fix this issue, align the eventing.in property of the methods"

PR #832, opened 2022-02-09
chore: introduce ClassMessageType

octonato:
Spin-off from #796

This PR includes some improvements that I added to #796 that I will would like to keep anyway. Namely, the split between `ClassMessageType` and `ProtoMessageType`. 

Although we are not supporting POJO for the moment, the `ClassMessageType` is still useful for the codegen. We generated or manipulate types that are not derived from protos. Before we were faking them. We don't need to fake it anymore with `ClassMessageType`. 

Also `MessageType`, `ClassMessageType` and `ProtoMessageType` are much better than `FullyQualifiedName`.


johanandren (approved):
LGTM (but massive so hard to know if I missed something)

octonato:
> LGTM (but massive so hard to know if I missed something)

Agree. It's massive and boring. Nothing much going on here, only type massage. 

I rebased because I had conflicts. All solved now. I will merge once green. Thanks

PR #833, opened 2022-02-10
feat(sbt-plugin): announce support for proto3 optional fields

raboof:
It doesn't seem like we need to treat them specially, we can just 'say' we support them.

I'm not sure we need to do anything at the Maven side, will check that next.


johanandren (approved):

raboof:
> I'm not sure we need to do anything at the Maven side, will check that next.

OK, after changing `<protocArtifact>com.google.protobuf:protoc:3.15.0:exe:${os.detected.classifier}</protocArtifact>` (should not be needed on `main` anymore) I get some other errors, but it looks like those are local to my machine, and @sebastian-alfers successfully tested this before.

sebastian-alfers:
Do we have a plan to release this?

PR #843, opened 2022-02-24
feat: generate proto descriptors from POJOs

octonato:
Generates proto descriptors from Java classes - very draft and almost working

* support only unary call in Action
* only support class with fields of type Long (we need to map all scalar types by hand, easy to fix, but tedious)
* very limited unit tests


samples/java-fibonacci-action-pojo/src/main/java/com/example/Main.java
```diff
@@ -0,0 +1,28 @@
...
+import com.example.fibonacci.FibonacciAction;
+import com.akkaserverless.javasdk.action.ActionPojoProvider;
+
+public final class Main {
+
+  private static final Logger LOG = LoggerFactory.getLogger(Main.class);
+
+  public static AkkaServerless createAkkaServerless() {
+    AkkaServerless akkaServerless = new AkkaServerless();
+    akkaServerless.register(ActionPojoProvider.of(FibonacciAction.class,__ -> new FibonacciAction()));
```
octonato, line 20: Registering is a bit clunky, but we can improve it later. We might find something.

johanandren (commented):
Haven't dug into the details but gut feeling is that this is much better than the previous ideas!

pvlugter:
Interesting to flip from protocol-first and code generation to code-first and protocol generation. So no annotations on methods currently, but they'll be expected? — to be able to specify other things, like HTTP mappings, and which parameter is the entity key, and so on.

octonato:
> Interesting to flip from protocol-first and code generation to code-first and protocol generation. So no annotations on methods currently, but they'll be expected? — to be able to specify other things, like HTTP mappings, and which parameter is the entity key, and so on.

Yes, I think we will have to bring some annotations to fine tune how we want the http endpoints to be shaped. Also to annotate the field that is expected to carry the entity id. 
I deliberately pick Actions as a starting point because is the most simple case and I could have something working without adding any new annotations, but as this evolves we will need to add custom annotations for sure. 

We could even consider annotations at component level (eg: `@Action`, `@EventSourcedEntity`) and have some auto registration for the components and eventually supporting some DI all together. That's just an idea. There much more to do before we reach such a point.

samples/java-fibonacci-action-pojo/src/main/java/com/example/fibonacci/FibonacciAction.java
```diff
@@ -0,0 +1,42 @@
+/* This code was generated by Akka Serverless tooling.
+ * As long as this file exists it will not be re-generated.
+ * You are free to make changes to this file.
```
patriknw, line 3: remove these comments to avoid confusion

samples/java-fibonacci-action-pojo/src/main/java/com/example/fibonacci/FibonacciAction.java
```diff
@@ -0,0 +1,42 @@
...
+    };
+    return isPerfectSquare.test(5*num*num + 4) || isPerfectSquare.test(5*num*num - 4);
+  }
+
+  private long nextFib(long num) {
+    double result = num * (1 + Math.sqrt(5)) / 2.0;
+    return Math.round(result);
+  }
+
+  public Effect<Number> nextNumber(Number number) {
```
patriknw, line 33: will it look for methods returning Effect, or what is the criteria for a method becoming a rpc?

sdk/java-sdk/src/main/scala/com/akkaserverless/javasdk/impl/ProtoDescriptorGenerator.scala
```diff
@@ -0,0 +1,125 @@
...
+        val output = returnParamType(method).getName
+
+        methodBuilder
+          .setName(method.getName.capitalize)
+          .setInputType(input)
+          .setOutputType(output)
+
+        serviceBuilder.addMethod(methodBuilder.build())
+      }
+
```
patriknw, line 107: nice

patriknw (commented):
looking good

samples/java-fibonacci-action-pojo/src/main/java/com/example/fibonacci/FibonacciAction.java
octonato: For the moment, I'm picking methods with one param and returning Effect. 

We can explore more options later. The goal now is to really get the minimal working.

octonato:
The last commit adds the missing serializer. It's now possible to run the Fibonacci POJO sample (code first, no protobuf descriptor) and receive the response back as expected. 

Needs a local build of course. 😄

octonato:
closing in favour of #991

Issue #718, opened 2021-11-18
Remove support for noReply effect


johanandren:
If you use it, it causes an error in the proxy, a call to a service should always result in a reply or a failure so we can remove the effect from the API?

Noticed by @hseeberger

franciscolopezsancho:
I'm testing this in local with the java-eventsourced-counter, that has noReply in an Action, and I don't see any errors. Could you please elaborate further? 
Is the acceptance criteria here just removing the Effect noReply from everywhere in API? From each entity and also from Action? 
@johanandren  @hseeberger

johanandren:
I don't remember the specific error @hseeberger saw, but even if there is no error I think it doesn't make sense. For a void type of action we use `google.protobuf.Empty` in all the samples so I think that should be the canonical way (and that also makes it visible in the descriptor that a method is a no-reply).

franciscolopezsancho:
Makes sense to me. I'll remove `noReply` from Entities and Action, and change for  `google.protobuf.Empty` in the current examples. I'll check the docs.

pvlugter:
Was the `noReply` for streaming?

franciscolopezsancho:
Just to be sure. We are talking about removing ,among others, [NoReplay](https://github.com/lightbend/akkaserverless-java-sdk/blob/3fb57278574016629745753775bb70760e8d2e0e/sdk/java-sdk/src/main/scala/com/akkaserverless/javasdk/impl/effect/SecondaryEffectImpl.scala#L82) as well, right? So, there will be no noReplay in any shape or form.

pvlugter:
Yeah, not clear what the original intent was. The entities all have `allowNoReply = false`. Maybe it was originally for `noReply` + side effects when streaming, which is only supported by actions, as the only way to just send side effects?

franciscolopezsancho:
Interesting. I had no idea what `allowNoReply` could be for.  Now I have one idea at least 👍

franciscolopezsancho:
I understand this is now ready to close. Would you agree @johanandren?

PR #844, opened 2022-03-04
removing noReply

franciscolopezsancho:
References #718

### DIFF
--- a/sdk/java-sdk/src/main/scala/com/akkaserverless/javasdk/impl/effect/SecondaryEffectImpl.scala
+++ b/sdk/java-sdk/src/main/scala/com/akkaserverless/javasdk/impl/effect/SecondaryEffectImpl.scala
@@ -40,20 +40,16 @@ sealed trait SecondaryEffectImpl {
           ClientAction(
             ClientAction.Action
               .Failure(com.akkaserverless.protocol.component.Failure(commandId, failure.description))))
-      case NoSecondaryEffectImpl =>
+      case NoSecondaryEffectImpl(_) =>
         throw new RuntimeException("No reply or forward returned by command handler!")
     }
   }
 }
 
-case object NoSecondaryEffectImpl extends SecondaryEffectImpl {
-  override def sideEffects: Vector[SideEffect] = Vector.empty
+case class NoSecondaryEffectImpl(sideEffects: Vector[SideEffect] = Vector.empty) extends SecondaryEffectImpl {
 
   override def addSideEffects(newSideEffects: Iterable[SideEffect]): SecondaryEffectImpl =
-    throw new RuntimeException(
-      "can't add side effect to NoSecondaryEffectImpl!"
-    ) //TODO review, this seems fishy. Shall we redesing the trait?
-
+    copy(sideEffects = sideEffects ++ newSideEffects)
 }
 
 final case class MessageReplyImpl[T](message: T, metadata: Metadata, sideEffects: Vector[SideEffect])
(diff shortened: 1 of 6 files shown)


sdk/java-sdk/src/main/scala/com/akkaserverless/javasdk/impl/effect/SecondaryEffectImpl.scala
```diff
@@ -43,21 +40,20 @@ sealed trait SecondaryEffectImpl {
...
     }
   }
 }
 
 case object NoSecondaryEffectImpl extends SecondaryEffectImpl {
   override def sideEffects: Vector[SideEffect] = Vector.empty
 
-  override def addSideEffects(sideEffects: Iterable[SideEffect]): SecondaryEffectImpl =
-    NoReply(sideEffects.toVector)
+  override def addSideEffects(newSideEffects: Iterable[SideEffect]): SecondaryEffectImpl =
```
franciscolopezsancho, line 52: I'm not sure what to do with NoSecondaryEffectImpl itself. Is the default value for `_secondaryEffect` in `EventSourceEntityEffectImpl` for example

` private var _secondaryEffect: SecondaryEffectImpl = NoSecondaryEffectImpl`

franciscolopezsancho:
I need to update the [akkaserverless-framework](https://github.com/lightbend/akkaserverless-framework)
For example [this](https://github.com/lightbend/akkaserverless-framework/blob/f420cd506649b7d312251231a3b2da74264edfc2/tck/src/main/scala/com/akkaserverless/tck/ActionTCK.scala#L114) test now wouldn't pass. I think I just need to delete those sort of tests . Those test testing `noReply` plus, of course, noReply method itself, such as [this](https://github.com/lightbend/akkaserverless-framework/blob/f420cd506649b7d312251231a3b2da74264edfc2/testkit/src/main/scala/com/akkaserverless/testkit/action/ActionMessages.scala#L53)

Would you agree?

johanandren:
I agree, and then that will also require a release of the proxy and update of the version here in the SDK before this PR can be merged (possibly some change in the JS SDK as well, not sure)

franciscolopezsancho:
local tests using https://github.com/lightbend/akkaserverless-framework/pull/1205/files pass. 

Pending PR above gets approved and with it, the new version of the framework is published. So, the CI tests running here --pointing to that new akkaserverless-framework image-- can succeed.

sdk/java-sdk/src/main/scala/com/akkaserverless/javasdk/impl/effect/SecondaryEffectImpl.scala
johanandren: I think it'd make more sense to make that an option/optional in the internal effect builder impls now that it is not possible to attach side effects to a no-reply, but maybe I'm missing something?

franciscolopezsancho:
> I agree, and then that will also require a release of the proxy and update of the version here in the SDK before this PR can be merged (possibly some change in the JS SDK as well, not sure)

Yep, I just found out that I have to update `javascript-sdk` for integration tests on `framework` to pass.

franciscolopezsancho:
javascript and framework ready
https://github.com/lightbend/akkaserverless-javascript-sdk/pull/255
https://github.com/lightbend/akkaserverless-framework/pull/1205

franciscolopezsancho:
I'll try again against the new kalix-proxy

Issue #830, opened 2022-02-08
Allow responding with a custom gRPC error code


raboof:
Right now, when you use `effects().error("text")`, this will always create an error with gRPC status code 2 UNKNOWN.

It might be nice to add API to allow users to send more specific error codes, such as INVALID_ARGUMENT, NOT_FOUND or UNAUTHENTICATED (https://github.com/grpc/grpc/blob/master/doc/statuscodes.md).

johanandren:
Sounds good, but I think if we should not allow the whole set of codes, a subset of (useful) ones would be better.

johanandren:
Because gRPC error codes are few and mostly related to explicit gRPC transport related errors: https://grpc.github.io/grpc/core/md_doc_statuscodes.html ). 

Out of the large number of HTTP codes there are also codes with specific technical/protocol meanings unrelated to what a deployed service does that does not ever make sense for a service to emit (405, 411, 426, 505 to mention a few).

If we want to allow arbitrary HTTP codes we will have to at least allow emitting a pair of the gRPC status and the HTTP status, or make sure we can map any HTTP status code emitted to a gRCP status code that makes sense.

raboof:
> when someone is developing an REST API, they are just free to use any error code they want. Why would we not do / offer the same?

This is a bit awkward since we currently share the implementation between REST and gRPC calls. Using 'symbolic' names would allow us to provide options that have both the HTTP and gRPC error code set. Of course a 'power api' that allows you to set both (or just one and default the other to 'unknown') could also be available.

ennru:
Implemented with #847

patriknw:
@jroper Has this been added to the documentation?

PR #847, opened 2022-03-09
feat: custom failure status code support

jroper:
Added support for custom status codes to both Java and Scala SDKs.

This is to match feature parity with https://github.com/lightbend/akkaserverless-javascript-sdk/pull/253.

### DIFF
--- a/sdk/java-sdk/src/main/java/com/akkaserverless/javasdk/action/Action.java
+++ b/sdk/java-sdk/src/main/java/com/akkaserverless/javasdk/action/Action.java
@@ -20,6 +20,7 @@ import com.akkaserverless.javasdk.DeferredCall;
 import com.akkaserverless.javasdk.Metadata;
 import com.akkaserverless.javasdk.SideEffect;
 import com.akkaserverless.javasdk.impl.action.ActionEffectImpl;
+import io.grpc.Status;
 
 import java.util.Collection;
 import java.util.Optional;
@@ -105,6 +106,16 @@ public abstract class Action {
        * @param <S> The type of the message that must be returned by this call.
        */
       <S> Effect<S> error(String description);
+
+      /**
+       * Create an error reply.
+       *
+       * @param description The description of the error.
+       * @param statusCode A custom gRPC status code.
+       * @return An error reply.
+       * @param <T> The type of the message that must be returned by this call.
+       */
+      <T> Effect<T> error(String description, Status.Code statusCode);
 
       /**
        * Create a message reply from an async operation result.
--- a/sdk/java-sdk/src/main/scala/com/akkaserverless/javasdk/impl/action/ActionEffectImpl.scala
+++ b/sdk/java-sdk/src/main/scala/com/akkaserverless/javasdk/impl/action/ActionEffectImpl.scala
@@ -16,13 +16,13 @@
 
 package com.akkaserverless.javasdk.impl.action
 
-import com.akkaserverless.javasdk.DeferredCall
-import com.akkaserverless.javasdk.Metadata
-import com.akkaserverless.javasdk.SideEffect
+import com.akkaserverless.javasdk.{ DeferredCall, Metadata, SideEffect }
 import com.akkaserverless.javasdk.action.Action
-
 import java.util
 import java.util.concurrent.CompletionStage
+
+import io.grpc.Status
+
 import scala.collection.immutable.Seq
 import scala.concurrent.ExecutionContext
 import scala.concurrent.Future
@@ -58,7 +58,11 @@ object ActionEffectImpl {
     protected def withSideEffects(sideEffects: Seq[SideEffect]): ForwardEffect[T] =
       copy(internalSideEffects = sideEffects)
   }
-  final case class ErrorEffect[T](description: String, internalSideEffects: Seq[SideEffect]) extends PrimaryEffect[T] {
+  final case class ErrorEffect[T](
+      description: String,
+      statusCode: Option[Status.Code],
+      internalSideEffects: Seq[SideEffect])
+      extends PrimaryEffect[T] {
     def isEmpty: Boolean = false
     protected def withSideEffects(sideEffects: Seq[SideEffect]): ErrorEffect[T] =
       copy(internalSideEffects = sideEffects)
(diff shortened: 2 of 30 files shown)


samples/java-valueentity-customer-registry/pom.xml
```diff
@@ -10,8 +10,7 @@
   <name>Customer Registry</name>
 
   <properties>
-    <!-- TODO Update to your own Docker repository or Docker Hub scope -->
-    <dockerImage>my-docker-repo/${project.artifactId}</dockerImage>
+    <dockerImage>jamesroper/${project.artifactId}</dockerImage>
```
raboof, line 13: Did you intentionally check this in?

sdk/java-sdk/src/main/java/com/akkaserverless/javasdk/GrpcStatus.java
```diff
@@ -0,0 +1,91 @@
...
+package com.akkaserverless.javasdk;
+
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * A gRPC status code.
+ */
+public enum GrpcStatus {
```
raboof, line 26: We could also use `io.grpc.Status.Code` rather than introducing our own type, but I'm fine with either.

Perhaps consider https://github.com/lightbend/akkaserverless-java-sdk/issues/830#issuecomment-1032837992

sdk/java-sdk/src/main/java/com/akkaserverless/javasdk/action/Action.java
```diff
@@ -106,6 +107,16 @@ public void _internalSetActionContext(Optional<ActionContext> context) {
...
 
+      /**
+       * Create an error reply.
+       *
+       * @param description The description of the error.
+       * @param status A custom gRPC status code.
+       * @return An error reply.
+       * @param <T> The type of the message that must be returned by this call.
+       */
+      <T> Effect<T> error(String description, GrpcStatus status);
```
raboof, line 118: I guess in the future we might introduce yet another variant to also be able to control the HTTP status code (when transcoding), but this is :+1:.

sdk/scala-sdk/src/main/scala/com/akkaserverless/scalasdk/valueentity/ValueEntity.scala
```diff
@@ -80,12 +78,14 @@ object ValueEntity {
...
        *   The description of the error.
+       * @param status
+       *   An optional gRPC status code.
        * @return
        *   An error reply.
        * @tparam T
        *   The type of the message that must be returned by this call.
        */
-      def error[T](description: String): Effect[T]
+      def error[T](description: String, status: Option[GrpcStatus] = None): Effect[T]
```
raboof, line 88: If we want to support controlling HTTP status codes through this API as well, that default parameter might make it harder to evolve in a binary compatible way - but I guess for the SDK we care mainly about source compatibility and not so much about bincompat?

raboof (commented):
Nice! Refs #830

Limitation of this implementation is that you can only specify an error code, but not add a customer 'message' or additional protobuf-serialized 'details' objects. Seems fine to me.

Some comments for consideration and one probably accidentally-committed change, otherwise LGTM!

sdk/scala-sdk/src/main/scala/com/akkaserverless/scalasdk/valueentity/ValueEntity.scala
jroper: For the most commonly used HTTP status codes, there's a corresponding gRPC status code that we map to it in the HTTP transcoding.

But yes, my thinking here was that source compatibility is the most important thing.

samples/java-valueentity-customer-registry/pom.xml
jroper: Hmm...

sdk/java-sdk/src/main/java/com/akkaserverless/javasdk/GrpcStatus.java
jroper: Yeah, I did think about that. Are we guaranteed to always have `io.grpc.Status.Code` on the classpath? If `akka-grpc` switches to an `akka-http` based client implementation, it might go away right?

jroper:
> Limitation of this implementation is that you can only specify an error code, but not add a customer 'message' or additional protobuf-serialized 'details' objects.

The existing `description` parameter that's passed is passed to the standard `grpc-message` trailer. There are no other defined trailer values to pass, however, it might make sense at some point to add an optional `Metadata` object to the protocol and hence to the SDK APIs that gets added to the trailers if present.

sdk/java-sdk/src/main/java/com/akkaserverless/javasdk/GrpcStatus.java
raboof: Since `io.grpc.Status.Code` is part of `grpc-api` (and not `grpc-core`), I expect `akka-grpc` to keep depending on this even when we switch to `akka-http` for the transport implementation.

raboof:
> > Limitation of this implementation is that you can only specify an error code, but not add a customer 'message' or additional protobuf-serialized 'details' objects.
> 
> The existing `description` parameter that's passed is passed to the standard `grpc-message` trailer.

Of course, forgot about that. In that case it might be neat to have an `error` overload that takes only the code (and infers the default message from that), but not essential.

> There are no other defined trailer values to pass, however, it might make sense at some point to add an optional `Metadata` object to the protocol and hence to the SDK APIs that gets added to the trailers if present.

Yeah, there's a convention to add a `grpc-status-details-bin` header but that's not yet widely documented (https://github.com/grpc/grpc/issues/24007)

ennru:
@jroper Do you intend to finish this, or should someone else take over?

jroper:
Yes I'll finish it.

sdk/java-sdk/src/main/java/com/akkaserverless/javasdk/GrpcStatus.java
jroper: Ok, I've switched to `io.grpc.Status.Code`.

sdk/java-sdk/src/main/scala/com/akkaserverless/javasdk/impl/action/ActionEffectImpl.scala
```diff
@@ -73,7 +77,11 @@ object ActionEffectImpl {
     def reply[S](message: S, metadata: Metadata): Action.Effect[S] = ReplyEffect(message, Some(metadata), Nil)
     def forward[S](serviceCall: DeferredCall[_, S]): Action.Effect[S] = ForwardEffect(serviceCall, Nil)
     def noReply[S](): Action.Effect[S] = NoReply(Nil)
-    def error[S](description: String): Action.Effect[S] = ErrorEffect(description, Nil)
+    def error[S](description: String): Action.Effect[S] = ErrorEffect(description, None, Nil)
+    def error[S](description: String, statusCode: Status.Code): Action.Effect[S] = {
+      if (statusCode.toStatus.isOk) throw new IllegalArgumentException("Cannot fail with a success status")
```
raboof, line 82: :+1:

sdk/java-sdk/src/main/scala/com/akkaserverless/javasdk/impl/action/ActionsImpl.scala
```diff
@@ -134,10 +134,11 @@ private[javasdk] final class ActionsImpl(
           .recover { case NonFatal(ex) =>
             handleUnexpectedException(service, command, ex)
           }
-      case ErrorEffect(description, sideEffects) =>
+      case ErrorEffect(description, status, sideEffects) =>
         Future.successful(
           ActionResponse(
-            ActionResponse.Response.Failure(Failure(description = description)),
+            ActionResponse.Response.Failure(
+              Failure(description = description, grpcStatusCode = status.map(_.value()).getOrElse(0))),
```
raboof, line 141: ok, `0` does not mean `SUCCESS` but 'unspecified' here, an will be converted to `UNKNOWN` in the proxy. Not super obvious but :+1:

raboof (approved):

PR #849, opened 2022-03-21
fix(local-proxy): avoid telemetry from our CI

raboof:
Missed this use of testcontainers earlier...

References #974


pvlugter (approved):
JavaScript SDK will need something similar?

Issue #851, opened 2022-03-22
Use recommended kafka config in java-customer-registry-kafka-quickstart docker-compose.yml


raboof:
In the [docs](https://developer.lightbend.com/docs/akka-serverless/projects/message-brokers.html#_use_a_kafka_broker_locally) we show how to create a `docker-compose.yml` that connects to Kafka, but our [customer-registry-sample](https://github.com/lightbend/akkaserverless-java-sdk/blob/main/samples/java-customer-registry-kafka-quickstart/docker-compose.yml) application still connects to a Google PubSub emulator.

Probably because we still use that for testing ourselves?

beritou:
I agree, `java-customer-registry-kafka` should demonstrate how to use Kafka locally.

Testing the Google PubSub emulator is a separate concern and is ideally handled elsewhere.

beritou:
I see this is already underway in https://github.com/lightbend/akkaserverless-java-sdk/pull/852

PR #852, opened 2022-03-22
fix(sample): run kafka sample locally with kafka

raboof:
Probably also needs changes in the circleci config

Refs #851


franciscolopezsancho:
We could delete this file and only use what's in the kafka folder to start the proxy, Kafka and zookeeper? It might not be easy to spot in the README.md that `docker compose up` is supposed to run from within the `./kafka` folder.

raboof:
> It might not be easy to spot in the README.md that `docker compose up` is supposed to run from within the `./kafka` folder.

Ah, indeed I completely missed that :D

raboof:
> Probably also needs changes in the circleci config

ah, looking closer we don't seem to test this sample in CI at all - probably want to fix that?

octonato:
> Do you mean like [this one](https://github.com/lightbend/kalix-jvm-sdk/blob/main/samples/java-customer-registry-kafka-quickstart/kafka/docker-compose.yml?rgh-link-date=2022-05-13T16%3A39%3A11Z)? or something else?

Yes. So we have it already. 

We can maybe move it out of the `kafka` folder to make it more visible and directly accessible? 

`mv kafka/docker-compose.yml docker-compose.yml`

octonato (approved):
LGTM

octonato:
@raboof, I think the best to do in this sample, is to move the `mv kafka/docker-compose.yml docker-compose.yml` and adapt the README. We don't need the docker to be inside `kafka` folder.

franciscolopezsancho:
superseded by https://github.com/lightbend/kalix-jvm-sdk/pull/1142

PR #854, opened 2022-03-24
bump: proxy 0.9.0

johanandren:
Not for merging quite yet, proxy is not deployed, just want to run tests to see if something pops up.


johanandren:
Hitting the view id limit in a few samples so perhaps we need to revisit that. :/

PR #856, opened 2022-03-28
chore: make sure to run CI tasks against kalix-main

jrudolph:
The circleci config will have to be changed again after the branch rename later


jrudolph:
Looking good, going ahead.

PR #861, opened 2022-03-28
chore: cycle circle CI cache

jrudolph:
It seems the cache is severely out of date and the maven plugin publish step
is taking many minutes to build just because of downloading dependencies.

CircleCI isn't able to cache by contents so, we need to cycle keys to refresh
the cache.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -50,7 +50,7 @@ commands:
     description: "Save dependency cache"
     steps:
       - save_cache:
-          key: deps-cache-01
+          key: deps-cache-02
           paths:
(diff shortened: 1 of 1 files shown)


jrudolph:
Haha, what can go wrong? It broke the samples...

jrudolph:
Not sure why it breaks, in any case the whole caching setup on CircleCI seems pretty useless on these multi-step projects that share many of the dependencies. We would either have to create a super-cache including everything that is needed upfront (a cache for a key is immutable once created), or smaller caches for each step (much duplication we need to pay for).

PR #864, opened 2022-03-28
rename: adapt to run against new proxy protocols

jrudolph:


octonato (approved):
LGTM

jrudolph:
Several issues right now:

 * we need to add snapshot repository to every project to make use of a snapshot proxy
 * tck seems to be broken but more likely because of postgres / yugabyte than because of the renaming it seems

jrudolph:
https://github.com/lightbend/kalix-proxy/pull/1306 will help with the TCK problems

jrudolph:
All tests green :fireworks:

PR #866, opened 2022-03-29
Workaround sample view

johanandren:
Explicitly set the view ids for the ones that get to long ids autogenerated


octonato (approved):
LGTM, but ....
Why is that all about? And why it gets fixed if we stop using `KalixFactory`? 🤔

johanandren:
@octonato We had to limit the view ids because of a limit on table names in the db and using these ids as part of the table names, we'll probably have to revisit that decision but to workaround the build failing samples with long view ids we set them  manually to shorter ones for now.

octonato:
Ok, I see. 

My bad, I missed the calls to `withViewId` and was only looking on the change about the factory. Make sense now.

PR #875, opened 2022-03-30
rename: change group id and maven artifacts

octonato:


raboof (approved):
needs an update in ./sbt-plugin/src/main/scala/kalix/sbt/KalixPlugin.scala and fix some conflicts, then should be :+1: I think

PR #883, opened 2022-03-30
rename: more build changes using kalix in the name

octonato:


project/ProtocolProject.scala
```diff
@@ -44,8 +44,8 @@ object PublicProtocolProject extends AutoPlugin {
       // this repo doesn't exist (yet) but we need this for publishing to Sonatype OSS
       scmInfo := Some(
         ScmInfo(
-          url("https://github.com/lightbend/akkaserverless-protocol"),
-          "scm:git@github.com:lightbend/akkaserverless-protocol.git")),
+          url("https://github.com/lightbend/kalix-protocol"),
+          "scm:git@github.com:lightbend/kalix-protocol.git")),
```
octonato, line 48: This is one is a surprise. Why we are pointing to a non-existent repo in GitHub? See comment on line 44. 

I'm keeping the status-quo, which is faking now that it comes from `kalix-protocol` repo. I think this should just be `kalix-jvm-sdk`, but first I want to dig why we have it.

raboof: I agree this should be `kalix-jvm-sdk` - I suspect this configuration is a remnant from when the sdk was still part of the (private) akkaserverless-framework repo.

octonato: ok, I will change this. Makes no sense now.

project/Dependencies.scala
```diff
@@ -116,14 +116,14 @@ object Dependencies {
     // FIXME: For now TCK protos have been copied and adapted into this project.
     //        Running the TCK is still meaningful as it runs the TCK check against the defined framework version.
     //        Eventually, the final form of protos from should be backported to the framework.
-    //        See https://github.com/lightbend/akkaserverless-java-sdk/issues/605
-    //  akkaslsTckProtocol % "protobuf-src",
-    //  "com.akkaserverless" % "akkaserverless-tck-protocol" % AkkaServerless.FrameworkVersion % "protobuf-src",
+    //        See https://github.com/lightbend/kalix-jvm-sdk/issues/605
+    //  kalixTckProtocol % "protobuf-src",
+    //  "io.kalix" % "kalix-tck-protocol" % Kalix.FrameworkVersion % "protobuf-src",
```
patriknw, line 121: Would be good to rename Framework to Proxy, but that should perhaps be a separate PR

patriknw (approved):
LGTM, with follow up suggestion

PR #886, opened 2022-03-30
rename: update antora docbuilder image

octonato:
This is now good to go. Image is published.

### DIFF
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -10,7 +10,7 @@ managed_attachments := ${src_managed}/modules/${module}/attachments
 managed_examples := ${src_managed}/modules/${module}/examples
 managed_partials := ${src_managed}/modules/${module}/partials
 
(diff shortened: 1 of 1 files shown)


octonato:
failing for yet unknown reason, that's for tomorrow now

pvlugter:
Failing on `error: Generator not found or failed to load. Try installing the '@antora/xref-validator' package.`

The SDK docs use the xref validation, which needs to be in the image. Should be included in its docker build already though.

pvlugter:
So the issue with the new doc builder image is dependencies. Given the way that node dependencies are often minimum versions or ranges, rebuilding the image has installed new dependencies for the xref-validator, which then fail. We need to update things (although xref-validator hasn't been changed for a long time, maybe there's something different for that now) and make it a reproducible build in some way — maybe not installing globally, but from a lock file instead.

For now, I've made the kalix doc builder image a copy of the old akkaserverless image, rather than rebuild it.

PR #887, opened 2022-03-31
chore: remove go_package

octonato:
Unrelated to renaming, but sending to kalix-main anyway. 

I don't think we need it in the tck files. Probably dates from the time where this and the proxy were one single project?


patriknw (approved):
LGTM

PR #888, opened 2022-03-31
rename: final pass to remove all leftovers

octonato:
Chased all leftovers. Links, texts (docs/readme), some references in proto files. 

The only thing left are the references from the go_package, but I send a separate PR (#887) to remove it completely.


patriknw (approved):
LGTM

Issue #501, opened 2021-09-24
Allow passing metadata in testkits


johanandren:
It will probably be interesting to be able to pass subject id, as well as headers and maybe other metadaata for advanced use cases, especially in actions, we should generate testkit methods taking a `Metadata` in addition to the request and set that up in the context.

franciscolopezsancho:
I'm adding `eventSubject` and `metadata` to the `TestkitActionContext`. That's straightforward, what I'm not sure is about the headers. Do you mean `ActionOptions.withForwardHeaders`?

johanandren:
When a service is configured to forward some http/grpc header it will show up in the user function as a metadata entry, mentioning it was just an explanation of what use cases it is important to pass metadata along with the command when testing a command handler.

johanandren:
The `CommandContext`s for the entities extend `MetadataContext` so they do have the metadata available. The `subjectId` is tied to the instance in the entities and called `entityId`, so that should stay as it is with something specified upon creation of the testkit rather than be possible to send on each request.

franciscolopezsancho:
I see. Ok, great. Thanks!

PR #892, opened 2022-04-01
added functionality and tests

franciscolopezsancho:
Allow passing metadata in testkits

References #501

### DIFF
--- a/samples/java-eventsourced-counter/src/test/java/com/example/actions/CounterJournalToTopicActionTest.java
+++ b/samples/java-eventsourced-counter/src/test/java/com/example/actions/CounterJournalToTopicActionTest.java
@@ -5,12 +5,16 @@
 package com.example.actions;
 
 import com.akkaserverless.javasdk.testkit.ActionResult;
+import com.akkaserverless.javasdk.testkit.impl.TestKitActionContext;
 import com.example.actions.CounterJournalToTopicAction;
 import com.example.actions.CounterJournalToTopicActionTestKit;
 import com.example.actions.CounterTopicApi;
 import com.example.domain.CounterDomain;
 import com.google.protobuf.Any;
 import com.google.protobuf.Empty;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Optional;
 import org.junit.Test;
 import static org.junit.Assert.*;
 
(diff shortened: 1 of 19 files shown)


samples/scala-eventsourced-counter/src/test/scala/com/example/actions/CounterJournalToTopicActionSpec.scala
```diff
@@ -15,6 +18,13 @@ class CounterJournalToTopicActionSpec extends AnyWordSpec with Matchers {
       result.reply shouldBe Increased(1)
     }
 
+    "handle command Increase with Metadata" in {
+      val context = new TestKitActionContext(metadataEntries = Map("myKey" -> Left("myValue")))
```
franciscolopezsancho, line 22: I'm not sure I like a Left here but if not then I can't verify at compile time whether it's a String or a java.nio.ByteBuffer. 
Please, have a look at `com.akkaserveless.scalasdk.testkit.impl.TestKitActionContext`.
Another caviat is that `com.akkaserveless.javasdk.testkit.impl.TestKitActionContext` doesn't have Either so its constructor has a different signature.

samples/java-eventsourced-counter/src/test/java/com/example/actions/CounterJournalToTopicActionTest.java
```diff
@@ -24,6 +28,14 @@ public void increaseTest() {
     assertEquals(1, result.getReply().getValue());
   }
 
+  @Test
+  public void increaseMetadataTest() {
+    TestKitActionContext context = new TestKitActionContext(new HashMap<String, String>(Map.of("myKey","myValue")));
```
johanandren, line 33: I think this is the wrong approach, Metadata is something that comes with each request, not something global across several calls.

I would have expected the user facing testkit API to add overloads to call methods specific metadata, like this:

```java
testKit.increase(ValueIncreased.newBuilder...build(), Metadata(stuff))
```

franciscolopezsancho: I see. Yep, now that you mention it, makes a lot of sense. I'll do that.

franciscolopezsancho:
This approach is not adecuate.

PR #894, opened 2022-04-04
chore: rename label

ennru:


.github/labeler.yml
```diff
@@ -1,6 +1,6 @@
 # https://github.com/marketplace/actions/labeler
 
-as-framework:
+kalix-proxy:
```
octonato, line 3: Sounds strange to have a label `kalix-proxy` in the SDK, no?

patriknw: The purpose of this label is to be a "team" label that we use in the aggregated project board. I agree that kalix-proxy might not be the best, but I don't have a better suggestion.

patriknw:
could you also change that label in https://github.com/lightbend/akkaserverless-java-sdk/blob/main/RELEASING.md while you are at it

octonato (approved):
LGTM

PR #895, opened 2022-04-05
Adding Metadata and eventSubject to ActionTestKit

franciscolopezsancho:
Aproach passing Metadata to the RPC call methods themselves. One test passing using the TestKit generated by the ActionTestKitGenerator.

References #501

### DIFF
--- a/codegen/java-gen/src/main/scala/com/lightbend/akkasls/codegen/java/ActionTestKitGenerator.scala
+++ b/codegen/java-gen/src/main/scala/com/lightbend/akkasls/codegen/java/ActionTestKitGenerator.scala
@@ -44,6 +44,7 @@ object ActionTestKitGenerator {
         "java.util.function.Function",
         "java.util.Optional",
         s"$packageName.$className",
+        "com.akkaserverless.javasdk.Metadata",
         "com.akkaserverless.javasdk.action.Action.Effect",
         "com.akkaserverless.javasdk.action.ActionCreationContext",
         "com.akkaserverless.javasdk.testkit.ActionResult",
@@ -64,11 +65,11 @@ object ActionTestKitGenerator {
         |
         |  private Function<ActionCreationContext, $className> actionFactory;
         |
-        |  private $className createAction() {
-        |    $className action = actionFactory.apply(new TestKitActionContext());
-        |    action._internalSetActionContext(Optional.of(new TestKitActionContext()));
+        |  private $className createAction(TestKitActionContext context) {
+        |    $className action = actionFactory.apply(context);
+        |    action._internalSetActionContext(Optional.of(context));
         |    return action;
-        |  };
+        |  }
         |
         |  public static $testKitClassName of(Function<ActionCreationContext, $className> actionFactory) {
         |    return new $testKitClassName(actionFactory);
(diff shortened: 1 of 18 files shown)


franciscolopezsancho:
@johanandren let me know how you find it, please.

samples/scala-eventsourced-counter/src/test/scala/com/example/actions/CounterJournalToTopicActionSpec.scala
```diff
@@ -15,6 +16,12 @@ class CounterJournalToTopicActionSpec extends AnyWordSpec with Matchers {
       result.reply shouldBe Increased(1)
     }
 
+    "handle command Increase with Metadata" in {
+      val testKit = CounterJournalToTopicActionTestKit(new CounterJournalToTopicAction(_))
+      val result = testKit.increaseConditional(ValueIncreased(1),Metadata.empty.set("myKey","myValue"))
```
franciscolopezsancho, line 21: Here's the UX

johanandren: 👍 make it so!

codegen/java-gen/src/main/scala/com/lightbend/akkasls/codegen/java/ActionTestKitGenerator.scala
```diff
@@ -135,8 +140,9 @@ object ActionTestKitGenerator {
     service.commands
       .map { command =>
         s"""|public ${selectOutputResult(command)} ${lowerFirst(command.name)}(${selectInputType(command)} ${lowerFirst(
-          command.inputType.protoName)}) {
-          |  ${selectOutputEffect(command)} effect = createAction().${lowerFirst(command.name)}(${lowerFirst(
+          command.inputType.protoName)}, Metadata metadata, Optional<String> eventSubject) {
```
johanandren, line 143: In "production" the "event subject" comes from the metadata entry `ce-subject`, I wonder if we should extract that like that from user provided Metadata here instead of having a specific parameter for actions?

And maybe no default if it is not set, as in a normal call to an action? I think it may be set for topic in, and always for eventing in, so for testing such actions a user can specify it through the passed Metadata.

johanandren (commented):
Looking good but I think we should skip the `eventSubject` parameter.

And then we'll want to be able to pass Metadata on the other component action handlers in test kits as well, not only actions.

codegen/java-gen/src/main/scala/com/lightbend/akkasls/codegen/java/ActionTestKitGenerator.scala
franciscolopezsancho: I'm happy to hear that the event subject comes from the metadata in prod. It just didn't look good that parameter constructor. Cool, I'll extract if there's a key `ce-subject` otherwise won't be set. Meaning `TestKitActionContext.eventSubject()` would return `None` (or `Optional.empty()`) when no `ce-subject` in its metadata.

johanandren: Yep, makes sense. For entities it will be a bit different since they _are_ the subject and it will never change for the same one, I don't think they will ever see a `ce-subject` in the metadata, and if they do it does not imply the same thing as for an action.

johanandren (approved):
LGTM, we'll need the same for the other component type testkits.

PR #902, opened 2022-04-12
Adding metadata testkit es

franciscolopezsancho:
References #501

### DIFF
--- a/samples/scala-eventsourced-counter/src/main/scala/com/example/domain/Counter.scala
+++ b/samples/scala-eventsourced-counter/src/main/scala/com/example/domain/Counter.scala
@@ -25,7 +25,22 @@ class Counter(context: EventSourcedEntityContext) extends AbstractCounter {
         .thenReply(_ => Empty())
     }
 
-   import com.akkaserverless.javasdk.impl.DeferredCallImpl
+  override def increaseWithConditional(
+      currentState: CounterState,
+      increaseValue: example.IncreaseValue): EventSourcedEntity.Effect[Empty] =
+    if (increaseValue.value < 0) {
+      effects.error("Value must be a zero or a positive number")
+    } else {
+      println(commandContext.metadata)
+      val valIncrease = if (commandContext.metadata.get("myKey") == Some("myValue")){
+        ValueIncreased(2 * increaseValue.value) 
+      } else {
+        ValueIncreased(increaseValue.value)
+      }
+      effects
+        .emitEvent(valIncrease)
+        .thenReply(_ => Empty())
+    }
 
   override def increaseWithSideEffect(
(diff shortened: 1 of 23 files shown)


samples/scala-eventsourced-counter/src/main/scala/com/example/domain/Counter.scala
```diff
@@ -25,7 +25,22 @@ class Counter(context: EventSourcedEntityContext) extends AbstractCounter {
...
     }
 
-   import com.akkaserverless.javasdk.impl.DeferredCallImpl
+  override def increaseWithConditional(
+      currentState: CounterState,
+      increaseValue: example.IncreaseValue): EventSourcedEntity.Effect[Empty] =
+    if (increaseValue.value < 0) {
+      effects.error("Value must be a zero or a positive number")
+    } else {
+      println(commandContext.metadata)
```
johanandren, line 34: runaway println
```suggestion
```

sdk/java-sdk-testkit/src/main/scala/com/akkaserverless/javasdk/testkit/impl/TestKitEventSourcedEntityCommandContext.scala
```diff
@@ -30,6 +30,11 @@ final class TestKitEventSourcedEntityCommandContext(
     override val metadata: Metadata = Metadata.EMPTY)
     extends CommandContext
     with InternalContext {
+
+  def this(metadat: Metadata) {
+    this(metadata = metadat)
```
johanandren, line 35:
```suggestion
  def this(metadata: Metadata) {
    this(metadata = metadata)
```

johanandren (commented):
Looking good, two tiny things.

sdk/java-sdk-testkit/src/main/scala/com/akkaserverless/javasdk/testkit/impl/TestKitEventSourcedEntityCommandContext.scala
franciscolopezsancho: Thanks for picking this up Johan but unfortunately this wouldn't work because the compiler gets an ambiguous reference for the constructor. Doesn't know if `this` refers to the very same constructor or the constructor with 5 parameters.

samples/scala-eventsourced-counter/src/main/scala/com/example/domain/Counter.scala
franciscolopezsancho: Nice catch. Thanks!

johanandren (approved):
LGTM!

Issue #78, opened 2021-07-01
Document how to use codegen maven archetype with Apple M1 architecture


kikiya:
Small changes required in `pom.xml` also user will need to install `protoc` locally in order to use the maven archetype. 

This needs to be documented properly.

johanandren:
Upstream fixes in akka-grpc in progress in https://github.com/akka/akka-grpc/pull/1477

For the record the easiest path right now is probably using rosetta and an x86 JDK.

johanandren:
Upstream issue for the docker plugin to support cross building: https://github.com/fabric8io/docker-maven-plugin/issues/1502

PR #905, opened 2022-04-13
feat(scala): Support for building Kalix images on Mac m1 #78

johanandren:
References #78

A somewhat sketchy solution, but seems to work so maybe worth it?

If we decide it is, needs to be applied to all samples and the quickstart g8 template.

### DIFF
--- a/samples/scala-customer-registry-quickstart/build.sbt
+++ b/samples/scala-customer-registry-quickstart/build.sbt
@@ -10,6 +10,7 @@ enablePlugins(KalixPlugin, JavaAppPackaging, DockerPlugin)
 dockerBaseImage := "docker.io/library/adoptopenjdk:11-jre-hotspot"
 dockerUsername := sys.props.get("docker.username")
 dockerRepository := sys.props.get("docker.registry")
+dockerUpdateLatest := true
 dockerBuildCommand := {
   if (sys.props("os.arch") != "amd64") {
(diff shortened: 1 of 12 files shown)


johanandren:
@raboof what do you think about this somewhat hacky solution? (m1/arm64 needs a different docker command to cross build the image for amd64)

raboof:
> @raboof what do you think about this somewhat hacky solution? (m1/arm64 needs a different docker command to cross build the image for amd64)

Yeah I think it makes sense. Would be neat if this could be an upstream feature of native-packager, where you would set the 'target architecture' (default 'inherit' or somesuch) and it'd do the right thing, but failing that this seems reasonable ;)

raboof:
I guess we need something similar for javascript and the Maven plugin then? :/

johanandren:
The maven plugin I have already solved m1 support for in #https://github.com/lightbend/kalix-jvm-sdk/pull/904 so it's only sbt we need to figure it out for.

Javascript is a whole different story with more blockers before docker becomes the main problem https://github.com/lightbend/kalix-javascript-sdk/issues/99

johanandren:
Created an upstream issue/feature request for the plugin: https://github.com/sbt/sbt-native-packager/issues/1503

johanandren:
I'll go ahead and copy this to the other samples and the g8 template then.

ennru (approved):
LGTM.

PR #906, opened 2022-04-14
Adding metadata testkit value entity

franciscolopezsancho:
Better to be reviewed after https://github.com/lightbend/akkaserverless-java-sdk/pull/902 gets merged because it's based on that branch. 
References #xxxx

PR #919, opened 2022-04-22
Adding metadata testkit value entity

franciscolopezsancho:
References #906


codegen/java-gen/src/main/scala/kalix/codegen/java/EventSourcedEntityTestKitGenerator.scala
```diff
@@ -44,6 +44,7 @@ object EventSourcedEntityTestKitGenerator {
         "java.util.ArrayList",
```
johanandren, line 44: PR says value entity but this is event sourced entity testkit, rebase miss somehow or why is this in here?

johanandren (commented):
Still looks like it contains a lot of different changes (ActionTestKit/ESTestkit) did something go wrong with rebase?

codegen/java-gen/src/main/scala/kalix/codegen/java/EventSourcedEntityTestKitGenerator.scala
franciscolopezsancho: you're right. The thing is that this branches out from this [PR](https://github.com/lightbend/kalix-jvm-sdk/pull/902) that's already merged. Where all the EventSourced changes are included. I wrongly assumed none of those EventSourced would show as a difference now that they are in the main branch. I'll have a look.

johanandren (approved):

johanandren:
Meh, I accidentally hit enter when rewriting the commit message here so it became truncated _and_ probably a merge commit rather than squash. 🤦

franciscolopezsancho:
I just bought a new keyboard and is happening to me all the time :) Thanks for merging it in any shape or form.

PR #921, opened 2022-04-22
feat: reliable timers API

octonato:
TODO

- [x] javadoc and scaladoc for new API
- [x] refactor java-reliable-timers sample 
- [x] java doc snippets

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -69,7 +69,7 @@ lazy val sdkJava = project
       "Kalix Java SDK",
       "-noqualifier",
       "java.lang"),
-    Compile / akkaGrpcGeneratedSources := Seq(AkkaGrpc.Server),
+    Compile / akkaGrpcGeneratedSources := Seq(AkkaGrpc.Server, AkkaGrpc.Client),
     Compile / akkaGrpcGeneratedLanguages := Seq(AkkaGrpc.Scala), // FIXME should be Java, but here be dragons
     // We need to generate the java files for things like entity_key.proto so that downstream libraries can use them
     // without needing to generate them themselves
--- a/docs/src/modules/java/pages/timers.adoc
+++ b/docs/src/modules/java/pages/timers.adoc
@@ -1,0 +1,170 @@
+= Timers
+:page-supergroup-java-scala: Language
+
+Timers allow for scheduling calls in the future. For example, to verify that some process have been completed or not.
+
+Timers are persisted in the https://docs.kalix.io/reference/glossary.html#proxy[Kalix proxy] and are guaranteed to run at least once. 
+
+When a timer is triggered, the scheduled call is executed. If successfully executed, the timer completes and is automatically removed.
+In case of a failure, the timer is rescheduled with a delay of 3 seconds. This process repeats until the call succeeds.
+
+You can schedule a timer for any service method in Kalix, but you can only create a timer from within an Action and by passing a `DefferedCall`. 
+
+**Timer features**:
+
+* Timers are guaranteed to run at least once.
+* Timers can be scheduled to run at any time in the future.
+* Timers can be cancelled.
+* Timers are automatically removed once successfully completed.
+* Timers are re-scheduled in case of failures.
+
+**Timer limitations**: 
+
+* Timers can only be scheduled from within an Action.
+* Timers can only call other components (eg: Actions and Entities), therefore external service calls must be wrapped by an Action in the deployed service.
+
+To demonstrate its functionality, let's consider an Ordering Service composed of a xref:value-entity.adoc[Value Entity] and an xref:actions.adoc[Action]. The Action will work as a xref:actions-as-controller.adoc[Controller] receiving calls and delegating to the Order Entity. 
+
+Users can place an order, but the order must be confirmed within a period of time. You can compare it with an Ordering Food application where the restaurant needs to confirm if it can accept the order. If no confirmation is sent within some pre-defined period of time, the order is automatically cancelled. 
+
+== Order Entity
+Let's have a look on how the Order Entity can be implemented.
+
+[.tabset]
+Java::
++
+.src/main/java/com/example/domain/Order.java
+[source,java]
+----
+include::example$java-reliable-timers/src/main/java/com/example/domain/Order.java[tag=order]
+----
+<1> The first method to look at is the `placeOrder`. It's basically the creation of an order. Note that we set the `placed` field to true.
+<2> When confirming an Order, we must ensure that the Order was created before. 
+<3> If the Order was never created, we return `NOT_FOUND`.
+<4> Cancelling an Order that was never placed also returns `NOT_FOUND`.
+<5> While cancelling an already confirmed order returns `INVALID_ARGUMENT`.
+<6> Finally, if the Order is placed, but not confirmed, the cancel method resets the order to the `emptyState`. This is the equivalent of deleting it.
+
+Scala::
++
+.src/main/scala/com/example/domain/Order.scala
+[source,scala]
+----
+include::example$scala-reliable-timers/src/main/scala/com/example/domain/Order.scala[tag=order]
+----
+<1> The first method to look at is the `placeOrder`. It's basically the creation of an order. Note that we set the `placed` field to true.
+<2> When confirming an Order, we must ensure that the Order was created before. 
+<3> If the Order was never created, we return `NOT_FOUND`.
+<4> Cancelling an Order that was never placed also returns `NOT_FOUND`.
+<5> While cancelling an already confirmed order returns `INVALID_ARGUMENT`.
+<6> Finally, if the Order is placed, but not confirmed, the cancel method resets the order to the `emptyState`. This is the equivalent of deleting it.
+
+== Order Action and Timers
+
+As said before, the `OrderAction` will act as a controller intercepting incoming messages, running some logic and then calling the Order Entity.
+
+=== Scheduling a timer
+
+We will first look at `OrderAction.placeOrder`. Before delegating the request to the Order Entity, the Action creates a timer.
+
+[.tabset]
+Java::
++
+.src/main/java/com/example/actions/OrderAction.java
+[source,java]
+----
+include::example$java-reliable-timers/src/main/java/com/example/actions/OrderAction.java[tags=timers;place-order]
+----
+<1> First we generate a random identifier for the Order. We will use it for identifying the Order, but also as unique name for the timer.
+<2> We call the timers API to register a new timer. Note that it returns `CompletionStage<Done>`. A successul completion means that Kalix registered the timer. 
+<3> Order number is used to generate a unique name for the timer.
+<4> We set the delay we want for the timer to trigger. 
+<5> We scheduled call to `OrderAction.expire` method. We will cover it in a while.
+<6> We build a request for the Order entity. 
+<7> Finally, we build an `asyncReply` by composing the `timerResgistration` CompletionStage with a call to place the order. We access the Order entity through `components.order` and we call `execute()` to run the call. 
+
+Scala::
++
+.src/main/java/com/example/actions/OrderAction.scala
+[source,scala]
+----
+include::example$scala-reliable-timers/src/main/scala/com/example/actions/OrderAction.scala[tags=timers;place-order]
+----
+<1> First we generate a random identifier for the Order. We will use it for identifying the Order, but also as unique name for the timer.
+<2> We call the timers API to register a new timer. Note that it returns `Future[Done]`. A successul completion means that Kalix registered the timer. 
+<3> Order number is used to generate a unique name for the timer.
+<4> We set the delay we want for the timer to trigger. 
+<5> We scheduled call to `OrderAction.expire` method. We will cover it in a while.
+<6> Next we define a method that will make a call to the Order entity.
+<7> We access the Order entity through `components.order` and we call `execute()` to run the call. 
+<8> Finally, we build an `asyncReply` by chaining the `timerResgistration` and the `placeOrder` futures.
+
+In a nutshell, we first requested Kalix to register a timer. When it completes, we know that the timer is persisted and will run at the specified time. We then proceed by placing the order.
+
+The sequence of actions is important here. If we had called the entity first and then registered the timer, the Order could have been placed and the timer registration could have failed due to some network issue for example. In such as case, we would end up with an Order without an expiration timer.
+
+But the inverse is also true. There is still the risk of registering the timer and then failing to place the Order. However, the implementation of the `expire` method can take that into account.
+
+=== Handling the timer call
+
+Let's have a look at the `OrderAction.expire` method implementation.
+
+[.tabset]
+Java::
++
+.src/main/java/com/example/actions/OrderAction.java
+[source,java]
+----
+include::example$java-reliable-timers/src/main/java/com/example/actions/OrderAction.java[tags=timers;expire-order]
+----
+<1> When the `OrderAction` receives the expiration call, it immediately tries to cancel the Order. We use the `execute()` method to run it. This method returns a `CompletionStage`.
+<2> If the `CompletionStage` completes successfuly, we are all good and we can simply return `Empty.getDefaultInstance()`. Since this method is returning normally, the timer will be considered as executed and will be removed from Kalix.
+<3> On the other hand, if the `CompletionStage` completes with a failure, we must decide if we will recover the call or not. If we recover, the timer will be considered as completed. If we will let the call fail, the timer will be re-scheduled. 
+
+Scala::
++
+.src/main/java/com/example/actions/OrderAction.scala
+[source,scala]
+----
+include::example$scala-reliable-timers/src/main/scala/com/example/actions/OrderAction.scala[tags=timers;expire-order]
+----
+<1> When the `OrderAction` receives the expiration call, it immediately tries to cancel the Order. We use the `execute()` method to run it. This method returns a `Future`.
+<2> If the `Future` completes successfuly, we are all good and we can simply return `Empty.defaultInstance`. Since this method is returning normally, the timer will be considered as executed and will be removed from Kalix.
+<3> On the other hand, if the `Future` completes with a failure, we must decide if we will recover the call or not. If we recover, the timer will be considered as completed. If we will let the call fail, the timer will be re-scheduled. 
+
+We have seen the `Order.cancel` implementation and we know that if we get a `NOT_FOUND`, it means that either the Order never existed or it was already cancelled (deleted). Or, we may get an `INVALID_ARGUMENT` error, meaning that the order has been confirmed in the meantime. In both cases, we can consider that the timer has become obsolete and don't need to be rescheduled, therefore we recover the call.
+
+For all other possible errors, the call to `OrderAction.expire` will fail and the timer will be re-scheduled. 
+
+IMPORTANT: Whenever we implement a method that is called from a timer, we need carefully handle errors inside that method. Failing to do so may cause the timer to keep re-scheduling. Therefore, we should ensure that any failure is properly handled and only propagated if the intention is to re-try the call.
+
+=== Cancelling a timer
+
+Next, we can have a look at `OrderAction.confirm` and `OrderAction.cancel` implementations. They are very similar. The only difference being the method they call on the Order entity. 
+
+[.tabset]
+Java::
++
+.src/main/java/com/example/actions/OrderAction.java
+[source,java]
+----
+include::example$java-reliable-timers/src/main/java/com/example/actions/OrderAction.java[tags=timers;confirm-cancel-order]
+----
+<1> We first call the Order entity to execute the command.
+<2> If it succeeds, we remove the timer. 
+
+Scala::
++
+.src/main/java/com/example/actions/OrderAction.scala
+[source,scala]
+----
+include::example$scala-reliable-timers/src/main/scala/com/example/actions/OrderAction.scala[tags=timers;confirm-cancel-order]
+----
+<1> We first call the Order entity to execute the command.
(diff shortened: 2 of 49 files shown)


docs/src/modules/java/pages/timers.adoc
```diff
@@ -0,0 +1,152 @@
+= Timers
+:page-supergroup-java-scala: Language
+
+Timers allow for scheduling calls in the future. For example, to verify that some process have been completed or not.
+
+Timers are persisted in the https://docs.kalix.io/reference/glossary.html#proxy[Kalix proxy] and are guaranteed to run at least once. 
+
+When a timer is triggered, the scheduled call is executed. If successfully excuted, the timer completes and is automatically removed. In case of a failure, the timer is rescheduled to run once more after 3 seconds. This process repeats until the call succeeds.
```
octonato, line 8: The 3 seconds backoff is hard-coded in the config, but that is super arbitrary. We can go with this default for now (or change it), but we should consider adding it as part of the timer registration.

samples/scala-reliable-timers/README.md
```diff
@@ -0,0 +1,61 @@
...
+```
+
+For further details see [Running a service locally](https://docs.kalix.io/developing/running-service-locally.html) in the documentation.
+
+## Exercise the service
+
+With both the proxy and your application running, any defined endpoints should be available at `http://localhost:9000`. In addition to the defined gRPC interface, each method has a corresponding HTTP endpoint. Unless configured otherwise (see [Transcoding HTTP](https://docs.kalix.io/java/writing-grpc-descriptors-protobuf.html#_transcoding_http)), this endpoint accepts POST requests at the path `/[package].[entity name]/[method]`. For example, using `curl`:
+
+```shell
+curl -XPOST -H "Content-Type: application/json" localhost:9000/com.example.CounterService/GetCurrentCounter -d '{"counterId": "foo"}'
```
octonato, line 36: note to self, change it to call the Order service

samples/scala-reliable-timers/src/main/scala/com/example/actions/OrderAction.scala
```diff
@@ -0,0 +1,122 @@
...
+    "order-expiration-timer-" + orderNum.number
+
+  override def placeOrder(orderRequest: OrderRequest): Action.Effect[OrderNumber] = {
+
+    val orderNumber = OrderNumber(UUID.randomUUID().toString) // <1>
+
+    val timerRegistration: Future[Done] = // <2>
+      timers.startSingleTimer(
+        name = timerName(orderNumber), // <3>
+        delay = 5.minutes, // <4>
```
octonato, line 43: This is made 5 minutes to make it more realistic for the docs, but not so convenient if someone wants to play around with the sample 🤷

samples/scala-reliable-timers/src/test/scala/com/example/actions/OrderActionSpec.scala
```diff
@@ -0,0 +1,52 @@
...
+// This class was initially generated based on the .proto definition by Kalix tooling.
+//
+// As long as this file exists it will not be overwritten: you can maintain it yourself,
+// or delete it so it is regenerated as needed.
+
+class OrderActionSpec
+    extends AnyWordSpec
+    with Matchers {
+
+  "OrderAction" must {
```
octonato, line 20: The annoying thing about the Action testkit is that we can't use `components` and `timers`, but this Action use it a lot. 

Now we generate the tests skeleton for the user, but they don't have the means to implement it and they will have to keep this file because of the codegen.

octonato (commented):
Scala sample and docs can be reviewed for now. 

Need to refactor the Java sample to use the Order model and add java/scaladoc.

samples/scala-reliable-timers/src/main/scala/com/example/actions/OrderAction.scala
```diff
@@ -0,0 +1,122 @@
...
+
+import java.util.UUID
+
+import com.example.domain.Order
+import io.grpc.Status
+import io.grpc.StatusRuntimeException
+import org.slf4j.LoggerFactory
+
+// FIXME: we need to provide a default dispatcher for users to use
+import scala.concurrent.ExecutionContext.Implicits.global
```
octonato, line 11: Scala users deserve something better for this.

sdk/scala-sdk/src/main/scala/kalix/scalasdk/action/Action.scala
```diff
@@ -181,6 +186,23 @@ abstract class Action {
     _actionContext = context
   }
 
+  /**
+   * Returns a [[TimerScheduler]] that can be used to schedule further in time.
+   */
+  final def timers: TimerScheduler = {
+
+    val javaActionContextImpl =
+      actionContext("Timers can only be scheduled or cancelled when handling a message.") match {
```
johanandren, line 195: Not sure if it is a huge problem, but, this will limit composability a bit, that you can't do something like:

```scala
val result = asyncTask().flatMap { result => 
  if (result.something) timers.startSingleTimer(some call)
  else Future.successful(whatever)
}
effects().asyncReply(result)
```

Maybe it makes more sense to create the `TimerSchedulerImpl` up front and pass through some internal field of the `ActionContext`? Or simply just always creating it on action creation as a final val.

sdk/java-sdk/src/main/scala/kalix/javasdk/timer/TimerSchedulerImpl.scala
```diff
@@ -0,0 +1,62 @@
...
+import kalix.javasdk.impl.AnySupport
+import kalix.javasdk.impl.DeferredCallImpl
+import kalix.javasdk.impl.GrpcClients
+import kalix.timers.timers.Call
+import kalix.timers.timers.Name
+import kalix.timers.timers.SingleTimer
+import kalix.timers.timers.TimerService
+
+/** INTERNAL API */
+private[kalix] class TimerSchedulerImpl(anySupport: AnySupport, system: ActorSystem) extends TimerScheduler {
```
johanandren, line 36: "If you liked it, make it final"

samples/scala-reliable-timers/src/main/scala/com/example/actions/OrderAction.scala
johanandren: Why, what is wrong with global (or one the user defined/decided on)?

Usually in Akka applications you should use the Akka dispatchers so that futures triggered by actors etc isn't staying around if you terminate the actor system, but for a Kalix app that is less important as the lifecycle is more tied to the JVM which will be killed on stop.

samples/scala-reliable-timers/src/test/scala/com/example/actions/OrderActionSpec.scala
johanandren: Yeah, timers will be easy to do in the testkit as well, as they can just be a stored list of scheduled timers to assert happened or not after a call. I don't think that should be held up by missing support for async.exec-calls, which are quite different, not having a solution.

That would mean a dummy timer service impl in the testkit action runner with access through the ActionTestKit

docs/src/modules/java/pages/timers.adoc
johanandren: To me "once more" reads like it only repeats once which then the next paragraph argues with, I know you love a good argument but maybe replace that and the next section with "this repeats every 3 seconds until the call succeeds"

sdk/scala-sdk/src/main/scala/kalix/scalasdk/action/Action.scala
johanandren: Creating it/injecting up front (maybe through `ActionContext`) would also make testkitting it easier.

build.sbt
```diff
@@ -69,7 +69,7 @@ lazy val sdkJava = project
       "Kalix Java SDK",
       "-noqualifier",
       "java.lang"),
-    Compile / akkaGrpcGeneratedSources := Seq(AkkaGrpc.Server),
+    Compile / akkaGrpcGeneratedSources := Seq(AkkaGrpc.Server, AkkaGrpc.Client), // FIXME we should only have client for TimerService
```
patriknw, line 72: > FIXME we should only have client for TimerService

Wouldn't that be difficult? why is that important?

docs/src/modules/java/pages/timers.adoc
```diff
@@ -0,0 +1,152 @@
...
+
+* Timers are guaranteed to run at least once.
+* Timers can be scheduled to run at any time in the future.
+* Timers can be cancelled.
+* Timers are automatically removed once successfully completed.
+* Timers are re-scheduled in case of failures.
+
+**Timer limitations**: 
+
+* Timers run from the Kalix proxy and therefore won't be run if the service is scaled down to zero instances.
```
patriknw, line 22: Do we talk about "Kalix proxy" in the documentation? I don't think that is something we should describe too much for the user. 

Currently, it's not possible to scale to zero. Later that might be possible, but do we need to document that now?

docs/src/modules/java/pages/timers.adoc
```diff
@@ -0,0 +1,152 @@
...
+----
+
+Scala::
++
+.src/main/scala/com/example/domain/Order.scala
+[source,scala]
+----
+include::example$scala-reliable-timers/src/main/scala/com/example/domain/Order.scala[tag=order]
+----
+<1> The fi'rst method to look at is the `placeOrder`. It's basically the creation of an order. Note that we set the `placed` field to true.
```
patriknw, line 49: typo "fi'rst" ?

docs/src/modules/java/pages/timers.adoc
```diff
@@ -0,0 +1,152 @@
...
+<1> First we generate a random identifier for the Order. We will use it for identifying the Order, but also as unique name for the timer.
+<2> We call the timers API to register a new timer. Note that it returns `Future[Done]`. A successul completion means that Kalix registered the timer. 
+<3> Order number is used as the name for the timer.
+<4> We set the delay we want for the timer to trigger. 
+<5> We scheduled call to `OrderAction.expire` method. We will cover it in a while.
+<6> Next we define a method that will make a call to the Order entity.
+<7> We access the Order entity through `components.order` and we call `execute()` to run the call. 
+<8> Finally, we build an `asyncReply` by chaining the `timerResgistration` and the `placeOrder` futures.
+
+In a nutshell, we first requested Kalix to register a timer. This is an asynchronous call to the Kalix proxy. When it completes, we know that the timer is persisted and will run at the specified time. We then proceed by placing the order.
```
patriknw, line 89: > asynchronous call to the Kalix proxy

isn't that too much implementation internals?

samples/scala-reliable-timers/src/main/scala/com/example/actions/OrderAction.scala
```diff
@@ -0,0 +1,122 @@
...
+    val orderNumber = OrderNumber(UUID.randomUUID().toString) // <1>
+
+    val timerRegistration: Future[Done] = // <2>
+      timers.startSingleTimer(
+        name = timerName(orderNumber), // <3>
+        delay = 5.minutes, // <4>
+        deferredCall = components.orderAction.expire(orderNumber) // <5>
+      )
+
+    def placeOrder: Future[OrderNumber] = // <6>
```
patriknw, line 47: `def placeOrder()` since it has a side effect

sdk/java-sdk/src/main/scala/kalix/javasdk/impl/GrpcClients.scala
```diff
@@ -90,6 +90,9 @@ final class GrpcClients(system: ExtendedActorSystem) extends Extension {
     }
   }
 
+  def getProxyGrpcClient[T](serviceClass: Class[T]): T =
+    getGrpcClient(serviceClass, "localhost", 9000)
```
patriknw, line 94: localhost as we talked about?

sdk/scala-sdk/src/main/scala/kalix/scalasdk/timer/TimerScheduler.scala
```diff
@@ -0,0 +1,87 @@
...
+
+  /**
+   * TODO - document
+   */
+  override def startSingleTimer[I, O](
+      name: String,
+      delay: FiniteDuration,
+      deferredCall: DeferredCall[I, O]): Future[Done] = {
+
+    val timerServiceClient = GrpcClients(system).getProxyGrpcClient(classOf[TimerService])
```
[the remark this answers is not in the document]
patriknw: are you sure? note that this is a new method `getProxyGrpcClient`

samples/scala-reliable-timers/src/main/scala/com/example/actions/OrderAction.scala
octonato: In our examples, we are just transforming data (map/flatMap), so global might be fine, but are we ok in using it for calling external services? Or should we just tell users to defined their own execution context if they do more then map/flatMap?

build.sbt
octonato: I think it will be difficult, indeed. 
This was more a remind to myself to see if something is possible.

docs/src/modules/java/pages/timers.adoc
octonato: We mention the proxy a few times in the docs and we even have a glossary entry for it. 
I can remove the mention, no problem.

samples/scala-reliable-timers/src/main/scala/com/example/actions/OrderAction.scala
patriknw: `global` is just an ordinary ForkJoinPool so that should be fine. Exposing Akka's default execution context could be slightly more efficient because one less thread pool, but maybe not worth it if it becomes additional API that the user has to learn. If we could put Akka's execution context in scope by default that might be nice and convenient.

sdk/java-sdk/src/main/scala/kalix/javasdk/impl/GrpcClients.scala
octonato: Yep, that's what I was mentioning today and that's what I mean that I needed to change something in the SDK.

sdk/scala-sdk/src/main/scala/kalix/scalasdk/action/Action.scala
octonato: Good catch!

I think that injecting the client upfront should work. 

About composability, we will have the same issue with components. We should also break it up there. If possible.

samples/scala-reliable-timers/src/main/scala/com/example/actions/OrderAction.scala
octonato: What about blocking calls? 

My initial idea was to expose Akka's default and blocking through the action context.

build.sbt
octonato: I removed it. As you said, that's not important.

sdk/scala-sdk/src/main/scala/kalix/scalasdk/action/Action.scala
octonato: @johanandren, I think you have already faced that issue before and came out with a workaround.

See https://github.com/lightbend/kalix-jvm-sdk/blob/530c676ad904f40ac434642fdbc19df9994291d9/sdk/java-sdk/src/main/scala/kalix/javasdk/impl/action/ActionRouter.scala#L160-L162

We can't have it as a val in Action. Action is an open interface that anyone can implement and we need to have control over it as we need the actor system and the any support at hand. 

We could add the `timers()` method to the `ActionCreationContext`, but I think it's convenient to have it as a method in Action, like `Action.effects()` because it make it discoverable and because users will interact with timers and effects, so makes sense to keep them next to each other. 

Anyway, I think your idea of only setting it once works. I guess you added it for components, no?

sdk/java-sdk/src/main/scala/kalix/javasdk/impl/timer/TimerSchedulerImpl.scala
```diff
@@ -58,7 +58,7 @@ private[kalix] final class TimerSchedulerImpl(anySupport: AnySupport, system: Ac
 
   def cancel(name: String): CompletionStage[Done] = {
     val timerServiceClient = GrpcClients(system).getProxyGrpcClient(classOf[TimerService])
-    timerServiceClient.remove(Name(name)).asJava.thenApply(_ => Done)
+    timerServiceClient.remove(StringValue(name)).asJava.thenApply(_ => Done)
```
octonato, line 61: I have changed it long time ago in the proxy, but somehow I still had this in my local build and didn't see the error. 🤷

sdk/scala-sdk/src/main/scala/kalix/scalasdk/action/Action.scala
patriknw: was this fixed or still remaining?

sdk/scala-sdk/src/main/scala/kalix/scalasdk/impl/timer/TimerSchedulerImpl.scala
```diff
@@ -0,0 +1,77 @@
...
+import kalix.scalasdk.timer.TimerScheduler
+import kalix.timers.timers.Call
+import kalix.timers.timers.SingleTimer
+import kalix.timers.timers.TimerService
+
+/** INTERNAL API */
+private[kalix] final class TimerSchedulerImpl(anySupport: AnySupport, system: ActorSystem) extends TimerScheduler {
+
+  /**
+   * TODO - document
```
patriknw, line 42: remove

sdk/scala-sdk/src/main/scala/kalix/scalasdk/impl/timer/TimerSchedulerImpl.scala
```diff
@@ -0,0 +1,77 @@
...
+        deferredCallImpl.fullServiceName,
+        deferredCallImpl.methodName,
+        Some(anySupport.encodeScala(deferredCall.message)))
+
+    val singleTimer = SingleTimer(name, Some(call), Some(ProtoDuration(delay.toJava)))
+    timerServiceClient.addSingle(singleTimer).map(_ => Done)(ExecutionContext.parasitic)
+  }
+
+  /**
+   * TODO - document
```
patriknw, line 71: remove

patriknw (commented):
looking good

samples/java-reliable-timers/src/it/java/com/example/OrderIntegrationTest.java
```diff
@@ -0,0 +1,63 @@
...
+  /**
+   * Use the generated gRPC client to call the service through the Kalix proxy.
+   */
+  private final OrderService client;
+
+  public OrderIntegrationTest() {
+    client = testKit.getGrpcClient(OrderService.class);
+  }
+
+  @Test
```
johanandren, line 36:
```suggestion
  @Test
  @Ignore("to be implemented")
```

samples/scala-reliable-timers/src/test/scala/com/example/actions/OrderActionSpec.scala
johanandren: I expect we will add testkit support before releasing the SDK with the timers feature? I created https://github.com/lightbend/kalix-jvm-sdk/issues/957 to track that

sdk/java-sdk/src/main/java/kalix/javasdk/action/Action.java
```diff
@@ -61,6 +64,14 @@ public final Effect.Builder effects() {
     return ActionEffectImpl.builder();
   }
 
+  /** Returns a {@link TimerScheduler} that can be used to schedule further in time. */
+  public final TimerScheduler timers() {
+    ActionContextImpl impl =
+        (ActionContextImpl)
+            actionContext("Timers can only be scheduled or cancelled when handling a message.");
+    return new TimerSchedulerImpl(impl.anySupport(), impl.system());
```
johanandren, line 72: As I said earlier, I think we should create it upfront so that it is always available from the ActionContextImpl, we can then still make the public API for the user this method that delegates to it. That way we can easily inject a stateful pretend-timer in the testkit.

But we can ofc also do that refactoring once we work on testkit support.

sdk/scala-sdk/src/main/scala/kalix/scalasdk/action/Action.scala
johanandren: I think we can leave it be as is and revisit refactoring/pulling it into the ActionContextImpl to be able to provide a pretend timer in the testkit.

johanandren (approved):
LGMT given Patriks small notes and that we follow up with testkit support before releasing.

sdk/scala-sdk/src/main/scala/kalix/scalasdk/action/Action.scala
octonato: As far as I understand, the point originally raised is not about being able to replace the timer for the testkit, but if it's safe to access the actionContext in a map/flatMap/etc. And it's safe for the same reason it's safe to to call `components()` in a map/flatMap.

We can create the TimerScheduler beforehand and set it in the ActionContext, but the fact that we can only access it inside a command handler remains.

samples/java-reliable-timers/src/it/java/com/example/OrderIntegrationTest.java
octonato: We add it to the codegen, but only for unit tests. I will create an issue to track it.

octonato:
All last suggestions implemented. I will merge when green. 

Moving to timers assertion in testkit #957

PR #922, opened 2022-04-25
chore: more clear error message when accessing 'components'

octonato:
This came to the surface when working on the Timers API. 

If a user calls `components()` from outside a command handler, they get an exception with message: "ActionContext is only available when handling a message."

Not so helpful since the user don't see they are accessing the `actionContext`. 

(sending in separated PR to not clutter the Timers API PR)


octonato:
For the record, this PR has two commits. The first one (5a603f7) is the change itself that needs to be reviewed, the second contains only the generated files.

pvlugter (approved):

PR #924, opened 2022-04-25
Codegen: Mark JUnit tests as ignored

ennru:
Marks the unimplemented JUnit test methods with `@Ignored` so they get reported as "skipped", instead of as "success".

References https://github.com/lightbend/kalix/issues/6600


octonato (approved):
LGTM

patriknw:
For Scala we could do the same by using `pending`

PR #926, opened 2022-04-26
fix: better names in suggested Java test code

ennru:
When working with the Maven archetype, I found the suggested variable names inconsistent with what things are called in other places.
I believe this makes the generated unit test code easier applicable.


octonato (approved):
LGTM

PR #928, opened 2022-04-26
chore: more convenient Effect.error API

octonato:
The scala API to report errors is not so convenient because it forces the user to always wrap the status code in a Option. 

It was defined as 

```scala
def error[S](description: String, statusCode: Option[Status.Code] = None):
```

Users use it as:

```scala
error("my error message")
```

or 

```scala
error("my error message", Some(statusCode))
```

So, instead, we should have 

```scala
error("my error message")
error("my error message", statusCode) // no wrapping
```

The optional status code was recently introduced and it was released in a M1. So, probably fine to just change the method signature without deprecation cycle.

### DIFF
--- a/sdk/scala-sdk/src/main/scala/kalix/scalasdk/action/Action.scala
+++ b/sdk/scala-sdk/src/main/scala/kalix/scalasdk/action/Action.scala
@@ -105,14 +105,26 @@ object Action {
        *
        * @param description
        *   The description of the error.
-       * @param statusCode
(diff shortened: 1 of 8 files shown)


pvlugter (approved):
LGTM. Overloaded method, same as the Java API, does look better.

patriknw (approved):
LGTM

patriknw:
Breaking API change, but that's fine together with the rename anyway.

PR #933, opened 2022-04-28
Codegen: Mark ScalaTest tests as pending

ennru:
Fixes https://github.com/lightbend/kalix/issues/6600

Scala counterpart to #924 and #926

### DIFF
--- a/codegen/scala-gen/src/main/scala/kalix/codegen/scalasdk/impl/ActionTestKitGenerator.scala
+++ b/codegen/scala-gen/src/main/scala/kalix/codegen/scalasdk/impl/ActionTestKitGenerator.scala
@@ -111,13 +111,15 @@ object ActionTestKitGenerator {
     val testCases = service.commands.map { cmd =>
       s""""handle command ${cmd.name}" in {\n""" +
       (if (cmd.isUnary || cmd.isStreamOut)
-         s"""|  val testKit = ${actionClassName}TestKit(new $actionClassName(_))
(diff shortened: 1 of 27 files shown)


octonato (commented):
I think it will be better to use `pending` as it express better the situation.

codegen/scala-gen/src/main/scala/kalix/codegen/scalasdk/impl/ActionTestKitGenerator.scala
```diff
@@ -137,12 +137,12 @@ object ActionTestKitGenerator {
           |
           |  "${actionClassName}" must {
           |
-          |    "have example test that can be removed" in {
-          |      val testKit = ${actionClassName}TestKit(new $actionClassName(_))
+          |    "have example test that can be removed" pending {
```
patriknw, line 140: This is not the right syntax. It should be
```
in {
  pending
```

codegen/scala-gen/src/main/scala/kalix/codegen/scalasdk/impl/ActionTestKitGenerator.scala
```diff
@@ -137,12 +137,12 @@ object ActionTestKitGenerator {
...
+          |      val service = ${actionClassName}TestKit(new $actionClassName(_))
           |      // use the testkit to execute a command
```
patriknw, line 142: Shall we have a TODO comment here to hint that this test can be implemented based on the following template and that the `pending` should be removed?

patriknw (approved):
LGTM

octonato (approved):
LGTM

PR #937, opened 2022-04-29
fix: Include build timestamp in image tag, #934

### LINKED ISSUE #934: build of the service should generate version number
Building the image should have a new version so that `kalix service deploy` (`mvn deploy`) actually updates the service with the changes.

We can use something similar as the Akka Platform guide examples.

patriknw:
* so that each deploy will update the service, without having to
  change the version in the pom.xml

If we agree this is the right approach I will update all samples and archetypes.

I considered using https://github.com/git-commit-id/git-commit-id-maven-plugin but two concerns with that:
* it requires a git repository, and doesn't have a fallback if not in git
* LGPL license may be scary

For sbt we already use sbt-dynver for the version number.

Fixes #934


ennru (approved):
LGTM.

PR #938, opened 2022-04-29
fix: Use an absolute resource path to find user-function.desc on classpath

johanandren:
Follow up to #6499


ennru (approved):
What a nice `/`!
LGTM.

PR #940, opened 2022-04-29
chore: re-ordering navigation menu

octonato:
I think we should re-order the items in the navigation menu.

We currently have:

CleanShot 2022-04-29 at 15 44 16@2x

We have a few "Implementing..." items, then some Action related pages and then "Implementing Views"

I suggest the following:

CleanShot 2022-04-29 at 15 43 46@2x

First all the items about how to "Implement" something. 
First the entities, then the view. This is all about manipulating data.

Then followed by "Implementing Actions" and its two more specific usages: "Actions as Controllers" and "Publishing and Subscribing".


ennru (approved):
LGTM.

Issue #942, opened 2022-04-30
Provide execution context to be used by user code


octonato:
It came to my attention that we don't have an execution  context that users can use when working with Futures. 

That's mostly needed for Actions since it's the place we recommend users to deal with async calls.

We could also offer two ExecutionContext. One for general usage and one for blocking operations.

_Originally posted by @octonato in https://github.com/lightbend/kalix-jvm-sdk/pull/921#discussion_r861937779_

johanandren:
Is there a reason they should not use the default global one in Scala?

PR #947, opened 2022-05-03
feat: implicitly available ExecutionContext for Actions in Scala

octonato:
As discussed in https://github.com/lightbend/kalix-jvm-sdk/pull/921#discussion_r862790158

This makes Akka's default dispatcher implicitly available inside Action. The drawback is that we can only access it when handling message as it depends on `actionContext`.

References #942

### DIFF
--- a/sdk/scala-sdk/src/main/scala/kalix/scalasdk/action/Action.scala
+++ b/sdk/scala-sdk/src/main/scala/kalix/scalasdk/action/Action.scala
@@ -158,7 +158,7 @@ abstract class Action {
    * Note that this ExecutionContext is only available when handling a message. It will throw an exception if accessed
    * from constructor.
    */
-  implicit final def executionContext: ExecutionContext = {
+  implicit lazy val executionContext: ExecutionContext = {
     actionContext("ExecutionContext is only available when handling a message") match {
       case ScalaActionContextAdapter(actionContext: ActionContextImpl) => actionContext.system.dispatcher
       // should not happen as we always need to pass ScalaActionContextAdapter(ActionContextImpl)
--- a/sdk/scala-sdk/src/main/scala/kalix/scalasdk/action/Action.scala
+++ b/sdk/scala-sdk/src/main/scala/kalix/scalasdk/action/Action.scala
@@ -17,10 +17,14 @@
 package kalix.scalasdk.action
 
 import scala.collection.immutable.Seq
+import scala.concurrent.ExecutionContext
 import scala.concurrent.Future
+
 import kalix.scalasdk.{ DeferredCall, Metadata, SideEffect }
 import kalix.scalasdk.impl.action.ActionEffectImpl
 import io.grpc.Status
+import kalix.javasdk.impl.action.ActionContextImpl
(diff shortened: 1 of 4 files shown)


sdk/scala-sdk/src/main/scala/kalix/scalasdk/action/Action.scala
```diff
@@ -153,14 +157,31 @@ object Action {
...
   @volatile
   private var _actionContext: Option[ActionContext] = None
 
+  /**
+   * An ExecutionContext to use when composing Futures inside Actions.
+   *
+   * Note that this ExecutionContext is only available when handling a message. It will throw an exception if accessed
+   * from constructor.
+   */
+  implicit final def executionContext: ExecutionContext = {
```
patriknw, line 171: I think this is a good (convenient) default and it's still possible to override if needed for some reason.

patriknw (approved):
LGTM

sdk/scala-sdk/src/main/scala/kalix/scalasdk/action/Action.scala
```diff
@@ -153,14 +157,31 @@ object Action {
...
+    actionContext("ExecutionContext is only available when handling a message") match {
```
patriknw, line 172: but will this work in for example a `flatMap`? shouldn't it be a `lazy val` instead of `def`? Would be good if it worked for all cases except from the constructor.

johanandren: I think it will break for flatMap with another map inside, because that will execute and access the context after the method has returned and the action context cleared.

Is it really important that the ec is from the actor system, could it be a predefined implicit def returning the global ec instead, and always be available with no caveats?

octonato: The action context is never cleared. See my comment here: https://github.com/lightbend/kalix-jvm-sdk/pull/921#discussion_r862882928

patriknw: The exception message "ExecutionContext is only available when handling a message" is not completely correct. The only case it will not work is in the constructor.

I would prefer that we eagerly set it together with the first or all times we set actionContext. The reason is that I'm not sure this will work if you for example make a `deferredCall.execute.flatMap(result => ...)`

johanandren: Ah, did not remember the context is never cleared, then I think this will be fine, except for calling from the constructor but is there really a scenario where that makes any sense?

johanandren (approved):

sdk/scala-sdk/src/main/scala/kalix/scalasdk/action/Action.scala
patriknw: ok, ignore my comment about eagerly set. The exception message might also make sense for the end user, they don't think about threads (as I do).

patriknw (approved):
LGTM

PR #948, opened 2022-05-03
chore: update testcontainers

jrudolph:
We've seen hanging tests while pulling docker images in CI regularly. 1.16.0 had a fix that mentioned pulling, so maybe updating to a more recent version helps...


octonato (approved):
LGTM

PR #951, opened 2022-05-05
chore: regenerate Main class for customer registry

### LINKED ISSUE #945: customer-registry contains Main generated with different SDK version
the current customer-registry has a `Main.java` with a comment with `/*`, but when you re-generate it it has a comment with `//`. We should probably make sure it's generated with the correct version of the SDK?

octonato:
fixes #945


raboof (approved):
great improvement!

Issue #957, opened 2022-05-10
Support for asserting timers in testkit


johanandren:
> Yeah, timers will be easy to do in the testkit as well, as they can just be a stored list of scheduled timers to assert happened or not after a call. I don't think that should be held up by missing support for async.exec-calls, which are quite different, not having a solution.

> That would mean a dummy timer service impl in the testkit action runner with access through the ActionTestKit

_Originally posted by @johanandren in https://github.com/lightbend/kalix-jvm-sdk/pull/921#discussion_r862665266_

PR #965, opened 2022-05-12
feat: testkit timers support

octonato:
References #957

### DIFF
--- a/samples/scala-reliable-timers/src/test/scala/com/example/actions/OrderActionSpec.scala
+++ b/samples/scala-reliable-timers/src/test/scala/com/example/actions/OrderActionSpec.scala
@@ -1,61 +1,47 @@
 package com.example.actions
 
-import com.example.OrderStatus
+import scala.concurrent.Future
+
+import com.example.OrderService
 import com.google.protobuf.empty.Empty
-import kalix.scalasdk.action.Action
-import kalix.scalasdk.testkit.ActionResult
+import kalix.scalasdk.testkit.MockRegistry
+import org.scalamock.scalatest.AsyncMockFactory
+import org.scalatest.concurrent.ScalaFutures
 import org.scalatest.matchers.should.Matchers
-import org.scalatest.wordspec.AnyWordSpec
+import org.scalatest.wordspec.AsyncWordSpec
 
 // This class was initially generated based on the .proto definition by Kalix tooling.
 //
 // As long as this file exists it will not be overwritten: you can maintain it yourself,
 // or delete it so it is regenerated as needed.
 
-class OrderActionSpec
-    extends AnyWordSpec
-    with Matchers {
+class OrderActionSpec extends AsyncWordSpec with Matchers with ScalaFutures with AsyncMockFactory {
 
   "OrderAction" must {
 
-    "have example test that can be removed" in {
-      val service = OrderActionTestKit(new OrderAction(_))
-      pending
-      // use the testkit to execute a command
-      // and verify final updated state:
-      // val result = service.someOperation(SomeRequest)
-      // verify the reply
-      // result.reply shouldBe expectedReply
-    }
-
     "handle command PlaceOrder" in {
-      val service = OrderActionTestKit(new OrderAction(_))
-          pending
-      // val result = service.placeOrder(OrderRequest(...))
-    }
 
-    "handle command Confirm" in {
-      val service = OrderActionTestKit(new OrderAction(_))
-          pending
-      // val result = service.confirm(OrderNumber(...))
-    }
+      val orderService = mock[OrderService]
+      (orderService.placeOrder _)
+        .expects(*)
+        .returning(Future.successful(Empty()))
 
-    "handle command Cancel" in {
-      val service = OrderActionTestKit(new OrderAction(_))
-          pending
-      // val result = service.cancel(OrderNumber(...))
-    }
+      val mockRegistry = MockRegistry.withMock(orderService)
 
-    "handle command Expire" in {
-      val service = OrderActionTestKit(new OrderAction(_))
-          pending
-      // val result = service.expire(OrderNumber(...))
-    }
+      val service = OrderActionTestKit(new OrderAction(_), mockRegistry)
+      val result =
+        service.placeOrder(OrderRequest(item = "Pizza Margherita", quantity = 3))
 
-    "handle command GetOrderStatus" in {
-      val service = OrderActionTestKit(new OrderAction(_))
-          pending
-      // val result = service.getOrderStatus(OrderNumber(...))
+      val timer = result.nextSingleTimerDetails
+      timer.name should startWith("order-expiration-timer")
+      val deferredCall = timer.deferredCallDetails[OrderRequest, Empty]
+
+      deferredCall.serviceName shouldBe "com.example.actions.Order"
+      deferredCall.methodName shouldBe "Expire"
+
+      val callResult = result.asyncResult.futureValue
+      callResult.isError shouldBe false
+      callResult.reply.number shouldNot be(null)
     }
 
   }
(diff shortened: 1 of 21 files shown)


samples/scala-reliable-timers/src/test/scala/com/example/actions/OrderActionSpec.scala
```diff
@@ -12,50 +10,16 @@ import org.scalatest.wordspec.AnyWordSpec
...
-          pending
-      // val result = service.expire(OrderNumber(...))
-    }
-
-    "handle command GetOrderStatus" in {
-      val service = OrderActionTestKit(new OrderAction(_))
-          pending
-      // val result = service.getOrderStatus(OrderNumber(...))
+      val timer = result.nextSingleTimerDetails[OrderNumber, Empty]
+      timer.name should startWith("order-expiration-timer")
```
octonato, line 22: In this example, we can't test much more than this. The problem is that after setting a timer, we do a call to the Order entity component and component calls are not supported by the testkit. 

I'm not so convinced about timers support in the testkit anymore. The limitations we have for the testkit will very block the users in most of the cases.

Of course, we can have actions that simply add a timer, but I do expect the most often than not, timers will be used in combination with component calls. Making unit tests impossible.

The reason this test is passing is because the component call is done in the flatMap of the timer result. So, it doesn't interfere with the tests. 

The other usage of the timers API we have in this Order sample can't even be tested. We cancel the timer when confirming or cancelling an order. But we first call the entity and then cancel the timer. So, the test will fail when calling the component.

sdk/java-sdk/src/main/java/kalix/javasdk/action/Action.java
```diff
@@ -16,6 +16,7 @@
 
 package kalix.javasdk.action;
 
+import akka.actor.ActorSystem;
```
octonato, line 19: leftover, will remove

sdk/java-sdk/src/main/java/kalix/javasdk/action/Action.java
```diff
@@ -60,16 +62,21 @@ public void _internalSetActionContext(Optional<ActionContext> context) {
     actionContext = context;
   }
 
+  /** INTERNAL API */
+  public void _internalSetTimerScheduler(TimerScheduler timerScheduler) {
+    this.timerScheduler = Optional.of(timerScheduler);
+  }
```
octonato, line 68: injectable TimerScheduler

sdk/scala-sdk-testkit/src/main/scala/kalix/scalasdk/testkit/ActionResult.scala
```diff
@@ -48,6 +48,12 @@ trait ActionResult[T] {
    */
   def forwardedTo: DeferredCallDetails[_, T]
 
+  def singleTimerDetails: Seq[SingleTimerDetails[_, _]]
+
+  def nextSingleTimerDetails[I, O]: SingleTimerDetails[I, O]
+
+  def timerCancellations: Seq[String]
```
octonato, line 55: `next` variant for the timer because we can ask and cast to the types from the deferred call.

`timerCancellations` are just strings so can be just iterated without extra help.

sdk/scala-sdk/src/main/scala/kalix/scalasdk/action/Action.scala
```diff
@@ -20,13 +20,13 @@ import scala.collection.immutable.Seq
 import scala.concurrent.ExecutionContext
 import scala.concurrent.Future
 
+import akka.actor.ActorSystem
```
octonato, line 23: leftover, will remove

sdk/scala-sdk/src/main/scala/kalix/scalasdk/action/Action.scala
```diff
@@ -154,21 +154,13 @@ abstract class Action {
...
-  implicit lazy val executionContext: ExecutionContext = {
-    actionContext("ExecutionContext is only available when handling a message") match {
-      case ScalaActionContextAdapter(actionContext: ActionContextImpl) => actionContext.system.dispatcher
-      // should not happen as we always need to pass ScalaActionContextAdapter(ActionContextImpl)
-      case other =>
-        throw new RuntimeException(
-          s"Incompatible ActionContext instance. Found ${other.getClass}, expecting ${classOf[ActionContextImpl].getName}")
-    }
-  }
+  implicit lazy val executionContext: ExecutionContext = ExecutionContext.global
```
octonato, line 163: I started by inject the ActorSystem, but won't work because we don't have one when using the testkit, so need to be `ExCtxt.global`. @johanandren, you will like it, no? :-)

We need the execution context because timers return Future and chances are that users will flatMap with something else. But, as explained above, that something else will probably be a call to component. 

So, using `ExecutionContext.global` kind of solve the problem, but only for a limit number of cases. 

On top of that, the idea of making a `Materializer` always available, as [discussed here](https://github.com/lightbend/kalix-jvm-sdk/pull/947#discussion_r863834641) becomes problematic because of the testkit.

octonato (commented):
Still very early stages, but submitting it already to highlight the issues I found. 

I'm not so convinced anymore that this is something we should have. It's too limited. Better to simply say: to test `timers` and `components` one should write an integration tests.

samples/scala-reliable-timers/src/test/scala/com/example/actions/OrderActionSpec.scala
octonato: in retrospect, I don't think that a method `nextSingleTimerDetails[In, Out]` makes much sense in the context here. 

`In/Out` are the types of the deferred call, not the timer. So maybe better to have an expected method in the `SingleTimerDetails` to check if the DeferredCall it holds has the expect type.

octonato:
I re-worked this a bit. 

Still draft and Java API is missing. Looking for feedback on the Timers Testkit API before I continue.

sdk/scala-sdk-testkit/src/main/scala/kalix/scalasdk/testkit/impl/TestKitTimerScheduler.scala
```diff
@@ -0,0 +1,59 @@
...
+import kalix.javasdk.impl.DeferredCallImpl
+import kalix.scalasdk.DeferredCall
+import kalix.scalasdk.impl.ScalaDeferredCallAdapter
+import kalix.scalasdk.testkit.SingleTimerDetails
+import kalix.scalasdk.timer.TimerScheduler
+
+class TestKitTimerScheduler extends TimerScheduler {
+
+  private var _singleTimers: Seq[SingleTimerDetails] = Seq.empty
+  private var _timerCancellations: Seq[String] = Seq.empty
```
johanandren, line 32: Probably needs to be `@volatile` since mutable and _maybe_ written to from future/cs callbacks (even if the opposite, set timer, do async stuff, is probably more common)

sdk/scala-sdk-testkit/src/main/scala/kalix/scalasdk/testkit/impl/TestKitTimerScheduler.scala
```diff
@@ -0,0 +1,59 @@
...
+    _singleTimers = _singleTimers :+ TestKitSingleTimerDetails(name, delay, TestKitDeferredCall(deferredCallImpl))
+
+    Future.successful(Done)
+  }
+
+  override def cancel(name: String): Future[Done] = {
+    _timerCancellations = _timerCancellations :+ name
+    Future.successful(Done)
+  }
+}
```
johanandren, line 59: API looks good to me

johanandren (approved):

octonato:
I'm closing this PR as it is too old and stale. It turns out that when testing timers, it's much better to write integration tests. So maybe not that interesting after all.

PR #966, opened 2022-05-14
fix: correctly pass Action and View options in the Java SDK

raboof:


ennru:
Scala compilation fails https://app.circleci.com/pipelines/github/lightbend/kalix-jvm-sdk/3507/workflows/a70034aa-2ac1-4b13-9179-cdfddb653a94/jobs/43295?invite=true#step-108-1270

sdk/java-sdk/src/main/scala/kalix/javasdk/impl/action/ActionsImpl.scala
```diff
@@ -47,9 +47,18 @@ final class ActionService(
     val factory: ActionFactory,
     override val descriptor: Descriptors.ServiceDescriptor,
     override val additionalDescriptors: Array[Descriptors.FileDescriptor],
-    val anySupport: AnySupport)
+    val anySupport: AnySupport,
+    val actionOptions: Option[ActionOptions])
```
octonato, line 51: Maybe we could use `ActionOptions.defaults` instead of wrapping with a `Option`

octonato:
@raboof, I fixed this PR and did a force push on your branch because I rebased it with main.

sdk/java-sdk/src/test/scala/kalix/javasdk/impl/action/ActionHandlerSpec.scala
```diff
@@ -68,7 +70,8 @@ class ActionHandlerSpec
   private val anySupport = new AnySupport(Array(ActionspecApi.getDescriptor), this.getClass.getClassLoader)
 
   def create(handler: ActionRouter[_]): Actions = {
-    val service = new ActionService(_ => handler, serviceDescriptor, Array(), anySupport)
+    val actionFactory: ActionFactory = _ => handler
+    val service = new ActionService(actionFactory, serviceDescriptor, Array(), anySupport, None)
```
octonato, line 74: For some reason, after adding the `Option[ActionOptions]`, the type inference stopped to work for `ActionFactory`. That's why I'm had to type it explicitly. 🤷

octonato (approved):
LGTM, but I did one commit in this PR (ea1745634f432d1f9940e223f7bc5e7719268772), so I won't merge it myself.

codegen/scala-gen/src/main/scala/kalix/codegen/scalasdk/impl/SourceGenerator.scala
```diff
@@ -139,7 +139,7 @@ object SourceGenerator {
       protoFileName = "",
       name = "",
       protoPackage = packageString,
-      javaPackageOption = None,
+      javaPackageOption = Some(packageString),
```
raboof, line 142: I know I did this myself, but looking at it again with fresh eyes it doesn't necessarily make sense to me anymore: if we initialize it with `Some`, does it need to be an `Option` at all?

raboof (commented):
Thanks for finishing this up. we should probably verify if the `javaPackageOption` is really needed here, otherwise LGTM.

codegen/scala-gen/src/main/scala/kalix/codegen/scalasdk/impl/SourceGenerator.scala
octonato: I think we should keep it. Or change in another PR in any case. 

As far as I can tell, when we fill it from a descriptor, it can be empty: 
https://github.com/lightbend/kalix-jvm-sdk/blob/c5c537144006d425a3087b28aa9a772dfb0376fe/codegen/core/src/main/scala/kalix/codegen/MessageType.scala#L177-L178

PR #973, opened 2022-05-18
quickstart docs: improve styling

ennru:
* add command line styling to more commands in the Quickstart docs for Java and Scala
* align the Java and Scala pages more
* fix an `akkaserverless` leftover


raboof (approved):
(didn't generate the docs, but looks good from the diff!)

PR #975, opened 2022-05-18
quickstart: add emptyState in instructions

ennru:
Cover adding the `emptyState` implementation in the Customer Registry quickstart instructions for Java and Scala.


johanandren (approved):
LGTM

PR #981, opened 2022-05-23
feat: ACL support

johanandren:
References https://github.com/lightbend/kalix/issues/6808

Started out by adding support for picking up a `kalix_proxy.proto` with defaults for the entire Kalix service.
Will add to maven archetype as well, and _maybe_ do docs in this PR a bit later.


samples/scala-doc-snippets/src/main/proto/com/example/kalix_policy.proto
```diff
@@ -0,0 +1,14 @@
...
+package com.example;
+
+import "kalix/annotations.proto";
+
+// Allow other Kalix services deployed in the same project to access the components of this
+// Kalix service, but disallow access from the internet. This can be overridden explicitly
+// per component or method using annotations.
+option (kalix.file).acl = {
+  allow: { principal: ALL }
+  deny: { principal: INTERNET }
```
jroper, line 13: This could be simplified (currently at least) to:

```
allow: { service: "*" }
```

jroper:
Hey @johanandren, we've both done some work documenting ACLs here, me for the project generally, you for the Java/Scala APIs. I think it does make sense to have documentation in both places, but there's probably some unnecessary overlap. The general documentation can probably be more abstract, while the docs here more concrete. Take a look here:

https://github.com/lightbend/kalix-docs/pull/1339

Any thoughts? Perhaps the stuff around development and testing that I've written should be moved here.

johanandren:
Removed the parts that are already in the main doc PR and added a little bit about principal access and testkit support that is being added in #988

samples/java-customer-registry-views-quickstart/src/it/java/customer/view/CustomerByEmailIntegrationTest.java
```diff
@@ -54,7 +54,7 @@ public void findByEmail() throws Exception {
         CustomerViewModel.ByEmailRequest.newBuilder().setEmail("foo@example.com").build();
 
     // // the view is eventually updated
-    await().ignoreExceptions().atMost(10, SECONDS).until(() -> viewClient.getCustomer(req).toCompletableFuture()
+    await().ignoreExceptions().atMost(20, SECONDS).until(() -> viewClient.getCustomer(req).toCompletableFuture()
```
johanandren, line 57: Unrelated but kept failing for every push :/

jroper (approved):

PR #988, opened 2022-05-26
feat(acl): self identity, and ACL APIs

jroper:
This does a number of things:

* Ensures that calls to self are authenticated using the self identification token passed by the discovery call.
* Ensures that calls to other services have this services identity attached (only applies in dev/test, since in prod the service mesh handles this).
* Adds an API to provide access to the principals associated with a request via the request metadata.
* Adds settings to the testkit to allow starting the proxy with ACLS disabled, and also to set the name of the service that the proxy should run as in order to support authenticating cross service calls.
* Adds an additional method to the testkit to get a gRPC client authenticated as a given principal.

I haven't yet looked at the Scala SDK, only the Java, and I haven't written tests for this yet. This is depending on a snapshot build of the proxy, so we need to wait for a release.


jroper:
@johanandren take a look at this.

jroper:
It may make sense to merge our two PRs together.

sdk/java-sdk/src/main/scala/kalix/javasdk/impl/GrpcClients.scala
```diff
@@ -71,37 +74,67 @@ final class GrpcClients(system: ExtendedActorSystem) extends Extension {
...
+    getGrpcClient(serviceClass, service, port = 80, remoteAddHeader)
 
   /** Local gRPC clients point to services (user components or Kalix services) in the same deployable */
   private def getLocalGrpcClient[T](serviceClass: Class[T]): T = {
-    selfServiceName match {
-      case Some("localhost") => getGrpcClient(serviceClass, "localhost", selfPort.getOrElse(9000))
-      case Some(selfName)    => getGrpcClient(serviceClass, selfName)
+    proxyHostname match {
+      case Some("localhost") => getGrpcClient(serviceClass, "localhost", selfPort.getOrElse(9000), localAddHeader)
+      case Some(selfName)    => getGrpcClient(serviceClass, selfName, 80, localAddHeader)
```
johanandren, line 109: Not strictly related to this PR, more of a future improvement perhaps: if we pass proxy port to the info as well rather than the guess-logic here it will make running locally a bit less with docker in any scenario where the ports are changed (multiple services for example).

johanandren (commented):
LGTM 

but we can't merge until 1.0.7 is out

Edit: and Scala API as well ofc

sdk/java-sdk/src/main/scala/kalix/javasdk/impl/GrpcClients.scala
jroper: Agreed.

jroper: Although, one of the challenges with docker running locally is that the proxy might not know the port that the user function needs to access it on, since this is specified by a port-forward.

jroper:
@johanandren I've added the equivalent Scala SDK API, and some tests for accessing the principal metadata.

What I haven't tested, that would be really good to test, is the testkit, to ensure we are setting the principal headers correctly in that and configuring the proxy correctly, but, I'm not sure what the best approach for writing those tests are. We can't make them regular tests because they depend on docker, and our regular tests as run in circle don't have Docker available currently -  I think, though maybe that does work with testcontainers? I don't know. What we might need to do is create an integration test scope, and put the tests there, and then either run them when we run the TCK tests, or create a new job to run them.

jroper:
Ok, so the failure seen now in the TCK tests is interesting.

The way I've done the dev-mode principal identification is, if there's no headers, it's assumed that the request is from the internet. But this can be overridden using a `impersonate-kalix-service` header. The result is that when the TCK invokes commands, there's a `_kalix-src: internet` header in the metadata, and the TCK is failing on the presence of that.

So, maybe we should update the TCK to tolerate those headers? Or even require them? Not sure...

johanandren:
We've pushed quite a bit of the test coverage to the samples (because of convenience), so I think we could do the same for ACLs

A few of the Java samples has integration tests that are run by the CI, but none of the Scala ones (fixmes all over there), we should probably re-enable Scala ones as well, I've created a separate issue for that (https://github.com/lightbend/kalix-jvm-sdk/issues/1002) 

For the TCK, gut feeling is that we could make it ignore the header for now, it's not clear to me that there is anything ACL-related that should go in the TCK.

johanandren:
PR:ed ignoring principal header (actually all metadata) in TCK in https://github.com/lightbend/kalix-proxy/pull/1479

johanandren (approved):
LGTM

PR #992, opened 2022-05-31
chore: add auto label for spring-sdk

octonato:


.github/labeler.yml
```diff
@@ -4,15 +4,20 @@ kalix-proxy:
   - '**/*'
 
 java-sdk:
-  - sdk/java/**/*
+  - sdk/java-sdk/**/*
+  - sdk/java-sdk-testkit/**/*
```
octonato, line 8: it seems that we had it wrong for java-sdk and scala-sdk

.github/labeler.yml
```diff
@@ -4,15 +4,19 @@ kalix-proxy:
...
   - maven-java/**/*
   - codegen/java-gen/**/*
-  - codegen/java-gen-compilation-tests/**/*
```
octonato, line 10: this one doesn't exist anymore

PR #998, opened 2022-06-02
feat: transpose RequestMapping to proto http options

octonato:
References https://github.com/lightbend/kalix/issues/6859

Still very minimal, no validation so far, but enough for preliminary tests and moving with proxy work.


octonato:
Oh, realising now that we can also update the sample in that same PR. Will update it now.

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/ProtoDescriptorGenerator.scala
```diff
@@ -108,12 +110,54 @@ object ProtoDescriptorGenerator {
...
+
+            // those below are not supported in by the proto http annotations
+            case RequestMethod.OPTIONS | RequestMethod.HEAD | RequestMethod.TRACE =>
+              throw new IllegalArgumentException(s"Unsupported Http method: $httpMethod")
+          }
+
+        }
+        .getOrElse {
+          // non-annotated methods get an url like in gRPC, eg: full-qualified-name/method-name
+          ruleBuilder.setPost(s"/${component.getName}/${method.getName}")
```
octonato, line 143: hmm, wondering now why we are even bothering doing that. 
Would we get this already for free from the proto?

octonato: Well, I think the initial thought was to provide an easier name for the user and not so long as the fully-qualified name. 

But when implementing, I decide to use the FQN to avoid any conflict, but now I realise that it doesn't make sense to add it if we use the FQN.

octonato:
> I rebased this PR on top of #999 to avoid conflicts.

And I got conflicts. Tried to fix in GitHub UI and now I have compile time errors! Yeah!

johanandren (approved):
Looks like a good start

PR #1000, opened 2022-06-02
chore: use Collections.empty()

octonato:
Follow-up to #991


johanandren (approved):
LGTM but I think there was one more in a default method in an interface.

PR #1004, opened 2022-06-07
Some support for spring annotations

jroper:
This is what I have so far for supporting `@PostRequest`/`@GetRequest` etc. The request body mapping doesn't seem to be working, not sure why, but right with this method:

```java
  @GetMapping("/echo/{number}")
  public Effect<Number> echoNumber(@PathVariable Long number) {
    return effects().reply(new Number(number));
  }
```

I can do this:

```
$ curl http://localhost:9000/echo/20
{"value":20}
```

While this method:

```java
  @PostMapping("/echo")
  public Effect<Message> echoMessage(@RequestParam("add") String add, @RequestBody Message msg) {
    return effects().reply(new Message(msg.value + add));
  }
```

is failing:

```
$ curl http://localhost:9000/echo -H "Content-Type: application/json" -d '{"value":"hello"}'
Invalid Any type.
```

### DIFF
--- a/sdk/java-sdk/src/main/java/kalix/javasdk/Kalix.java
+++ b/sdk/java-sdk/src/main/java/kalix/javasdk/Kalix.java
@@ -124,7 +124,7 @@ public final class Kalix {
           new ResolvedActionFactory(actionFactory, anySupport.resolveServiceDescriptor(descriptor));
 
       ActionService service =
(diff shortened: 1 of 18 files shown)


jroper:
By the way, my code is quite poorly structured, there's a lot of stuff in action specific classes that will need to be made more general.

octonato:
This one is superseded by #1006. Closing it.

PR #1006, opened 2022-06-07
Some support for spring annotations

octonato:
Supersedes #1004

@jroper, I fixed the conflict and worked a little further on it. I decide to not push directly on your branch because I touch quite some code. 

The main change is that I removed the previous `Serializer` we had and I merged the `MessageInDecoder` and `MessageOutEncoder` into one single class, `MessageCodec`. Then, all code depends on it and `AnySupport` is just the protobuf variant of it. See 66bdaf9edee52d9be007304dcc8bcdcb560288e0

If I understand it correctly, we don't need anything that is in `AnySupport` for the Spring SDK because we either use Json, or we keep the message as proto until we dynamically extract the data to pass to the Java method. So, basically, the code is now either using `AnySupport` or `SpringSdkMessageCodec`. 

Well, that means that we won't be able to mix generated proto message with json. But we can review this later.

We will need more testing around the parsing and such. I couldn't manage to run the spring-java-fibonacci-action sample, for instance. I got two different kind of errors, but didn't investigate it further.

First, it seems that the parsing requires that there is at least one param in the path for a POST. I know, this is far from complete. 

Second, I get the following message when deploying: 
`com.example.fibonacci.NextNumberRequest.num: "num" is already defined in "com.example.fibonacci.NextNumberRequest"`. 

`num` is a path param that I added to workaround the first issue. Anyway, I can investigate it further tomorrow. 

Concretely, I ask you to review 66bdaf9edee52d9be007304dcc8bcdcb560288e0 and eventually merge this PR and continue on top of it. If you don't agree with my changes, then just continue on your original PR and we sync tomorrow morning.


sdk/java-sdk/src/main/java/kalix/javasdk/Kalix.java
```diff
@@ -119,31 +117,20 @@ public Kalix registerAction(
...
       ActionFactory resolvedActionFactory =
           new ResolvedActionFactory(actionFactory, anySupport.resolveServiceDescriptor(descriptor));
 
-      ActionService service =
-          new ActionService(resolvedActionFactory, descriptor, additionalDescriptors, anySupport);
-
-      services.put(descriptor.getFullName(), system -> service);
-
-      return Kalix.this;
+      return registerAction(resolvedActionFactory, anySupport, descriptor, additionalDescriptors);
```
octonato, line 123: I think we only need a `ResolvedActionFactory` when using protobuf. The Spring SDK won't need this.

sdk/java-sdk/src/main/java/kalix/javasdk/Kalix.java
```diff
@@ -379,11 +369,19 @@ public Kalix register(ViewProvider provider) {
...
    */
   public Kalix register(ActionProvider provider) {
-    return lowLevel.registerAction(
-        provider::newRouter,
-        provider.serviceDescriptor(),
-        provider.additionalSerializers(),
-        provider.additionalDescriptors());
+
+    // FIXME: type inference complaining about lack of type param in ActionProvider<>
+    Optional<MessageCodec> codecOpt = provider.alternativeCodec();
```
octonato, line 374: inference is broken here, not big issue, but could be nice to fix that later.

sdk/java-sdk/src/main/java/kalix/javasdk/action/ActionProvider.java
```diff
@@ -40,7 +38,7 @@
 
   Descriptors.FileDescriptor[] additionalDescriptors();
 
-  default Map<Class<?>, Serializer> additionalSerializers() {
-    return Collections.emptyMap();
+  default Optional<MessageCodec> alternativeCodec() {
+    return Optional.empty();
```
octonato, line 42: We will use `AnySupport` when this is empty. Otherwise, we use the alternative codec defined by the provider.

sdk/java-sdk/src/main/scala/kalix/javasdk/impl/AnySupport.scala
```diff
@@ -397,16 +381,8 @@ class AnySupport(
         ScalaPbAny(BytesPrimitive.fullName, primitiveToBytes(BytesPrimitive, byteString))
 
       case other =>
-        // last chance: try to find a suitable JacksonProto serializer
-        additionalSerializers
```
octonato, line 401: this was added to support serialization using protobuf generated from Jackson lib. We don't need it anymore.

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/action/ActionIntrospector.scala
```diff
@@ -0,0 +1,91 @@
...
+
+    new ActionDescription(fileDescriptor, serviceDescriptor, methods)
+  }
+
+}
+
+class ActionDescription[A <: Action](
+    val fileDescriptor: Descriptors.FileDescriptor,
+    val serviceDescriptor: Descriptors.ServiceDescriptor,
+    val methods: Map[String, ActionMethod])
```
octonato, line 91: I renamed it to `ActionDescription`. It was previously `SpringSdkActionFactory`. We don't need another factory because we have the `Provider` already. 

We could eventually call it `ComponentDescription` and have something like `ComponentMethod` instead of `ActionMethod`. I think it will be the same for all the components, no?

sdk/spring-sdk/src/test/scala/kalix/springsdk/impl/SpringSDKTestRunner.scala
```diff
@@ -0,0 +1,28 @@
...
+
+import kalix.javasdk.Kalix
+import kalix.springsdk.action.EchoAction
+import kalix.springsdk.action.ReflectiveActionProvider
+
+// Temporary class just for testing
+object SpringSDKTestRunner extends App {
+  val kalix = new Kalix()
+  kalix.register(ReflectiveActionProvider.of(classOf[EchoAction], _ => new EchoAction))
+  kalix.start().toCompletableFuture.get()
```
octonato, line 27: Refactor this test runner to use the register method. This is important because the previous version was not detecting an issue with the `ResolvedActionFactory`. 

The issue is now solved because we won't use a `ResolvedActionFactory` when the provider has a alternative `MessageCodec`.

octonato (commented):
left some components to explain some changes I did

jroper:
> If I understand it correctly, we don't need anything that is in AnySupport for the Spring SDK because we either use Json, or we keep the message as proto until we dynamically extract the data to pass to the Java method. So, basically, the code is now either using AnySupport or SpringSdkMessageCodec.

It's possible that this is the case, but I'm not 100% sure. In particular, I don't know how serialization of forwards/effects should/will be handled.

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/action/ActionIntrospector.scala
jroper: Yes probably - a lot of this code was written just to be able to test the message and HTTP annotation generators against a proxy, I did the minimum to get actions working.

PR #1012, opened 2022-06-09
feat(spring-sdk): Value Entity support

octonato:


sdk/java-sdk/src/main/java/kalix/javasdk/valueentity/ValueEntity.java
```diff
@@ -56,6 +58,17 @@ public void _internalSetCommandContext(Optional<CommandContext> context) {
...
+  /** INTERNAL API */
+  public void _internalSetCurrentState(S state) {
+    currentState = Optional.ofNullable(state);
+  }
+
+  protected final S currentState() {
+    return currentState.orElseThrow(
+        () ->
+            new IllegalStateException("Current state can only available when handling a command."));
+  }
```
octonato, line 70: We may need to introduce this new API, because otherwise it will be quite alien to have entity methods with Rest annotations, but with params that are not participating on the REST API.

For example: 

```java
 @PostMapping("/user/{id}/create")
    public ValueEntity.Effect<Done> createUser(User user, @RequestBody CreateUser createUser) {
      return effects().reply(Done.instance);
    }
```

In that example, `User` can only be passed by Kalix and doesn't not participate on the REST API.

sdk/java-sdk/src/main/scala/kalix/javasdk/impl/valueentity/ValueEntitiesImpl.scala
```diff
@@ -167,12 +164,12 @@ final class ValueEntitiesImpl(system: ActorSystem, val services: Map[String, Val
 
           val serializedSecondaryEffect = effect.secondaryEffect match {
             case MessageReplyImpl(message, metadata, sideEffects) =>
-              MessageReplyImpl(service.anySupport.encodeJava(message), metadata, sideEffects)
+              MessageReplyImpl(service.messageCodec.encodeScala(message), metadata, sideEffects)
```
octonato, line 167: We have been encoding with Java and then converting to Java, like in: `ScalaPbAny.fromJavaProto(service.anySupport.encodeJava(newState)`

This was the only place where we were not wrapping. Just using Scala now, should be ok.

sdk/spring-sdk/src/test/java/kalix/springsdk/valueentity/RestAnnotatedValueEntities.java
```diff
@@ -0,0 +1,41 @@
...
+
+import kalix.javasdk.valueentity.ValueEntity;
+import kalix.springsdk.annotations.EntityKey;
+import org.springframework.web.bind.annotation.PostMapping;
+import org.springframework.web.bind.annotation.RequestBody;
+import org.springframework.web.bind.annotation.RequestMapping;
+
+public class RestAnnotatedValueEntities {
+
+  @EntityKey("id")
```
johanandren, line 27: Could it make sense a convention with a `entityKey` path parameter being the id so that this annotation is optional and only needed if some other name is used, less noise (but a bit more magic)?

johanandren (commented):
Lookin' good

sdk/java-sdk/src/main/java/kalix/javasdk/valueentity/ValueEntity.java
johanandren: Can we have intermediate component baseclasses in the Spring SDK rather than extend the core Java SDK with features only needed for the Spring SDK?

octonato: I consider this option as well, but the drawback is that users will have access to the same classes of the Java SDK. 
We will have to duplicate them (not an issue) and then we will have: `kalix.javasdk.valueentity.ValueEntity` and  `kalix.springsdk.valueentity.ValueEntity` to import. 

But may not be a major issue:
* we can emit a nice message showing that the wrong type was used. Also, if users pick the Java SDK one, they won't have access to the state, so they should realise it quite soon. If they try have a method receiving the State, it won't work neither. 
* I think we will need a `@KalixComponent` annotation anyway. If possible, we should restrict it's usage to some Spring SDK interface. If not possible, we can do it with annotations processors at compile time. I'm not talking about generated code, but to fail compilation if we see that `@KalixComponent` is being used in a class that is one of the allowed ones.

sdk/spring-sdk/src/test/java/kalix/springsdk/valueentity/RestAnnotatedValueEntities.java
octonato: The problem is that if people get it wrong, they won't know it until deploying and trying to hit the entity.

Maybe we could enforce it as well. Like any entity type must have a paths containing `entityKey` or have the annotation.

octonato: Let's explore this on another PR. We can probably enforce it and eventually catch it at compile time with annotation processors.

Btw, at least for entities, we will need some annotation at type level. We need to have the `entityType` string. Can be `@ValueEntity("foo")` or generic one `@KalixComponent("foo")`.

octonato:
Oh, this closed because I changed the name of the branch. I will send a new soon.

PR #1013, opened 2022-06-09
fix: More restrictive ACLs by default in quickstarts

johanandren:
Docs already claim this but it wasn't completely added to all archetypes and quickstarts

This puts a default ACL allowing cross-service calls but not access from the outside, and then specifically opens up individual gRPC services to public access

References https://github.com/lightbend/kalix/issues/6808

### DIFF
--- a/samples/java-customer-registry-kafka-quickstart/src/main/proto/customer/kalix_policy.proto
+++ b/samples/java-customer-registry-kafka-quickstart/src/main/proto/customer/kalix_policy.proto
@@ -1,4 +1,4 @@
-// This is the default ACLs for all components of this Kalix Service
+// This is the default Access Control List (ACL) for all components of this Kalix Service
 syntax = "proto3";
 
 package customer;
@@ -8,6 +8,7 @@ import "kalix/annotations.proto";
 // Allow all other Kalix services deployed in the same project to access the components of this
 // Kalix service, but disallow access from the internet. This can be overridden explicitly
 // per component or method using annotations.
+// Documentation at https://docs.kalix.io/services/using-acls.html
 option (kalix.file).acl = {
   allow: { service: "*" }
 };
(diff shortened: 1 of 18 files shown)


samples/java-customer-registry-kafka-quickstart/src/main/proto/customer/kalix_policy.proto
```diff
@@ -0,0 +1,13 @@
+// This is the default ACLs for all components of this Kalix Service
```
ennru, line 1:
```suggestion
// This is the default Access Control List (ACL) for all components of this Kalix Service
```
Maybe write it out in these files?

samples/java-customer-registry-kafka-quickstart/src/main/proto/customer/kalix_policy.proto
```diff
@@ -0,0 +1,13 @@
...
+syntax = "proto3";
+
+package customer;
+
+import "kalix/annotations.proto";
+
+// Allow all other Kalix services deployed in the same project to access the components of this
+// Kalix service, but disallow access from the internet. This can be overridden explicitly
+// per component or method using annotations.
```
ennru, line 10:
```suggestion
// per component or method using annotations.
// Documentation at https://docs.kalix.io/services/using-acls.html
```

ennru (approved):
Add some more guidance.
LGTM.

kheyron-oscar-bou:
Excuse me, but for cyber-resilient systems one of most stablished principles are "security by design" and "security by default". If Kalyx services manage PII, regulations like EU GDPR legally mandates "privacy by default".
So basically services should not be accesible "by default" from the outside, i.e. Internet principals (and, applying the principle at the most restrictive level, it could be not accesible until explicitely define an ACL allowing access from other project services or Internet).

HTH

johanandren:
Hi, @kheyron-oscar-bou, Kalix services are not accessible from the public internet at all until explicitly exposed. 

What the new ACL support adds is a more fine grained control over what components of a service that becomes available when a deployed Kalix service is exposed with a public route.

What this PR changes is a restrictive default in quickstart and project templates so that each gRPC Service needs explicit opening up to the public internet even if the deployed service itself is exposed.

PR #1015, opened 2022-06-10
feat(spring-sdk): Value Entity support

octonato:

PR #1018, opened 2022-06-10
chore: specific spring sdk components

octonato:
Follow up on: https://github.com/lightbend/kalix-jvm-sdk/pull/1015#discussion_r894468350

This is working more or less ok. We have thin layer of Spring SDK components and some basic runtime validation to ensure that users don't use Java SDK components classes. 

So far so good, but when coding against this we get IntelliJ offering both component versions and that's not great. But we can probably do something with Java Modules. I don't have experience with it, but worth trying to block direct usage of Java SDK components classes.

### DIFF
--- a/samples/spring-java-fibonacci-action/src/main/java/com/example/fibonacci/FibonacciAction.java
+++ b/samples/spring-java-fibonacci-action/src/main/java/com/example/fibonacci/FibonacciAction.java
@@ -1,6 +1,6 @@
 package com.example.fibonacci;
 
-import kalix.javasdk.action.Action;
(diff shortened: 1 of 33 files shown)


octonato:
One DX issue that came to my attention when hacking a sample app, is that users now need to know that they have to type `Action.Effect` or `ValueEntity.Effect`. And again, they should pay attention to use the ones from the Spring SDK.

In the codegen based SDKs, the methods are generated for the user with the types already imported and in place for them.

octonato:
closing this, I don't think is something we need for now.

PR #1019, opened 2022-06-13
feat(codegen): avoid touching regenerated files if nothing has changed

jrudolph:


octonato (approved):
LGTM

raboof (approved):
nice

jrudolph:
> Does it work when the file does not exist yet?

No :) But I hopefully now fixed it...

pvlugter (approved):

jrudolph:
Btw. the purpose behind this was to speed up subsequent maven builds which otherwise recompile everything all the time. Turns out, maven still recompiles everything if only a single source file changed, so this change only helps a bit. However, you can turn on incremental compilation in maven using:

```
-Dmaven.compiler.useIncrementalCompilation=false
```

And yes, `false`actually turns incremental compilation on (see https://issues.apache.org/jira/browse/MCOMPILER-209).

Issue #1016, opened 2022-06-10
EntityKey - where should it be required/allowed?


octonato:
_Originally posted by @octonato in https://github.com/lightbend/kalix-jvm-sdk/pull/1015#discussion_r894470949_

Relates to https://github.com/lightbend/kalix-jvm-sdk/issues/1014

After experimenting a bit with `@EntityKey`, I believe that we should only have it as a type level annotation. 

For methods (ElementType.METHOD), I think it will be much more intuitive if we simply say that there is only one place where the user can add this annotation. The only reason that I see to allow it at method level would be to do the following:

(pseudo java code)

```java
@RequestMapping("/foo")
class Foo
  @EntityKey("idOne")
  @PostRequest("/{idOne}/one")
  String methodOne(...)

  @EntityKey("idTwo")
  @PostRequest("/{idTwo}/two")
  String methodTwo(...)
```

I don't see a strong reason to allow it. At the end, this is only the name of a field. Why would we offer so much flexibility? Which feature would we be enabling? 

Users can still add the `{id}` to paths in methods, like in, although I think the best is to add `/foo/{id}` at type level and not repeat yourself.

```java
@EntityKey("id")
@RequestMapping("/foo")
class Foo
  @PostRequest("/{id}/one")
  String methodOne(...)

  @PostRequest("/{id}/two")
  String methodTwo(...)
```

For parameters (ElementType.PARAMETER), I believe the idea was to combine it with `@RequestParam` and query strings. In that case, I think we could wait with that feature. 

I think it makes lots of sense to enforce a convention that entity keys must be part of a the path. It's intuitive, it's simple. We can start small and re-evaluate the need later.

PR #1020, opened 2022-06-13
feat(spring-sdk): basic value changes subscription

octonato:
References https://github.com/lightbend/kalix/issues/6859

Working in progress, not working yet for unknown reasons. 🤷 

Also, intention is to get a basic evening.in for Value Entiies, so we can move on to Views.

### DIFF
--- a/sdk/spring-sdk/src/main/java/kalix/springsdk/annotations/Entity.java
+++ b/sdk/spring-sdk/src/main/java/kalix/springsdk/annotations/Entity.java
@@ -23,5 +23,6 @@ import java.lang.annotation.*;
 @Documented
 public @interface Entity {
   String entityType();
+
   String[] entityKey();
 }
--- a/sdk/spring-sdk/src/main/java/kalix/springsdk/annotations/Subscribe.java
+++ b/sdk/spring-sdk/src/main/java/kalix/springsdk/annotations/Subscribe.java
@@ -18,7 +18,6 @@ package kalix.springsdk.annotations;
 
 import java.lang.annotation.*;
 
-
 public @interface Subscribe {
 
   @Target(ElementType.METHOD)
@@ -27,5 +26,4 @@ public @interface Subscribe {
   @interface ValueEntity {
     Class<? extends kalix.javasdk.valueentity.ValueEntity<?>> value();
   }
-
 }
--- a/sdk/spring-sdk/src/main/java/kalix/springsdk/valueentity/ReflectiveValueEntityProvider.java
+++ b/sdk/spring-sdk/src/main/java/kalix/springsdk/valueentity/ReflectiveValueEntityProvider.java
@@ -42,20 +42,18 @@ public class ReflectiveValueEntityProvider<S, E extends ValueEntity<S>>
   private final Descriptors.ServiceDescriptor serviceDescriptor;
   private final ComponentDescription componentDescription;
 
-  public static <S, E extends ValueEntity<S>> ReflectiveValueEntityProvider<S, E> of(Class<E> cls, Function<ValueEntityContext, E> factory) {
-    return new ReflectiveValueEntityProvider<>(
-         cls, factory, ValueEntityOptions.defaults());
+  public static <S, E extends ValueEntity<S>> ReflectiveValueEntityProvider<S, E> of(
+      Class<E> cls, Function<ValueEntityContext, E> factory) {
+    return new ReflectiveValueEntityProvider<>(cls, factory, ValueEntityOptions.defaults());
   }
 
   public ReflectiveValueEntityProvider(
-
-      Class<E> entityClass,
-      Function<ValueEntityContext, E> factory,
-      ValueEntityOptions options) {
+      Class<E> entityClass, Function<ValueEntityContext, E> factory, ValueEntityOptions options) {
 
     Entity annotation = entityClass.getAnnotation(Entity.class);
     if (annotation == null)
-      throw new IllegalArgumentException("Value Entity [" + entityClass.getName() + "] is missing '@Entity' annotation");
+      throw new IllegalArgumentException(
+          "Value Entity [" + entityClass.getName() + "] is missing '@Entity' annotation");
 
     this.entityType = annotation.entityType();
 
--- a/sdk/spring-sdk/src/test/java/kalix/springsdk/testmodels/valueentity/ValueEntitiesTestModels.java
+++ b/sdk/spring-sdk/src/test/java/kalix/springsdk/testmodels/valueentity/ValueEntitiesTestModels.java
@@ -25,7 +25,9 @@ import org.springframework.web.bind.annotation.RequestMapping;
 
 public class ValueEntitiesTestModels {
 
-  @Entity( entityKey = {"userId", "cartId"}, entityType = "user")
+  @Entity(
+      entityKey = {"userId", "cartId"},
+      entityType = "user")
   @RequestMapping("/user/{userId}/{cartId}")
   public static class PostWithEntityKeys extends ValueEntity<User> {
     @Override


octonato:
Ok, it seems that this is not working because the proxy is not yet capable of sending json payload wrapped in Any when streaming events (projections).

sdk/spring-sdk/src/main/java/kalix/springsdk/annotations/Subscribe.java
```diff
@@ -0,0 +1,31 @@
...
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+public @interface Subscribe {
+
+  @Target(ElementType.METHOD)
+  @Retention(RetentionPolicy.RUNTIME)
+  @interface ValueEntity {
+    String entityType();
+  }
```
octonato, line 30: I would like to experiment with this new naming. If we feel that is more intuitive, we should change the proto annotations to reflect it. 

Basically, when annotating a class with this, we end-up with: 

```java
@Subscribe.ValueEntity(entityType = "counter")
  public Effect<Done> changes(@RequestBody CounterState counterState) {
    logger.info("Counter subscriber: counter id '{}' is '{}'", counterState.id, counterState.value);
    return effects().reply(Done.instance);
  }
```

I believe this reads better. "Subscribe to the ValueEntity with entityType defined as 'counter'"

If we would make it similar to the current protobuf annotations, it will be something like:
`@EventingIn.ValueEntity(source = "counter")`

octonato: The idea is to have the following:

```java
@Subscribe.ValueEntity(entityType = "counter")

@Subscribe.EventSourcedEntity(entityType = "counter")

@Subscribe.Topic(name = "counter")
```
and then 

```java
@Publish.Topic(name = "counter")
```

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/Introspector.scala
```diff
@@ -41,12 +46,16 @@ object Introspector {
...
-    val messageDescriptors = dynamicRestMethods.map { method =>
-      grpcService.addMethod(method.method)
-      method.descriptor
+    val messageDescriptors = methodsInfo.flatMap { methodInfo =>
+      grpcService.addMethod(methodInfo.grpcMethod)
+      // Input message descriptor can be a None. That happens when the desired type is Any.
+      // In such case, we should not add an `Any` as a message to the generated proto file
+      // And we can't properly filter it out because that's a DescriptorProto and it has no package information.
+      // Therefore, our best option is to use a None to exclude it.
+      methodInfo.inputMessageDescriptor
```
octonato, line 58: I wish I could have found a better solution. Not obvious why it's a None. Info is spread through different parts of the code.

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/ProtoDescriptorGenerator.scala
```diff
@@ -43,7 +44,11 @@ object ProtoDescriptorGenerator {
 
     protoBuilder.addDependency("google/protobuf/any.proto")
     protoBuilder.addService(service)
-    messages.foreach(protoBuilder.addMessageType)
+    messages
+      .filterNot { desc =>
+        desc.getName == JavaPbAny.getDescriptor.getFullName
+      }
```
octonato, line 50: Leftover, I will remove. We can't filter it in a reliable way. 

This is filtered already by using `None`, btw

johanandren (commented):
Lookin' good, added some comments.

sdk/spring-sdk/src/test/java/kalix/springsdk/testmodels/subscriptions/SubscriptionsTestModels.java
```diff
@@ -0,0 +1,41 @@
...
+  public static class SubscribeToValueEntityAction extends Action {
+
+    @Subscribe.ValueEntity(entityType = "ve-counter")
+    @PostMapping("/message/one")
+    public Action.Effect<Message> messageOne(@RequestBody Message message) {
+      return effects().reply(message);
+    }
+
+    @Subscribe.ValueEntity(entityType = "ve-counter")
+    @PostMapping("/message/two")
```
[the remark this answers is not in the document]
octonato: Correct,  I'm aware of it. That's a small regression. 

On the first version we had, if the method were not annotated, we would create a POST using `full.qualified.ServiceName/MethodName`. 

We have an issue to solve that.

sdk/spring-sdk/src/test/scala/kalix/springsdk/impl/IntrospectionSuite.scala
```diff
@@ -41,6 +40,18 @@ trait IntrospectionSuite extends Matchers {
...
+  def findEventSource(desc: ComponentDescription, methodName: String): EventSource = {
+    val grpcMethod = desc.serviceDescriptor.findMethodByName(methodName)
+    val methodOptions = grpcMethod.toProto.getOptions.getExtension(kalix.Annotations.method)
+    methodOptions.getEventing.getIn
+  }
+
+  def assertSubscriptionMethod(method: ComponentMethod) = {
+//    val field = findField(method, fieldName)
+//    val fieldOption = field.toProto.getOptions.getExtension(kalix.Annotations.field)
+//    fieldOption.getEntityKey shouldBe true
```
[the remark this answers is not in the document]
octonato: Oeps!

sdk/spring-sdk/src/main/java/kalix/springsdk/annotations/Subscribe.java
octonato: Maybe, when subscribing to entities from another service we will want to have another kind of annotations.

johanandren: Yes, enforcing at runtime/startup is what I meant, I don't think the type system can help at compile time.

Can't we enforce the value entity / es enttity type of consumption anyway by looking at what type the entity subscribed to is?

octonato: We can enforce at compile time, but then we need `@Subscribe.ValueEntity` and `@Subscribe.EventSourcedEntity`. 
Or by introducing a common trait, like `PersistentEntity` (too exclude replicated ones) and having a single annotation: `@Subscribe.Entity`

However, I think we should have two distinct ones so we can block the usage of  `@Subscribe.EventSourcedEntity` at type level. 

Another reason for distinct annotations is that it will make clear that we can't subscribe to any kind of entities, but only for VE and ES.

sdk/spring-sdk/src/test/java/kalix/springsdk/testmodels/subscriptions/SubscriptionsTestModels.java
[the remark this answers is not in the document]
octonato: For the record: https://github.com/lightbend/kalix-jvm-sdk/issues/1022

sdk/spring-sdk/src/main/java/kalix/springsdk/annotations/Subscribe.java
octonato: I tried it out. It works.

sdk/spring-sdk/src/main/java/kalix/springsdk/annotations/Entity.java
```diff
@@ -16,13 +16,12 @@
...
 
-@Target({ElementType.TYPE})
+@Target(ElementType.TYPE)
 @Retention(RetentionPolicy.RUNTIME)
-public @interface EntityKey {
-  String[] value() default {};
+@Documented
+public @interface Entity {
+  String entityType();
+  String[] entityKey();
```
octonato, line 26: @johanandren, I replaced EntityKey by this one. 

Defining an Entity requires the entityKey and entityType, always. So easier if user needs to deal with only one annotation. And both fields are mandatory. 

Also, as discussed here: https://github.com/lightbend/kalix-jvm-sdk/issues/1016
I don't think we should allow entityKeys in different places.

sdk/spring-sdk/src/main/java/kalix/springsdk/annotations/Subscribe.java
```diff
@@ -0,0 +1,31 @@
...
+
+
+public @interface Subscribe {
+
+  @Target(ElementType.METHOD)
+  @Retention(RetentionPolicy.RUNTIME)
+  @Documented
+  @interface ValueEntity {
+    Class<? extends kalix.javasdk.valueentity.ValueEntity<?>> value();
+  }
```
octonato, line 29: Keep distinct annotations so we can enforce that the entity is either `ValueEntity` or `EventSourcedEntity`.

sdk/spring-sdk/src/main/java/kalix/springsdk/valueentity/ReflectiveValueEntityProvider.java
```diff
@@ -42,22 +42,27 @@
...
-      Class<E> cls,
+
+      Class<E> entityClass,
       Function<ValueEntityContext, E> factory,
       ValueEntityOptions options) {
-    this.entityType = entityType;
+
+    Entity annotation = entityClass.getAnnotation(Entity.class);
+    if (annotation == null)
+      throw new IllegalArgumentException("Value Entity [" + entityClass.getName() + "] is missing '@Entity' annotation");
```
octonato, line 58: This validation won't be needed when we have autowire. We won't build a reflective provider if the entity is not annotated with `@Entity`. 

On the other hand, we may need to validate that `entityType` and `entityKey` are not empty string, but this can be validated when wiring things up and not on each provider.

johanandren: For the API-first services we don't require entityType but synthesise it if not specified, so I guess we need to deal with that as well if we don't already,

johanandren: Maybe I misremembered and we always require it there as well.

octonato: We do require it. We can't create one based on class name because users may change the class name and then they lose the connection to the existing journal.

PR #1021, opened 2022-06-14
chore: minor improvement to updatePomVersions.sh script

octonato:


updatePomVersions.sh
```diff
@@ -33,6 +29,15 @@ sbt publishM2;publishLocal
   git checkout */pom.xml
 )
 
+if [ "$1" = "all" ]; then
+  PROJ="samples/*java-*"
+elif [ $1 ]; then
+  PROJ=$1  
+else 
+  PROJ=""
```
raboof, line 37: what happens in this case? should we throw an error and show how to invoke the script?

octonato: When we choose to not pass an argument, we only publish the artifacts locally. 

But I think we should merge this script with the `publishLocalM2` one. 

So, no argument, only publish local. With arguments, updated the sample to run locally.

I never use the `publishLocalM2` script because I want to update the pom.xml and open the project in IntelliJ. 

Let me update this PR by merging both scripts.

franciscolopezsancho: Would make sense just delete `publishLocalM2` and use only `updatePomVersion`?

octonato: yep, I think we should have only one script. It's just that I keep postponing the task. I will do it one day. 😄

octonato:
Reviving this PR. 

Since I created it (very long time ago), the state of our scripts changed a lot. 

This PR is now: 
* introduces `publishLocally.sh` which also runs `sbt publishLocal;publishM2`
* removes old `publishLocalM2.sh`
* ` updateSamplesVersions.sh` that can update individual samples for local testing

publishLocally.sh
```diff
@@ -1,16 +1,24 @@
+
 # this script updates all maven projects versions (maven-java and samples) to align with the current sdk version
 # if SDK_VERSION env var is defined, it will use it, otherwise it will take the version from sbt
 # after running this script, you may run local tests or simply send a PR with the updates. 
 
 # you can pass the path to a sample to only change the version for single sample, 
 # eg: ./updatePomVersions.sh samples/java-eventsourced-counter
+# or you can pass 'all' to it and all java samples (pom.xml) will be updated
+# eg: ./updatePomVersions.sh all
```
efgpinto, line 9: In need of update? since it's not ùpdatePomVersions.sh` anymore..

efgpinto (approved):
small comment. LGTM otherwise.

publishLocally.sh
octonato: Well, the whole docs were wrong. Updated now.

efgpinto: indeed 👍

PR #1030, opened 2022-06-22
ci: add untested samples to workflow

pvlugter:
The `scala-eventsourced-counter` sample was reported as not working (in a slack discussion).

A couple of samples have jobs that are not included in the CI workflow. There's a verification that each sample has a CI job, but not whether these are in the workflow and actually run. May need a little more trickery for that verification.

Adding the missing samples, one of which should fail.

### DIFF
--- a/samples/java-eventsourced-counter/src/main/proto/com/example/actions/counter_topic.proto
+++ b/samples/java-eventsourced-counter/src/main/proto/com/example/actions/counter_topic.proto
@@ -48,9 +48,6 @@ service CounterJournalToTopic {
   } 
 
   rpc IncreaseConditional (com.example.domain.ValueIncreased) returns (Increased) { 
-    option (kalix.method).eventing.in = { 
-      event_sourced_entity: "counter" 
-    };
     option (kalix.method).eventing.out = {
       topic:  "counter-events"
     };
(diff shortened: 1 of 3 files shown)


pvlugter:
Updated the broken event sourced counter samples. Removed the double subscription to value increased events — the entity will double based on metadata, and emit just value increased. Leaving the action method for testing metadata when calling method directly. Have also enabled the integration tests for the java version of this sample. Left the FIXME note about eventing integration tests. Unclear to me why scala projects have their integration tests under `test`, rather than `it`, but left that as is as well.

pvlugter:
Updated the verification check that all samples are tested, with both testing that the directory exists somewhere in the CI config, and that the job name appears twice — for the job definition and included in the workflow.

samples/java-eventsourced-counter/src/main/proto/com/example/actions/counter_topic.proto
```diff
@@ -48,9 +48,6 @@ service CounterJournalToTopic {
   } 
 
   rpc IncreaseConditional (com.example.domain.ValueIncreased) returns (Increased) { 
-    option (kalix.method).eventing.in = { 
-      event_sourced_entity: "counter" 
-    };
```
johanandren, line 53: Aha, so two eventing in methods for the same event?

johanandren (approved):

PR #1032, opened 2022-06-22
chore: update link to kalix forum in release issue template

efgpinto:
Just noticed the release issue template stills points to the old forum.. I guess we can update that.

### DIFF
--- a/docs/release-issue-template.md
+++ b/docs/release-issue-template.md
@@ -38,7 +38,7 @@ You can see the proxy version on prod [on grafana](https://lightbendcloud.grafan
 
 ### Announcements
 
-- [ ] Announce in [the forum](https://discuss.lightbend.com/c/kalix/40)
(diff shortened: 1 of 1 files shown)


docs/release-issue-template.md
```diff
@@ -38,7 +38,7 @@ You can see the proxy version on prod [on grafana](https://lightbendcloud.grafan
 
 ### Announcements
 
-- [ ] Announce in [the forum](https://discuss.lightbend.com/c/kalix/40)
-  - tag with Kalix, releases and announcement
+- [ ] Announce in [the forum](https://discuss.kalix.io/)
+  - tag with releases and announcement
```
efgpinto, line 42: Don't think these tags already exist on the new forum but probably worth creating them apart from the `kalix` one?

johanandren (approved):

Issue #955, opened 2022-05-06
Status code not being mapped to http code


octonato:
The following method returns NOT_FOUND

```scala
  override def confirm(currentState: OrderState, confirmRequest: example.ConfirmRequest): ValueEntity.Effect[Empty] =
    if (currentState.placed) // <2>
      effects
        .updateState(currentState.copy(confirmed = true))
        .thenReply(Empty.defaultInstance)
    else
      effects
        .error(
          s"No order found for ${confirmRequest.orderNumber}", 
          Status.Code.NOT_FOUND) // <3>
```

When calling it with CURL, we get: 

```bash
curl -XPOST -v -H "Content-Type: application/json" http://localhost:9000/com.example.OrderService/Confirm -d '{ "order_number": "b6c47a60-e7fe-4632-af9d-f59f65467b8e" }'
Note: Unnecessary use of -X or --request, POST is already inferred.
*   Trying 127.0.0.1:9000...
* Connected to localhost (127.0.0.1) port 9000 (#0)
> POST /com.example.OrderService/Confirm HTTP/1.1
> Host: localhost:9000
> User-Agent: curl/7.79.1
> Accept: */*
> Content-Type: application/json
> Content-Length: 58
>
* Mark bundle as not supporting multiuse
< HTTP/1.1 500 Internal Server Error
< Server: akka-http/10.2.9
< Date: Fri, 06 May 2022 18:26:47 GMT
< Content-Type: text/plain; charset=UTF-8
< Content-Length: 55
<
* Connection #0 to host localhost left intact
No order found for b6c47a60-e7fe-4632-af9d-f59f65467b8e%
```

Note that it returns 500, not 404 as one would expect.

johanandren:
That's surprising, I was close to 100% sure that the HTTP transcoding does exactly that error code mapping for us here: https://github.com/lightbend/kalix-proxy/blob/main/proxy/core/src/main/scala/kalix/proxy/http/HttpEndpointMethod.scala#L152

johanandren:
Is the right gRPC code returned if not transcoded? Maybe the code is not passed correctly in the reply/error at all from the SDK?

johanandren:
I tried this out with the current quickstart both locally (proxy 1.0.1 which the quickstart pulls in) and actually deployed to kalix and I cannot reproduce. 

With gRPC I get `NotFound/5` and curl/HTTP `NotFound/404` as expected.

PR #1036, opened 2022-06-28
fix: propagate GRPC status code to response for entities

efgpinto:
References #955 

This was working correctly but only for actions, not for entities.


sdk/java-sdk/src/main/scala/kalix/javasdk/impl/effect/SecondaryEffectImpl.scala
```diff
@@ -38,7 +38,8 @@ sealed trait SecondaryEffectImpl {
         Some(
           ClientAction(
             ClientAction.Action
-              .Failure(kalix.protocol.component.Failure(commandId, failure.description))))
+              .Failure(kalix.protocol.component
+                .Failure(commandId, failure.description, grpcStatusCode = failure.status.map(_.value()).getOrElse(0)))))
```
johanandren, line 42: 👍

johanandren (approved):
LGTM!

PR #1037, opened 2022-06-30
chore: allow override of proxy version by config

efgpinto:
The goal is to allow external config of the proxy version dependency. One example would be when trying to run the integration tests of an app using the SDK but against a specific proxy version.


patriknw (approved):
LGTM

johanandren (approved):

PR #1039, opened 2022-06-30
feat: Ignore eventing.in methods

patriknw:
* bump: kalix-proxy 1.0.12
* skip codegen of methods marked with `ignore: true` in Actions and Views
* they will be handled by the proxy, without calling the user function


codegen/java-gen/src/main/scala/kalix/codegen/java/ActionTestKitGenerator.scala
```diff
@@ -141,9 +143,9 @@ object ActionTestKitGenerator {
   }
 
   def generateServices(service: ModelBuilder.ActionService): String = {
-    require(!service.commands.isEmpty, "empty `commands` not allowed")
```
patriknw, line 144: I don't know why we had this `require` in the testkit for Java? It's not in the testkit for Scala.
With this additional filter of ignore methods it can be possible to mark all methods as ignore (even though it doesn't make sense).

codegen/java-gen/src/main/scala/kalix/codegen/java/ActionTestKitGenerator.scala
```diff
@@ -141,9 +143,9 @@ object ActionTestKitGenerator {
...
+    val commands = service.commands.filterNot(_.ignore)
```
johanandren, line 146: Isn't that previous validation still valid? If an action has no methods something is not right. Maybe also only ignore-methods because that wouldn't make sense either.

johanandren (approved):
Looks good, one question about the previous validation now removed.

codegen/java-gen/src/main/scala/kalix/codegen/java/ActionTestKitGenerator.scala
patriknw: I don't know why we had this require in the testkit for Java? It's not in the testkit for Scala and not in the actual action generation.
With this additional filter of ignore methods it can be possible to mark all methods as ignore (even though it doesn't make sense).

johanandren: Ah, it's in the testkit, not in the action generator. I had a feeling we did something like that in the action generator after some user feedback. But then it's fine removing I guess.

patriknw:
There is a bug in the proxy for the View validation when using `ignore: true`. I'm fixing it.

pvlugter:
> Do we need to adjust the JavaScript/TypeScript SDK to support this?

Yes, will be similar changes. The codegen is using an older approach, but since it was all in the same project originally, it's still similar:

- https://github.com/lightbend/kalix-javascript-sdk/blob/main/codegen/core/src/main/scala/io/kalix/codegen/ModelBuilder.scala
- https://github.com/lightbend/kalix-javascript-sdk/blob/main/codegen/js-gen/src/main/scala/io/kalix/codegen/js/ActionServiceSourceGenerator.scala
- and so on ...

Docs are maybe not as developed, but have the ignore events example for subscribing:

https://github.com/lightbend/kalix-javascript-sdk/blob/main/docs/src/modules/javascript/pages/actions-publishing-subscribing.adoc

samples/java-eventsourced-customer-registry/src/main/proto/customer/view/customer_view.proto
```diff
@@ -67,15 +67,18 @@ service CustomerByName {
...
+      event_sourced_entity: "customers"
+      ignore: true
      };
      option (kalix.method).view.update = {
        table: "customers"
-       transform_updates: true // <6>
+       transform_updates: true
      };
   };
+  // end::ignore-other-events[]
```
octonato, line 81: It's kind of awkward that we need to add 4 annotations for a method that is here just to be ignored. 
Are the `view.update` annotations needed? 

A service level annotation saying that it's partial will be handy.

patriknw: Good point. It works fine without the update annotation for ignore methods. I have removed them in https://github.com/lightbend/kalix-jvm-sdk/pull/1039/commits/2bf36dc83a0d4df9f7175191f2dd94cd4103fc61

PR #1040, opened 2022-07-01
feat(spring-sdk): view support

octonato:
References https://github.com/lightbend/kalix/issues/6859

PR #1046, opened 2022-07-06
Spring json based views

johanandren:
Continues on top of #1040

Depends on snapshot of proxy from json-view-PR so won't pass PR validation until we release that

 * Adds jackson-to-proto schema generation so we can annotate views with a type declaring how the actual json in views are parsed into indexes from updates and for query validation

Lots to do still.


johanandren:
Sample project now starts up and proxy from json view PR discovers and fails with: 

```
{"timestamp":"2022-07-08T15:41:04.825Z","thread":"kalix-akka.actor.default-dispatcher-2","logger":"kalix.javasdk.impl.DiscoveryImpl","message":"Error reported from Kalix system: KLX-01122 Invalid JSON annotations in com.example.CustomerByNameView.GetCustomer: Defined json_body_input_field 'json_body' does not exist on the query request message type 'com.example.GetCustomerRequest'\n\nAt CustomerByNameView.proto","context":"default","severity":"ERROR"}
```

Because the query does not have any request body but it always adds the annotation. First up next week is to fix that.

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/reflection/KalixMethod.scala
```diff
@@ -0,0 +1,197 @@
...
+ *
+ * @param callable
+ *   Is this actually a method that will ever be called or just there for metadata annotations?
+ */
+case class SpringRestServiceMethod(
+    classMapping: Option[RequestMapping],
+    mapping: RequestMapping,
+    javaMethod: Method,
+    params: Seq[RestMethodParameter],
+    callable: Boolean = true)
```
octonato, line 91: In which case will that be false? 

As far I understand it, methods that only exist for metadata are covered by `VirtualServiceMethod`. Those are synthetic methods that we will create if the `@Subscription` annotation is added at type level, meaning that there is no transformation and therefore no method to call in the user function.

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/ComponentDescriptor.scala
```diff
@@ -0,0 +1,340 @@
...
+
+  private def buildSyntheticMessageAndExtractors(
+      nameGenerator: NameGenerator,
+      serviceMethod: SpringRestServiceMethod,
+      httpRule: HttpRule.Builder,
+      entityKeys: Seq[String] = Seq.empty): (DescriptorProto, Map[Int, ExtractorCreator]) = {
+
+    // FIXME the name depends on order of calls to name generator which could be problematic
+    // (for example sharding passing the request to a different Kalix node where the name was generated
+    // to have 2 instead of 1 in the msg name because of class path ordering not being consistent?)
```
octonato, line 194: Good point. 
I wasn't happy with the NameGenerator using a global counter. It feels weird and unsafe. 
Shall append a hash build from method name and all it's params names? Maybe worth to check how the `Method` hashCode is build.

octonato: hmm, not unique enough: 

```java
/**
     * Returns a hashcode for this {@code Method}.  The hashcode is computed
     * as the exclusive-or of the hashcodes for the underlying
     * method's declaring class name and the method's name.
     */
    public int hashCode() {
        return getDeclaringClass().getName().hashCode() ^ getName().hashCode();
    }
```

Overloaded methods will have the same hashCode

johanandren: Ok, wont methods named the same in different components end up with the same request message name and clash then?

octonato: They will be isolated by the service descriptor. Each java class has its own generated descriptor containing all the messages. In that case, I think we are fine.

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/reflection/ReflectionUtils.scala
```diff
@@ -0,0 +1,29 @@
...
+
+import java.lang.reflect.Method
+import java.util
+
+object ReflectionUtils {
+
+  private implicit val stringArrayOrdering: Ordering[Array[String]] =
+    Ordering.fromLessThan(util.Arrays.compare[String](_, _) < 0)
+  implicit val methodOrdering =
+    Ordering.by((m: Method) => (m.getName, m.getReturnType.getName, m.getParameterTypes.map(_.getName)))
```
octonato, line 27: We are still not protecting against nodes with different versions during a rolling upgrade.

johanandren: At least now this won't produce a broken cluster that stays broken once the roll completed.

octonato:
> @octonato one annoying thing with merging this is that it will fail all other CI builds on the branch since we have no proxy release with the json schema, maybe we can publish a proxy snapshot somehow?

I think we can cut a proxy version, no? What is you concern? That we will publish a version that has some new protocol that may still change? 

We can create a tag suffixed with RC1 or M1, or even some hash.

johanandren:
Found issues in the proxy view json support after writing that, so need to fix those first.

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/ProtoDescriptorGenerator.scala
```diff
@@ -46,6 +49,10 @@ object ProtoDescriptorGenerator {
     messages.foreach(protoBuilder.addMessageType)
 
     // finally build all final descriptor
-    Descriptors.FileDescriptor.buildFrom(protoBuilder.build, dependencies)
+    val fd = Descriptors.FileDescriptor.buildFrom(protoBuilder.build, dependencies)
+    if (logger.isDebugEnabled) {
+      logger.debug("Generated file descriptor for service [{}]: {}", name, ProtoDescriptorRenderer.toString(fd))
```
octonato, line 54: This render is doing a great job, but this log puts the first line of the descriptor together with the main logging message. After that the rest appears beautifully formatted. 

I'm sure you did it to drive me crazy. It was fun, but can we add a break line now? 😆 

```suggestion
      logger.debug("Generated file descriptor for service [{}]: \n{}", name, ProtoDescriptorRenderer.toString(fd))
```

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/ProtoDescriptorRenderer.scala
```diff
@@ -0,0 +1,106 @@
...
+import scala.collection.mutable
+
+object ProtoDescriptorRenderer {
+
+  def toString(fileDescriptor: FileDescriptor): String = {
+    // not water tight but better than the default non-protobuf Proto-toString format
+    val proto = fileDescriptor.toProto
+    val builder = new mutable.StringBuilder()
+    proto.getMessageTypeList.forEach { messageType =>
+      builder ++= "message "
```
octonato, line 30: Or maybe better to start the render with a break line? 
```suggestion
      builder ++= "\n"
      builder ++= "message "
```

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/ViewDescriptorFactory.scala
```diff
@@ -0,0 +1,208 @@
...
+    builder.setView(view)
+  }
+
+  private def valueEntityStateClassOf(valueEntityClass: Class[_]): Class[_] = {
+    valueEntityClass.getGenericSuperclass
+      .asInstanceOf[ParameterizedType]
+      .getActualTypeArguments
+      .head
+      .asInstanceOf[Class[_]]
+  }
```
octonato, line 207: We can generify this one as 'getFirstTypeParam' or something like that. 
We are doing the same for the View type param

octonato (approved):
LGTM

I left a few comments here and there, but all minor stuff. We can move to merging this and create follow-up PRs.

PR #1049, opened 2022-07-11
bump: update proxy to 1.0.13

pvlugter:


pvlugter:
Expected to fail a couple of integration tests that use side effects.

pvlugter:
Hmm, so the event sourced counter samples are failing on the integration test, where the method side effects to itself:

```
{"timestamp":"2022-07-11T05:08:52.372Z","thread":"kalix-akka.actor.default-dispatcher-8","logger":"kalix.javasdk.impl.DiscoveryImpl","message":"Error reported from Kalix system:  Service failed to start because of an unexpected error: UNAVAILABLE: Network closed for unknown reason","context":"default","severity":"ERROR"}
[info] CounterServiceIntegrationSpec:
[info] CounterService
[info] - must handle side effect that adds the initial input multiplied by two *** FAILED *** (506 milliseconds)
[info]   The future returned an exception of type: io.grpc.StatusRuntimeException, with message: UNAVAILABLE: Network closed for unknown reason. (CounterServiceIntegrationSpec.scala:33)
```

Reintroducing the readiness delay of 3 seconds fixes this (recently reduced to 100 ms for dev proxy).

Seems odd that we rely on that delay here, so could be something to follow up on.

pvlugter:
Assuming it works because it delays the eventing startup. And eventing is failing during start? I'll look more closely at this tomorrow.

pvlugter:
So the testkit assumes google pubsub, which is not started for these integration tests, and so it fails when starting eventing. Would be good to add a more useful error message to say what failed to start — just has an unavailable exception currently. It passed earlier because the eventing startup was delayed enough that the test could run and finish first.

Start the google pubsub emulator for the CI test in the way it used to be for old integration tests, using a docker compose file. Reverted the readiness delay.

There's also already a FIXME for adding integration tests that use eventing.

PR #1051, opened 2022-07-13
feat(spring-sdk): auto-registration of kalix components

octonato:
* KalixServer moved to SDK project 
* samples now using it instead of own KalixServer version

References https://github.com/lightbend/kalix/issues/7366


samples/spring-customer-registry-views-quickstart/src/main/java/com/example/Main.java
```diff
@@ -1,11 +1,14 @@
...
 
+import kalix.springsdk.impl.KalixServer;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.boot.SpringApplication;
 import org.springframework.boot.autoconfigure.SpringBootApplication;
+import org.springframework.context.annotation.Import;
 
 @SpringBootApplication
+@Import(KalixServer.class)
```
octonato, line 11: Spring Boot will scan files under package `com.example` and sub-pacakges. This is the package of the main function. We need to import the KalixServer like that so Spring can find it and start it.

sdk/java-sdk/src/main/java/kalix/javasdk/Kalix.java
```diff
@@ -337,7 +337,7 @@ public <D extends ReplicatedData, E extends ReplicatedEntity<D>> Kalix register(
    *
    * @return This stateful service builder.
    */
-  public <S, E extends ValueEntity<S>> Kalix register(ValueEntityProvider<S, E> provider) {
+  public Kalix register(ValueEntityProvider<?, ?> provider) {
```
octonato, line 340: Leftover. 
I got some issue with inference and reflection, but I think I can revert this change now.

samples/spring-customer-registry-views-quickstart/src/main/java/com/example/Main.java
johanandren: So this is what a user will need to do as well in their project? Could we ship the `Application` instead and have zero extras on top of the components themselves, or is that a no-no in Spring users eyes?

octonato: You mean to ship the Main for the user? I'm not sure if that's a good idea. They may need to tweak their applications by adding extra scan paths and other configurations.

My idea was to offer this as part of the template, with a clear comment that this import must be kept around. Users will most probably start from our template. 

Users not using the template will have to copy a few bits from the templates pom.xml to have it working properly. In that case, I think it's ok for them to copy that extra line.

johanandren: Sounds good, I'm not quite familiar with what Spring people expect, I thought it was maybe bootstrapping everything for them unless they needed some specifics. (Creating a main just to call `SpringApplication.run(Main.class, args)` seems kinda boilerplatey, but I get it if you need to do other bootstrapping)

sdk/spring-sdk/src/test/java/kalix/springsdk/impl/SpringSDKTestRunner.java
```diff
@@ -1,37 +0,0 @@
...
-
-import kalix.javasdk.Kalix;
-import kalix.springsdk.action.ReflectiveActionProvider;
-import kalix.springsdk.testmodels.action.CounterSubscriber;
-import kalix.springsdk.testmodels.valueentity.Counter;
-import kalix.springsdk.valueentity.ReflectiveValueEntityProvider;
-
-import java.util.concurrent.ExecutionException;
-
-public class SpringSDKTestRunner {
```
octonato, line 27: I remove this class that was here for local, manual tests.

For the moment, it's being replaced by  `kalix.springsdk.wriing.action.App`. The goal is to have many of those small apps and then write small tests that run then and do so assertions, see `ActionWiringTest`.

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/KalixServer.scala
```diff
@@ -0,0 +1,223 @@
...
+
+  val kalixBeanFactory = new DefaultListableBeanFactory(applicationContext)
+
+  @PostConstruct
+  def init(): Unit = {
+    logger.info("Starting Kalix Server!")
+
+    // TODO: it should not be allowed to annotate Kalix components with Spring stereotypes
+    //  otherwise it will be possible to inject kalix components everywhere and that won't work as expected
+    // therefore, we should check all beans in ApplicationContext and fail if any of them implements a Kalix component
```
octonato, line 135: I still need to prove that it will be possible to do this kinds of blockade. I hope we can.

johanandren (commented):
Doesn't look too crazy IMO (but I'm also not familiar with the Spring injection/lifecycle parts).

sdk/spring-sdk/src/main/scala/kalix/springsdk/KalixConfiguration.scala
```diff
@@ -0,0 +1,54 @@
...
+import kalix.springsdk.impl.KalixServer
+import kalix.springsdk.impl.KalixServer.ActionCreationContextFactoryBean
+import kalix.springsdk.impl.KalixServer.ValueEntityContextFactoryBean
+import kalix.springsdk.impl.KalixServer.ViewCreationContextFactoryBean
+import org.springframework.context.ApplicationContext
+import org.springframework.context.annotation.Bean
+import org.springframework.context.annotation.Configuration
+
+@Configuration
+class KalixConfiguration(applicationContext: ApplicationContext) {
```
octonato, line 31: This KalixConfiguration will 'teach' Spring which bits it needs to include in the application context. 

The first part is the `KalixServer` itself. Next to it, we need  ThreadLocal backed FactoryBeans to provide the different creation contexts when needed. That's the hack!

sdk/spring-sdk/src/main/scala/kalix/springsdk/KalixConfiguration.scala
```diff
@@ -0,0 +1,54 @@
...
+  def actionCreationContextFactoryBean: ActionCreationContextFactoryBean =
+    new ActionCreationContextFactoryBean(threadLocalActionContext)
+
+  @Bean
+  def valueEntityContext: ValueEntityContextFactoryBean =
+    new ValueEntityContextFactoryBean(threadLocalValueEntityContext)
+
+  @Bean
+  def viewCreationContext: ViewCreationContextFactoryBean =
+    new ViewCreationContextFactoryBean(threadLocalViewContext)
```
octonato, line 53: None of those should be available to users, but only accessible internally. If possible, we should block access to them.

octonato: hmm, I think we are fine here. They won't be available in the application context that users will have access.

octonato:
> Doesn't look too crazy IMO (but I'm also not familiar with the Spring injection/lifecycle parts).

What sounds crazy/scary for me is that we are using thread local and then let the read happens by a code that we don't control. I don't expect Spring to switch threads when wiring components, so it's probably safe. Still, it doesn't feel good. 

If we find another way of doing that we should refactor it. For now, I'm fine in going with it.

sdk/spring-sdk/src/test/resources/logback.xml
```diff
@@ -25,8 +25,8 @@
     <appender-ref ref="JSON-STDOUT"/>
   </appender>
 
-  <logger name="akka" level="INFO"/>
-  <logger name="kalix" level="INFO"/>
+  <logger name="akka" level="DEBUG"/>
+  <logger name="kalix" level="DEBUG"/>
```
efgpinto, line 29: Did you actually want to change to debug? Or is it a leftover from troubleshooting?

efgpinto (approved):
From my very limited understanding, this LGTM.

samples/spring-customer-registry-views-quickstart/src/main/java/customer/api/Address.java
```diff
@@ -0,0 +1,18 @@
...
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+public class Address {
+  public String street;
+  public String city;
+
+  @JsonCreator
+  public Address(@JsonProperty("street") String street, @JsonProperty("city") String city) {
```
johanandren, line 14: I think this Java compiler flag avoids having to repeat the property names: https://github.com/akka/akka-platform-guide/blob/main/docs-source/docs/modules/microservices-tutorial/examples/shopping-cart-service-java/pom.xml#L175

johanandren (approved):
LGTM

samples/spring-customer-registry-views-quickstart/src/main/java/customer/api/Address.java
octonato: Possible, I even think that `@JsonProperty` should not be needed at all. If I'm not mistaking there was something related with Spring itself. I think I wrote it in a commnet somewhere. I will check it.

sdk/spring-sdk/src/test/resources/logback.xml
octonato: I don't remember to have changed this, but I think it's good to keep it like that for now. For instance, the generated proto descriptors are printed to the logs. It's good to have it during development. We can remove it later on.

octonato:
Ok, I finally got this working. The `spring-customer-registry-views-quickstart` sample were failing because we create to `Customers` name 'Johanna' and then query by name was causing an error. Silly, but took time to find out that this was the reason as we only see an internal server error.

Anyway, I will merge it now.

PR #1052, opened 2022-07-13
feat: emptyState defaults to 'null' in java sdk

octonato:
Defaults the `emptyState` to `null` for the Java API. This is most probably what a Java developer will expect. 

This change only affects `ValueEntity`, `EventSourcedEntity` and `View`. `ReplicatedEntity` is left as is because the `emptyData` cannot be `null`. 

This idea came out while implementing Views in the Spring SDK. Actually, it doesn't make much sense to have an empty state for views. Either we have a row in the index or we have nothing.


johanandren (approved):
LGTM (but needs formatting)

PR #1053, opened 2022-07-14
bump: update proxy to 1.0.14

octonato:


raboof (commented):
I thought we don't want to update the version of the proxy used in the SDK (and samples) before that version of the proxy had made it to prod, and according to https://lightbendcloud.grafana.net/d/2n4jVuw7z/prod-kalix-metrics?orgId=1 prod is still on 1.0.12.

If we're really sure there is no risk that the SDK would use proxy features that are new in 1.0.14 then we could merge this already - but in that case perhaps there's no urgent need to?

octonato:
You are right, I will close this one to avoid accidental merges. 

We do need it on the spring sdk branch, so I will open another PR for it.

octonato:
Btw, we updated it already to 1.0.13 in #1049. 

It's hard to say if that's a problem or not. Same for the 1.0.14.

johanandren (approved):

pvlugter:
It should usually be fine to have the updates slightly out-of-sync, given protocol compatibility. Things shouldn't break for a protocol addition, with default values on one side, ignored unknown fields on the other, although a new feature may not do anything until the proxy is updated.

PR #1055, opened 2022-07-14
chore: avoiding calling through the mesh to the proxy

franciscolopezsancho:
This is the first attempt to avoid calling the proxy through DNS. 

Maybe the better way to do this is to change https://github.com/lightbend/kalix-proxy/blob/9215b87b92eb46f6fae443c3ef99468130af360c/proxy/core/src/main/resources/reference.conf#L5-L6

and add only one param which is the port kind of 
```
  proxy-port = 9000
  proxy-port = ${?KALIX_PROXY_PORT}
```
and then consume that from this SDK here https://github.com/lightbend/kalix-jvm-sdk/blob/1a00df65e569679804502124a67fdab2988b876d/sdk/java-sdk/src/main/scala/kalix/javasdk/impl/GrpcClients.scala#L109

For that I would need to deal with the deployment yaml files which so far I don't know how to test. 
Do you think this could be the first part of a two-step solution?

References https://github.com/lightbend/kalix-proxy/issues/1469

### DIFF
--- a/sdk/java-sdk/src/main/scala/kalix/javasdk/impl/DiscoveryImpl.scala
+++ b/sdk/java-sdk/src/main/scala/kalix/javasdk/impl/DiscoveryImpl.scala
@@ -70,6 +70,14 @@ class DiscoveryImpl(system: ActorSystem, services: Map[String, Service]) extends
    * Discover what components the user function wishes to serve.
    */
   override def discover(in: ProxyInfo): scala.concurrent.Future[Spec] = {
+    log.info(
+      "Received discovery call from [{} {}] at [{}]:[{}] supporting Kalix protocol {}.{}",
+      in.proxyName,
+      in.proxyVersion,
+      in.internalProxyHostname,
+      in.proxyPort,
+      in.protocolMajorVersion,
+      in.protocolMinorVersion)
     if (isVersionProbe(in)) {
       // only (silently) send service info for hybrid proxy version probe
       Future.successful(Spec(serviceInfo = Some(serviceInfo)))
@@ -78,23 +86,21 @@ class DiscoveryImpl(system: ActorSystem, services: Map[String, Service]) extends
       val proxyTerminatedPromise = if (in.devMode) Promise.successful[Done](Done) else Promise[Done]()
       proxyTerminatedRef.getAndSet(proxyTerminatedPromise).trySuccess(Done)
 
-      log.info(
-        "Received discovery call from [{} {}] at [{}] supporting Kalix protocol {}.{}",
-        in.proxyName,
-        in.proxyVersion,
-        in.proxyHostname,
-        in.protocolMajorVersion,
-        in.protocolMinorVersion)
       log.debug(s"Supported sidecar entity types: {}", in.supportedEntityTypes.mkString("[", ",", "]"))
 
       val unsupportedServices = services.values.filterNot { service =>
         in.supportedEntityTypes.contains(service.componentType)
       }
 
-      val grpcClients = GrpcClients.get(system)
+      val grpcClients = GrpcClients(system)
       // pass the deployed name of the service on to GrpcClients for cross component calls
-      GrpcClients.get(system).setProxyHostname(in.proxyHostname)
-
+      if (in.internalProxyHostname.isEmpty) {
+        // for backward compatibiliy with proxy 1.0.14 or older
+        grpcClients.setProxyHostname(in.proxyHostname)
+      } else {
+        grpcClients.setProxyHostname(in.internalProxyHostname)
+      }
+      grpcClients.setProxyPort(in.proxyPort)
       grpcClients.setIdentificationInfo(in.identificationInfo)
 
       if (unsupportedServices.nonEmpty) {
--- a/sdk/java-sdk/src/main/scala/kalix/javasdk/impl/GrpcClients.scala
+++ b/sdk/java-sdk/src/main/scala/kalix/javasdk/impl/GrpcClients.scala
@@ -61,7 +61,7 @@ final class GrpcClients(system: ExtendedActorSystem) extends Extension {
   private val log = LoggerFactory.getLogger(classOf[GrpcClients])
 
   @volatile private var proxyHostname: Option[String] = None
-  @volatile private var selfPort: Option[Int] = None
+  @volatile private var proxyPort: Option[Int] = None
   @volatile private var identificationInfo: Option[IdentificationInfo] = None
   private implicit val ec: ExecutionContext = system.dispatcher
   private val clients = new ConcurrentHashMap[Key, AnyRef]()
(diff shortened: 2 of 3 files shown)


sdk/java-sdk/src/main/scala/kalix/javasdk/impl/GrpcClients.scala
```diff
@@ -107,8 +107,12 @@ final class GrpcClients(system: ExtendedActorSystem) extends Extension {
   /** Local gRPC clients point to services (user components or Kalix services) in the same deployable */
   private def getLocalGrpcClient[T](serviceClass: Class[T]): T = {
     proxyHostname match {
-      case Some("localhost") => getGrpcClient(serviceClass, "localhost", selfPort.getOrElse(9000), localAddHeader)
-      case Some(selfName)    => getGrpcClient(serviceClass, selfName, 80, localAddHeader)
+      case Some("localhost") => {
+        getGrpcClient(serviceClass, "localhost", selfPort.getOrElse(9000), localAddHeader)
+      }
+      case Some(selfName) => {
+        getGrpcClient(serviceClass, "localhost", selfPort.getOrElse(8013), localAddHeader)
```
johanandren, line 114: I think it would be better if this decision was moved to the proxy and handed to the SDK through the `ProxyInfo` message (defined in discovery.proto in the proxy)

johanandren: (just like hos the proxyHostName is today)

franciscolopezsancho: Didn't see that coming. Thanks!

franciscolopezsancho:
Depends on https://github.com/lightbend/kalix-proxy/pull/1599

sdk/java-sdk/src/main/scala/kalix/javasdk/impl/GrpcClients.scala
```diff
@@ -106,11 +106,12 @@ final class GrpcClients(system: ExtendedActorSystem) extends Extension {
...
   /** Local gRPC clients point to services (user components or Kalix services) in the same deployable */
   private def getLocalGrpcClient[T](serviceClass: Class[T]): T = {
-    proxyHostname match {
-      case Some("localhost") => getGrpcClient(serviceClass, "localhost", selfPort.getOrElse(9000), localAddHeader)
-      case Some(selfName)    => getGrpcClient(serviceClass, selfName, 80, localAddHeader)
-      case None =>
-        throw new IllegalStateException("Self service name not set by proxy at discovery, too old proxy version?")
+    (proxyHostname.get, proxyPort.get) match {
+      case (host: String, port: Int) =>
+        getGrpcClient(serviceClass, host, port, localAddHeader)
```
franciscolopezsancho, line 111: `host` should be always `localhost`

sdk/java-sdk/src/main/scala/kalix/javasdk/impl/DiscoveryImpl.scala
```diff
@@ -71,6 +71,7 @@ class DiscoveryImpl(system: ActorSystem, services: Map[String, Service]) extends
    */
   override def discover(in: ProxyInfo): scala.concurrent.Future[Spec] = {
     if (isVersionProbe(in)) {
+      log.info("this version is probe. The service info is: [{}]", serviceInfo)
```
octonato, line 74: "this version is probe" is not something that users will understand. I don't understand it myself. :-)

I think we emit a log similar to what we are logging in the `else` branch of this `if`.

octonato (approved):
LGTM, with a minor comment about logging

franciscolopezsancho:
Before we merge this PR, [this](https://github.com/lightbend/kalix-proxy/pull/1599) needs to be released

sdk/java-sdk/src/main/scala/kalix/javasdk/impl/DiscoveryImpl.scala
```diff
@@ -70,23 +70,22 @@ class DiscoveryImpl(system: ActorSystem, services: Map[String, Service]) extends
...
    */
   override def discover(in: ProxyInfo): scala.concurrent.Future[Spec] = {
+    log.info(
+      "Received discovery call from [{} {}] at [{}]:[{}] supporting Kalix protocol {}.{}",
+      in.proxyName,
+      in.proxyVersion,
+      in.internalProxyHostname,
+      in.proxyPort,
+      in.protocolMajorVersion,
+      in.protocolMinorVersion)
```
octonato, line 80: 👍

sdk/java-sdk/src/main/scala/kalix/javasdk/impl/GrpcClients.scala
```diff
@@ -104,13 +103,14 @@ final class GrpcClients(system: ExtendedActorSystem) extends Extension {
...
-      case Some("localhost") => getGrpcClient(serviceClass, "localhost", selfPort.getOrElse(9000), localAddHeader)
-      case Some(selfName)    => getGrpcClient(serviceClass, selfName, 80, localAddHeader)
-      case None =>
-        throw new IllegalStateException("Self service name not set by proxy at discovery, too old proxy version?")
+    (proxyHostname, proxyPort) match {
+      case (Some(hostname), Some(port)) =>
+        getGrpcClient(serviceClass, hostname, port, localAddHeader)
+      case _ =>
+        throw new IllegalStateException(
+          "Service proxy hostname and port are not set by proxy at discovery, too old proxy version?")
```
octonato, line 113: As discussed on the stand-up today, we still need to support the old code because a user can update their SDK version, but still have the docker-compose using an old version of the proxy. 

When that happens, users will probably revert their SDK version believing that the SDK is broken. It will only broken in dev mode though. 

So, we need to add two cases to this pattern matching.

```suggestion
    (proxyHostname, proxyPort) match {
      case (Some(hostname), Some(port)) =>
        getGrpcClient(serviceClass, hostname, port, localAddHeader)
      case (Some("localhost"), None) => getGrpcClient(serviceClass, "localhost", selfPort.getOrElse(9000), localAddHeader)
      case (Some(selfName), None)    => getGrpcClient(serviceClass, selfName, 80, localAddHeader)
      case _ =>
        throw new IllegalStateException(
          "Service proxy hostname and port are not set by proxy at discovery, too old proxy version?")
```

Basically, if port is not set (old proxy behavior), we fallback to the old strategy.

Same for the javascript sdk

sdk/java-sdk/src/main/scala/kalix/javasdk/impl/DiscoveryImpl.scala
```diff
@@ -78,23 +86,21 @@ class DiscoveryImpl(system: ActorSystem, services: Map[String, Service]) extends
...
       val unsupportedServices = services.values.filterNot { service =>
         in.supportedEntityTypes.contains(service.componentType)
       }
 
-      val grpcClients = GrpcClients.get(system)
+      val grpcClients = GrpcClients(system)
       // pass the deployed name of the service on to GrpcClients for cross component calls
-      GrpcClients.get(system).setProxyHostname(in.proxyHostname)
-
+      if (in.internalProxyHostname == "") {
```
octonato, line 97:
```suggestion
      if (in.internalProxyHostname.isEmpty) {
```

octonato (approved):
LGTM

We still need to wait for a proxy release to get this merged. We are depending on the new protocol file that brings in `internalProxyHostname`

octonato:
Commits are squashed and then rebased on top of #1061. 

We should merge this one after #1061 and before releasing.

octonato:
This one is ready for merge. Will do when green

PR #1079, opened 2022-08-26
feat: add topic subscription

franciscolopezsancho:
References https://github.com/lightbend/kalix/issues/7406


sdk/spring-sdk/src/main/java/kalix/springsdk/annotations/Subscribe.java
```diff
@@ -26,4 +27,13 @@
   @interface ValueEntity {
     Class<? extends kalix.javasdk.valueentity.ValueEntity<?>> value();
   }
+
+  @Target(ElementType.METHOD)
+  @Retention(RetentionPolicy.RUNTIME)
+  @Documented
+  @interface Topic {
+    String value();
+    // @AliasFor("value")
```
franciscolopezsancho, line 36: WIP

franciscolopezsancho: I tested the alias with the following as shown [here](https://www.baeldung.com/spring-aliasfor-annotation#explicit-within) but it doesn't work. I get `value` empty if I set the `name`.

``` 
    @AliasFor("name")
    String value() default "";

    @AliasFor("value")
    String name() default "";
```

As far as I understand this only works if Spring instantiates the bean which is not our case because we use reflection. As you mention @octonato in a previous call and I can see now in the code.

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/ActionDescriptorFactory.scala
```diff
@@ -48,11 +50,23 @@ private[impl] object ActionDescriptorFactory extends ComponentDescriptorFactory
...
+          .withKalixOptions(kalixOptions)
+      }
+
     val serviceName = nameGenerator.getName(component.getSimpleName)
     ComponentDescriptor(
       nameGenerator,
       serviceName,
       component.getPackageName,
-      springAnnotatedMethods ++ subscriptionMethods)
+      springAnnotatedMethods ++ subscriptionMethods ++ subscriptionTopicMethods)
```
franciscolopezsancho, line 70: I should change `subscriptionMethods` for `subscriptionValueEntityMethods`.

sdk/spring-sdk/src/main/java/kalix/springsdk/annotations/Subscribe.java
octonato: yeah, I think you are right. The alias won't work in our case. 

We can leave it out for now and eventually add our own alias annotation in the future if we find it will be useful.

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/ActionDescriptorFactory.scala
octonato: yep, I think it will increase clarity

octonato:
@franciscolopezsancho, watch out for the PR title. Better to have something like: "feat: add topic subscription" :-)

octonato (commented):
This is looking good. 

I don't think there is much to add except an integration tests. It's hard for us to test that the gRPC is correctly generated using unit tests.

franciscolopezsancho:
I can't find other integration tests for Spring-SDK @octonato . How do you find I finish with this PR, with ValueEntities and EventSourcedEntities subscriptions and then we add integration tests for the Spring SDK in a different PR?

octonato:
I will drop here quickly what we discussed in the meeting. 

We need to add an option attribute for the `consumer_group`, but we can do that later. 

More pressing is to find a solution to differentiate distinct payload and dispatch to the right method. 

For example, a Topic or events from ES, may have different payload. On the Java side, a user will want to have them typed, eg: `Foo`, `Bar`, `Baz`. 

But the proxy will see different methods all receiving `Any` and won't know which one to dispatch to. 

Discussed solution is to pass to the proxy only one catch all method, like: `SubscribeTo{TopicName}(Any)`. Proxy send all payloads to that single method and is the SDK that is responsible to dispatch to the right Java methods.

Once we have clear what we should do to solve this, we can think how we can support the new `ignore` option. Possibly with a list of typeUrls to ignore.

octonato:
Relates to: https://github.com/lightbend/kalix/issues/7407

franciscolopezsancho:
Just to make sure we are on the same page. Scenario.
```
Action A {
   @Subscribe.Topic("x")
   public Action.Effect<Foo> methodF(Foo foo)
   
   @Subscribe.Topic("x")
   public Action.Effect<Bar> methodB(Bar  bar)

}
```

This will generate the following .proto
```
rpc OnXMessage(Any) returns (Any)
```
This generates - through the SDK - a router in the User function listening to /onXMessage endpoint that parses `Any` to `Bar` or `Foo` and sends it to `methodB` or `methodF` respectively.
The proxy generates the scafolding to consume that topic and when a message arrives it routes it to OnXMessage UserFunction. 

Later on, we add something in the proxy to ignore messages that aren't Foo or Bar.

franciscolopezsancho:
Unpack the Any (json) I'll modify the [ReflectiveActionRouter](https://github.com/lightbend/kalix-jvm-sdk/blob/spring-sdk/sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/action/ReflectiveActionRouter.scala)

octonato:
> Just to make sure we are on the same page. Scenario.
> 
> ```
> Action A {
>    @Subscribe.Topic("x")
>    public Action.Effect<Foo> methodF(Foo foo)
>    
>    @Subscribe.Topic("x")
>    public Action.Effect<Bar> methodB(Bar  bar)
> 
> }
> ```
> 
> This will generate the following .proto
> 
> ```
> rpc OnXMessage(Any) returns (Any)
> ```
> 
> This generates - through the SDK - a router in the User function listening to /onXMessage endpoint that parses `Any` to `Bar` or `Foo` and sends it to `methodB` or `methodF` respectively. The proxy generates the scafolding to consume that topic and when a message arrives it routes it to OnXMessage UserFunction.
> 
> Later on, we add something in the proxy to ignore messages that aren't Foo or Bar.

Well described. This is the current plan.

franciscolopezsancho:
superseeded by https://github.com/lightbend/kalix-jvm-sdk/pull/1089

Issue #1078, opened 2022-08-26
Calls to same entity type sequential


johanandren:
Noted in the forums https://discuss.kalix.io/t/multiple-requests-at-the-same-time-multi-threading-concern/106/12 and verified by me just now, sneaking in a blocking Thread.sleep in an entity command handler highlights that something is off with how calls to different entity ids of the same entity type are handled.

Easily repeated by taking the Java shopping-cart quickstart, add a Thread.sleep surounded by before/after logging in the addItem command handler, run with local proxy and fire off parallel:ish requests:

```bash
for i in 1 2 3 4 5 6 7 8; do grpcurl --plaintext -d '{"cart_id": "cart'$i'", "product_id": "akka-tshirt", "name": "Akka t-shirt", "quantity": 3}' localhost:9000  shopping.cart.api.ShoppingCart/AddItem & done
```

johanandren:
#1081 fixes this, but the forum post will still have problems with pool starvation and will need to re-configure dispatchers in some way, possibly both separating the internal dispatcher from the component running one (undoing the default `internal-dispatcher` in the JVM SDK) and re-sizing the threadpool for the blocking work.

We do not currently have any documentation on how to do this, but the config would go under the namespaced Akka config `kalix.system.akka`.

patriknw:
> re-configure dispatchers in some way, possibly both separating the internal dispatcher from the component running one (undoing the default internal-dispatcher in the JVM SDK) and re-sizing the threadpool for the blocking work

That is out of scope of what should be done in a Kalix Event Sourced Entity. We should simply recommend against it. If some blocking or async work needs to be done in the command handler that should be moved to an Action that can spawn that task on separate thread pool and forward the result to the entity.

PR #1081, opened 2022-08-29
fix: Ensure streaming calls from proxy are parallel 

johanandren:
Refs #1078

Calls to stream out methods from the same client are sequential with Akka HTTP/gRPC
by default, this adds an async boundary to allow them to run in parallel.


johanandren:
Yup, this is all we need.

Issue #1045, opened 2022-07-06
Report unique SDK names for Scala and Java


johanandren:
At discovery the SDK tells the proxy what SDK it is, it currently reports the same "kalix-java-sdk" for both Java and Scala. We should report "kalix-java-sdk", "kalix-scala-sdk" (and in the future "kalix-spring-sdk") instead.

efgpinto:
Just so I get a better understading of the rational/benefit here, is it because although at the moment Java and Scala SDK are always released together, we might want to have separate releases in the future? Are there other benefits?

ennru:
We'd still have everything in one jar, but get data which taste of the JVM SDK gets used.

patriknw:
@efgpinto note that there is also a check in the proxy for valid SDKs: https://github.com/lightbend/kalix-proxy/blob/3a8714ebefae96fdda2c89938bfd91984c81889a/proxy/core/src/main/scala/kalix/proxy/DiscoveryManager.scala#L688

johanandren:
"scala" is already in there though, so that check should be fine.

efgpinto:
Exactly, I've noticed that while testing locally. On the other hand, once this one is fixed it will lead to a different issue where the proxy won't be able to warn of a new scala sdk version but I'll open a different issue to discuss that in the meantime.

efgpinto:
The implementation for this is complete. I think I'll keep this issue open until this is released and we can actually confirm on Grafana the different SDKs flavours being reported correctly.

efgpinto:
Closing now as it looks to be working as expected (any scala app from sdk 1.0.6+ should be reported as such). From [dashboard](https://lightbendcloud.grafana.net/d/2n4jVuw7z/prod-kalix-metrics?orgId=1&viewPanel=119):
<img width="618" alt="Screenshot 2022-09-08 at 21 45 05" src="https://user-images.githubusercontent.com/1105359/189222766-3427a2d2-c51d-4daf-97fa-7d736c5d77cc.png">

PR #1082, opened 2022-08-29
chore: report spring sdk info

efgpinto:
Refs #1045 

Since we've done it recently for scala vs java, might as well get it out of the way for Spring.


efgpinto:
Seems to work as expected locally.

octonato (approved):
LGTM

PR #1086, opened 2022-08-30
feat: Adding publishing action

franciscolopezsancho:
References https://github.com/lightbend/kalix/issues/7405


sdk/spring-sdk/src/main/java/kalix/springsdk/annotations/Subscribe.java
```diff
@@ -26,4 +27,11 @@
   @interface ValueEntity {
     Class<? extends kalix.javasdk.valueentity.ValueEntity<?>> value();
   }
+
+  @Target(ElementType.METHOD)
```
johanandren, line 31: Note the work in progress on allowing this at service/class level as well, can probably be a follow up but just so you know. See PR over here: https://github.com/lightbend/kalix-proxy/pull/1622/files#diff-c1390e2e16a5479ebfa65460b72ddb4985714f0c0a3a9727be20e50b1af22502R46

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/ActionDescriptorFactory.scala
```diff
@@ -48,11 +52,35 @@ private[impl] object ActionDescriptorFactory extends ComponentDescriptorFactory
...
+      .filter(hasTopicSubscription)
+      .sorted // make sure we get the methods in deterministic order
+      .map { method =>
+        val subscriptionOptions = eventingInForTopic(method)
+        val kalixOptions =
+          kalix.MethodOptions.newBuilder().setEventing(subscriptionOptions).build()
+
+        KalixMethod(RestServiceMethod(method))
+          .withKalixOptions(kalixOptions)
+      }
```
johanandren, line 65: Should probably validate that no two method accept the same event as parameter and subscribe to the same topic (since routing is done based on event type)

sdk/spring-sdk/src/main/java/kalix/springsdk/annotations/Subscribe.java
octonato: I suggested to remove the `ElementType.TYPE`, but didn't think about the case you mentioned. 

That will be indeed much convenient to annotated only on the class.

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/ActionDescriptorFactory.scala
octonato: Good idea. 

We have another related issue that came up in our discussion  today. See https://github.com/lightbend/kalix/issues/7407

franciscolopezsancho: Shall we take into account that the user may want to do two different things in two different methods both pointing to the same topic and consuming the same type of object?

franciscolopezsancho:
I messed up with a rebase I'll open a new PR

franciscolopezsancho:
superseded by https://github.com/lightbend/kalix-jvm-sdk/pull/1089

PR #1084, opened 2022-08-30
feat(spring-sdk): add support for event sourced entities

efgpinto:
Refs https://github.com/lightbend/kalix/issues/7360

Validations in place for event handler methods:
- they must be annotated with `@EventHandler`, be public, have a single parameter and a return type that matched the enclosing class type parameter
- there cannot be two handlers with the same type parameter (as in, each event type must have only one handler)

PR #1088, opened 2022-09-01
chore: spring sdk integration tests

octonato:
Add integration tests (with testcontainers) for the Spring SDK. 

Once we merge this, we should remove the `samples/spring-wiring-tests`. Not removing now to avoid conflicts with #1084. 

Also, when we are ready to remove `samples/spring-wiring-tests`, we must expand the tests to include ValueEntity and Views. For now, I'm only add a simple test for Action. The goal is just to create the infra we need.


octonato:
> Does it need something in the CI-config as well or does that already run IT test for all modules?

You are right. We need to do `sbt it:test` for the spring sdk. I will change

octonato:
@johanandren, added it to CI. It will now run `IntegrationTest/test` for all projects, but only the SpringSDK have integration tests. 

If we add other IT tests in other submodules we will be covered already.

octonato:
> @johanandren, added it to CI. It will now run `IntegrationTest/test` for all projects, but only the SpringSDK have integration tests.
> 
> If we add other IT tests in other submodules we will be covered already.

That didn't work out. To use testcontainers, we need to change the setup of the job. To avoid changing the main test job, I moved the spring IT tests to its own job.

.circleci/config.yml
```diff
@@ -17,7 +17,7 @@ commands:
           parent: spring-sdk # we need to revert this when merging back into main
 
   install-java-11:
-    description: install openjdk-11
+    description: install openjdk-11t
```
octonato, line 20: Oeps!

johanandren (approved):
LGTM once that IT runs and passes :)

PR #1089, opened 2022-09-02
feat(spring-sdk): pub and sub with mixing spring annotations

franciscolopezsancho:
References publishing 
https://github.com/lightbend/kalix/issues/7363

Superseding PR https://github.com/lightbend/kalix-jvm-sdk/pull/1086
References subscription 
 https://github.com/lightbend/kalix/issues/7362
 https://github.com/lightbend/kalix-jvm-sdk/pull/1079

### DIFF
--- a/sdk/spring-sdk/src/it/resources/logback.xml
+++ b/sdk/spring-sdk/src/it/resources/logback.xml
@@ -1,0 +1,38 @@
+<configuration>
+    <appender name="STDOUT" class="ch.qos.logback.core.ConsoleAppender">
+        <encoder>
+            <pattern>%date{ISO8601} %-5level %logger - %msg%n</pattern>
+        </encoder>
+    </appender>
+
+    <appender name="JSON-STDOUT" target="System.out" class="ch.qos.logback.core.ConsoleAppender">
+        <encoder class="ch.qos.logback.core.encoder.LayoutWrappingEncoder">
+            <layout class="kalix.javasdk.logging.LogbackJsonLayout">
+                <timestampFormat>yyyy-MM-dd'T'HH:mm:ss.SSSX</timestampFormat>
+                <timestampFormatTimezoneId>Etc/UTC</timestampFormatTimezoneId>
+                <appendLineSeparator>true</appendLineSeparator>
+                <jsonFormatter class="ch.qos.logback.contrib.jackson.JacksonJsonFormatter">
+                    <!-- Don't use prettyPrint true in production, but can be useful for local development -->
+                    <prettyPrint>false</prettyPrint>
+                </jsonFormatter>
+            </layout>
+        </encoder>
+    </appender>
+
+    <appender name="ASYNC-JSON-STDOUT" class="ch.qos.logback.classic.AsyncAppender">
+        <queueSize>8192</queueSize>
+        <neverBlock>true</neverBlock>
+        <appender-ref ref="JSON-STDOUT"/>
+    </appender>
+
+    <logger name="akka" level="DEBUG"/>
+    <logger name="kalix" level="DEBUG"/>
+    <logger name="org.springframework.web" level="DEBUG"/>
+    <logger name="akka.http" level="INFO"/>
+    <logger name="io.grpc" level="INFO"/>
+
+    <root level="INFO">
+        <!--        <appender-ref ref="STDOUT"/>-->
+        <appender-ref ref="STDOUT"/>
+    </root>
+</configuration>
--- a/sdk/spring-sdk/src/it/resources/logback.xml
+++ b/sdk/spring-sdk/src/it/resources/logback.xml
@@ -25,8 +25,8 @@
         <appender-ref ref="JSON-STDOUT"/>
     </appender>
 
-    <logger name="akka" level="DEBUG"/>
-    <logger name="kalix" level="DEBUG"/>
+    <logger name="akka" level="INFO"/>
+    <logger name="kalix.springsdk" level="DEBUG"/>
     <logger name="org.springframework.web" level="DEBUG"/>
     <logger name="akka.http" level="INFO"/>
     <logger name="io.grpc" level="INFO"/>
--- a/sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/ActionDescriptorFactory.scala
+++ b/sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/ActionDescriptorFactory.scala
@@ -65,30 +65,20 @@ private[impl] object ActionDescriptorFactory extends ComponentDescriptorFactory 
           .withKalixOptions(kalixOptions)
       }
 
-    def uniqueTopicSubscriptions(subscriptions: Seq[KalixMethod]): Seq[KalixMethod] =
-      groupByTopic(subscriptions)
-        .filter { case (topic, kMethods) => kMethods.size == 1 }
-        .flatMap { case (topic, kMethods) =>
-          kMethods
-        }
-        .toSeq
-
-    def combineSubscriptionsSameTopic(subscriptions: Seq[KalixMethod]): Seq[KalixMethod] = {
+    def checkNotTopicDuplication(subscriptions: Seq[KalixMethod]): Seq[KalixMethod] = {
+      def groupByTopic(methods: Seq[KalixMethod]): Map[String, Seq[KalixMethod]] = {
+        val withTopicIn = methods.filter(kalixMethod =>
+          kalixMethod.methodOptions.exists(option =>
+            option.hasEventing && option.getEventing.hasIn && option.getEventing.getIn.hasTopic))
+        //Assuming there is only one topic, therefore head is as good as any other
+        withTopicIn.groupBy(m => m.methodOptions.head.getEventing.getIn.getTopic)
+      }
       groupByTopic(subscriptions).collect {
         case (topic, kMethods) if kMethods.size > 1 =>
-          val representative = kMethods.head
-          val rs: RestServiceMethod = representative.serviceMethod.asInstanceOf[RestServiceMethod]
-          KalixMethod(RestServiceMethod(rs.javaMethod, Some("On" + topic.capitalize + "KalixSyntheticMethod")))
-            .withKalixOptions(representative.methodOptions)
-      }.toSeq
-    }
-
-    def groupByTopic(methods: Seq[KalixMethod]): Map[String, Seq[KalixMethod]] = {
-      val withTopicIn = methods.filter(kalixMethod =>
-        kalixMethod.methodOptions.exists(option =>
-          option.hasEventing && option.getEventing.hasIn && option.getEventing.getIn.hasTopic))
-      //Assuming there is only one topic, therefore head is as good as any other
-      withTopicIn.groupBy(m => m.methodOptions.head.getEventing.getIn.getTopic)
+          throw InvalidComponentException(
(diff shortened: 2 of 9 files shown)


sdk/spring-sdk/src/test/java/kalix/springsdk/testmodels/subscriptions/PubSubTestModels.java
```diff
@@ -39,6 +42,44 @@ public Action.Effect<Message> messageTwo(Message message) {
...
+  }
+
+  public static class RestWithPublishToTopicAction extends Action {
+
+    @PostMapping("/message/{msg}")
+    @Publish.Topic("foobar")
+    public Effect<Message> messageOne(@PathVariable String msg) {
+      return effects().reply(new Message(msg));
+    }
+  }
```
efgpinto, line 81: Could it make sense to add an additional RestAnnotated-only method to one of these examples just to have a test asserting that such method gets ignored when it comes to the pubsub logic maybe?

efgpinto (commented):
I'm still going through the component descriptor logic which is a bit more complex but thought it could be good to drop these questions right away in case I'm missing some context.

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/ActionDescriptorFactory.scala
```diff
@@ -48,11 +53,91 @@ private[impl] object ActionDescriptorFactory extends ComponentDescriptorFactory
...
       nameGenerator,
       serviceName,
       component.getPackageName,
-      springAnnotatedMethods ++ subscriptionMethods)
+      filterAndAddKalixOptions(
+        springAnnotatedMethods,
+        publicationTopicMethods) ++ subscriptionValueEntityMethods ++ combineSubscriptionsSameTopic(
+        subscriptionTopicMethods) ++ uniqueTopicSubscriptions(subscriptionTopicMethods) ++ removeDuplicates(
+        springAnnotatedMethods,
+        publicationTopicMethods))
```
octonato, line 141: This block could benefit from some formatting to make it easier to read.

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/ActionDescriptorFactory.scala
```diff
@@ -28,15 +32,16 @@ import kalix.springsdk.impl.reflection.RestServiceMethod
...
   override def buildDescriptorFor(component: Class[_], nameGenerator: NameGenerator): ComponentDescriptor = {
-
+    //we shoudl merge from here
     val springAnnotatedMethods =
       RestServiceIntrospector.inspectService(component).methods.map { serviceMethod =>
         validateRestMethod(serviceMethod.javaMethod)
         KalixMethod(serviceMethod)
       }
 
+    //TODO make sure no subscription shoudl be exposed via REST.
```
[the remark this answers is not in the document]
octonato: Yep, we can leave it for another PR. However, if we decide so, we should create a ticket to not loose track.

sdk/spring-sdk/src/main/java/kalix/springsdk/annotations/Subscribe.java
```diff
@@ -17,6 +17,7 @@
 package kalix.springsdk.annotations;
 
 import java.lang.annotation.*;
+import org.springframework.core.annotation.AliasFor;
```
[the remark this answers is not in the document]
octonato:
```suggestion
```

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/ActionDescriptorFactory.scala
```diff
@@ -48,11 +53,91 @@ private[impl] object ActionDescriptorFactory extends ComponentDescriptorFactory
...
+        name.charAt(0).toUpper + name.substring(1)
+      }
+      groupByTopic(subscriptions).collect {
+        case (topic, kMethods) if kMethods.size > 1 =>
+          val representative = kMethods.head
+          val rs: RestServiceMethod = representative.serviceMethod.asInstanceOf[RestServiceMethod]
+          KalixMethod(
+            RestServiceMethod(rs.javaMethod, Some("On" + firstToUpper(topic) + firstToUpper(rs.javaMethod.getName))))
+            .withKalixOptions(representative.methodOptions)
+      }.toSeq
```
octonato, line 87: As discussed on call today, that won't work. The proxy will send `Any` payloads to the user function and it will dispatch them all to the same Java method. We need to keep around a map to route back the individual messages to the right java methods. Something that take the typeUrl into consideration.

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/ComponentDescriptorFactory.scala
```diff
@@ -41,12 +49,40 @@ private[impl] object ComponentDescriptorFactory {
...
   def eventingInForValueEntity(javaMethod: Method): Eventing = {
     val entityType = findValueEntityType(javaMethod)
     val eventSource = EventSource.newBuilder().setValueEntity(entityType).build()
     Eventing.newBuilder().setIn(eventSource).build()
   }
 
+  def eventingInForTopic(javaMethod: Method): Eventing = {
+    val topicName = findSubTopicName(javaMethod)
+    val consumerGroup = findSubConsumerGroup(javaMethod)
+    val eventSource = EventSource.newBuilder().setTopic(topicName).setConsumerGroup(consumerGroup).build()
```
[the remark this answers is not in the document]
octonato: Yep, you are right, but we need to check in the proxy if we are not already creating consumer groups per action.

octonato (commented):
Looking good, but we still need a solution to route incoming `Any` to the write java method using the typeUrl

sdk/spring-sdk/src/main/java/kalix/springsdk/annotations/Subscribe.java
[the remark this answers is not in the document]
franciscolopezsancho: yes

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/ActionDescriptorFactory.scala
franciscolopezsancho: done

sdk/spring-sdk/src/test/scala/kalix/springsdk/impl/ActionDescriptorFactorySpec.scala
```diff
@@ -231,6 +236,79 @@ class ActionDescriptorFactorySpec extends AnyWordSpec with ComponentDescriptorSu
...
+      assertDescriptor[SubscribeToTwoTopicsAction] { desc =>
+        val methodOne = desc.methods("OnTopicXYZMethodOne")
+        methodOne.requestMessageDescriptor.getFullName shouldBe JavaPbAny.getDescriptor.getFullName
+
+        val eventSourceOne = findKalixMethodOptions(desc, "OnTopicXYZMethodOne").getEventing.getIn
+        eventSourceOne.getTopic shouldBe "topicXYZ"
+        val rule = findHttpRule(desc, "OnTopicXYZMethodOne")
+
+        rule.getPost shouldBe
+        "/kalix.springsdk.testmodels.subscriptions.PubSubTestModels.SubscribeToTwoTopicsAction/OnTopicXYZMethodOne"
```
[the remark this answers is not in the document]
franciscolopezsancho: done

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/ActionDescriptorFactory.scala
octonato: > Note that we can't rely on the jsonType alone for deserializing as that would open up a security hole when consuming from a topic (classic Jackson remote-code exploit, imagine malicious user passing some class name that can be executed), it must always be decided by an actual type handed to the SDK.

We don't rely on the typeUrl to deserialize. We simply take whatever payload is sent and we deserialize to whatever type that method is expecting. 

The typeUrl will only be used as an indication to find which Java method should receive the payload. 

Still, the problem with refactoring of state and events persists. We should instead, at least for the Spring SDK, use `@JsonTypeInfo`. If available, we should use it in the typeUrl instead of the FQCN.

octonato:
In our last discussion (me and Cisco) we came to the conclusion that we can take a simple path for now. 

We will allow one method per topic subscription per Action. If users want to receive different types, they can use `@JsonTypoInfo` and `@JsonSubTypes` annotations. Like in: 

```java
@JsonTypeInfo(use = JsonTypeInfo.Id.NAME, property = "type")
@JsonSubTypes({
    @JsonSubTypes.Type(value = Bar.class, name = "bar"),
    @JsonSubTypes.Type(value = Baz.class, name = "baz")
})
public interface Foo {
}

public class Baz implements Foo {
  final public String value;
  @JsonCreator
  public Baz(@JsonProperty String value) {
    this.value = value;
  }
}

public class Bar implements Foo {
  final public int value;
  @JsonCreator
  public Bar(@JsonProperty int value) {
    this.value = value;
  }
}
```

The gRPC method will receive a `Any`, but the SDK will resolve it to `Foo` (common interface) and the user will get a `Bar` or a `Baz`. 

This approach is simple enough and idiomatic enough for now. And is also a good solution for ES Events. 

Later, we can have distinct methods for the EventSourced events, but can be a later addition.

sdk/spring-sdk/src/it/resources/logback.xml
```diff
@@ -0,0 +1,38 @@
...
+    </appender>
+
+    <appender name="ASYNC-JSON-STDOUT" class="ch.qos.logback.classic.AsyncAppender">
+        <queueSize>8192</queueSize>
+        <neverBlock>true</neverBlock>
+        <appender-ref ref="JSON-STDOUT"/>
+    </appender>
+
+    <logger name="akka" level="DEBUG"/>
+    <logger name="kalix" level="DEBUG"/>
```
octonato, line 29:
```suggestion
    <logger name="akka" level="INFO"/>
    <logger name="kalix.springsdk" level="DEBUG"/>
```
We probably only need debug for the springsdk part.

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/ActionDescriptorFactory.scala
```diff
@@ -48,11 +53,75 @@ private[impl] object ActionDescriptorFactory extends ComponentDescriptorFactory
...
+        val withTopicIn = methods.filter(kalixMethod =>
+          kalixMethod.methodOptions.exists(option =>
+            option.hasEventing && option.getEventing.hasIn && option.getEventing.getIn.hasTopic))
+        //Assuming there is only one topic, therefore head is as good as any other
+        withTopicIn.groupBy(m => m.methodOptions.head.getEventing.getIn.getTopic)
+      }
+      groupByTopic(subscriptions).collect {
+        case (topic, kMethods) if kMethods.size > 1 =>
+          throw InvalidComponentException(
+            s"topic: '$topic' it is used in multiple @Subscription.Topic annotations. Each @Subscription. Topic must point to a different topic")
```
octonato, line 79:
```suggestion
            s"topic: '$topic' it is used in multiple @Subscription.Topic annotations. Each @Subscription.Topic must point to a different topic")
```

octonato (approved):
LGTM, with a few minor changes

PR #1090, opened 2022-09-02
chore: emptyState can be null - Optional(null)

octonato:
In #1052, we make it possible to have an emptyState set to `null`. That's convenient and even expected for java developers. 

However, on first command, one may want to check if the state was already created with `if (currentState() == null)`. 

Because we were using `Optional.ofNullable(state)`, we still have an empty Optional that will later fail with "Current state is only available when handling a command."

I'm opting to have a `Optional(null)` here because we then preserve the behavior that `currentState()` can only be called inside the command handler.

### DIFF
--- a/sdk/java-sdk/src/main/java/kalix/javasdk/valueentity/ValueEntity.java
+++ b/sdk/java-sdk/src/main/java/kalix/javasdk/valueentity/ValueEntity.java
@@ -32,6 +32,8 @@ public abstract class ValueEntity<S> {
 
   private Optional<S> currentState = Optional.empty();
 
+  private boolean handlingCommands = false;
+
   /**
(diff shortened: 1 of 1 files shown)


octonato:
Oh, I think I'm mixing APIs. 

We got this: 

```java
Caused by: java.lang.NullPointerException: null
	at java.base/java.util.Objects.requireNonNull(Objects.java:221)
	at java.base/java.util.Optional.<init>(Optional.java:107)
	at java.base/java.util.Optional.of(Optional.java:120)
```

octonato:
Ok, this is now updated and we need to keep a flag to track if we are constructing or handling commands.

johanandren (approved):

Issue #1091, opened 2022-09-02
spring-sdk: view need a currentState method


octonato:
We missed this one. 

Our sample project is not using transformation and that's probably why we didn't see it. But if need to get a handle on the current view state to apply view updates.

efgpinto:
Completed in https://github.com/lightbend/kalix-jvm-sdk/pull/1092

PR #1092, opened 2022-09-02
feat(spring-sdk): provide access to current state in views

efgpinto:
References #1091 

This MR:
- adds access to current state in the context of a view
- adds IT tests with valueentity, view and transformed view


sdk/java-sdk/src/main/java/kalix/javasdk/view/View.java
```diff
@@ -24,6 +24,8 @@
 public abstract class View<S> {
 
   private Optional<UpdateContext> updateContext = Optional.empty();
+  private Optional<S> currentState = Optional.empty();
```
johanandren, line 27: I think this makes much less sense than for entities since it is not a state per view like the state for an entity, it's the state for a given subject/entity id update being processed right now. 

So passing it as a parameter would be a more reasonable API for view update methods IMO

efgpinto: I don't have a strong opinion on this.. happy to change it to a parameter for the views.
In that case, I guess we also want to enforce the signature of the annotated methods to be `something(TransformedUser, User)`. Would you still call it `currentState` btw? Or is there a better name then? 

 @octonato any thoughts here on using a param instead of `currentState()`?

johanandren: In the protocol-first samples we just call it `state`

octonato: May work for Views indeed. As long as we don't allow it to be updated from REST calls. We can't have two `@RequestBody`. 

I don't think we should restrict it to always receive the state. We can accept two signatures: 
* `onChange(User): TransformedUser`
* `onChange(TransformedUser, User): TransformedUser`

For Views based on ValueEntities, it will be convenient to simply have `onChange(User): TransformedUser`

The only drawback of using an extra method arg, is that it diverge from the other APIs in the Spring SDK. But worth evaluating it, IMO.

octonato: For event sourcing, two signatures may be interesting as well, for example:

* `onUserCreated(UserCreated): UserView`
* `onUserEmailUpdated(UserView, EmailUpdated): UserView`

efgpinto:
Updated now to allow for passing the `state`by param. Some tests failing with the Jackson parameters issue, will revisit tomorrow.

sdk/spring-sdk/src/it/java/com/example/wiring/SpringSdkWiringIntegrationTest.java
```diff
@@ -143,7 +188,7 @@ public void verifyFindUsersByEmail() {
             new IsEqual("jane.example.com"));
   }
 
-  //  @Test
+  // @Test
```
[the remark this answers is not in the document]
octonato: no, it wasn't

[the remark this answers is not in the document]
octonato: that might explain why it was green 😆

samples/spring-wiring-tests/src/main/java/wiring/view/UserWithVersion.java
```diff
@@ -0,0 +1,17 @@
...
+    public final String email;
+    public final int version;
+
+    @JsonCreator
+    public UserWithVersion(@JsonProperty String name, @JsonProperty String email, @JsonProperty int version) {
+        this.name = name;
+        this.email = email;
+        this.version = version;
+    }
+}
```
octonato, line 17: we can delete `spring-wiring-test` now, no?

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/reflection/KalixMethod.scala
```diff
@@ -95,9 +95,9 @@ case class VirtualServiceMethod(component: Class[_], methodName: String, inputTy
  * Build from methods annotated with @Subscription. Those methods are not annotated with Spring REST annotations, but
  * they become a REST method at the end.
  */
-case class RestServiceMethod(javaMethod: Method) extends AnyServiceMethod {
+case class RestServiceMethod(javaMethod: Method, inputTypeParamIndex: Int = 0) extends AnyServiceMethod {
```
octonato, line 98: ahah!

octonato (approved):
LGTM

I did a quick review. Good to have another ✅ before merging

sdk/spring-sdk/src/it/java/com/example/wiring/SpringSdkWiringIntegrationTest.java
octonato, line 191:
```suggestion
  @Test
```

samples/spring-wiring-tests/src/main/java/wiring/view/UserWithVersion.java
efgpinto: removed now

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/reflection/KalixMethod.scala
efgpinto: what was that? :D

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/view/ReflectiveViewRouter.scala
```diff
@@ -17,12 +17,15 @@
 package kalix.springsdk.impl.view
 
 import com.google.protobuf.any.{ Any => ScalaPbAny }
-import kalix.javasdk.Metadata
+import kalix.javasdk.{ JsonSupport, Metadata }
```
franciscolopezsancho, line 20: two lines, one line? :)

sdk/spring-sdk/src/test/java/kalix/springsdk/testmodels/view/ViewTestModels.java
```diff
@@ -194,4 +216,36 @@ public User emptyState() {
...
+
+    @Subscribe.ValueEntity(UserEntity.class)
+    public UpdateEffect<TransformedUser> onChange(User user, TransformedUser userView) {
+      return effects().updateState(new TransformedUser(userView.name, user.email));
+    }
+
+    @Query("SELECT * FROM users_view WHERE email = :email")
+    @PostMapping("/users/by-email")
+    public TransformedUser getUserByEmail(@RequestBody ByEmail byEmail) {
+      return null;
```
franciscolopezsancho, line 248: I checked and this is working but I fail to understand how returning `null` can work here.

sdk/spring-sdk/src/test/scala/kalix/springsdk/impl/ViewDescriptorFactorySpec.scala
```diff
@@ -16,15 +16,20 @@
...
-import kalix.springsdk.testmodels.view.ViewTestModels.TransformedUserView
-import kalix.springsdk.testmodels.view.ViewTestModels.UserByEmailWithGet
-import kalix.springsdk.testmodels.view.ViewTestModels.UserByEmailWithPost
-import kalix.springsdk.testmodels.view.ViewTestModels.UserByEmailWithPostRequestBodyOnly
-import kalix.springsdk.testmodels.view.ViewTestModels.UserByNameEmailWithPost
-import kalix.springsdk.testmodels.view.ViewTestModels.UserByNameStreamed
-import kalix.springsdk.testmodels.view.ViewTestModels.ViewWithNoQuery
-import kalix.springsdk.testmodels.view.ViewTestModels.ViewWithSubscriptionsInMixedLevels
-import kalix.springsdk.testmodels.view.ViewTestModels.ViewWithTwoQueries
+import kalix.springsdk.testmodels.view.ViewTestModels.{
```
franciscolopezsancho, line 19: format?

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/reflection/KalixMethod.scala
octonato: I think I added one `h` to much

It was intended to be: Aha! 

Like in: Aha! Now we have a solution for skipping the first arg and extracting the next one.

.circleci/config.yml
```diff
@@ -1016,23 +1016,6 @@ jobs:
...
-      # note: `copy-from-workspace` requires `publish-local` to be run first as declared in workflow section below
-      - copy-from-workspace
-      - set-sdk-version
-      - run:
-          name: Spring SDK wiring tests
-          command: |
-            cd samples/spring-wiring-tests
-            echo "Running mvn with SDK version: '$SDK_VERSION'"
-            # no testing for now, only 'compile'
-            mvn -Dkalix-sdk.version=$SDK_VERSION verify -Pit
```
octonato, line 1035: 💯

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/reflection/KalixMethod.scala
efgpinto: Aha! Now I get it! :D

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/ViewDescriptorFactory.scala
```diff
@@ -90,21 +90,36 @@ private[impl] object ViewDescriptorFactory extends ComponentDescriptorFactory {
...
               case Some(other) =>
                 throw InvalidComponentException(
                   s"All update methods must return the same type, but [${method.getName}] returns [${valueEntityClass.getName}] while a previous update method returns [${other.getName}]")
               case None => previousValueEntityClass = Some(valueEntityClass)
             }
-            // FIXME validate that transform method accepts value entity state type
+
+            method.getParameterTypes.toList match {
+              case p1 :: Nil if p1 != valueEntityEventClass =>
+                throw InvalidComponentException(
```
franciscolopezsancho, line 108: maybe not worthy but as they are the same error message would be beneficial to have a val with this exception. The idea behind this would be for the one coming next looking at this code not to have to read the whole exception to see if they have anything different. 
Feel free to skip this comment.

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/view/ReflectiveViewRouter.scala
```diff
@@ -31,19 +34,46 @@ class ReflectiveViewRouter[S, V <: View[S]](view: V, componentMethods: Map[Strin
...
+          .asInstanceOf[View.UpdateEffect[S]]
+      case (2, Some(s)) =>
+        componentMethod.method.get
+          .invoke(view, s, params.head)
+          .asInstanceOf[View.UpdateEffect[S]]
+      case (2, None) =>
+        componentMethod.method.get
+          .invoke(view, null, params.head)
+          .asInstanceOf[View.UpdateEffect[S]]
+      case (n, _) => // this shouldn't really be reached
```
franciscolopezsancho, line 74: :) Famous last words.

franciscolopezsancho (commented):
LGTM

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/view/ReflectiveViewRouter.scala
efgpinto: not sure what you mean :S would you elaborate?

sdk/spring-sdk/src/test/java/kalix/springsdk/testmodels/view/ViewTestModels.java
efgpinto: The implementation on the Query methods is not used. The proxy responds to the resquests directly so the body of the methods are never actually used. We might have a way to allow the user to keep it abstract in the future but we don't at the moment.

octonato: This method is never called, we only use it to tell the proxy how to create the index and how to respond. 

This is completely executed on the proxy side.

We will have to think how we want to solve this later. Users will be also puzzled by this.

sdk/spring-sdk/src/test/scala/kalix/springsdk/impl/ViewDescriptorFactorySpec.scala
efgpinto: fixed now, although I think we should probably rely this to the scala fmt

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/view/ReflectiveViewRouter.scala
efgpinto: yeah :D

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/ViewDescriptorFactory.scala
octonato: Not a `val`, but a `def`.

A `val` will create the exception eagerly.

efgpinto: I was in doubt about having 2 different msgs or not.. I will do this on a follow up PR.. just not to drag this one more.

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/view/ReflectiveViewRouter.scala
efgpinto: ah, you mean two separate lines?

I mean, I guess this should be automatic.. otherwise it's a pain.. I think this is fine if no checks are triggered. I checked that we have other instances in the code of using this format.

PR #1094, opened 2022-09-06
chore: do not allow stream-in calls in Spring SDK

octonato:
We previously decided that we won't support stream-in calls, but somehow I forgot to push the changes I had to block it. These changes were supported to go in #1087


He-Pin:
Why not just take 1 from it and compose with idleTimeout？

octonato:
> Why not just take 1 from it and compose with idleTimeout？

The problem is more related with how someone will stream data to it from a client. 

The proper way would be to switch to websocket and start to stream. On Kalix side, we would need to consume the websocket stream and delivery it through gRPC to the user function. It's certainly possible, but we don't believe this is common enough to worth the effort. Users willing to stream-in data will have the option to use gRPC directly.

efgpinto (approved):
LGTM

PR #1095, opened 2022-09-07
feat: adding maven archetype spring

franciscolopezsancho:
References https://github.com/lightbend/kalix/issues/7364


maven-java/kalix-maven-archetype-spring-value-entity/README.md
```diff
@@ -0,0 +1,27 @@
...
+
+For the latest release see [GitHub releases](https://github.com/lightbend/kalix-jvm-sdk/releases).
+
+To kickstart a project on Linux and macOS:
+
+```shell
+mvn \
+  archetype:generate \
+  -DarchetypeGroupId=io.kalix \
+  -DarchetypeArtifactId=kalix-maven-archetype-spring-value-entity \
```
octonato, line 15: The value entity archetype we have doesn't contain `-value-entity`. 

The goal is to have a short name for our default archetype. It happens to be one have has a value entity already encoded.

maven-java/kalix-maven-archetype-spring-value-entity/README.md
```diff
@@ -0,0 +1,27 @@
+# kalix-maven-archetype
+
+This archetype can be used to generate a project suitable for the development of a Service using a Value Entity with [Kalix](https://docs.kalix.io).
+
+**The Maven project created from the archetype will contain a Spring Value entity.**
```
octonato, line 5: It is still a Kalix Value Entity. 

Maybe we should say: 

```suggestion
**The Maven project created from the archetype will contain a Value Entity encoded using the Kalix Spring SDK.**
```

maven-java/kalix-maven-archetype-spring-value-entity/src/main/resources/archetype-resources/README.md
```diff
@@ -0,0 +1,88 @@
...
+
+```shell
+mvn compile exec:exec
+```
+
+With both the proxy and your application running, any defined endpoints should be available at `http://localhost:9000`. 
+
+FIXME update for spring
+
+* Create a customer with:
```
octonato, line 43: Our archetypes use a simple Counter model, not Customer registry. 

it should only contain the Value Entity, not Views. The goal is to make is a simple as possible.

octonato: @ennru / @beritou , this have been discussed in the past, but maybe worth to revisit it.

If I remember correctly, the reason why we added a counter protobuf in our gRPC archetype is that we have to teach people about protobuf, how to get started with our annotations and have some structure to start from. 

For the SpringSDK, it may make more sense to not have a model at all because the target audience knows already how to code in Java. Once we have the documentation in place, users can start from an empty archetype and code their model in Java right away. 

WDYT?

maven-java/kalix-maven-archetype-spring-value-entity/src/main/resources/archetype-resources/src/main/resources/logback.xml
```diff
@@ -0,0 +1,37 @@
...
+        <appender-ref ref="JSON-STDOUT"/>
+    </appender>
+
+    <logger name="akka" level="INFO"/>
+    <logger name="kalix" level="DEBUG"/>
+    <logger name="akka.http" level="INFO"/>
+    <logger name="io.grpc" level="INFO"/>
+
+    <root level="INFO">
+<!--        <appender-ref ref="STDOUT"/>-->
```
octonato, line 34:
```suggestion
```

octonato (commented):
Looking good, but we should use a `Counter` Value Entity or not have a pre-defined model at all. 

I asked for @ennru and @beritou ideas on the subject.

maven-java/kalix-maven-archetype-spring-value-entity/README.md
franciscolopezsancho: After our discussion @octonato, deciding to clean the content of the archetype the comment above is no longer relevant. I'll use a short name as you suggested.

maven-java/kalix-maven-archetype-spring/pom.xml
```diff
@@ -0,0 +1,62 @@
...
+    <artifactId>kalix-maven-archetype-spring</artifactId>
+    <version>1.0.5</version>
+    <packaging>maven-archetype</packaging>
+    <parent>
+        <groupId>io.kalix</groupId>
+        <artifactId>kalix-maven-java</artifactId>
+        <version>1.0.5</version>
+    </parent>
+
+    <name>Kalix Maven Archetype (Spring Value entity)</name>
```
octonato, line 13:
```suggestion
    <name>Kalix Maven Archetype (Spring SDK)</name>
```

maven-java/kalix-maven-archetype-spring/README.md
```diff
@@ -0,0 +1,24 @@
...
+
+For the latest release see [GitHub releases](https://github.com/lightbend/kalix-jvm-sdk/releases).
+
+To kickstart a project on Linux and macOS:
+
+```shell
+mvn \
+  archetype:generate \
+  -DarchetypeGroupId=io.kalix \
+  -DarchetypeArtifactId=kalix-maven-archetype-spring
```
octonato, line 13: I'm in doubt about the naming. 

The other archetypes start with `kalix-maven-archetype-` and then a categorisation. That makes sense because we make the choice for one or other type of component. 

While here, is an archetype for a specific Kalix SDK. 
My first thought was: `kalix-spring-maven-archetype` or `kalix-spring-boot-maven-archetype`. 

I wonder why we call it maven-archetype. It's not that we can call `mvn archetype` and have something else that is not a `maven-archetype`. 

Anyway, my current ideas are:

* `kalix-spring-boot-archetype` (my preferred)
* `kalix-spring-boot-maven-archetype` - maybe too vervose
* `kalix-maven-archetype-spring-boot` - also verbose
* `kalix-maven-archetype-spring` (currently in this PR)

WDYT?

maven-java/kalix-maven-archetype-spring/src/main/resources/archetype-resources/README.md
```diff
@@ -0,0 +1,56 @@
...
+
+
+## Designing
+
+To understand the Kalix concepts that are the basis for this example, see [Designing services](https://docs.kalix.io/services/development-process.html) in the documentation.
+
+
+## Developing
+
+This project contains the framework to create a Kalix application by adding Kalix components. To understand more about these components, see [Developing services]https://docs.kalix.io/services/). Spring-SDK is an experimental feature and so far there is no [official](https://docs.kalix.io/) documentation.
```
octonato, line 11: Good to mention that is an experimental feature. 👍 

We can add here links to the samples. At least, people willing to play with it will have some example to look at. 

We can link the samples folder and ask people to look for samples starting with `spring-`. As such as we don't need to keep updating this file each time we add a new sample.

octonato (commented):
Two more suggestions. 

Not sure about the archetype name though.

maven-java/kalix-maven-archetype-spring/src/main/resources/archetype-resources/README.md
franciscolopezsancho: Good point with adding the samples in general and in particular not adding any specific sample. Sehr gut!

maven-java/kalix-maven-archetype-spring/README.md
franciscolopezsancho: I agree the most accurate is `kalix-spring-boot-maven-archetype` but may be a bit too verbose. Let's save some reading time and use `kalix-spring-boot-archetype`.

octonato: yep, let's pick `kalix-spring-boot-archetype`

maven-java/kalix-spring-boot-archetype/src/main/resources/archetype-resources/pom.xml
```diff
@@ -0,0 +1,297 @@
...
+            </goals>
+          </execution>
+        </executions>
+        <configuration>
+          <dockerImage>${dockerImage}:${dockerTag}</dockerImage>
+          <mainClass>${mainClass}</mainClass>
+          <integrationTestSourceDirectory>src/it/java</integrationTestSourceDirectory>
+        </configuration>
+      </plugin>
+
```
octonato, line 226:
```suggestion
      <plugin>
       <groupId>org.springframework.boot</groupId>
       <artifactId>spring-boot-maven-plugin</artifactId>
      </plugin>
```

This will allow us to start this app like a Spring Boot app using `mvn spring-boot:run`

maven-java/kalix-spring-boot-archetype/src/main/resources/archetype-resources/README.md
```diff
@@ -0,0 +1,56 @@
...
+To start the proxy, run the following command from this directory:
+
+```shell
+docker-compose up
+```
+
+To start the application locally, the `exec-maven-plugin` is used. Use the following command:
+
+```shell
+mvn compile exec:exec
```
octonato, line 36:
```suggestion
mvn spring-boot:run
```

octonato (commented):
We are almost there. 

The pom.xml is not exactly the same we have in the samples. See comment about spring-boot plugin and how to start the app. 

And the Main file is being place under `src/main/java/customer`. That's a leftover from previous iteration. Normally, the package should be the one that the user chose for the groupId. Check the other archetype to see how this is made.

octonato (approved):
Great!
Working now perfectly. Cool to a nice simple archetype.

PR #1099, opened 2022-09-08
feat: adding Subcribe to Event Sourced Entitites

franciscolopezsancho:
References https://github.com/lightbend/kalix/issues/7361

### DIFF
--- a/sdk/spring-sdk/src/main/java/kalix/springsdk/annotations/Subscribe.java
+++ b/sdk/spring-sdk/src/main/java/kalix/springsdk/annotations/Subscribe.java
@@ -27,7 +27,7 @@ public @interface Subscribe {
     Class<? extends kalix.javasdk.valueentity.ValueEntity<?>> value();
   }
 
-  @Target({ElementType.TYPE, ElementType.METHOD})
+  @Target(ElementType.METHOD)
   @Retention(RetentionPolicy.RUNTIME)
   @Documented
   @interface EventSourcedEntity {
--- a/sdk/spring-sdk/src/main/java/kalix/springsdk/annotations/Subscribe.java
+++ b/sdk/spring-sdk/src/main/java/kalix/springsdk/annotations/Subscribe.java
@@ -27,6 +27,13 @@ public @interface Subscribe {
     Class<? extends kalix.javasdk.valueentity.ValueEntity<?>> value();
   }
 
+  @Target({ElementType.TYPE, ElementType.METHOD})
+  @Retention(RetentionPolicy.RUNTIME)
+  @Documented
+  @interface EventSourcedEntity {
+    Class<? extends kalix.javasdk.eventsourcedentity.EventSourcedEntity<?>> value();
+  }
+
   @Target(ElementType.METHOD)
   @Retention(RetentionPolicy.RUNTIME)
   @Documented
--- a/sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/reflection/KalixMethod.scala
+++ b/sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/reflection/KalixMethod.scala
@@ -95,11 +95,11 @@ case class VirtualServiceMethod(component: Class[_], methodName: String, inputTy
  * Build from methods annotated with @Subscription. Those methods are not annotated with Spring REST annotations, but
  * they become a REST method at the end.
  */
-case class RestServiceMethod(javaMethod: Method, optMethodName: Option[String] = None) extends AnyServiceMethod {
+case class RestServiceMethod(javaMethod: Method) extends AnyServiceMethod {
 
   val inputType: Class[_] = javaMethod.getParameterTypes()(0)
 
-  override def methodName: String = optMethodName.getOrElse(javaMethod.getName)
+  override def methodName: String = javaMethod.getName
 
   override def requestMethod: RequestMethod = RequestMethod.POST
   override def javaMethodOpt: Option[Method] = Some(javaMethod)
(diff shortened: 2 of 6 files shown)


sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/reflection/KalixMethod.scala
```diff
@@ -95,11 +95,11 @@ case class VirtualServiceMethod(component: Class[_], methodName: String, inputTy
  * Build from methods annotated with @Subscription. Those methods are not annotated with Spring REST annotations, but
  * they become a REST method at the end.
  */
-case class RestServiceMethod(javaMethod: Method, optMethodName: Option[String] = None) extends AnyServiceMethod {
```
franciscolopezsancho, line 98: not needed anymore. This was lingering after we decided not to pursue creating synthetic methods when adding @Subscribe.Topic. Instead we decided to forbid multiple @Subscribe.Topic for the same topic.

sdk/spring-sdk/src/test/scala/kalix/springsdk/impl/ActionDescriptorFactorySpec.scala
```diff
@@ -198,6 +202,42 @@ class ActionDescriptorFactorySpec extends AnyWordSpec with ComponentDescriptorSu
...
+        val eventSourceTwo = findKalixMethodOptions(desc, "MethodTwo").getEventing.getIn
+        eventSourceTwo.getEventSourcedEntity shouldBe "counter"
+
+        val ruleTwo = findHttpRule(desc, "MethodTwo")
+        ruleTwo.getPost shouldBe "/kalix.springsdk.testmodels.subscriptions.PubSubTestModels.SubscribeToEventSourcedEntityAction/MethodTwo"
+
+      }
+    }
+
+    "fail if has both Event Sourced Entity Subscription and REST annotations" in {
```
franciscolopezsancho, line 236: I assumed we don't want the users to add REST annotations themselves in combination with the @Subscribe.EventSourcedEntity. I think that this is because we already provide such an endpoint with @Subscribe.EventSourcedEntity but I think there is more to it.

sdk/spring-sdk/src/test/scala/kalix/springsdk/impl/ActionDescriptorFactorySpec.scala
```diff
@@ -33,13 +33,17 @@ import kalix.springsdk.testmodels.action.ActionsTestModels.PutWithoutParam
...
+import kalix.springsdk.testmodels.subscriptions.PubSubTestModels.{
+  PublishToTopicAction,
+  RestAnnotatedSubscribeToEventSourcedEntityAction,
+  RestAnnotatedSubscribeToValueEntityAction,
+  RestWithPublishToTopicAction,
+  SubscribeToEventSourcedEntityAction,
+  SubscribeToTopicAction,
+  SubscribeToTwoTopicsAction,
+  SubscribeToValueEntityAction
+}
```
octonato, line 46: Does that comes from IntelliJ or scalafmt? 
And if from Intellij, isn't scalafmt reverting it?

sdk/spring-sdk/src/test/scala/kalix/springsdk/impl/ActionDescriptorFactorySpec.scala
```diff
@@ -198,6 +202,42 @@ class ActionDescriptorFactorySpec extends AnyWordSpec with ComponentDescriptorSu
         assertRequestFieldJavaType(method, "one", JavaType.STRING)
       }
     }
+    "generate mappint with Event Sourced Entity Subscription annotation" in {
```
octonato, line 205:
```suggestion
    "generate mapping with Event Sourced Entity Subscription annotation" in {
```

sdk/spring-sdk/src/test/scala/kalix/springsdk/impl/ActionDescriptorFactorySpec.scala
octonato: Correct. 

And on top of this, I think that they are automatically blocked by the ACL. They are meant for internal use.

sdk/spring-sdk/src/main/java/kalix/springsdk/annotations/Subscribe.java
```diff
@@ -27,6 +27,13 @@
     Class<? extends kalix.javasdk.valueentity.ValueEntity<?>> value();
   }
 
+  @Target({ElementType.TYPE, ElementType.METHOD})
```
octonato, line 30:
```suggestion
  @Target(ElementType.METHOD)
```
I think we should have only `ElementType.METHOD`. 

The reason why we have `TYPE` for ValueEntities is becaue we can have a View without transformation. When used without transformation, there is no method to transform the state from `Foo` to `ViewFoo` and then the annotation can be added at class level. 

For event sources, it will always be from one or many events to some other derived view state. So, by definition we need to have event handler methods for each event.

octonato (commented):
Looking good. I added some suggestions and typo fixes.

It will be good to extend the integration tests. We have already ES, but we could add a view subscribing to its events.

sdk/spring-sdk/src/test/scala/kalix/springsdk/impl/ActionDescriptorFactorySpec.scala
franciscolopezsancho: again?? darn! I change it a couple of times back but it seems IntelliJ has beaten me

sdk/spring-sdk/src/main/java/kalix/springsdk/annotations/Subscribe.java
franciscolopezsancho: understood 👍

sdk/spring-sdk/src/test/scala/kalix/springsdk/impl/ActionDescriptorFactorySpec.scala
franciscolopezsancho: let me check who is the culprit here. Apart from me.

franciscolopezsancho: It was Intellij. Disabled it 
Screenshot 2022-09-09 at 13 53 02

octonato (approved):
LGTM

One test failed, but due to some transient error download some artifacts. Let's merge when green.

PR #1100, opened 2022-09-08
feat(spring-sdk): streaming views

octonato:
Tests are not yet passing. Will fixed it tomorrow.

### DIFF
--- a/sdk/spring-sdk/src/it/java/com/example/wiring/SpringSdkWiringIntegrationTest.java
+++ b/sdk/spring-sdk/src/it/java/com/example/wiring/SpringSdkWiringIntegrationTest.java
@@ -18,18 +18,30 @@ package com.example.wiring;
 
 import com.example.Main;
 import com.example.wiring.actions.echo.Message;
+import com.example.wiring.valueentities.user.User;
 import kalix.springsdk.KalixConfigurationTest;
+import org.hamcrest.core.IsEqual;
 import org.junit.jupiter.api.Assertions;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.extension.ExtendWith;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.boot.test.context.SpringBootTest;
 import org.springframework.context.annotation.Import;
+import org.springframework.http.HttpStatus;
+import org.springframework.http.ResponseEntity;
 import org.springframework.test.context.TestPropertySource;
 import org.springframework.test.context.junit.jupiter.SpringExtension;
 import org.springframework.web.reactive.function.client.WebClient;
+import reactor.core.publisher.Flux;
 
 import java.time.Duration;
+
+import static java.time.temporal.ChronoUnit.SECONDS;
+
+import java.util.concurrent.TimeUnit;
+
+import static org.awaitility.Awaitility.await;
+
 import java.util.List;
 import java.util.stream.Collectors;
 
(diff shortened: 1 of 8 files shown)


sdk/spring-sdk/src/it/java/com/example/wiring/SpringSdkWiringIntegrationTest.java
```diff
@@ -102,4 +114,75 @@ public void verifyCounterEventSourcedWiring() {
...
     Assertions.assertEquals("\"200\"", counterGet);
   }
+
+  @Test
+  public void verifyFindUsersByEmail() {
+
+    ResponseEntity<String> response =
+        webClient
+            .post()
+            .uri("/user/jane/jane/jane.example.com")
```
efgpinto, line 124: I think this is why the test is failing

```suggestion
            .uri("/user/jane/jane.example.com/jane")
```

sdk/spring-sdk/src/it/java/com/example/wiring/SpringSdkWiringIntegrationTest.java
```diff
@@ -102,4 +114,75 @@ public void verifyCounterEventSourcedWiring() {
...
+  }
+
+  //  @Test
+  public void verifyFindUsersByNameStreaming() {
+
+    { // joe 1
+      ResponseEntity<String> response =
+          webClient
+              .post()
+              .uri("/user/user1/joe/joe1.example.com")
```
octonato, line 153:
```suggestion
              .uri("/user/user1/joe1.example.com/joe")
```

sdk/spring-sdk/src/it/java/com/example/wiring/SpringSdkWiringIntegrationTest.java
```diff
@@ -102,4 +114,75 @@ public void verifyCounterEventSourcedWiring() {
...
+              .block(timeout);
+
+      Assertions.assertEquals(HttpStatus.OK, response.getStatusCode());
+    }
+
+    { // joe 2
+      ResponseEntity<String> response =
+          webClient
+              .post()
+              .uri("/user/user2/joe/joe2.example.com")
```
octonato, line 165:
```suggestion
              .uri("/user/user2/joe2.example.com/joe")
```

sdk/spring-sdk/src/it/java/com/example/wiring/views/UsersByName.java
```diff
@@ -0,0 +1,37 @@
...
+import org.springframework.web.bind.annotation.GetMapping;
+import reactor.core.publisher.Flux;
+
+@Table("users_by_name")
+@Subscribe.ValueEntity(UserEntity.class)
+public class UsersByName extends View<User> {
+
+  @GetMapping("/users/by_name/{name}")
+  @Query("SELECT * FROM users_by_name WHERE name = :name")
+  public Flux<User> getUsers(String name) {
```
efgpinto, line 34: Just crossed my mind that in the future it could make sense to also allow the user to use a Collection instead of Flux.. and in that case it would no longer be a stream response but a wrapped the result (I think we support this correctly as well with a wrapper?)

efgpinto (approved):
LGTM

sdk/spring-sdk/src/it/java/com/example/wiring/views/UsersByName.java
octonato: In the gRPC SDKs this is done by using a wrapper, like in: 

```protobuf
message CustomersResponse {
  repeated domain.CustomerState results = 1; 
}
```

What we could do for the Spring SDK, is to generate a synthetic proto message when the method returns a Java collection. The synthetic message will be the wrapper and the users don't need to deal with it. 

The Json response on the side of the caller, will be just a Json Array.

octonato: Let's keep the ticket current ticket open. We can add this on follow up PRs.

PR #1101, opened 2022-09-09
chore: merge main into spring sdk 

octonato:
Getting it in closed to `main` before we merge the `spring-sdk` into `main`

Also because I think it will make it easier if all samples move to 1.0.6 first. A PR from spring-sdk to main will probably fail.

### DIFF
--- a/sdk/java-sdk/src/main/scala/kalix/javasdk/impl/action/ActionsImpl.scala
+++ b/sdk/java-sdk/src/main/scala/kalix/javasdk/impl/action/ActionsImpl.scala
@@ -265,6 +265,7 @@ private[javasdk] final class ActionsImpl(
               // user stream failed with an "unexpected" error
               handleUnexpectedException(service, in, ex)
             }
+            .async
         } catch {
           case NonFatal(ex) =>
             // command handler threw an "unexpected" error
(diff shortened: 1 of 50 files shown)


sdk/java-sdk/src/main/scala/kalix/javasdk/impl/action/ActionsImpl.scala
```diff
@@ -265,6 +265,7 @@ private[javasdk] final class ActionsImpl(
               // user stream failed with an "unexpected" error
               handleUnexpectedException(service, in, ex)
             }
+            .async
```
franciscolopezsancho, line 268: Is this useful because it will allow each service/route to run upon a dedicated actor? Therefore each incoming request gets isolated by service type? i.e. Action, View, ESE, RESE, and VE?

franciscolopezsancho (commented):
Apart from the comment
LGTM

efgpinto:
Reminder to not squash this when merging

sdk/java-sdk/src/main/scala/kalix/javasdk/impl/action/ActionsImpl.scala
efgpinto: For the record if you're curious, this comes from [here](https://github.com/lightbend/kalix-jvm-sdk/issues/1078) and there was a nice and informative thread [here](https://typesafe.slack.com/archives/CQUASEUDV/p1661512698803409). But to quote Johan, I think the summary was:
> One client shared between all actors for all entity ids on the node -> one connection, all request sequenced.

PR #1107, opened 2022-09-12
chore: fix release issue link

efgpinto:
No issue.. just noted a broken link in the release issue.

(we still have a operations dashboard but I think this one is more high-level so seems more suited to check the versions)


ennru (approved):
LGTM.

PR #1111, opened 2022-09-14
chore: update poms updating script to work for spring samples

efgpinto:
Noticed when doing https://github.com/lightbend/kalix-jvm-sdk/issues/1106


octonato (approved):
lgtm

PR #1113, opened 2022-09-15
chore: adding new event-sourced sample with Spring SDK

efgpinto:
Refs https://github.com/lightbend/kalix/issues/7473

Ended up going with the most simple sample of all with the Counter. Let me know if you think if would be worth to add a more complex one.

### DIFF
--- a/samples/spring-eventsourced-counter/README.md
+++ b/samples/spring-eventsourced-counter/README.md
@@ -34,7 +34,24 @@ mvn spring-boot:run
 
 With both the proxy and your application running, once you have defined endpoints they should be available at `http://localhost:9000`. 
 
+### Examples
 
+- increase (or create) a counter named `hello` with value `10`
+```shell
+curl -i -XPOST localhost:9000/counter/hello/increase/10
+```
+
+- retrieve the value of a counter named `hello`
+```shell
+curl -i -XGET localhost:9000/counter/hello
+```
+
+- multiply existing counter named `hello` by value `5`
+```shell
+curl -i -XPOST localhost:9000/counter/hello/multiply/5
+```
+
+### Deploy
 
 To deploy your service, install the `kalix` CLI as documented in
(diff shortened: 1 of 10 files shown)


samples/spring-eventsourced-counter/README.md
```diff
@@ -0,0 +1,51 @@
...
+docker-compose up
+```
+
+To start the application locally, the `exec-maven-plugin` is used. Use the following command:
+
+```shell
+mvn spring-boot:run
+```
+
+With both the proxy and your application running, once you have defined endpoints they should be available at `http://localhost:9000`.
```
franciscolopezsancho, line 35: would it be useful to add some examples to POST and GET so the user can execute them right away?

samples/spring-eventsourced-counter/pom.xml
```diff
@@ -0,0 +1,307 @@
...
+      <artifactId>junit-jupiter-api</artifactId>
+      <version>5.5.2</version>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.junit.jupiter</groupId>
+      <artifactId>junit-jupiter</artifactId>
+      <version>5.5.2</version>
+      <scope>test</scope>
+    </dependency>
```
franciscolopezsancho, line 304:
```suggestion
```

samples/spring-eventsourced-counter/README.md
efgpinto: wondered that myself but ended up not adding them.. think it would be useful though. Let me add some.

samples/spring-eventsourced-counter/pom.xml
efgpinto: Actually we need that dependency but not the `junit-jupiter-api` above.. Updated now.

franciscolopezsancho (approved):
LGTM

PR #1115, opened 2022-09-16
feat(spring-sdk): subscribe to event sourced entities

octonato:
References: https://github.com/lightbend/kalix/issues/7361


sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/ViewDescriptorFactory.scala
```diff
@@ -31,98 +33,46 @@ import kalix.springsdk.impl.reflection.ReflectionUtils
...
-        // create a virtual method
-        val methodOptionsBuilder = kalix.MethodOptions.newBuilder()
+      if (hasTypeLevelValueEntitySubs)
+        subscriptionForTypeLevelValueEntity(component, tableType, tableName, tableProtoMessageName)
+      else if (hasMethodLevelValueEntitySubs)
+        subscriptionForMethodLevelValueEntity(component, tableType, tableName, tableProtoMessageName)
+      else if (hasMethodLevelEventSourcedEntitySubs)
+        eventSourcedEntitySubscription(component, tableType, tableName, tableProtoMessageName)
+      else
+        Seq.empty
```
octonato, line 74: this block was getting too big, so I move the code in their own methods.

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/ViewDescriptorFactory.scala
```diff
@@ -136,10 +86,12 @@ private[impl] object ViewDescriptorFactory extends ComponentDescriptorFactory {
         .filter(_.javaMethod.getAnnotation(classOf[Query]) != null)
       if (annotatedMethods.isEmpty)
         throw new IllegalArgumentException(
-          "No method annotated with @Query found. " +
-          "Views should have a method annotated with @Query")
+          s"No valid query method found in class ${component.getName}. " +
+          "Views should have a method annotated with @Query and annotated exposed by a REST annotation")
```
octonato, line 90: I had a method with the Query annotation, but was getting the message "No method annotated with @Query found". Misleading. 

Also, I think we should first check for the `@Query` annotation and then for the other methods. Once we read the `Query` method, we know the view type and can then validate the other methods. 

But we can improve it in other iterations.

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/ViewDescriptorFactory.scala
```diff
@@ -206,6 +158,161 @@ private[impl] object ViewDescriptorFactory extends ComponentDescriptorFactory {
...
+          case p1 :: Nil if p1 == tableType =>
+            invalidComponentException(method, "Subscription method only has the view type.")
+
+          case p1 :: p2 :: Nil if p1 != tableType =>
+            invalidComponentException(method, s"Subscription method first param must be view type [$tableType]")
+
+          case p1 :: p2 :: Nil if p1 == tableType && p2 == tableType =>
+            invalidComponentException(method, "Subscription method receives twice the view type [$tableType]")
+          case _ => // happy days, dev did good with the signature
+        }
```
octonato, line 210: In retrospect, I think the idea of allowing the state to be part of the method was not so great. There are so many situations that users can get it wrong. I don't think that giving this option is helping the users. 

I think we should consider going to the original idea of having a `viewState` method.

sdk/spring-sdk/src/it/java/com/example/wiring/eventsourcedentities/counter/CreateEmployee.java
```diff
@@ -0,0 +1,31 @@
...
+public class CreateEmployee {
+  public final String name;
+  public final String email;
+
+  @JsonCreator
+  public CreateEmployee(@JsonProperty String name, @JsonProperty String email) {
+    this.name = name;
+    this.email = email;
+  }
+}
```
efgpinto, line 31: I don't think this one is being used for the IT tests, is it?

sdk/spring-sdk/src/it/java/com/example/wiring/views/CountersByValue.java
```diff
@@ -0,0 +1,47 @@
...
+  //  public Counter onEvent(Counter counter, CounterEvent event) {
+  //
+  //    if (event instanceof ValueIncreased) {
+  //      return counter.onValueIncreased((ValueIncreased) event);
+  //    } else if (event instanceof ValueMultiplied) {
+  //      return counter.onValueMultiplied((ValueMultiplied) event);
+  //    } else {
+  //      return counter;
+  //    }
+  //  }
```
efgpinto, line 45: Is this meant to be kept commented?

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/ViewDescriptorFactory.scala
efgpinto, line 90:
```suggestion
          "Views should have a method annotated with @Query and exposed by a REST annotation")
```

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/ViewDescriptorFactory.scala
efgpinto: Yeah, kind of agree with that.. also, it would be consistent with the remaining component types.

efgpinto (commented):
Left some chore comments but LGTM otherwise.

sdk/spring-sdk/src/test/java/kalix/springsdk/testmodels/view/ViewTestModels.java
```diff
@@ -210,15 +174,9 @@ public static class UserByNameStreamed extends View<User> {
...
     }
-
-    @Override
-    public User emptyState() {
-      return null;
-    }
   }
 
   @Table(value = "users_view")
-  @Subscribe.ValueEntity(UserEntity.class)
```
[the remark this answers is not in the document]
octonato: The tests were passing, but because of the wrong reason. It came to the surface when I started to check the error message.

sdk/spring-sdk/src/it/java/com/example/wiring/eventsourcedentities/counter/CreateEmployee.java
octonato: Good catc. No, it's not. I will remove it

sdk/spring-sdk/src/it/java/com/example/wiring/views/CountersByValue.java
octonato: No, and that's probably why the tests are not passing.

I don't remember why I commented it out. 🤷

efgpinto (approved):
LGTM

PR #1117, opened 2022-09-16
feat: use JsonTypeInfo to fill typeUrl

octonato:
References https://github.com/lightbend/kalix/issues/7453

### DIFF
--- a/sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/SpringSdkMessageCodec.scala
+++ b/sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/SpringSdkMessageCodec.scala
@@ -16,6 +16,9 @@
 
 package kalix.springsdk.impl
 
+import java.util.concurrent.ConcurrentHashMap
+import java.util.concurrent.ConcurrentMap
+
 import com.fasterxml.jackson.annotation.JsonSubTypes
 import com.fasterxml.jackson.annotation.JsonTypeName
 import com.google.protobuf.any.{ Any => ScalaPbAny }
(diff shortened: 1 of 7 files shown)


sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/SpringSdkMessageCodec.scala
```diff
@@ -30,10 +32,59 @@ private[springsdk] object SpringSdkMessageCodec extends MessageCodec {
...
+
+      val directParents = clz.getSuperclass +: clz.getInterfaces
+
+      directParents.foldLeft(acc) { case (acc, clz) =>
+        if (clz == null) acc // happens when we reach the bottom, ie: Object.getSuperclass == null
+        else annotatedParents(clz, acc)
+      }
+    }
+
+    // straightforward case: class is annotated with JsonTypeName
```
franciscolopezsancho, line 60: I really like these comments. They show the intent which is not often obvious. I think we would be better off with more of them. Would it be worthy to discuss this as a team?

sdk/spring-sdk/src/test/scala/kalix/springsdk/impl/SpringSdkMessageCodecSpec.scala
```diff
@@ -0,0 +1,196 @@
...
+
+  }
+}
+class SpringSdkMessageCodecSpec extends AnyWordSpec with Matchers {
+
+  "The SpringSdkMessageCodec" should {
+
+    "by default FQCN for typeUrl (java)" in {
+      val encoded = SpringSdkMessageCodec.encodeJava(SimpleClass("abc", 10))
+      encoded.getTypeUrl shouldBe "json.kalix.io/kalix.springsdk.impl.SpringSdkMessageCodecSpec$SimpleClass"
```
franciscolopezsancho, line 95: maybe using here the variable `kalix.javasdk.KALIX_JSON` to show where this comes from?

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/SpringSdkMessageCodec.scala
```diff
@@ -30,10 +32,59 @@ private[springsdk] object SpringSdkMessageCodec extends MessageCodec {
...
+    if (value.getClass.getAnnotation(classOf[JsonTypeName]) != null) {
+      value.getClass.getAnnotation(classOf[JsonTypeName]).value()
+    } else {
+      // otherwise needs to scan hierarchy until we find JsonSubTypes annotations
+      // in a parent class or trait
+      val parents = annotatedParents(value.getClass, Seq.empty)
+      if (parents.isEmpty)
+        value.getClass.getName
```
patriknw, line 68: I haven't followed what you have discussed for the type hint, but I'm not sure we should use class names.
* It's a source of many of the security problems with Jackson.
* It's volatile for refactoring changes.
* It's not interoperable with non-java clients.

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/SpringSdkMessageCodec.scala
octonato: I don't think that's worth. It's more a matter of style and taste.

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/SpringSdkMessageCodec.scala
octonato: Maybe some more context...

We have `typeUrl` using the FQCN all the time. Any state serialized to proto any will have it. Events as well. 

In principle, it will never be a problem for state (famous last words). The reason is that we don't use this information to deserialize anything. It's there because we need to fill in the `typeUrl`. 

When we deserialize the json to some java type, we know already, based on the the user code, which type it should be. User can move their classes around and it will still work. As long as the class and json aligns. 

The `typeUrl` became an issue for the events because now we need this information to be able to dispatch the events to the right methods inside the Spring SDK. See https://github.com/lightbend/kalix/issues/7512

Now we have all interest to provide a stable `typeUrl`.

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/SpringSdkMessageCodec.scala
```diff
@@ -16,24 +16,84 @@
...
+            val subTypeAnn = parent.getAnnotation(classOf[JsonSubTypes])
+            subTypeAnn.value().find(_.value() == messageClass)
+          }
+        }.headOption
+
+        ann
+          .map { a =>
+            // if more than one name, we pick the first one for the typeUrl
+            // otherwise, default to `name`
+            a.names().headOption.getOrElse(a.name())
```
octonato, line 91: This is more complicate bit that we will have to address after https://github.com/lightbend/kalix/issues/7512

In a nutshell, the jackson annotation allow json payloads with different type hints to be mapped to a single class. That's why they have this `names` array field in the annotation. 

On our side, when generating the mapping between typeUrl and Java method (in https://github.com/lightbend/kalix/issues/7512), we will need to generate as many as typeUrl entries as type names we find in `names`.

efgpinto (approved):
I'm not too versed in these Json subtypes but otherwise LGTM.

octonato:
Me neither! :-)

PR #1118, opened 2022-09-19
feat(spring-sdk): add support for cross-component calls

efgpinto:
References https://github.com/lightbend/kalix/issues/7370

In this PR, we'll start with a limited functionality in terms of DeferredCalls:
- [x] define kalixClient api (at least for the most usual methods and within the same service)
- [x] generated dummy deferredCalls to be used with `.execute`directly in tests
- [x] discover proxy address and port automatically and inject into webclient
- [x] add integrations tests
- [x] update a sample to use a cross-component call

### DIFF
--- a/sdk/spring-sdk/src/it/java/com/example/wiring/actions/echo/ShortenedEchoAction.java
+++ b/sdk/spring-sdk/src/it/java/com/example/wiring/actions/echo/ShortenedEchoAction.java
@@ -35,7 +35,7 @@ public class ShortenedEchoAction extends Action {
   @GetMapping("/echo/message/{msg}/short")
   public Effect<Message> stringMessage(@PathVariable String msg) {
     var shortenedMsg = msg.replaceAll("[AEIOUaeiou]", "");
-    var deferredCall = kalixClient.getCall("/echo/message/" + shortenedMsg, Message.class);
-    return effects().asyncReply(deferredCall);
+    var result = kalixClient.get("/echo/message/" + shortenedMsg, Message.class).execute();
+    return effects().asyncReply(result);
   }
 }
--- a/sdk/spring-sdk/src/it/java/com/example/wiring/actions/echo/ShortenedEchoAction.java
+++ b/sdk/spring-sdk/src/it/java/com/example/wiring/actions/echo/ShortenedEchoAction.java
@@ -1,0 +1,41 @@
+/*
+ * Copyright 2021 Lightbend Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.example.wiring.actions.echo;
+
+import kalix.javasdk.action.Action;
+import kalix.javasdk.action.ActionCreationContext;
+import kalix.springsdk.impl.KalixClient;
+import org.springframework.web.bind.annotation.GetMapping;
+import org.springframework.web.bind.annotation.PathVariable;
+
+public class ShortenedEchoAction extends Action {
+
+  private ActionCreationContext ctx;
+  private KalixClient kalixClient;
+
+  public ShortenedEchoAction(ActionCreationContext ctx, KalixClient kalixClient) {
+    this.ctx = ctx;
+    this.kalixClient = kalixClient;
+  }
+
+  @GetMapping("/echo/message/{msg}/short")
+  public Effect<Message> stringMessage(@PathVariable String msg) {
+    var shortenedMsg = msg.replaceAll("[AEIOUaeiou]", "");
+    var deferredCall = kalixClient.get("/echo/message/" + shortenedMsg, Message.class);
+    return effects().asyncReply(deferredCall.execute());
+  }
+}
--- a/sdk/spring-sdk/src/it/java/com/example/wiring/actions/echo/ShortenedEchoAction.java
+++ b/sdk/spring-sdk/src/it/java/com/example/wiring/actions/echo/ShortenedEchoAction.java
@@ -35,7 +35,7 @@ public class ShortenedEchoAction extends Action {
   @GetMapping("/echo/message/{msg}/short")
   public Effect<Message> stringMessage(@PathVariable String msg) {
     var shortenedMsg = msg.replaceAll("[AEIOUaeiou]", "");
-    var deferredCall = kalixClient.get("/echo/message/" + shortenedMsg, Message.class);
-    return effects().asyncReply(deferredCall.execute());
+    var deferredCall = kalixClient.getCall("/echo/message/" + shortenedMsg, Message.class);
+    return effects().asyncReply(deferredCall);
   }
 }
--- a/sdk/spring-sdk/src/it/java/com/example/wiring/actions/echo/ShortenedEchoAction.java
+++ b/sdk/spring-sdk/src/it/java/com/example/wiring/actions/echo/ShortenedEchoAction.java
@@ -18,7 +18,7 @@ package com.example.wiring.actions.echo;
 
 import kalix.javasdk.action.Action;
 import kalix.javasdk.action.ActionCreationContext;
-import kalix.springsdk.impl.KalixClient;
+import kalix.springsdk.KalixClient;
 import org.springframework.web.bind.annotation.GetMapping;
 import org.springframework.web.bind.annotation.PathVariable;
 
--- a/sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/KalixClient.scala
+++ b/sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/KalixClient.scala
@@ -16,51 +16,71 @@
 
 package kalix.springsdk.impl
 
-import kalix.javasdk.DeferredCall
-import kalix.javasdk.impl.{ DeferredCallImpl, MetadataImpl }
+import com.google.protobuf.ExperimentalApi
+import org.springframework.http.{ HttpHeaders, MediaType }
 import org.springframework.web.reactive.function.client.WebClient
 
-trait KalixClient {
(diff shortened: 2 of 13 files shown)


efgpinto:
Opening for review now. API is not actually in its final form because we have decided to hide the `execute()` away from the user for now so that will be defined at the next stage.

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/KalixClient.scala
```diff
@@ -0,0 +1,86 @@
...
+
+trait KalixClient {
+  //def post[P, R](uri: String, body: P, returnType: Class[R]): DeferredCall[P, R]
+  //def get[R](uri: String, resp: Class[R]): DeferredCall[_, R] // FIXME not sure what to use by default
+
+  @ExperimentalApi
+  def postCall[P, R](uri: String, body: P, returnType: Class[R]): CompletionStage[R]
+
+  @ExperimentalApi
+  def getCall[R](uri: String, resp: Class[R]): CompletionStage[R]
```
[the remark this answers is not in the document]
octonato: Now that I think about it one more time, maybe we can have already a `post` and `get` returning a `DeferredCall`. 

But instead of returning `DeferredCallImpl` we return `RestDeferredCallImpl`. The `execute` method for `RestDeferredCallImpl` we run the web client call. 

If we take that path, we won't break any code. The API will be 'stable' already. We can still mark them with `@ApiMayChange` though. 

My point is, we will be able to cut a release for now, even if the transcoding is not yet in place. Obviously, `forwards` and `timers` won't work. We will need to have some clear error message about it. Also document that they are not yet supported.

octonato (commented):
Looking good. 

I realise that we should already return a `DeferredCall` even if it won't work for now. See my comments below.

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/KalixClient.scala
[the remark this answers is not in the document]
efgpinto: I see what you mean, that would be better indeed. Let me try

sdk/spring-sdk/src/it/java/com/example/wiring/actions/echo/ShortenedEchoAction.java
```diff
@@ -0,0 +1,41 @@
...
+  public ShortenedEchoAction(ActionCreationContext ctx, KalixClient kalixClient) {
+    this.ctx = ctx;
+    this.kalixClient = kalixClient;
+  }
+
+  @GetMapping("/echo/message/{msg}/short")
+  public Effect<Message> stringMessage(@PathVariable String msg) {
+    var shortenedMsg = msg.replaceAll("[AEIOUaeiou]", "");
+    var deferredCall = kalixClient.getCall("/echo/message/" + shortenedMsg, Message.class);
+    return effects().asyncReply(deferredCall);
```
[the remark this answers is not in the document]
efgpinto: updated in last commit

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/KalixClient.scala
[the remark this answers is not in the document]
efgpinto: Updated in last version.

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/KalixServer.scala
```diff
@@ -231,50 +234,43 @@ class KalixServer(applicationContext: ApplicationContext, config: Config) {
...
+      clz,
+      context => {
+        if (hasContextConstructor(clz, classOf[ActionCreationContext]))
           threadLocalActionContext.set(context)
-          kalixBeanFactory.getBean(clz)
-        })
-    else
-      ReflectiveActionProvider.of(clz, _ => kalixBeanFactory.getBean(clz))
+
+        val grpcClients = GrpcClients.get(context.materializer().system)
```
octonato, line 243:
```suggestion
        val grpcClients = GrpcClients(context.materializer().system)
```

The `.get` is usually for Java APIs.

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/KalixClient.scala
```diff
@@ -0,0 +1,91 @@
...
+
+  @ApiMayChange
+  def post[P, R](uri: String, body: P, returnType: Class[R]): DeferredCall[P, R]
+
+  @ApiMayChange
+  def get[R](uri: String, resp: Class[R]): DeferredCall[_, R]
+
+}
+
+final class RestKalixClientImpl extends KalixClient {
```
octonato, line 39:
```suggestion
/**
 * INTERNAL API
 */
final class RestKalixClientImpl extends KalixClient {
```

Everything that is in package `impl` should be consider internal api, but also good to add some explicit message agout it.

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/KalixClient.scala
```diff
@@ -0,0 +1,91 @@
...
+import kalix.javasdk.{ DeferredCall, Metadata }
+import org.springframework.http.{ HttpHeaders, MediaType }
+import org.springframework.web.reactive.function.client.WebClient
+
+import java.util.concurrent.CompletionStage
+import scala.concurrent.{ Future, Promise }
+import scala.jdk.FutureConverters._
+import scala.concurrent.ExecutionContext.Implicits.global
+
+trait KalixClient {
```
octonato, line 29: I think this should move to `kalix.springsdk` package. 

It's an API that we are providing for users delight. 😄

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/KalixClient.scala
```diff
@@ -0,0 +1,91 @@
...
+
+  @ApiMayChange
+  def get[R](uri: String, resp: Class[R]): DeferredCall[_, R]
+
+}
+
+final class RestKalixClientImpl extends KalixClient {
+
+  private class RestDeferredCallImpl[P, R](asyncCall: () => CompletionStage[R]) extends DeferredCall[P, R] {
+    override def message(): P = ???
```
octonato, line 42: Here is a place where we could add some clear error message to the users. 

This method will probably be called if we try to send this DeferredCall to the proxy. At some point, we will need to convert it to the a proto messages that is put on the wire.

We can throw an IllegalArgumentException saying that REST based DeferredCalls are not yet supported in forwards and timers.

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/KalixClient.scala
```diff
@@ -0,0 +1,91 @@
...
+  def post[P, R](uri: String, body: P, returnType: Class[R]): DeferredCall[P, R]
+
+  @ApiMayChange
+  def get[R](uri: String, resp: Class[R]): DeferredCall[_, R]
+
+}
+
+final class RestKalixClientImpl extends KalixClient {
+
+  private class RestDeferredCallImpl[P, R](asyncCall: () => CompletionStage[R]) extends DeferredCall[P, R] {
```
octonato, line 41:
```suggestion
/**
 * INTERNAL API
 */
  private class RestDeferredCallImpl[P, R](asyncCall: () => CompletionStage[R]) extends DeferredCall[P, R] {
```

We can also make it a final case class, like DeferredCallImpl

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/KalixClient.scala
octonato: But we will need to test it out. We must make sure that the error will be propagated to user.

octonato: Hmm, maybe better to not solve it with a runtime error, but let the compiler block it. 

We can for instance, keep `RestDeferredCallImpl`, but not let it extends `DeferredCall` for now. We keep the type though so users code will stay: `client.post("/url", Input, Output.class).execute()`. 

But users won't be able to pass it to `forward` and `timer`.

When we add the local transcoding we make it extend `DeferredCall` and everything will fall in its place.

octonato (approved):
LGMT, after some small changes.

@efgpinto, I realise that it will be easier for us if we don't make `RestDeferredCallImpl` a `DeferredCall` in this PR, but only when we add support for transcoding. See my comments below.

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/KalixClient.scala
efgpinto: > Here is a place where we could add some clear error message to the users.

Yeah, I thought the same and had a IllegalArgumentException there but it was not really called anywhere and the user would still get the default unknown effect. In any case, agree that blocking at compile time is better.

So, to do that, the `KalixClient` API needs to return something different than a DeferredCall as well. Made it return a `RestDeferredCall` to avoid leaking the Impl.

(thanks for the suggestions Renato 🙇 )

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/KalixClient.scala
efgpinto: absolutely, will update. thanks Renato.

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/KalixClient.scala
octonato: Correct, there is a chance that users will assign the `RestDeferredCall` to a variable, but that's the price for using experimental stuff. :-)

octonato: Btw, let's keep it as internal API any.

efgpinto:
Updated now with the suggestions but still having issues when running in dev. So, not ready for merge.

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/KalixClient.scala
efgpinto: Got it, updated. Thanks again for all the tips 🙇

efgpinto:
For the record, tested it in dev, it was not working initially but #1122 fixed that. Tried both a version with identification headers and without and both worked (this only allows local calls to the same service) so agreed we can go without the headers being added. We will take care of that when working on ACLs and/or cross-service calls.

PR #1119, opened 2022-09-21
chore: ApiMayChange for currentState() methods

octonato:
Those methods were added to support the Spring SDK. At some point in the future, we may need or prefer to have them moved to Spring SDK specific component classes. 

I had previously a PR on this: https://github.com/lightbend/kalix-jvm-sdk/pull/1018, but closed because I was not so happy with the directly. 

I would like to add this annotation just for good measure and give us the option to move then later.


efgpinto (approved):
This also reminded me that we talked about revisiting the view state (similar to this vs passing as param as currently).

LGTM anyway.

PR #1120, opened 2022-09-21
chore(ci): add missing sample and run samples after it tests

efgpinto:
Follow up of https://github.com/lightbend/kalix-jvm-sdk/pull/1113

Just noticed that the sample was not being run. Also, since we have the IT tests for spring-sdk I think it makes sense to only run the samples after the IT tests pass. Thoughts?


octonato (approved):
Yep, it's good to have a first barrier.

PR #1121, opened 2022-09-21
chore: bump proxy version

octonato:
Seems that SDK docker-compose were point to a quite old proxy version. 

We do need to ensure that 1.0.20 is already in prod before we merge this though.


octonato:
I confirm that 1.0.20 is in prod already.

johanandren (approved):

PR #1122, opened 2022-09-21
fix: configure spring apps to not start embedded webserver

octonato:
References https://github.com/lightbend/kalix/issues/7515

This is a temporary solution. We need an extra piece of configuration to tell Spring to not start the embedded server. 

A Kalix Spring application using the webflux starter. We need to keep it for now because of the reactive webclient it brings in. 

However, we may want to provide our own starter, see: https://docs.spring.io/spring-boot/docs/current/reference/html/features.html#features.developing-auto-configuration

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -114,8 +114,7 @@ object Dependencies {
   val springDeps = Seq(
     jacksonDataFormatProto,
     "org.springframework.boot" % "spring-boot" % SpringVersion,
-    ("org.springframework.boot" % "spring-boot-starter-webflux" % SpringVersion)
-      .exclude("org.springframework.boot", "spring-boot-starter-tomcat"))
+    "org.springframework.boot" % "spring-boot-starter-webflux" % SpringVersion)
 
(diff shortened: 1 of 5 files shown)


project/Dependencies.scala
```diff
@@ -114,8 +114,7 @@ object Dependencies {
   val springDeps = Seq(
     jacksonDataFormatProto,
     "org.springframework.boot" % "spring-boot" % SpringVersion,
-    ("org.springframework.boot" % "spring-boot-starter-webflux" % SpringVersion)
-      .exclude("org.springframework.boot", "spring-boot-starter-tomcat"))
```
octonato, line 118: This was previously adding a dependency on `spring-boot-starter-web` and then excluding tomcat. 

At a later point, we changed it to `spring-boot-starter-webflux` because we need access to the reactive WebClient. The change brought in the embedded netty that we see now starting along side akka-http. 

We can't simply exclude the dependency because the WebClient needs it, so we need to disable it by configuring: `spring.main.web-application-type=none`

PR #1128, opened 2022-09-23
feat: add spring sample to demo ES views

octonato:
https://github.com/lightbend/kalix/issues/7473

PR #1130, opened 2022-09-23
chore: move testkit to its own package

octonato:


octonato:
will break the import on the new sample in https://github.com/lightbend/kalix-jvm-sdk/pull/1128, will require a rebase here or there depending which one gets merged first.

efgpinto (approved):
LGTM

PR #1132, opened 2022-09-26
chore: label changes on samples as Documentation

efgpinto:
When releasing, some changes we did in `samples` ended up on the Java SDK section of the Release Notes. I think it would make more sense to label those as documentation automatically.

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -21,3 +21,4 @@ spring-sdk:
 
 Documentation:
   - docs/**/*
+  - samples/**/*


efgpinto:
On a second thought, I think there were probably changes to the javasdk folder for some reason and that led to that labeling. Still think that changes to samples fit well under Documentation section.

octonato (approved):
LGTM

Agree, we can treat samples as general documentation. Doesn't matter which sample it is.

PR #1135, opened 2022-09-26
feat: add unit test kit for Spring SDK

octonato:
References https://github.com/lightbend/kalix/issues/7439

Still need to add the javadocs, but opening it already for review to get feedback on the API.

TODOs
- [x] add javadoc
- [x] add tests on existing samples

### DIFF
--- a/sdk/java-sdk-testkit/src/main/java/kalix/javasdk/testkit/impl/EventSourcedEntityEffectsRunner.java
+++ b/sdk/java-sdk-testkit/src/main/java/kalix/javasdk/testkit/impl/EventSourcedEntityEffectsRunner.java
@@ -64,6 +64,7 @@ public abstract class EventSourcedEntityEffectsRunner<S> {
     EventSourcedEntity.Effect<R> effectExecuted;
     try {
       entity._internalSetCommandContext(Optional.of(commandContext));
+      entity._internalSetCurrentState(this._state);
       effectExecuted = effect.get();
       this.events.addAll(EventSourcedResultImpl.eventsOf(effectExecuted));
     } finally {
@@ -73,6 +74,7 @@ public abstract class EventSourcedEntityEffectsRunner<S> {
       entity._internalSetEventContext(Optional.of(new TestKitEventSourcedEntityEventContext()));
       for (Object event : EventSourcedResultImpl.eventsOf(effectExecuted)) {
         this._state = handleEvent(this._state, event);
+        entity._internalSetCurrentState(this._state);
       }
     } finally {
       entity._internalSetEventContext(Optional.empty());
--- a/sdk/spring-sdk-testkit/src/main/java/kalix/springsdk/testkit/ValueEntityTestKit.java
+++ b/sdk/spring-sdk-testkit/src/main/java/kalix/springsdk/testkit/ValueEntityTestKit.java
@@ -101,7 +101,7 @@ public class ValueEntityTestKit<S, E extends ValueEntity<S>> {
 
   /**
    * The call method can be used to simulate a call to the ValueEntity. The passed java lambda
-   * should return an ValueEntity.Effect. The Effect is interpreted into a ValueEntityResult that
+   * should return a ValueEntity.Effect. The Effect is interpreted into a ValueEntityResult that
    * can be used in test assertions.
    *
    * @param func A function from ValueEntity to ValueEntity.Effect.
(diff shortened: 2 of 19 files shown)


sdk/spring-sdk-testkit/src/test/java/springsdk/testmodels/actions/EchoActionTest.java
```diff
@@ -0,0 +1,32 @@
...
+import org.junit.jupiter.api.Test;
+
+public class EchoActionTest {
+
+  @Test
+  public void testEchoCall() {
+    ActionTestkit<EchoAction> actionUnitTestkit = ActionTestkit.of(ctx -> new EchoAction());
+    ActionResult<String> result = actionUnitTestkit.call(echoAction -> echoAction.echo("Hey"));
+    Assertions.assertEquals(result.getReply(), "Hey");
+  }
```
octonato, line 31: We don't need extensive tests. We are already relying on the existing test infra.

sdk/spring-sdk-testkit/src/test/java/springsdk/testmodels/eventsourced/CounterEventSourcedEntity.java
```diff
@@ -0,0 +1,44 @@
...
+    else return effects().emitEvent(new Increased(value)).thenReply(__ -> "Ok");
+  }
+
+  public Effect<String> doubleIncreaseBy(Integer value) {
+    if (value < 0) return effects().error("Can't increase with a negative value");
+    else {
+      Increased event = new Increased(value);
+      return effects().emitEvents(List.of(event, event)).thenReply(__ -> "Ok");
+    }
+  }
```
octonato, line 37: This test is important. We must ensure that the test infra is set the current state after applying each event. 

That was not the case before, because the gRPC generated code has always the state as parameter of the event handler. For the Spring SDK, users are expected to use call `currentState` inside each event handler. As such, the `EventSourcedEntityEffectsRunner` must take that info account and we want to have this test to prove that we are doing the right thing.

sdk/java-sdk-testkit/src/main/java/kalix/javasdk/testkit/impl/EventSourcedEntityEffectsRunner.java
```diff
@@ -64,6 +64,7 @@ protected <R> EventSourcedResult<R> interpretEffects(
     EventSourcedEntity.Effect<R> effectExecuted;
     try {
       entity._internalSetCommandContext(Optional.of(commandContext));
+      entity._internalSetCurrentState(this._state);
```
octonato, line 67: Set the state on the entity when running the testkit runner.

sdk/java-sdk-testkit/src/main/java/kalix/javasdk/testkit/impl/EventSourcedEntityEffectsRunner.java
```diff
@@ -73,6 +74,7 @@ protected <R> EventSourcedResult<R> interpretEffects(
       entity._internalSetEventContext(Optional.of(new TestKitEventSourcedEntityEventContext()));
       for (Object event : EventSourcedResultImpl.eventsOf(effectExecuted)) {
         this._state = handleEvent(this._state, event);
+        entity._internalSetCurrentState(this._state);
```
octonato, line 77: Than, on each applied event, we have a new state that we need to make available to the entity for the next event handler call.

sdk/spring-sdk-testkit/src/main/java/kalix/springsdk/testkit/ValueEntityTestKit.java
```diff
@@ -0,0 +1,118 @@
...
+    ValueEntityResultImpl<Reply> result = new ValueEntityResultImpl<>(effect);
+    if (result.stateWasUpdated()) {
+      this.state = (S) result.getUpdatedState();
+    }
+    return result;
+  }
+
+  /**
+   * The call method can be used to simulate a call to the ValueEntity. The passed java lambda
+   * should return an ValueEntity.Effect. The Effect is interpreted into a ValueEntityResult that
```
efgpinto, line 104:
```suggestion
   * should return a ValueEntity.Effect. The Effect is interpreted into a ValueEntityResult that
```

sdk/spring-sdk-testkit/src/test/java/springsdk/testmodels/eventsourced/CounterEventSourcedEntity.java
efgpinto: nice catch 👍

efgpinto (approved):
LGTM. Found some typos.

PR #1137, opened 2022-09-27
ci: automate PR for sdk version bump

efgpinto:
As draft for now to be able to test on CI

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,6 +2,7 @@ version: 2.1
 
 orbs:
   build-tools: circleci/build-tools@2.9.0
+  github-cli: circleci/github-cli@1.0.2
 
 commands:
   checkout-and-merge-to-main:
@@ -208,6 +209,20 @@ jobs:
           command: |
             make -C docs validate-links
       - save_deps_cache
+
+  create-pr-sdk-version-update:
+    docker:
+      - image: 'cimg/base:stable'
+    steps:
+      - checkout
+      - attach_workspace:
+          at: ~/
+      - github-cli/setup
+      - run:
+          name: Create PR to update all sdk versions (samples, sbt and maven)
+          command: |
+            ./updateAllSdkVersions.sh
+            gh pr create -B main -t "Auto PR - Update " -b "Please review, and MERGE COMMIT if okay. [$CIRCLE_SHA1]($CIRCLE_BUILD_URL)" || echo "No changes"; exit 0
 
   tests:
     docker:
@@ -1356,6 +1371,18 @@ workflows:
           filters: # version tags only
             tags:
               only: /^v.*/
+            branches:
+              ignore: /.*/
+          requires:
+            # we only publish the docs if all publishing jobs are successful, otherwise we have a 'broken' release
+            - publish
+            - publish-maven
+            - publish-tck
+
+      - create-pr-sdk-version-update:
+          filters: # version tags only
+            tags:
+              only: test_pr_creation #FIXME change this to /^v.*/ once tested
             branches:
               ignore: /.*/
           requires:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -221,8 +221,19 @@ jobs:
       - run:
           name: Create PR to update all sdk versions (samples, sbt and maven)
           command: |
-            ./updateAllSdkVersions.sh
-            gh pr create -B main -t "Auto PR - Update " -b "Please review, and MERGE COMMIT if okay. [$CIRCLE_SHA1]($CIRCLE_BUILD_URL)" || echo "No changes"; exit 0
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -209,6 +210,31 @@ jobs:
...
+      - run:
+          name: Create PR to update all sdk versions (samples, sbt and maven)
+          command: |
+            chmod +x ./updateSdkReleasedVersion.sh
+            bash ./updateSdkReleasedVersion.sh
+            
+            BRANCH=bump-sdk-versions-$SDK_VERSION
+            git checkout -b $BRANCH
+            git config user.name "Kalix Bot"
+            git config user.email "kalix.team@lightbend.com" #FIXME use a proper email
```
efgpinto, line 231: Waiting to check what email we want to use here

ennru:
This may affect https://github.com/lightbend/kalix-jvm-sdk/pull/1021

efgpinto:
> This may affect #1021

Yeah, somewhat related but shouldn't conflict. I've decided to a different script because the existing one does more than needed (installing the version locally).. but we could also opt to make this new one parameterizable and make the other one depend on this. I deemed it may not be worth the effort but open for discussion.

.circleci/config.yml
```diff
@@ -209,6 +210,31 @@ jobs:
...
+    steps:
+      - github-cli/setup
+      - checkout
+      - restore_deps_cache
+      - setup_sbt
+      - set-sdk-version
+      - run:
+          name: Create PR to update all sdk versions (samples, sbt and maven)
+          command: |
+            chmod +x ./updateSdkReleasedVersion.sh
```
efgpinto, line 225: not sure why it required this, it should be runable without it I believe.. but CI complained about it when I tested.

updateSdkReleasedVersion.sh
```diff
@@ -0,0 +1,20 @@
...
+fi
+
+echo ">>> Updating pom versions to $SDK_VERSION"
+find ./samples -type f -name "*pom.xml" -print0 | xargs -0 sed -i -e "s/<kalix-sdk.version>\(.*\)<\/kalix-sdk.version>/<kalix-sdk.version>$SDK_VERSION<\/kalix-sdk.version>/"
+
+echo ">>> Updating sbt plugins to $SDK_VERSION"
+find ./samples -type f -name "*plugins.sbt" -print0 | xargs -0 sed -i -e "s/System.getProperty(\"kalix-sdk.version\", \".*\"))/System.getProperty(\"kalix-sdk.version\", \"$SDK_VERSION\"))/"
+
+echo ">>> Updating maven plugin to $SDK_VERSION"
+cd maven-java && mvn versions:set -DnewVersion="$SDK_VERSION"
```
efgpinto, line 20: I guess at some point we could probably consider having a `scripts` folder instead..

franciscolopezsancho: I think this makes sense. In `kalix-proxy` we have one `bin` folder for that.

franciscolopezsancho:
Combining both scripts into one I also think is a good idea. Maybe it is better to do it now than put it off and perhaps forget it.

franciscolopezsancho:
If you decide to merge both I'll wait and test it afterwards.

efgpinto:
Updated now, added some usage examples to PR description for clarity.

Also, made the `updatePomVersions.sh` depend on this one... TBH, I think that one should probably go but I was losing a bit too much time to think about the params we would need on this new one to support installing the local version vs just updating it.. So I'm not sure we need to do that here since we also have https://github.com/lightbend/kalix-jvm-sdk/pull/1021 open.

Unless someone has already a clear idea of what the this command could look like to support both features, I'm tempted to go with this version so we have the auto-PR thing in place and then try to re-org these 3 scripts (`updateSdkVersions`, `updatePomVersions` and `publishLocalM2`).

Thoughts?

efgpinto:
@johanandren IIUC you're making an sdk release soon. If you want to have a look at this one prior to that, hopefully it can make the release a bit easier.

johanandren:
@efgpinto failed because of bot not having permissions:

https://app.circleci.com/pipelines/github/lightbend/kalix-jvm-sdk/4386/workflows/a6b43db1-3885-4519-add4-0dc9fec1c74d/jobs/67064

efgpinto:
> @efgpinto failed because of bot not having permissions:
> 
> https://app.circleci.com/pipelines/github/lightbend/kalix-jvm-sdk/4386/workflows/a6b43db1-3885-4519-add4-0dc9fec1c74d/jobs/67064

Thanks for reviewing and testing Johan. Seems like the bot's permissions got downgraded, re-run solved that but the created [PR](https://github.com/lightbend/kalix-jvm-sdk/pull/1177) was not perfect. Will have a look.

PR #1139, opened 2022-09-29
ci: run sbt prior to Fossa anaylze

ennru:
Refs https://github.com/lightbend/kalix/issues/7448


mblund (approved):
It could be that we don't need to do the makePom step ourself, just ensure that sbt is installed. But it cannot hurt to do it.

For reference: https://github.com/fossas/fossa-cli/blob/master/docs/references/strategies/languages/scala/sbt.md

PR #1151, opened 2022-10-03
chore: disable descriptor source loading for SpringSDK

octonato:
Found this while working on the ACL and thought it was better to send as an independent PR first


efgpinto (approved):
LGTM

PR #1152, opened 2022-10-04
fix: include kalix_policy.proto even if source info is disabled

octonato:


sdk/java-sdk/src/main/scala/kalix/javasdk/impl/DiscoveryImpl.scala
```diff
@@ -250,10 +250,17 @@ class DiscoveryImpl(system: ActorSystem, services: Map[String, Service], sdkName
 object DiscoveryImpl {
 
   private[impl] def fileDescriptorSet(services: Iterable[Service], userDescPath: String, log: Logger) = {
-    val descriptorsWithSource = loadDescriptorsWithSource(userDescPath, log)
+
+    val descriptors = loadFileDescriptors(userDescPath, log)
```
octonato, line 254: all descriptors we can find in user_function.desc. Doesn't matter if sourceCodeInfo is enabled or not.

sdk/java-sdk/src/main/scala/kalix/javasdk/impl/DiscoveryImpl.scala
```diff
@@ -307,9 +312,7 @@ object DiscoveryImpl {
             .parseFrom(stream)
             .getFileList
             .asScala
-            .collect {
-              case file if file.hasSourceCodeInfo => file.getName -> file
-            }
+            .collect { case file => file.getName -> file }
```
octonato, line 315: hmm, should be a map. I will change it

efgpinto (approved):
LGTM

PR #1155, opened 2022-10-05
feat: ACL support for Spring SDK

octonato:
References https://github.com/lightbend/kalix/issues/7368

### DIFF
--- a/sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/AclDescriptorFactory.scala
+++ b/sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/AclDescriptorFactory.scala
@@ -1,0 +1,122 @@
+/*
+ * Copyright 2021 Lightbend Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package kalix.springsdk.impl
+
+import com.google.protobuf.DescriptorProtos
+import com.google.protobuf.Descriptors
+import kalix.PrincipalMatcher
+import kalix.{ Acl => ProtoAcl }
+import kalix.springsdk.annotations.Acl
+import kalix.springsdk.impl.ProtoDescriptorGenerator.dependencies
+import org.slf4j.LoggerFactory
+import kalix.{ Annotations => KalixAnnotations }
+
+object AclDescriptorFactory {
+
+  private val logger = LoggerFactory.getLogger(classOf[AclDescriptorFactory.type])
+
+  val invalidAnnotationUsage =
+    "Invalid annotation usage. Matcher has both 'principal' and 'service' defined. " +
+    "Only one is allowed."
+
+  private def validateMatcher(matcher: Acl.Matcher) = {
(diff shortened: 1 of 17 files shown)


sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/KalixServer.scala
```diff
@@ -178,22 +179,38 @@ class KalixServer(applicationContext: ApplicationContext, config: Config) {
...
 
-  // TODO: users may define their Kalix components in other packages as well and then use @ComponentScan
-  // to let Spring find them. We should also look for @ComponentScan in the Main class and collect any
-  // scan package declared there
-  val packagesToScan = springBootMain.map(_.getClass.getPackageName).toSet
+    // TODO: we need to verify how reliable is this method of trying to find the main class.
+    //  Would work for any jdk? Should we care?
+    val bootClassOpt =
+      Try(Class.forName(System.getProperty("sun.java.command"))).toOption
+        .filter { cls => cls.getAnnotation(classOf[SpringBootApplication]) != null }
```
octonato, line 191: Here we used to pick the Main class from the application context, but the version that Spring adds to the context is one enhanced by CGLIB and doesn't not have the `@Acl` annotation. 

This workaround to find out the main class may not be JDK dependent so we will need to very this and eventually find a more robust solution.

octonato (commented):
This is not working yet. I'm getting an error on the proxy. Probably the descriptor that we are sending is not correct. 

I will park it for a while now. May get back to this Friday or Monday.

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/KalixServer.scala
octonato: I did some local testing with a few different JDKs and `sun.java.command` seems to be available each time. Although I didn't find a official documentation affirming that (didn't search much neither).

Also, it only works if the pom explicitly declares the main class, which is the case for us. 

If user calls mvn and pass the main class we don't get the same result. 
For example, calling it with:

```java
 mvn compile exec:java -Dexec.mainClass="com.example.demo.DemoApplication"
```

we have as output 

```java
org.codehaus.plexus.classworlds.launcher.Launcher compile exec:java -Dexec.mainClass=com.example.demo.DemoApplication
```

Given that it works with the JDK we use in the container and given that our pom is well configured, we can assume it will work, but if we can find a better solution for this it will be better.

octonato (commented):
This is ready for review. 
In this PR, I'm only adding the default ACL. In follow up PR I will add support for ACL annotations on classes and methods.

efgpinto (approved):
Left some chore comments. Looks good to me but maybe good to have someone else ✅  as well.

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/AclDescriptorFactory.scala
```diff
@@ -0,0 +1,125 @@
...
+      throw new IllegalArgumentException(invalidAnnotationUsage)
+  }
+
+  private def deriveProtoAnnotation(aclJavaAnnotation: Acl): ProtoAcl = {
+
+    aclJavaAnnotation.allow().foreach(matcher => validateMatcher(matcher))
+    aclJavaAnnotation.deny().foreach(matcher => validateMatcher(matcher))
+
+    aclJavaAnnotation.allow().foreach(matcher => validateMatcher(matcher))
+    aclJavaAnnotation.deny().foreach(matcher => validateMatcher(matcher))
```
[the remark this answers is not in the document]
octonato: I was just trying to make sure that it will be truly validated. 😆 
Thanks for spotting it.

octonato:
I had to change how I look up for the original main class. My previous strategy would never work with integration tests because the main class is either `sbt.ForkMain` or some other class from maven's surefire plugin. 

Instead, I'm using Spring to scan the classpath and search for the original main class that the users defines for their applications.

efgpinto (approved):
LGTM again

PR #1156, opened 2022-10-06
fix: java 17 for spring samples and maven archetypes

aludwiko:
References #xxxx


lightbend-cla-validator:
Hi @aludwiko,

Thank you for your contribution! We really value the time you've taken to put this together.

We see that you have signed the Lightbend Contributors License Agreement before, however, the CLA has changed since you last signed it.
Please review the new CLA and sign it before we proceed with reviewing this pull request:

[https://www.lightbend.com/contribute/cla](https://www.lightbend.com/contribute/cla)

samples/spring-eventsourced-customer-registry/src/main/java/customer/api/CustomerEvent.java
```diff
@@ -4,12 +4,23 @@
...
+public sealed interface CustomerEvent {
+
+  record CustomerCreated(String email, String name, Address address) implements CustomerEvent {
+  }
+
+  record NameChanged(String newName) implements CustomerEvent {
+  }
+
+  record AddressChanged(Address address) implements CustomerEvent {
+  }
```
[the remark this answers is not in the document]
octonato: One thing we can try out with sealed class is to check if we can remove the `@JsonSubTypes`. I will expect that Jackson support for records will be smart enough to find out all the subtypes. 

We still need to give a type hint name for each subtypes, but that can be done with `@JsonTypeName`

samples/spring-eventsourced-customer-registry/pom.xml
```diff
@@ -123,7 +123,7 @@
               <name>${dockerImage}:%l</name>
               <build>
                 <!-- Base Docker image which contains jre-->
-                <from>docker.io/library/adoptopenjdk:${jdk.target}-jre-hotspot</from>
+                <from>docker.io/library/eclipse-temurin:${jdk.target}-alpine</from>
```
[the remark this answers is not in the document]
aludwiko: yeah, it turns out that adoptopenjdk is deprecated: https://hub.docker.com/_/adoptopenjdk 
```
This image is officially deprecated in favor of [the eclipse-temurin image](https://hub.docker.com/_/eclipse-temurin/), and will receive no further updates after 2021-08-01 (Aug 01, 2021). Please adjust your usage accordingly.
```

samples/spring-eventsourced-customer-registry/src/main/java/customer/api/CustomerEvent.java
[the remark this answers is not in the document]
aludwiko: Unfortunately, it's possible at this point:
https://github.com/FasterXML/jackson-databind/issues/3596
https://github.com/FasterXML/jackson-databind/pull/3549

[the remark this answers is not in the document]
octonato: oh, a pity! We will do it in the future.

samples/spring-eventsourced-customer-registry/pom.xml
```diff
@@ -11,12 +12,12 @@
   <properties>
 
     <!-- TODO Update to your own Docker repository or Docker Hub scope -->
-    <dockerImage>my-docker-repo/${project.artifactId}</dockerImage>
+    <dockerImage>aludwiko/${project.artifactId}</dockerImage>
```
octonato, line 15: needs to be reverted

samples/spring-fibonacci-action/pom.xml
```diff
@@ -11,12 +11,12 @@
   <properties>
 
     <!-- TODO Update to your own Docker repository or Docker Hub scope -->
-    <dockerImage>my-docker-repo/${project.artifactId}</dockerImage>
+    <dockerImage>aludwiko/${project.artifactId}</dockerImage>
```
octonato, line 14:
```suggestion
    <dockerImage>my-docker-repo/${project.artifactId}</dockerImage>
```

samples/spring-eventsourced-customer-registry/pom.xml
octonato, line 15:
```suggestion
    <dockerImage>my-docker-repo/${project.artifactId}</dockerImage>
```

samples/spring-fibonacci-action/pom.xml
aludwiko: fixes

johanandren (approved):
LGTM sans the hardcoded versions and the docker-hub account placeholders.

Note that we will need to bump the Java QuickStart doc pages as well once this is released as they now say JDK 11 is or later is the requirement but after this the quickstarts require 17:
https://docs.kalix.io/java/quickstart/cr-value-entity-java.html#_before_you_begin

I wonder if we should bump the Scala samples as well (separate PR), not that it changes anything for the Scala code but to get the same base JVM image for all JVM SDK services (and potential runtime improvements GC etc).

samples/spring-eventsourced-counter/src/main/java/com/example/CounterEvent.java
```diff
@@ -0,0 +1,10 @@
+package com.example;
+
+public sealed interface CounterEvent {
```
aludwiko, line 3: @octonato should we put `@JsonSubTypes` annotations here?

samples/spring-eventsourced-customer-registry/src/main/java/customer/view/CustomerView.java
```diff
@@ -35,25 +25,20 @@ public static Optional<CustomerView> onEvent(CustomerView state, CustomerEvent e
 
     if (state == null) {
```
aludwiko, line 26: @octonato how about changing this to sth like:
```
public static Optional<CustomerView> onEvent(CustomerView state, CustomerEvent event) {

    // the only event we can receive when state is null is the CustomerCreated
    if (state == null && event instanceof CustomerCreated created) {
      return Optional.of(new CustomerView(created.email(), created.name(), created.address()));
    } else if (event instanceof NameChanged nameChanged) {
      return Optional.of(state.withName(nameChanged.newName()));
    } else if (event instanceof AddressChanged addressChanged) {
      return Optional.of(state.withAddress(addressChanged.address()));
    } else {
      throw new IllegalStateException("Not handled event: " + event);
    }
  }
```

aludwiko:
> sans the hardcoded versions and the docker-hub account placeholders.
yeah, my bad, fixed.

samples/spring-eventsourced-customer-registry/src/main/java/customer/view/CustomerView.java
octonato: sth? You mean switch statement? 

Francisco's PR on event handler should have solved this. We just forgot to update the samples. It should be possible now to have overloaded `onEvent` methods receiving each a concrete type. Or simply different method names. Doesn't matter much.

samples/spring-eventsourced-counter/src/main/java/com/example/CounterEvent.java
octonato: Yes, I think this is needed. Didn't we have this already? That surprises me. This can not work without the annotation.

octonato:
> LGTM sans the hardcoded versions and the docker-hub account placeholders.
> 
> Note that we will need to bump the Java QuickStart doc pages as well once this is released as they now say JDK 11 is or later is the requirement but after this the quickstarts require 17: https://docs.kalix.io/java/quickstart/cr-value-entity-java.html#_before_you_begin
> 
> I wonder if we should bump the Scala samples as well (separate PR), not that it changes anything for the Scala code but to get the same base JVM image for all JVM SDK services (and potential runtime improvements GC etc).

The quickstart for the customer register is still using jdk 11. This PR is only changing the Spring samples. 

However, we should change the quickstart with template (archetype) page, 
https://docs.kalix.io/java/quickstart-template.html. The new archetype is using 17. Maybe we should revert it for now. And only concentrate on the Spring one. 

If we move the java (gRPC) archetype, than we can also move all the samples all together.

octonato:
To make my point, using Java 17 is interesting for the Spring SDK because users can use records and sealed interfaces. 

That's less relevant for the gRPC one because we don't control the Java classes.

samples/spring-eventsourced-customer-registry/src/main/java/customer/view/CustomerView.java
aludwiko: ok

octonato:
@aludwiko, can we revert the grpc archetypes and do it in another PR? Then we can update all the Java gRPC samples. And than another PR for the Scala ones.

octonato:
Ok, to update the docs, we need to change a variable in the Antora Makefile. That will change the JDK version for all pages that mention it. So, either we change it for all samples or we only do it for the Spring ones now. Since we don't have Spring docs yet, I think it's fine

aludwiko:
@octonato grpc archetypes reverted

octonato (approved):
Great! Waiting for green build to merge

beritou:
> I wonder why they decided to make the access to the fields look like a method call, eg: `name()`. Why a direct access to a public final method isn't good?

I also don't like the look of how fields are accessed.

Maybe this was done to emphasize the fact that these are ["getter methods"](https://www.baeldung.com/java-record-keyword#2-getters)? 🤷

PR #1158, opened 2022-10-07
Initial two-service-sample for service-to-service-eventing

johanandren:
Very early push, depends on a snapshot build of the proxy and then a snapshot of the SDK including the latest proxy from the service to service branch as well as running the proxy locally directly rather than in docker

Adds a sample of service-to-service:
* event publishing from scala-eventsourced-customer-registry 
* subscribing in a new scala-eventsourced-customer-registry-subscriber


samples/scala-eventsourced-customer-registry/src/main/proto/customer/api/direct_customer_events.proto
```diff
@@ -0,0 +1,50 @@
...
+        // skip/filter events that there is no transform method for (AddressChanged)
+        ignore_unknown: true
+    };
+    option (kalix.service).eventing.out.direct.event_stream_id = "customer_events";
+    // limit access to only be allowed from specific services
+    // option (kalix.service).acl.allow = { service: "customer-registry-views" };
+
+    // transform methods for turning internal event types to public API
+    rpc ProcessCustomerCreated(domain.CustomerCreated) returns (Created) { }
+    rpc ProcessCustomerNameChanged(domain.CustomerNameChanged) returns (NameChanged) {}
```
patriknw, line 48: rename to Transform instead of Process?

patriknw (commented):
Nice

samples/scala-eventsourced-customer-registry-subscriber/src/main/proto/customer/api/customer_api.proto
```diff
@@ -0,0 +1,23 @@
+// Copyright 2021 Lightbend Inc.
```
He-Pin, line 1: not 2022?

johanandren:
Switched to view because that makes sense and is nicely observable to get updates from the outside.

docs/src/modules/java/pages/service-to-service.adoc
```diff
@@ -0,0 +1,111 @@
+include::ROOT:partial$include.adoc[]
+
+= Service to Service Eventing
+:page-supergroup-java-scala: Language
+
+Kalix provides broker less at-least-once event delivery across Kalix services through the Service to Service eventing.
```
patriknw, line 6: is it "brokerless" or "broker less"?

docs/src/modules/java/pages/service-to-service.adoc
```diff
@@ -0,0 +1,111 @@
...
++
+[source,scala,indent=0]
+.src/main/scala/customer/api/CustomerEventsServiceAction.scala
+----
+include::example$scala-eventsourced-customer-registry/src/main/scala/customer/api/CustomerEventsServiceAction.scala[tag=transform]
+----
+
+== Event Consumer
+
+The consumer can be an action or a view, annotated with `(kalix.service).eventing.in.direct` with a `service` identifying the publishing service, and the `event_stream_id` to subscribe to.
```
patriknw, line 65: do we use first upper case in other places? Action, View, Event Sourced Entity

samples/java-eventsourced-customer-registry-subscriber/src/main/proto/customer/view/customer_view.proto
```diff
@@ -0,0 +1,69 @@
...
+  option (kalix.codegen) = {
+    view: {}
+  };
+
+  // consume events published by scala-eventsourced-customer-registry/CustomerEventsServiceAction
+  option (kalix.service).eventing.in.direct = { // <1>
+    // end::view[]
+    // Normally the Kalix service deployment name
+    // (but localhost + hardcoded port here to be able to run both locally)
+    service: "localhost:9000"
```
patriknw, line 39: will this become a problem if users download the source and try to deploy it?

patriknw: maybe not a problem since it's not a quickstart sample (yet)

patriknw (approved):
LGTM, maybe some nitpicks...

docs/src/modules/java/pages/service-to-service.adoc
He-Pin: brokerless

samples/java-eventsourced-customer-registry-subscriber/src/main/proto/customer/view/customer_view.proto
johanandren: Yes, the would have to change it to deploy (but changing that's easier than changing the other things to get it running locally, with the proxy port etc).

docs/src/modules/java/pages/service-to-service.adoc
johanandren: CI spellcheck disagrees :/ 

But I think you are right, so I'll see about some allow-list for that spellcheck/lang tool.

PR #1159, opened 2022-10-07
7407 adding subscrition to class and ignore option

franciscolopezsancho:
References https://github.com/lightbend/kalix/issues/7407


sdk/spring-sdk/src/main/java/kalix/springsdk/annotations/Subscribe.java
```diff
@@ -80,5 +89,15 @@
...
+
+    /**
+     * If the message is not of interest for consumption the method for that input message
+     * type can be marked with `ignore: true`.
+     * Enabling is more efficient than implementing the method.
+     * IMPORTANT: methods can't use this property. They are always ignore false
+     *
+     * FIXME agree with the strategy above. What even means that
+     */
+    boolean ignore() default false;
```
efgpinto, line 101: I think it would probably make sense to be a bit more specific? Maybe `ignoreUnknown` similar to what we have in the proxy protocol?

```suggestion
    boolean ignoreUnknown() default false;
```

franciscolopezsancho:
superseded by https://github.com/lightbend/kalix-jvm-sdk/pull/1167

PR #1161, opened 2022-10-07
chore: use overloaded event handlers in views

octonato:
First commit is the main change 73ade90db9b89fe72d79600dadd2545d23d440fd

Second commit is just some clean-up of copyright headers 4924c6816de2cdd5e233f277d49c24c4f7b218e0


samples/spring-customer-registry-views-quickstart/src/main/java/customer/api/Address.java
```diff
@@ -1,7 +1,4 @@
 package customer.api;
-/*
- * Copyright (C) 2009-2022 Lightbend Inc. <https://www.lightbend.com>
- */
```
octonato, line 4: we usually don't add copyright headers on samples. And certainly not from 2009 for Kalix.

efgpinto (approved):
LTGM

PR #1165, opened 2022-10-12
chore: remove option to pass view state update handler method

octonato:
References https://github.com/lightbend/kalix/issues/7616


sdk/java-sdk/src/main/java/kalix/javasdk/view/View.java
```diff
@@ -41,6 +47,33 @@ public void _internalSetUpdateContext(Optional<UpdateContext> context) {
...
+   * @throws IllegalStateException if accessed outside a handler method
+   */
+  @ApiMayChange
+  protected final S viewState() {
+    // user may call this method inside a command handler and get a null because it's legal
+    // to have emptyState set to null.
+    if (handlingUpdates) return viewState.orElse(null);
+    else
+      throw new IllegalStateException("Current state is only available when handling updates.");
+  }
```
octonato, line 75: If someone has ideas for a better name for this method...

I thought about `currentViewRow` but found it too technical. 

I was also thinking that we should rename `currentState` to `entityState` in the Entities. WDYT?

sdk/spring-sdk/README.md
```diff
@@ -25,6 +25,7 @@ Inside the generated project you can find a `README.md` file with further instru
 ## Features
 
 As the Spring SDK is more recent than their gRPC-first counterparts, not all features are supported at this time. However, there's already a lot to play with:
+
```
octonato, line 28: Some formatting here, as VSCode Markdown linting was complaining.

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/reflection/KalixMethod.scala
```diff
@@ -89,35 +90,32 @@ case class VirtualServiceMethod(component: Class[_], methodName: String, inputTy
...
 
   override def requestMethod: RequestMethod = RequestMethod.POST
-  override def javaMethodOpt: Option[Method] = Some(javaMethod)
+  override def javaMethodOpt: Option[Method] = None
 
-  val pathTemplate = buildPathTemplate(javaMethod.getDeclaringClass.getName, methodName)
+  val pathTemplate = buildPathTemplate(componentName, methodName)
+
+  val streamIn: Boolean = false
+  val streamOut: Boolean = false
```
octonato, line 107: I think there were some suboptimal stuff here. 

A `CombinedSubscriptionServiceMethod` should...

* always have the `inputType` set to proto Any, that's what it will be at the end
* it can't have a javaMethod because is a synthetic one, we were using the one from the `representative`, but that's the wrong one.
* it can't be streamIn nor streamOut since it's a subscription one

Anyway, by removing the option to pass the view state in the update method, there are a bunch of code here that becomes obsolete and gives us the opportunity to clean it up.

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/reflection/KalixMethod.scala
```diff
@@ -89,35 +90,32 @@ case class VirtualServiceMethod(component: Class[_], methodName: String, inputTy
...
  * Build from methods annotated with @Subscription. Those methods are not annotated with Spring REST annotations, but
  * they become a REST method at the end.
  */
-case class SubscriptionServiceMethod(javaMethod: Method, inputTypeParamIndex: Int = 0)
-    extends AnyJsonRequestServiceMethod {
+case class SubscriptionServiceMethod(javaMethod: Method) extends AnyJsonRequestServiceMethod {
 
   val methodName = javaMethod.getName
-  val inputType: Class[_] = javaMethod.getParameterTypes()(inputTypeParamIndex)
+  val inputType: Class[_] = javaMethod.getParameterTypes()(0)
```
octonato, line 118: actually, I should have simply call `.head`. It's should always be one input.

efgpinto (approved):
LGTM

PR #1166, opened 2022-10-12
chore: adds default ACL to all samples

octonato:


samples/java-eventsourced-counter/src/main/proto/com/example/kalix_policy.proto
```diff
@@ -0,0 +1,14 @@
...
+
+import "kalix/annotations.proto";
+
+// Allow all other Kalix services deployed in the same project to access the components of this
+// Kalix service, but disallow access from the internet. This can be overridden explicitly
+// per component or method using annotations.
+// Documentation at https://docs.kalix.io/services/using-acls.html
+option (kalix.file).acl = {
+  allow: { service: "*" }
+};
```
johanandren, line 14: We'd likely also want to open up the services of each sample to internet, if we don't rather want to allow that as default. If not it means that building a sample and deploying it and then exposing means it won't be accessible, which is not super nice for someone trying Kalix out the first time.

octonato (commented):
Almost all kalix default ACL files are not configured to allow traffic from the internet. 

The only ones that I didn't touch was the file from the doc snippets.

samples/java-shopping-cart-quickstart/src/main/proto/shopping/cart/kalix_policy.proto
```diff
@@ -1,14 +1,14 @@
 // This is the default Access Control List (ACL) for all components of this Kalix Service
 syntax = "proto3";
 
-package shopping.cart;
+package customer;
```
efgpinto, line 4: Previous version maybe?
All packages seem to be set to `customer` now (thanks copy-paste I suppose)... Not critical but if we want to keep them aligned you might want to revisit them.

johanandren:
For the quickstarts we already had a restrictive default and then service level ACLs opening up the public API, so I think we should keep those as is, and only add the permissive default policy to other sample projects.

samples/java-shopping-cart-quickstart/src/main/proto/shopping/cart/kalix_policy.proto
octonato: yep, I was copying and pasting stuff without even looking at the package name. 

I will revert and start over again.

efgpinto (approved):
LGTM

PR #1167, opened 2022-10-12
feat: Event subscription with ignore option #7407

franciscolopezsancho:
References https://github.com/lightbend/kalix/issues/7407

It partially the eventing in from this issue https://github.com/lightbend/kalix-jvm-sdk/issues/1169

### DIFF
--- a/sdk/spring-sdk/src/it/java/com/example/wiring/SpringSdkWiringIntegrationTest.java
+++ b/sdk/spring-sdk/src/it/java/com/example/wiring/SpringSdkWiringIntegrationTest.java
@@ -16,7 +16,6 @@
 
 package com.example.wiring;
 
-import akka.http.javadsl.model.StatusCode;
 import com.example.Main;
 import com.example.wiring.actions.echo.Message;
 import com.example.wiring.eventsourcedentities.counter.Counter;
@@ -107,6 +106,89 @@ public class SpringSdkWiringIntegrationTest {
                     .bodyToMono(Integer.class)
                     .block(timeout),
             new IsEqual(42 + 1));
+  }
+
+  @Test
+  public void verifyActionIsNotSubscribedToMultiplyAndRouterIgnores(){
+
+    webClient
+            .post()
+            .uri("counter/counterId2/increase/1")
+            .retrieve()
+            .bodyToMono(Integer.class)
+            .block(timeout);
+
+    webClient
+            .post()
+            .uri("counter/counterId2/multiply/2")
+            .retrieve()
+            .bodyToMono(Integer.class)
+            .block(timeout);
+
+    webClient
+            .post()
+            .uri("counter/counterId2/increase/1234")
+            .retrieve()
+            .bodyToMono(Integer.class)
+            .block(timeout);
+
+    Integer counterGet =
+            webClient.get().uri("/counter/counterId2").retrieve().bodyToMono(Integer.class).block(timeout);
+
+    //Before has time to process any action the counter has only the values from the POSTs
+    Assertions.assertEquals(1 * 2 + 1234, counterGet);
+
+    //Once the action IncreaseActionWithIgnore processes event 1234 it adds 1 more to the counter
+    await()
+            .atMost(10, TimeUnit.SECONDS)
+            .until(
+                    () ->
+                            webClient
+                                    .get()
+                                    .uri("/counter/counterId2")
+                                    .retrieve()
+                                    .bodyToMono(Integer.class)
+                                    .block(timeout),
+                    new IsEqual<Integer>(1 * 2 + 1234 + 1 ));
+  }
+
+  @Test
+  public void verifyViewIsNotSubscribedToMultiplyAndRouterIgnores() {
+
+    webClient
+            .post()
+            .uri("counter/counterId4/increase/1")
+            .retrieve()
+            .bodyToMono(Integer.class)
+            .block(timeout);
+    webClient
+            .post()
+            .uri("counter/counterId4/multiply/2")
+            .retrieve()
+            .bodyToMono(Integer.class)
+            .block(timeout);
+
+    Integer counterGet = webClient
+            .post()
+            .uri("counter/counterId4/increase/1")
+            .retrieve()
+            .bodyToMono(Integer.class)
+            .block(timeout);
+
+    Assertions.assertEquals(1 * 2 + 1, counterGet);
+
+    await()
+            .ignoreExceptions()
+            .atMost(10, TimeUnit.SECONDS)
+            .until(
+                    () -> webClient
+                            .get()
+                            .uri("/counters-ignore/by-value-with-ignore/2")
+                            .retrieve()
+                            .bodyToMono(Counter.class)
+                            .map(counter -> counter.value)
+                            .block(timeout),
+                    new IsEqual(1 + 1));
   }
 
   @Test
--- a/sdk/spring-sdk/src/main/java/kalix/springsdk/action/ReflectiveActionProvider.java
+++ b/sdk/spring-sdk/src/main/java/kalix/springsdk/action/ReflectiveActionProvider.java
@@ -24,6 +24,7 @@ import kalix.javasdk.action.ActionProvider;
 import kalix.javasdk.impl.MessageCodec;
 import kalix.javasdk.impl.action.ActionRouter;
 import kalix.springsdk.impl.ComponentDescriptor;
+import kalix.springsdk.impl.ComponentDescriptorFactory;
 import kalix.springsdk.impl.SpringSdkMessageCodec;
 import kalix.springsdk.impl.action.ReflectiveActionRouter;
 
@@ -76,7 +77,7 @@ public class ReflectiveActionProvider<A extends Action> implements ActionProvide
   @Override
   public ActionRouter<A> newRouter(ActionCreationContext context) {
     A action = factory.apply(context);
-    return new ReflectiveActionRouter<>(action, componentDescriptor.commandHandlers());
+    return new ReflectiveActionRouter<>(action, componentDescriptor.commandHandlers(), ComponentDescriptorFactory.findIgnore(action.getClass()));
   }
 
   @Override
--- a/sdk/spring-sdk/src/main/java/kalix/springsdk/annotations/Subscribe.java
+++ b/sdk/spring-sdk/src/main/java/kalix/springsdk/annotations/Subscribe.java
@@ -59,7 +59,7 @@ public @interface Subscribe {
    *       second one the event type.
    * </ul>
    */
-  @Target(ElementType.METHOD)
+  @Target({ElementType.METHOD, ElementType.TYPE})
   @Retention(RetentionPolicy.RUNTIME)
   @Documented
   @interface EventSourcedEntity {
@@ -68,10 +68,17 @@ public @interface Subscribe {
      * kalix.javasdk.eventsourcedentity.EventSourcedEntity EventSourcedEntity}.
      */
     Class<? extends kalix.javasdk.eventsourcedentity.EventSourcedEntity<?>> value();
+
+    /** This option is only available for classes. Using it in a method has no effect.
+     * If there is no method in the class whose input type matches the event type:
+     *   if ignoreUnknown is true the event is discarded.
+     *   if false, an Exception is raised.
+     * */
+    boolean ignoreUnkown() default false;
   }
 
   /** Annotation for subscribing to messages from a topic (i.e PubSub or Kafka topic). */
-  @Target(ElementType.METHOD)
+  @Target({ElementType.METHOD, ElementType.TYPE})
   @Retention(RetentionPolicy.RUNTIME)
   @Documented
   @interface Topic {
@@ -80,5 +87,13 @@ public @interface Subscribe {
 
     /** Assign the consumer group name to be used on the broker. */
     String consumerGroup() default "";
+
+    /**
+     * This option is only available for classes. Using it in a method has no effect.
+     * If there is no method in the class whose input type matches the event type:
+     *   if ignoreUnknown is true the event is discarded.
+     *   if false, an Exception is raised.
+     **/
+    boolean ignoreUnknown() default false;
   }
 }
--- a/sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/ActionDescriptorFactory.scala
+++ b/sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/ActionDescriptorFactory.scala
@@ -82,9 +82,7 @@ private[impl] object ActionDescriptorFactory extends ComponentDescriptorFactory 
         component.getMethods.sorted // make sure we get the methods in deterministic order
           .filter(hasActionOutput)
           .collect {
-            case method
-                if !subscriptionEventSourcedEntityMethods.exists(s =>
-                  s.serviceMethod.methodName == method.getName) => // individual annotated overrides class level annotation
+            case method => 
               val subscriptionOptions = eventingInForEventSourcedEntity(component)
               val kalixOptions =
                 kalix.MethodOptions.newBuilder().setEventing(subscriptionOptions).build()
--- a/sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/ActionDescriptorFactory.scala
+++ b/sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/ActionDescriptorFactory.scala
@@ -22,6 +22,7 @@ import kalix.springsdk.impl.ComponentDescriptorFactory.eventingInForEventSourced
 import kalix.springsdk.impl.ComponentDescriptorFactory.eventingInForTopic
 import kalix.springsdk.impl.ComponentDescriptorFactory.eventingInForValueEntity
 import kalix.springsdk.impl.ComponentDescriptorFactory.eventingOutForTopic
+import kalix.springsdk.impl.ComponentDescriptorFactory.hasActionOutput
 import kalix.springsdk.impl.ComponentDescriptorFactory.hasEventSourcedEntitySubscription
 import kalix.springsdk.impl.ComponentDescriptorFactory.hasTopicPublication
 import kalix.springsdk.impl.ComponentDescriptorFactory.hasTopicSubscription
@@ -41,7 +42,7 @@ private[impl] object ActionDescriptorFactory extends ComponentDescriptorFactory 
       messageCodec: SpringSdkMessageCodec,
       nameGenerator: NameGenerator): ComponentDescriptor = {
     //we should merge from here
-    val springAnnotatedMethods =
+    val syntheticMethods =
       RestServiceIntrospector.inspectService(component).methods.map { serviceMethod =>
         validateRestMethod(serviceMethod.javaMethod)
         KalixMethod(serviceMethod).withKalixOptions(buildJWTOptions(serviceMethod.javaMethod))
@@ -65,6 +66,9 @@ private[impl] object ActionDescriptorFactory extends ComponentDescriptorFactory 
       .filter(hasEventSourcedEntitySubscription)
       .sorted // make sure we get the methods in deterministic order
       .map { method =>
+        if (hasEventSourcedEntitySubscription(component))
+          throw InvalidComponentException(
+            s"You cannot use Subscribe annotation in both the methods and the class. You can do either one or the other.")
         val subscriptionOptions = eventingInForEventSourcedEntity(method)
         val kalixOptions =
           kalix.MethodOptions.newBuilder().setEventing(subscriptionOptions).build()
@@ -73,10 +77,30 @@ private[impl] object ActionDescriptorFactory extends ComponentDescriptorFactory 
           .withKalixOptions(kalixOptions)
       }
 
+    val subscriptionEventSourcedEntityClass =
+      if (hasEventSourcedEntitySubscription(component)) {
+        component.getMethods.sorted // make sure we get the methods in deterministic order
+          .filter(hasActionOutput)
+          .collect {
+            case method
+                if !subscriptionEventSourcedEntityMethods.exists(s =>
+                  s.serviceMethod.methodName == method.getName) => // individual annotated overrides class level annotation
+              val subscriptionOptions = eventingInForEventSourcedEntity(component)
+              val kalixOptions =
+                kalix.MethodOptions.newBuilder().setEventing(subscriptionOptions).build()
+
+              KalixMethod(SubscriptionServiceMethod(method))
+                .withKalixOptions(kalixOptions)
+          }
+      } else Array.empty[KalixMethod]
+
     val subscriptionTopicMethods = component.getMethods
       .filter(hasTopicSubscription)
       .sorted // make sure we get the methods in deterministic order
       .map { method =>
+        if (hasTopicSubscription(component))
+          throw InvalidComponentException(
+            s"You cannot use Subscribe annotation in both the methods and the class. You can do either one or the other.")
         val subscriptionOptions = eventingInForTopic(method)
         val kalixOptions =
           kalix.MethodOptions.newBuilder().setEventing(subscriptionOptions).build()
@@ -85,7 +109,21 @@ private[impl] object ActionDescriptorFactory extends ComponentDescriptorFactory 
           .withKalixOptions(kalixOptions)
       }
 
-    def combineByTopic(subscriptions: Seq[KalixMethod]): Seq[KalixMethod] = {
+    val subscriptionTopicClass: Array[KalixMethod] =
+      if (hasTopicSubscription(component)) {
+        component.getMethods.sorted // make sure we get the methods in deterministic order
+          .filter(hasActionOutput)
+          .collect { case method =>
+            val subscriptionOptions = eventingInForTopic(component)
+            val kalixOptions =
+              kalix.MethodOptions.newBuilder().setEventing(subscriptionOptions).build()
+
+            KalixMethod(SubscriptionServiceMethod(method))
+              .withKalixOptions(kalixOptions)
+          }
+      } else Array.empty[KalixMethod]
+
+    def combineByTopic(kalixMethods: Seq[KalixMethod]): Seq[KalixMethod] = {
       def groupByTopic(methods: Seq[KalixMethod]): Map[String, Seq[KalixMethod]] = {
         val withTopicIn = methods.filter(kalixMethod =>
           kalixMethod.methodOptions.exists(option =>
@@ -93,7 +131,8 @@ private[impl] object ActionDescriptorFactory extends ComponentDescriptorFactory 
         //Assuming there is only one topic annotation per method, therefore head is as good as any other
         withTopicIn.groupBy(m => m.methodOptions.head.getEventing.getIn.getTopic)
       }
-      groupByTopic(subscriptions).collect {
+
+      groupByTopic(kalixMethods).collect {
         case (topic, kMethods) if kMethods.size > 1 =>
           val methodsMap =
             kMethods.map { k =>
@@ -101,7 +140,6 @@ private[impl] object ActionDescriptorFactory extends ComponentDescriptorFactory 
               val typeUrl = messageCodec.typeUrlFor(inputType)
               (typeUrl, k.serviceMethod.javaMethodOpt.get)
             }.toMap
-
           KalixMethod(
             CombinedSubscriptionServiceMethod(
               component.getName,
@@ -127,20 +165,20 @@ private[impl] object ActionDescriptorFactory extends ComponentDescriptorFactory 
       }
     val serviceName = nameGenerator.getName(component.getSimpleName)
 
-    def filterAndAddKalixOptions(to: Seq[KalixMethod], from: Seq[KalixMethod]): Seq[KalixMethod] = {
-      val inCommon = to.flatMap(toAdd =>
+    def addKalixOptions(to: Seq[KalixMethod], from: Seq[KalixMethod]): Seq[KalixMethod] = {
+      val added = to.flatMap(toAdd =>
         from
           .filter { addingFrom =>
             addingFrom.serviceMethod.methodName.equals(toAdd.serviceMethod.methodName)
           }
           .map(addingFrom => toAdd.withKalixOptions(addingFrom.methodOptions)))
-      val unique = to
+      val unmatchedInTo = to
         .filter { toAdd =>
           !from.exists { addingFrom =>
             addingFrom.serviceMethod.methodName.equals(toAdd.serviceMethod.methodName)
           }
         }
-      inCommon ++ unique
+      added ++ unmatchedInTo
     }
 
     def removeDuplicates(springMethods: Seq[KalixMethod], pubSubMethods: Seq[KalixMethod]): Seq[KalixMethod] = {
@@ -154,10 +192,12 @@ private[impl] object ActionDescriptorFactory extends ComponentDescriptorFactory 
       serviceName,
       serviceOptions = AclDescriptorFactory.serviceLevelAclAnnotation(component),
       component.getPackageName,
-      filterAndAddKalixOptions(springAnnotatedMethods, publicationTopicMethods)
+      addKalixOptions(syntheticMethods, publicationTopicMethods)
       ++ subscriptionValueEntityMethods
-      ++ combineByES(component.getName, subscriptionEventSourcedEntityMethods, messageCodec)
+      ++ combineByES(subscriptionEventSourcedEntityMethods, messageCodec, component)
+      ++ combineByES(subscriptionEventSourcedEntityClass, messageCodec, component)
+      ++ combineByTopic(subscriptionTopicClass)
       ++ combineByTopic(subscriptionTopicMethods)
-      ++ removeDuplicates(springAnnotatedMethods, publicationTopicMethods))
+      ++ removeDuplicates(syntheticMethods, publicationTopicMethods))
   }
 }
--- a/sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/ActionDescriptorFactory.scala
+++ b/sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/ActionDescriptorFactory.scala
@@ -81,14 +81,13 @@ private[impl] object ActionDescriptorFactory extends ComponentDescriptorFactory 
       if (hasEventSourcedEntitySubscription(component)) {
         component.getMethods.sorted // make sure we get the methods in deterministic order
           .filter(hasActionOutput)
-          .collect {
-            case method => 
-              val subscriptionOptions = eventingInForEventSourcedEntity(component)
-              val kalixOptions =
-                kalix.MethodOptions.newBuilder().setEventing(subscriptionOptions).build()
+          .collect { case method =>
+            val subscriptionOptions = eventingInForEventSourcedEntity(component)
+            val kalixOptions =
+              kalix.MethodOptions.newBuilder().setEventing(subscriptionOptions).build()
 
-              KalixMethod(SubscriptionServiceMethod(method))
-                .withKalixOptions(kalixOptions)
+            KalixMethod(SubscriptionServiceMethod(method))
+              .withKalixOptions(kalixOptions)
           }
       } else Array.empty[KalixMethod]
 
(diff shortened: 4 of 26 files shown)


sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/action/ReflectiveActionRouter.scala
```diff
@@ -60,10 +67,13 @@ class ReflectiveActionRouter[A <: Action](action: A, commandHandlers: Map[String
...
     val typeUrl = message.payload().asInstanceOf[ScalaPbAny].typeUrl
-    val methodInvoker = componentMethod.lookupInvoker(typeUrl)
 
-    val response = methodInvoker.invoke(action, context).asInstanceOf[Flux[Action.Effect[_]]]
-    Source.fromPublisher(response)
+    componentMethod.lookupInvoker(typeUrl) match {
+      case Some(methodInvoker) =>
+        val response = methodInvoker.invoke(action, context).asInstanceOf[Flux[Action.Effect[_]]]
+        Source.fromPublisher(response)
+      case None => Source.empty()
```
franciscolopezsancho, line 75: idem

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/eventsourcedentity/ReflectiveEventSourcedEntityRouter.scala
```diff
@@ -56,10 +60,14 @@ class ReflectiveEventSourcedEntityRouter[S, E <: EventSourcedEntity[S]](
...
     val methodInvoker = commandHandler.lookupInvoker(inputTypeUrl)
 
-    entity._internalSetCurrentState(state)
+    methodInvoker match {
+      case Some(invoker) =>
+        entity._internalSetCurrentState(state)
+        invoker
+          .invoke(entity, invocationContext)
+          .asInstanceOf[EventSourcedEntity.Effect[_]]
+      case None => new EventSourcedEntityEffectImpl()
```
franciscolopezsancho, line 69: idem

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/valueentity/ReflectiveValueEntityRouter.scala
```diff
@@ -43,12 +44,18 @@ class ReflectiveValueEntityRouter[S, E <: ValueEntity[S]](
...
+    methodInvoker match {
+      case Some(invoker) =>
+        // pass current state to entity
+        entity._internalSetCurrentState(state);
+
+        // safe call: typeUrl2Methods has at least one entry - famous last words.
+        invoker
+          .invoke(entity, context)
+          .asInstanceOf[ValueEntity.Effect[_]]
+      case None => new ValueEntityEffectImpl
```
franciscolopezsancho, line 57: idem

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/CommandHandler.scala
```diff
@@ -38,25 +39,29 @@ case class CommandHandler(
...
+        lastParam.getAnnotation(classOf[JsonSubTypes]).value().exists { subType =>
+          inputTypeUrl == messageCodec.typeUrlFor(subType.value())
+        }
+      } else false
+    }
   }
 
-  def lookupInvoker(inputTypeUrl: String): MethodInvoker =
-    methodInvokers.getOrElse(inputTypeUrl, lookupMethodAcceptingSubType(inputTypeUrl))
+  def lookupInvoker(inputTypeUrl: String): Option[MethodInvoker] =
```
franciscolopezsancho, line 53: quite verbose. I'll try again more compact way to state this.

franciscolopezsancho:
AFAIK we didn't subscribe on Views on Type level for a reason. Do I remember correctly?

sdk/java-sdk/src/main/java/kalix/javasdk/action/Action.java
```diff
@@ -148,6 +148,22 @@ interface Builder {
        * @param <S> The type of the message that must be returned by this call.
        */
       <S> Effect<S> asyncEffect(CompletionStage<Effect<S>> futureEffect);
+
```
franciscolopezsancho, line 151: @johanandren what do you think?

sdk/java-sdk/src/main/java/kalix/javasdk/action/Action.java
```diff
@@ -148,6 +148,22 @@ interface Builder {
...
+      <S> Effect<S> empty();
+
+      /**
+       * Creates a response that ignores the current element of a subscription or a publication
+       * so that consumption or production can proceed to the next element
+       *
+       * @return
+       * @param <S>
+       */
+      <S> Effect<S> ignore();
```
franciscolopezsancho, line 166: @johanandren what do you think?

johanandren: Looks good, did you double check that it causes 404/NotFound for a regular request-response? If so I think we can document that as well.

franciscolopezsancho: I did not test it. Although I would expect a 200 OK with an empty response. I'll test it now

franciscolopezsancho: Crystal clear now 👍 . I'll have a look at the proxy as well.

sdk/java-sdk/src/main/scala/kalix/javasdk/impl/action/ActionEffectImpl.scala
```diff
@@ -81,6 +85,10 @@ object ActionEffectImpl {
       AsyncEffect(futureMessage.asScala.map(s => Builder.reply[S](s))(ExecutionContext.parasitic), Nil)
     def asyncEffect[S](futureEffect: CompletionStage[Action.Effect[S]]): Action.Effect[S] =
       AsyncEffect(futureEffect.asScala, Nil)
+    def empty[S](): Action.Effect[S] =
```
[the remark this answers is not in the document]
johanandren: I think we should skip the `empty` and only have `ignore`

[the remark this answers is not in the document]
franciscolopezsancho: I thought `empty` for an endpoint usage. Mainly because `ignore` doesn't convey meaning there. Ignore makes sense on a stream but I can't quite make sense out of it in a `resquest/response` scenario. As a user, if a want to say this action gives no effect back. Don't you think that using ignore implies a knowledge of how things work underneath? Maybe we can use `noReply` instead?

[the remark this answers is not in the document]
franciscolopezsancho: ok then. I'll leave it to the day we have a pure Spring SDK 👍

johanandren:
This one has a ton of deleted java-sdk test protos, was that intentional?

franciscolopezsancho:
it was not. I'll put it back.

sdk/java-sdk/src/main/java/kalix/javasdk/action/Action.java
franciscolopezsancho: https://github.com/lightbend/kalix-proxy/pull/1664

sdk/spring-sdk/src/it/java/com/example/wiring/SpringSdkWiringIntegrationTest.java
```diff
@@ -109,6 +108,91 @@ public void verifyCounterEventSourceSubscription() {
...
 
+  @Test
+  public void verifyActionIsNotSubscribedToMultiplyAndRouterIgnores(){
+
+     webClient
+            .post()
+            .uri("counter/counterId2/increase/1")
+            .retrieve()
+            .bodyToMono(Integer.class)
+            .block(timeout);
```
octonato, line 119:
```suggestion
    webClient
            .post()
            .uri("counter/counterId2/increase/1")
            .retrieve()
            .bodyToMono(Integer.class)
            .block(timeout);

```

Just a minor formatting thing. This should have be picked by the code formatter. Maybe the Spring SDK module doesn't have it configured?

sdk/spring-sdk/src/it/java/com/example/wiring/SpringSdkWiringIntegrationTest.java
```diff
@@ -109,6 +108,91 @@ public void verifyCounterEventSourceSubscription() {
...
+            .post()
+            .uri("counter/counterId2/increase/1234")
+            .retrieve()
+            .bodyToMono(Integer.class)
+            .block(timeout);
+
+    Integer counterGet =
+            webClient.get().uri("/counter/counterId2").retrieve().bodyToMono(Integer.class).block(timeout);
+
+    //Before has time to process any action things still work
```
octonato, line 137: not clear what this comment means

sdk/spring-sdk/src/it/java/com/example/wiring/SpringSdkWiringIntegrationTest.java
```diff
@@ -109,6 +108,91 @@ public void verifyCounterEventSourceSubscription() {
...
+    Assertions.assertEquals(1*2 + 1234, counterGet);
+
+    //Once it tries to process actions will fail
```
octonato, line 140: but this is not failing and we are calling the entity here

what am I missing?

sdk/spring-sdk/src/it/java/com/example/wiring/SpringSdkWiringIntegrationTest.java
```diff
@@ -109,6 +108,91 @@ public void verifyCounterEventSourceSubscription() {
...
+            .atMost(10, TimeUnit.SECONDS)
+            .until(
+                    () ->
+                            webClient
+                                    .get()
+                                    .uri("/counter/counterId2")
+                                    .retrieve()
+                                    .bodyToMono(Integer.class)
+                                    .block(timeout),
+                    new IsEqual<Integer>(1*2 + 1234 + 1 ));
```
octonato, line 151: I'm also puzzled by this. How and why this call is now returning the same result +1?

sdk/spring-sdk/src/it/java/com/example/wiring/SpringSdkWiringIntegrationTest.java
```diff
@@ -109,6 +108,91 @@ public void verifyCounterEventSourceSubscription() {
...
+            .uri("counter/counterId2/increase/1234")
+            .retrieve()
+            .bodyToMono(Integer.class)
+            .block(timeout);
+
+    Integer counterGet =
+            webClient.get().uri("/counter/counterId2").retrieve().bodyToMono(Integer.class).block(timeout);
+
+    //Before has time to process any action things still work
+    Assertions.assertEquals(1*2 + 1234, counterGet);
```
octonato, line 138: nitpicking formatting. 

Here are use spaces around `+`, but not around `*`. 

```suggestion
    Assertions.assertEquals(1 * 2 + 1234, counterGet);
```

sdk/spring-sdk/src/it/java/com/example/wiring/SpringSdkWiringIntegrationTest.java
octonato, line 151:
```suggestion
                    new IsEqual<Integer>(1 * 2 + 1234 + 1 ));
```

sdk/spring-sdk/src/it/java/com/example/wiring/SpringSdkWiringIntegrationTest.java
```diff
@@ -314,4 +398,16 @@ public void verifyFindUsersByNameStreaming() {
...
+  // GIVEN view with subscription on type level marked with ignore
+  // to event having two different subtypes,
+  // the view handling only one of them
+  // WHEN another event the can handle comes along
+  // THEN the router delivers the event
+  // WHEN the event not handled comes along
+  // THEN the router ignores the event
+  // WHEN another event the can handle comes along
+  // THEN the router delivers the event
+  // THEN the view has the final state
```
octonato, line 412: missing tests?

sdk/spring-sdk/src/main/java/kalix/springsdk/annotations/Subscribe.java
```diff
@@ -68,10 +68,13 @@
      * kalix.javasdk.eventsourcedentity.EventSourcedEntity EventSourcedEntity}.
      */
     Class<? extends kalix.javasdk.eventsourcedentity.EventSourcedEntity<?>> value();
+
+    /** This option is only available for classes. Using it in a method doesn't change the default. */
```
octonato, line 72: this docs is only saying that it won't work in methods, but it doesn't say what it means. 

We should explain what `ignoreUnknown` means here. 

Once we explain it, we can then say that it's only effective when used at type level and that the option is ignore on method level. Better not to say "doesn't change the default", but simply say is not applicable.

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/ActionDescriptorFactory.scala
```diff
@@ -41,7 +46,7 @@ private[impl] object ActionDescriptorFactory extends ComponentDescriptorFactory
       messageCodec: SpringSdkMessageCodec,
       nameGenerator: NameGenerator): ComponentDescriptor = {
     //we should merge from here
-    val springAnnotatedMethods =
+    val syntheticMethods =
```
octonato, line 49: 👍

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/CommandHandler.scala
```diff
@@ -27,7 +27,8 @@ case class CommandHandler(
     grpcMethodName: String,
     messageCodec: SpringSdkMessageCodec,
     requestMessageDescriptor: Descriptors.Descriptor,
-    methodInvokers: Map[String, MethodInvoker]) {
+    methodInvokers: Map[String, MethodInvoker],
+    ignoreUnknown: Boolean) {
```
octonato, line 31: I feel that this option should not be here, but only at router level. We can move it to the component descriptor and then pass to the router.

Can have two variants of `lookupInvoker`? 

`lookInvoker` can stay as in this PR, returning `Option`, but not deciding if it throws exception based on `ignoreUnknown`. The router should make that decision. 

Then we also introduce a strict lookup. Can be `getInvoker`. This method would not return an `Option`, but a non-wrapped invoker. We will use this method in the routers that always expect to find an invoker, like the entity routers.

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/ViewDescriptorFactory.scala
```diff
@@ -176,51 +179,76 @@ private[impl] object ViewDescriptorFactory extends ComponentDescriptorFactory {
...
-            throw InvalidComponentException(
-              s"All update methods must return the same type, but [${method.getName}] returns [${entityClass.getName}] while a previous update method returns [${other.getName}]")
-          case None => previousEntityClass = Some(entityClass)
-        }
+      def invalidComponentException(method: Method, extraMessage: String) = {
+        throw InvalidComponentException(
+          s"Method [${method.getName}] annotated with '@Subscribe.EventSourcedEntity' should either receive " +
+          "a single parameter of one of the event types or " +
+          s"two ordered parameters  of type [${tableType.getName}] and an event type. $extraMessage")
+      }
```
octonato, line 206: This is obsolete. We removed that option. Don't need to fix in this PR, I will create a ticket to clean this up.

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/ViewDescriptorFactory.scala
```diff
@@ -176,51 +179,76 @@ private[impl] object ViewDescriptorFactory extends ComponentDescriptorFactory {
...
-        addTableOptionsToUpdateMethod(tableName, tableProtoMessageName, methodOptionsBuilder, true)
+    def getEventing(method: Method, component: Class[_]): Eventing =
+      if (hasEventSourcedEntitySubscription(component)) eventingInForEventSourcedEntity(component)
+      else eventingInForEventSourcedEntity(method)
 
-        KalixMethod(SubscriptionServiceMethod(method))
-          .withKalixOptions(methodOptionsBuilder.build())
+    def findIgnores(method: Method, component: Class[_]): Boolean =
+      if (hasEventSourcedEntitySubscription(component)) findIgnore(component)
+      else findIgnore(method)
```
octonato, line 241: But we don't support it as method annotation, so why we need it?

I see that we are using it to construct `SubscriptionServiceMethod`, but that will probably generate gRPC method with a invalid annotation.

octonato (commented):
Looking good, but I think there is some room for simplification.

The fact that we passing an `ignoreUnknown` deep in the code is forcing us to change lots of code. And pass it as `false` in places that it doesn't even make sense to have this flag. 

I think we should move it up to the `ComponentDescriptor` and then passed it further to the routers may be interested.

sdk/spring-sdk/src/it/java/com/example/wiring/SpringSdkWiringIntegrationTest.java
franciscolopezsancho: It was misleading. Changed now.

sdk/spring-sdk/src/it/java/com/example/wiring/SpringSdkWiringIntegrationTest.java
franciscolopezsancho: misleading comment. Changed now

sdk/spring-sdk/src/it/java/com/example/wiring/SpringSdkWiringIntegrationTest.java
franciscolopezsancho: added as a comment at the beginning of the assertion

sdk/spring-sdk/src/main/java/kalix/springsdk/annotations/Subscribe.java
franciscolopezsancho: reworded and added functionality description

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/ViewDescriptorFactory.scala
franciscolopezsancho: now you know we don't pass it to the proxy

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/CommandHandler.scala
franciscolopezsancho: refactored

franciscolopezsancho: done now

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/ViewDescriptorFactory.scala
franciscolopezsancho: for your reference @octonato 
https://github.com/lightbend/kalix/issues/7733

sdk/java-sdk/src/main/scala/kalix/javasdk/impl/action/ActionEffectImpl.scala
```diff
@@ -68,6 +65,13 @@ object ActionEffectImpl {
       copy(internalSideEffects = sideEffects)
   }
 
+  final case class EmptyEffect[T](internalSideEffects: Seq[SideEffect]) extends PrimaryEffect[T] {
```
johanandren, line 68: I vote for calling it Ignore all the way to protocol, so `IgnoreEffect`

sdk/java-sdk/src/main/scala/kalix/javasdk/impl/action/ActionEffectImpl.scala
```diff
@@ -68,6 +65,13 @@ object ActionEffectImpl {
...
+    def isEmpty: Boolean = true
+    protected def withSideEffects(sideEffect: Seq[SideEffect]): EmptyEffect[T] =
+      copy(internalSideEffects = sideEffect)
```
johanandren, line 71: Do we really allow side effects to empty? I guess we can't easily protect against it with types, so shouldn it lead to an IllegalState runtime exception rather?

johanandren (commented):
I think it would be easier to review and a bit cleaner if you could extract the ignore effect addition into a separate PR and then do the spring ignore on top. Also for visibility since the ignore effect will be interesting for non Spring usage as well.

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/ViewDescriptorFactory.scala
octonato: I had already created one. I mentioned that I was about to do it. 
https://github.com/lightbend/kalix/issues/7708

franciscolopezsancho: Ok. I thought you may forget with all the spring tasks you where creating. I'll delete it.

sdk/java-sdk/src/main/scala/kalix/javasdk/impl/action/ActionEffectImpl.scala
franciscolopezsancho: left over from the previous implementation. Yes, that is better

sdk/java-sdk/src/main/scala/kalix/javasdk/impl/action/ActionEffectImpl.scala
franciscolopezsancho: I thought why not allow side effects, like an alert? But I'm fine with adding an exception or even trying to redefine the types. How about making `PrimaryEffect` extend `EmptyEffect`? And being EmptyEffect having only `def isEmpty`?

sdk/java-sdk/src/main/java/kalix/javasdk/action/Action.java
```diff
@@ -148,6 +148,17 @@ interface Builder {
...
       <S> Effect<S> asyncEffect(CompletionStage<Effect<S>> futureEffect);
+
+      /**
+       * Creates a response when used in a subscription or a publication that ignores the current element of consumption or
+       * production so that they can proceed to the next element
+       *
+       * Creates a response 404 when used in an endpoint.
+       *
+       * @return
+       * @param <S>
```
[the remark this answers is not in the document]
franciscolopezsancho: I'll have to check the proxy. When we reviewed the `handleUnary` it was filtered out, creating only a `ResultEnd` right? I would expect the same. I'll check it out.

sdk/java-sdk/src/main/scala/kalix/javasdk/impl/action/ActionEffectImpl.scala
franciscolopezsancho: done

sdk/java-sdk/src/main/scala/kalix/javasdk/impl/action/ActionEffectImpl.scala
franciscolopezsancho: Forget what I said. Never mind. I'll add the exception.

sdk/spring-sdk/src/it/java/com/example/wiring/SpringSdkWiringIntegrationTest.java
franciscolopezsancho: Nope, this was the placeholder for verifyViewIsNotSubscribedToMultiplyAndRouterIgnores. Deleted the comments

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/ActionDescriptorFactory.scala
```diff
@@ -57,7 +60,9 @@ private[impl] object ActionDescriptorFactory extends ComponentDescriptorFactory
         val kalixOptions =
           kalix.MethodOptions.newBuilder().setEventing(subscriptionOptions).build()
 
-        KalixMethod(SubscriptionServiceMethod(method))
+        KalixMethod(
+          SubscriptionServiceMethod(method)
+        ) //FIXME add as option because valueEntity subscription doesn't have meaning
```
octonato, line 65: I don't see that is the problem here. What needs to be fixed?

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/ActionDescriptorFactory.scala
```diff
@@ -69,39 +74,84 @@ private[impl] object ActionDescriptorFactory extends ComponentDescriptorFactory
...
+                if !subscriptionTopicMethods.exists(s =>
+                  s.serviceMethod.methodName == method.getName) => // individual annotated overrides class level annotation
+              val subscriptionOptions = eventingInForTopic(component)
+              val kalixOptions =
+                kalix.MethodOptions.newBuilder().setEventing(subscriptionOptions).build()
+
+              KalixMethod(SubscriptionServiceMethod(method))
+                .withKalixOptions(kalixOptions)
+          }
+      } else Array.empty[KalixMethod]
```
octonato, line 133: If I'm understanding it correctly, it's possible to mix type level annotations with method level annotations as long as they are not on the same topic/entity. Is that correct? 

Do we also support it in the gRPC SDK? 

Because if that's the case, we are introducing a bug here. If we set the type level one to `ignoreUnknown` that will also affect the individually annotated methods.

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/action/ReflectiveActionRouter.scala
```diff
@@ -62,15 +70,17 @@ class ReflectiveActionRouter[A <: Action](action: A, commandHandlers: Map[String
...
-        .invoke(action, context)
-        .asInstanceOf[Flux[Action.Effect[_]]]
+    val inputTypeUrl = message.payload().asInstanceOf[ScalaPbAny].typeUrl
 
-    Source.fromPublisher(response)
+    componentMethod.lookupInvoker(inputTypeUrl) match {
+      case Some(methodInvoker) =>
+        val response = methodInvoker.invoke(action, context).asInstanceOf[Flux[Action.Effect[_]]]
+        Source.fromPublisher(response)
+      case None if ignoreUnknown => Source.empty()
```
octonato, line 79: We need to check if that is legal or expected. 

I can imagine a subscription / publishing that for every incoming message produces one or more outgoing messages.

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/action/ReflectiveActionRouter.scala
```diff
@@ -43,11 +47,15 @@ class ReflectiveActionRouter[A <: Action](action: A, commandHandlers: Map[String
...
-      .invoke(action, context)
-      .asInstanceOf[Action.Effect[_]]
+    val methodInvoker = commandHandler.lookupInvoker(inputTypeUrl)
+
+    methodInvoker match {
+      case Some(invoker) =>
+        invoker
+          .invoke(action, context)
+          .asInstanceOf[Action.Effect[_]]
+      case None => ActionEffectImpl.Builder.ignore()
```
octonato, line 57: Usually topic/entity subscriptions are done in with unary calls. I'm not sure if it will work on streamed calls. Possibly. In any case, we need to protect this method with the `ignoreUnknown` flag.

The tests are passing because we are always ignoring. We should only ignore if the flag is true. We should have a test that fails.

octonato (commented):
I think a spot a problem with the mixing of type level and method level subscriptions. If that's what we want to support, we will need to do some more refactoring.

sdk/spring-sdk/src/main/java/kalix/springsdk/action/ReflectiveActionProvider.java
```diff
@@ -76,7 +77,7 @@ public Descriptors.ServiceDescriptor serviceDescriptor() {
   @Override
   public ActionRouter<A> newRouter(ActionCreationContext context) {
     A action = factory.apply(context);
-    return new ReflectiveActionRouter<>(action, componentDescriptor.commandHandlers());
+    return new ReflectiveActionRouter<>(action, componentDescriptor.commandHandlers(), ComponentDescriptorFactory.findIgnore(action.getClass()));
```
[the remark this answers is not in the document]
franciscolopezsancho: indeed

sdk/spring-sdk/src/it/java/com/example/wiring/SpringSdkWiringIntegrationTest.java
```diff
@@ -109,6 +108,92 @@ public void verifyCounterEventSourceSubscription() {
...
+            .bodyToMono(Integer.class)
+            .block(timeout);
+
+    Integer counterGet =
+            webClient.get().uri("/counter/counterId2").retrieve().bodyToMono(Integer.class).block(timeout);
+
+    //Before has time to process any action the counter has only the values from the POSTs
+    Assertions.assertEquals(1 * 2 + 1234, counterGet);
+
+    //Once the action IncreaseActionWithIgnore processes event 1234 it adds 1 more to the counter
```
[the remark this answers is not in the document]
franciscolopezsancho: done

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/ComponentDescriptorFactory.scala
```diff
@@ -69,21 +101,57 @@ private[impl] object ComponentDescriptorFactory {
...
+  def findIgnoreForEventSourcedEntity(method: Method): Boolean = {
+    val ann = method.getAnnotation(classOf[Subscribe.EventSourcedEntity])
+    ann.ignoreUnkown()
+  }
+
+  def findIgnore(clazz: Class[_]): Boolean = {
+    if (hasTopicSubscription(clazz)) findIgnoreForTopic(clazz)
+    else if (hasEventSourcedEntitySubscription(clazz)) findIgnoreForEventSourcedEntity(clazz)
+    else false
+  }
```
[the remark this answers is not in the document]
franciscolopezsancho: true. Changed and deleted unsed

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/ActionDescriptorFactory.scala
```diff
@@ -69,39 +74,84 @@ private[impl] object ActionDescriptorFactory extends ComponentDescriptorFactory
...
     val subscriptionTopicMethods = component.getMethods
       .filter(hasTopicSubscription)
       .sorted // make sure we get the methods in deterministic order
       .map { method =>
         val subscriptionOptions = eventingInForTopic(method)
         val kalixOptions =
           kalix.MethodOptions.newBuilder().setEventing(subscriptionOptions).build()
+        if (hasTopicSubscription(component)
+          && findSubscriptionTopicName(method) == findSubscriptionTopicName(component))
+          throw InvalidComponentException(s"method [$method] and class [$component] are subscribe to the same topic.")
```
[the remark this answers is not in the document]
franciscolopezsancho: changed to  `subscribing`

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/action/ReflectiveActionRouter.scala
franciscolopezsancho: wow. Totally missed that `if`. 

That's a good point. I tried to create that test but it turns out that we can't catch that exception because gets thrown in another thread. I can do a test but it is very ugly and will need a brand-new entity. Shall I?

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/ActionDescriptorFactory.scala
franciscolopezsancho: I think this is more than a bug. We are just checking for the type level `ignore` in the router. We should not allow different topic/entity then. This is a problem.

franciscolopezsancho: It turns out that the protocol first doesn't allow both annotations. Type and method at the same time. I'll follow that path.

franciscolopezsancho: done

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/action/ReflectiveActionRouter.scala
franciscolopezsancho: Sorry. This is what I reviewed with Johan.

johanandren: I don't remember that we talked about that. I think it is alway ok for a streaming out to complete with empty stream but best verify by actually trying that with the proxy (I can't find any test coverage of subscription with streaming in)

I could imagine some cool stuff you could do if streaming in is allowed though, `statefulMap` + `flatMapConcat` to aggregate incoming events and only publish outgoing based on the aggregated state type of stuff.

johanandren: That doesn't mean it should/must be supported in the spring sdk though

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/action/ReflectiveActionRouter.scala
franciscolopezsancho: regarding the testing created the following [issue](https://github.com/lightbend/kalix/issues/7756)

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/ActionDescriptorFactory.scala
```diff
@@ -85,23 +109,40 @@ private[impl] object ActionDescriptorFactory extends ComponentDescriptorFactory
...
 
-    def combineByTopic(subscriptions: Seq[KalixMethod]): Seq[KalixMethod] = {
+    val subscriptionTopicClass: Array[KalixMethod] =
+      if (hasTopicSubscription(component)) {
+        component.getMethods.sorted // make sure we get the methods in deterministic order
+          .filter(hasActionOutput)
+          .collect {
+            case method
+                if !subscriptionTopicMethods.exists(s =>
+                  s.serviceMethod.methodName == method.getName) => // individual annotated overrides class level annotation
```
octonato, line 119: we don't need this check anymore, correct?

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/action/ReflectiveActionRouter.scala
franciscolopezsancho: I've got confused then. I thought that we covered this case when talking about `ResultPart` and `ResultEnd` in the `EventingManager`. My bad, we only covered handleUnary then I suppose. In either case, @octonato and I decided to continue this in another PR. In part related to this https://github.com/lightbend/kalix-jvm-sdk/issues/1169
Are you fine with this @johanandren?

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/ActionDescriptorFactory.scala
franciscolopezsancho: correct

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/ActionDescriptorFactory.scala
```diff
@@ -73,10 +77,30 @@ private[impl] object ActionDescriptorFactory extends ComponentDescriptorFactory
...
       }
 
+    val subscriptionEventSourcedEntityClass =
+      if (hasEventSourcedEntitySubscription(component)) {
+        component.getMethods.sorted // make sure we get the methods in deterministic order
+          .filter(hasActionOutput)
+          .collect {
+            case method
+                if !subscriptionEventSourcedEntityMethods.exists(s =>
+                  s.serviceMethod.methodName == method.getName) => // individual annotated overrides class level annotation
```
octonato, line 87: also not needed

franciscolopezsancho: I specifically looked for it... and failed

octonato (approved):
LGTM

PR #1168, opened 2022-10-13
fix: KalixClient only available for injections on Actions

octonato:
References https://github.com/lightbend/kalix/issues/7642

Still draft because tests are missing.


octonato (commented):
It's easy to test it. We need a deployed application and then we need to send a request that will trigger the reflective provides.

I had to make a manual test and it is working as expected. 

As we progress with the SDK, we will need to find a way to run this kind of test. Preferably without needing a running proxy.

sdk/spring-sdk/src/main/scala/kalix/springsdk/KalixConfiguration.scala
```diff
@@ -18,13 +18,21 @@ package kalix.springsdk
...
 import org.springframework.context.annotation.Configuration
 import org.springframework.stereotype.Component
 
+object KalixConfiguration {
+  def beanPostProcessorErrorMessage(beanClass: Class[_]): String =
+    s"${beanClass.getName} is a Kalix component and is marked as a Spring bean for automatic wiring. " +
+    "Kalix components cannot be accessed directly and therefore cannot be wired into other classes. " +
+    "In order to interact with a Kalix component, you should call it using the provided KalixClient."
+}
+
```
octonato, line 35: move this to this method so we can easily test the exception message

PR #1191, opened 2022-10-19
Add placeholders for Spring docs

beritou:
Refs: https://github.com/lightbend/kalix/issues/7717

### DIFF
--- a/docs/src/modules/spring/pages/index.adoc
+++ b/docs/src/modules/spring/pages/index.adoc
@@ -1,0 +1,22 @@
+= Developing with Spring (beta)
+
+include::ROOT:partial$include.adoc[]
+
+The Spring SDK provides a code-first approach to developing Kalix applications so that a developer does not need to worry about learning protobuf upfront.
+
+This SDK offers a way to develop applications which should be familiar to someone who has worked with Spring applications by making use of annotations to assemble Kalix and its components.
+
+NOTE: If you're new to Kalix and the different types of entities that exist in a Kalix app, consider reading [this section](https://docs.kalix.io/services/programming-model.html#_what_is_a_kalix_service) before continuing. 
+
+
+== Getting Started
+
+How to get started
+
+== Features
+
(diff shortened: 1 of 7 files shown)


octonato (approved):
Thanks @beritou 

I think we can merge it already. And add more placeholders as we go. Tomorrow we will start adding the pages. I created issues for each of them.

octonato:
Were you planning to add more placeholders? Is that the reason why you made it a draft?

beritou:
> Were you planning to add more placeholders? Is that the reason why you made it a draft?

I made it a draft because I wasn't sure we would all agree on the directory structure right away.

I will mark this as ready for review and merge.

efgpinto (approved):
Lets go!

docs/src/modules/spring/pages/index.adoc
```diff
@@ -0,0 +1,22 @@
+= Developing with Spring (beta)
+
+include::ROOT:partial$include.adoc[]
+
+The Spring SDK provides a code-first approach to developing Kalix applications so that a developer does not need to worry about learning protobuf upfront.
+
```
octonato, line 6: This PR is already merged, but by re-reading this paragraph, I have now a second thought. Should we even mention protobuf in the Spring SDK? 

We may want to keep it as such for know because in this initial phase, the words protobuf and gRPC is not completely hidden from the user. They will probably run into it in other parts of the documentation. 

But we may need to revisit it. Anyway, I just want to drop it here so we have a short discussion.

WDYT, @beritou?

beritou: > I think this comes directly from the spring-sdk README

Yes it does.

> Should we even mention protobuf in the Spring SDK?

Nope!

> I would say more, even the term "code-first" if you think about it in isolation (without the context that there is some API-first counterparts) can be very surprising for someone who just lands on this page directly ("code-first as in are we supposed to skip design? or what does even mean?").

> So yeah, I think this might need some rewording in the future.

Great points and I completely agree. If you really think about it, this entire sentence can be removed.

Issue #1169, opened 2022-10-13
Reintroduce ignore/noReply effect


johanandren:
In #844 we removed the `NoReply` effect because we couldn't figure out when it is useful, and it is weird when used as an effect for a regular request-response action call.

Currently you can only ignore events on a message type level, ignoring a specific message type or ignoring all unknown message types, making the consumption of the event in continue for unknowns rather than stop and stay stuck until you deploy a new version supporting the message type.

While working on #1167 we realized that for usage with eventing in and out on the same method, so for example subscribing to events from an event sourced entity, transforming, and then publishing to a message broker, ignoring on a more fine granular level makes sense internally (based on the classes accepted by the spring action). There already is room for this in the protocol, with an `ActionResponse(ActionResponse.Response.Empty)`.

I think such a more fine granular level could make sense also for other types of eventing. 

Just to make one example: "publish shopping cart updates to this kafka topic, but only once they reach $100 so we can trigger a coupon to that customer", while ignoring and moving on with the source event stream for other updates.

We'd still need to handle the empty case for regular better in the proxy, maybe we could just turn that into 404/NotFound for HTTP/gRPC and filter such responses out from streaming replies.

patriknw:
Seems like this is done?

PR #1194, opened 2022-10-20
feat: Reintroduce ignore/noReply effect #1169

franciscolopezsancho:
References https://github.com/lightbend/kalix-jvm-sdk/issues/1169

### DIFF
--- a/sdk/java-sdk/src/main/java/kalix/javasdk/action/Action.java
+++ b/sdk/java-sdk/src/main/java/kalix/javasdk/action/Action.java
@@ -148,6 +148,13 @@ public abstract class Action {
        * @param <S> The type of the message that must be returned by this call.
        */
       <S> Effect<S> asyncEffect(CompletionStage<Effect<S>> futureEffect);
+
+      /**
+       * Ignore the current element and proceed with processing the next element if returned for an
+       * element from eventing in. If used as a response to a regular gRPC or HTTP request it is turned
+       * into a NotFound response.
+       */
+      <S> Effect<S> ignore();
     }
 
     /**
--- a/sdk/java-sdk/src/main/java/kalix/javasdk/action/Action.java
+++ b/sdk/java-sdk/src/main/java/kalix/javasdk/action/Action.java
@@ -153,6 +153,8 @@ public abstract class Action {
        * Ignore the current element and proceed with processing the next element if returned for an
        * element from eventing in. If used as a response to a regular gRPC or HTTP request it is turned
        * into a NotFound response.
+       * 
+       * Ignore is not allowed to have side effects added with `addSideEffects`
        */
       <S> Effect<S> ignore();
     }
--- a/sdk/java-sdk/src/main/scala/kalix/javasdk/impl/action/ActionEffectImpl.scala
+++ b/sdk/java-sdk/src/main/scala/kalix/javasdk/impl/action/ActionEffectImpl.scala
@@ -18,12 +18,10 @@ package kalix.javasdk.impl.action
 
 import kalix.javasdk.{ DeferredCall, Metadata, SideEffect }
 import kalix.javasdk.action.Action
+
 import java.util
 import java.util.concurrent.CompletionStage
-
 import io.grpc.Status
-
-import scala.collection.immutable.Seq
 import scala.concurrent.ExecutionContext
 import scala.concurrent.Future
 import scala.jdk.CollectionConverters._
@@ -58,6 +56,13 @@ object ActionEffectImpl {
     protected def withSideEffects(sideEffects: Seq[SideEffect]): ForwardEffect[T] =
       copy(internalSideEffects = sideEffects)
   }
+  final case class IgnoreEffect[T](internalSideEffects: Seq[SideEffect]) extends PrimaryEffect[T] {
+    def isEmpty: Boolean = true
+
+    protected def withSideEffects(sideEffect: Seq[SideEffect]): IgnoreEffect[T] = {
+      throw new IllegalArgumentException("adding side effects to is not allowed.")
+    }
+  }
   final case class ErrorEffect[T](
       description: String,
       statusCode: Option[Status.Code],
@@ -81,6 +86,8 @@ object ActionEffectImpl {
       AsyncEffect(futureMessage.asScala.map(s => Builder.reply[S](s))(ExecutionContext.parasitic), Nil)
     def asyncEffect[S](futureEffect: CompletionStage[Action.Effect[S]]): Action.Effect[S] =
       AsyncEffect(futureEffect.asScala, Nil)
+    def ignore[S](): Action.Effect[S] =
+      IgnoreEffect(Nil)
   }
 
   def builder(): Action.Effect.Builder = Builder
--- a/sdk/java-sdk/src/main/scala/kalix/javasdk/impl/action/ActionEffectImpl.scala
+++ b/sdk/java-sdk/src/main/scala/kalix/javasdk/impl/action/ActionEffectImpl.scala
@@ -16,12 +16,14 @@
 
 package kalix.javasdk.impl.action
 
-import kalix.javasdk.{ DeferredCall, Metadata, SideEffect }
+import kalix.javasdk.{DeferredCall, Metadata, SideEffect}
 import kalix.javasdk.action.Action
 
 import java.util
 import java.util.concurrent.CompletionStage
+
 import io.grpc.Status
+
 import scala.concurrent.ExecutionContext
 import scala.concurrent.Future
 import scala.jdk.CollectionConverters._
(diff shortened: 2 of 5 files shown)


sdk/java-sdk/src/main/scala/kalix/javasdk/impl/action/ActionEffectImpl.scala
```diff
@@ -58,6 +56,13 @@ object ActionEffectImpl {
     protected def withSideEffects(sideEffects: Seq[SideEffect]): ForwardEffect[T] =
       copy(internalSideEffects = sideEffects)
   }
+  final case class IgnoreEffect[T](internalSideEffects: Seq[SideEffect]) extends PrimaryEffect[T] {
```
johanandren, line 59: Is the `internalSideEffects` still needed when not allowed?

sdk/scala-sdk/src/main/scala/kalix/scalasdk/impl/action/ActionEffectImpl.scala
```diff
@@ -112,6 +111,16 @@ private[scalasdk] object ActionEffectImpl {
     }
   }
 
+  final case class IgnoreEffect[T](internalSideEffects: Seq[SideEffect]) extends PrimaryEffect[T] {
+    def isEmpty: Boolean = true
+    protected def withSideEffects(sideEffect: Seq[SideEffect]): IgnoreEffect[T] =
+      copy(internalSideEffects = sideEffect)
```
johanandren, line 117: Same fail early here to get the call site where you call it in the stack trace rather than somewhere else in internal code where we toJavaSdk it?

sdk/java-sdk/src/main/java/kalix/javasdk/action/Action.java
```diff
@@ -148,6 +148,13 @@ interface Builder {
        * @param <S> The type of the message that must be returned by this call.
        */
       <S> Effect<S> asyncEffect(CompletionStage<Effect<S>> futureEffect);
+
+      /**
+       * Ignore the current element and proceed with processing the next element if returned for an
+       * element from eventing in. If used as a response to a regular gRPC or HTTP request it is turned
+       * into a NotFound response.
+       */
```
johanandren, line 156:
```suggestion
       * 
       * Ignore is not allowed to have side effects added with `addSideEffects`
       */
```

and same in the Scala one

sdk/scala-sdk/src/main/scala/kalix/scalasdk/action/Action.scala
```diff
@@ -146,6 +145,14 @@ object Action {
...
       def asyncEffect[S](futureEffect: Future[Action.Effect[S]]): Action.Effect[S]
 
+      /**
+       * Ignore the current element and proceed with processing the next element if returned for an element from
+       * eventing in. If used as a response to a regular gRPC or HTTP request it is turned into a NotFound response.
+       *
+       * @return
+       * @param S
+       */
+      def ignore[S](): Action.Effect[S]
```
johanandren, line 155: It's not imperative ignore, more of a value, so no parenthesis's?

sdk/java-sdk/src/main/scala/kalix/javasdk/impl/action/ActionEffectImpl.scala
franciscolopezsancho: it is because is a PrimaryEffect

sdk/scala-sdk/src/main/scala/kalix/scalasdk/action/Action.scala
franciscolopezsancho: I wasn't sure. Yes, it makes sense.  I'll follow these two rules
- it has no side effects. See https://docs.scala-lang.org/style/method-invocation.html, see Arity-0 
- Also to hide the fact that it can be a method or a value. Following the uniform access syntax. See https://docs.scala-lang.org/glossary/index.html#:~:text=The%20uniform%20access%20principle%20states,call%20sites%20of%20parameterless%20functions.

sdk/scala-sdk/src/main/scala/kalix/scalasdk/impl/action/ActionEffectImpl.scala
franciscolopezsancho: right

octonato (approved):

sdk/java-sdk/src/main/scala/kalix/javasdk/impl/action/ActionEffectImpl.scala
```diff
@@ -40,24 +40,28 @@ object ActionEffectImpl {
     def internalSideEffects(): Seq[SideEffect]
     protected def withSideEffects(sideEffects: Seq[SideEffect]): Action.Effect[T]
   }
+
```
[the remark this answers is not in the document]
franciscolopezsancho: hehehe

johanandren (changes requested):
👆

sdk/scala-sdk/src/main/scala/kalix/scalasdk/action/Action.scala
```diff
@@ -146,6 +145,16 @@ object Action {
...
       def asyncEffect[S](futureEffect: Future[Action.Effect[S]]): Action.Effect[S]
 
+      /**
+       * Ignore the current element and proceed with processing the next element if returned for an element from
+       * eventing in. If used as a response to a regular gRPC or HTTP request it is turned into a NotFound response.
+       *
+       * Ignore is not allowed to have side effects added with `addSideEffects`
+       *
+       * @return
+       * @param S
```
johanandren, line 155:
```suggestion
```

sdk/java-sdk/src/main/scala/kalix/javasdk/impl/action/ActionEffectImpl.scala
```diff
@@ -68,6 +72,20 @@ object ActionEffectImpl {
       copy(internalSideEffects = sideEffects)
   }
 
+  final case class IgnoreEffect[T](internalSideEffects: Seq[SideEffect]) extends PrimaryEffect[T] {
```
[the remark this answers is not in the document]
franciscolopezsancho: so much better

sdk/java-sdk/src/main/scala/kalix/javasdk/impl/action/ActionEffectImpl.scala
```diff
@@ -68,6 +72,16 @@ object ActionEffectImpl {
...
 
+  final case class IgnoreEffect[T]() extends PrimaryEffect[T] {
+    def isEmpty: Boolean = true
+
+    override def internalSideEffects = Nil
+
+    protected def withSideEffects(sideEffect: Seq[SideEffect]): IgnoreEffect[T] = {
+      throw new IllegalArgumentException("adding side effects to is not allowed.")
+    }
+  }
```
johanandren, line 83: Ah I see, PrimaryEffect is invariant, you can do like this to keep it a singleton:

```suggestion
  def IgnoreEffect[T]() = IgnoreEffect.asInstanceOf[PrimaryEffect[T]] 
  case object IgnoreEffect extends PrimaryEffect[Nothing] {
    def isEmpty: Boolean = true

    override def internalSideEffects = Nil

    protected def withSideEffects(sideEffect: Seq[SideEffect]): PrimaryEffect[Nothing] = {
      throw new IllegalArgumentException("adding side effects to is not allowed.")
    }
  }
```

johanandren: Or just cast it in `ignore()` since that is the only place referencing it needing to assign a type param

franciscolopezsancho: I prefer your first solution. I think is more expressive.

sdk/scala-sdk/src/main/scala/kalix/scalasdk/impl/action/ActionEffectImpl.scala
```diff
@@ -112,6 +113,17 @@ private[scalasdk] object ActionEffectImpl {
...
   }
 
+  def IgnoreEffect[T](): PrimaryEffect[T] = IgnoreEffect.asInstanceOf[PrimaryEffect[T]]
+  final case object IgnoreEffect extends PrimaryEffect[Nothing] {
+    def isEmpty: Boolean = true
+    override def internalSideEffects = Nil
+
+    protected def withSideEffects(sideEffect: Seq[SideEffect]): PrimaryEffect[Nothing] =
+      throw new IllegalArgumentException("adding side effects to is not allowed.")
+    override def toJavaSdk = javasdk.impl.action.ActionEffectImpl.IgnoreEffect()
```
franciscolopezsancho, line 123: removed previous casting

johanandren (approved):
LGTM!

PR #1195, opened 2022-10-20
feat: support for service and method level ACL

octonato:
Adds support for ACL annotations on type level (service) and method level.

Also block the mix of `@Acl` annotations with `@Subscription`

Resolves https://github.com/lightbend/kalix/issues/7368

### DIFF
--- a/sdk/spring-sdk/src/main/java/kalix/springsdk/annotations/Acl.java
+++ b/sdk/spring-sdk/src/main/java/kalix/springsdk/annotations/Acl.java
@@ -18,26 +18,75 @@ package kalix.springsdk.annotations;
 
 import java.lang.annotation.*;
 
+
+/**
+ * Defines ACL configuration for a resource.
+ */
 @Target({ElementType.TYPE, ElementType.METHOD})
 @Retention(RetentionPolicy.RUNTIME)
 @Documented
 public @interface Acl {
 
+  /**
+   * Principals that are allowed to access this resource.
+   * An incoming request must have at least one principal associated with it in this list to be allowed.
+   */
   Matcher[] allow() default {};
 
+
+  /**
+   * After matching an allow rule, an incoming request that has at least one principal
+   * that matches a deny rule will be denied.
+   */
   Matcher[] deny() default {};
 
+
+  /**
+   * The gRPC status code to respond with when access is denied.
+   *
+   * By default, this will be 7 (permission denied), but alternatives might include 16 (unauthenticated) or 5 (not
+   * found). If 0, indicates that the code should be inherited from the parent (regardless of the inherit field).
+   *
+   * When HTTP transcoding is in use, this code will be translated to an equivalent HTTP status code.
+   */
   int denyCode() default 0;
 
+
+  /**
+   * A principal matcher that can be used in an ACL.
+   *
+   * A principal is a very broad concept. It can correlate to a person, a system, or a more abstract concept, such as
+   * the internet.
+   *
+   * A single request may have multiple principals associated with it, for example, it may have come from a particular
+   * source system, and it may have certain credentials associated with it. When a matcher is applied to the request,
+   * the request is considered to match if at least one of the principals attached to the request matches.
+   *
+   * Each Matcher can be configured either with a 'service' or a 'principal', but not both.
+   */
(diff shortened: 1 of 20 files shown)


octonato:
I realise that I didn't add any Javadoc to the ACL annotations. I will add on this PR. I'm moving it to draft because of that.

sdk/spring-sdk/src/main/java/kalix/springsdk/annotations/Acl.java
```diff
@@ -18,26 +18,75 @@
 
 import java.lang.annotation.*;
 
+
+/**
+ * Defines ACL configuration for a resource.
+ */
```
octonato, line 24: just copy from the Proxy proto

sdk/spring-sdk/src/main/java/kalix/springsdk/annotations/Acl.java
```diff
@@ -18,26 +18,75 @@
...
+   * A principal matcher that can be used in an ACL.
+   *
+   * A principal is a very broad concept. It can correlate to a person, a system, or a more abstract concept, such as
+   * the internet.
+   *
+   * A single request may have multiple principals associated with it, for example, it may have come from a particular
+   * source system, and it may have certain credentials associated with it. When a matcher is applied to the request,
+   * the request is considered to match if at least one of the principals attached to the request matches.
+   *
+   * Each Matcher can be configured either with a 'service' or a 'principal', but not both.
```
octonato, line 65: This is an extra line in the docs. In proto, we can use `oneof`, but in java annotations land there is no such a thing. So we need to document. 

We do validate it at runtime.

sdk/spring-sdk/src/main/java/kalix/springsdk/annotations/Acl.java
```diff
@@ -18,26 +18,75 @@
...
 
+
+  /**
+   * The gRPC status code to respond with when access is denied.
+   *
+   * By default, this will be 7 (permission denied), but alternatives might include 16 (unauthenticated) or 5 (not
+   * found). If 0, indicates that the code should be inherited from the parent (regardless of the inherit field).
+   *
+   * When HTTP transcoding is in use, this code will be translated to an equivalent HTTP status code.
+   */
```
octonato, line 51: This is something that I didn't realise before. The deny code are gRPC codes. I think we can leave it as it for now, but we may want to change it to http status code and then convert to proto under the hood.

efgpinto (approved):
LGTM, left 2 comments about details.

Also, was thinking if we should have a sample demonstrating the use of ACLs (we don't have for java/scala btw).

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/ActionDescriptorFactory.scala
```diff
@@ -152,11 +152,12 @@ private[impl] object ActionDescriptorFactory extends ComponentDescriptorFactory
...
-      filterAndAddKalixOptions(springAnnotatedMethods, publicationTopicMethods)
-      ++ subscriptionValueEntityMethods
-      ++ combineByES(component.getName, subscriptionEventSourcedEntityMethods, messageCodec)
-      ++ combineByTopic(subscriptionTopicMethods)
-      ++ removeDuplicates(springAnnotatedMethods, publicationTopicMethods))
+      filterAndAddKalixOptions(springAnnotatedMethods, publicationTopicMethods) ++
+      subscriptionValueEntityMethods ++
+      combineByES(component.getName, subscriptionEventSourcedEntityMethods, messageCodec) ++
+      combineByTopic(subscriptionTopicMethods) ++
+      removeDuplicates(springAnnotatedMethods, publicationTopicMethods))
```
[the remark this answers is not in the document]
octonato: That was not intentional.

The reason is that I moved this block to outside the params block because initially I was planning to run a `map` and I didn't want to map inside the param block.

If you try to assign this Seq concatenation to a val, you need to add the `++` at the end. So I did that. Later, I changed my mind and brought the seq concatenation back inside and forgot to revert the position of the `++`.  

I do prefer this variant though, but just a question of taste.

I will revert it because that will conflict with Cisco's PR.

PR #1196, opened 2022-10-20
chore: adding spring event-sourced shopping cart sample

efgpinto:
References https://github.com/lightbend/kalix/issues/7715

We agreed on using similar samples from the ones being used on the java/scala counterpart, so we were missing a shoppingcart ES sample to use in docs. Adding it here.

Have a look at the approach of Service vs Entity, trying to model things closer to what a Spring dev would expect.
~I might add the ES docs here as well or we can merge this independently. Will see.~ I will do that on a separate PR.

### DIFF
--- a/samples/spring-eventsourced-shopping-cart/pom.xml
+++ b/samples/spring-eventsourced-shopping-cart/pom.xml
@@ -1,0 +1,308 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+
+  <groupId>com.example</groupId>
+  <artifactId>spring-eventsourced-shopping-cart</artifactId>
+  <version>0.0.1-SNAPSHOT</version>
+  <packaging>jar</packaging>
+
+  <name>spring-eventsourced-shopping-cart</name>
+  <properties>
+
+    <!-- TODO Update to your own Docker repository or Docker Hub scope -->
+    <dockerImage>gcr.io/kalix-public/${project.artifactId}</dockerImage>
+    <dockerTag>${project.version}-${build.timestamp}</dockerTag>
+    <maven.build.timestamp.format>yyyyMMddHHmmss</maven.build.timestamp.format>
+    <mainClass>com.example.shoppingcart.Main</mainClass>
+
+    <jdk.target>17</jdk.target>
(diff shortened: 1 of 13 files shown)


efgpinto:
@octonato do you want to have a look at the structure and see if this is more or less what you had in mind?

samples/spring-eventsourced-shopping-cart/pom.xml
```diff
@@ -0,0 +1,304 @@
...
+      </plugin>
+
+      <plugin>
+        <groupId>io.kalix</groupId>
+        <artifactId>kalix-maven-plugin</artifactId>
+        <version>${kalix-sdk.version}</version>
+        <executions>
+          <execution>
+            <goals>
+              <goal>generate</goal>
```
octonato, line 212:
```suggestion
```

samples/spring-eventsourced-shopping-cart/src/main/java/com/example/shoppingcart/ShoppingCartService.java
```diff
@@ -0,0 +1,116 @@
...
+  // tag::itemAdded[]
+  @EventHandler
+  public ShoppingCart itemAdded(ShoppingCartEvent.ItemAdded itemAdded) {
+    var item = itemAdded.item();
+    var lineItem = updateItem(item, currentState());
+    List<LineItem> lineItems =
+        removeItemByProductId(currentState(), item.productId());
+    lineItems.add(lineItem);
+    lineItems.sort(Comparator.comparing(LineItem::productId));
+    return currentState().withItems(lineItems);
```
octonato, line 80: To emphasize that the `ShoppingCart` is the model and that `ShoppingCartService` is just a service around it, we can move this kind of logic to `ShoppingCart`. 

This is not possible in the gRPC SDKs, because the models are generated and we can't add logic to them. So users have to create anemic models with logic in the services. 

So the pattern that I had in mind was to move as much logic as possible to the model and have the service as a thin wrapper/facade between the model and kalix.

octonato (approved):
Looking good. If one suggestion. Good for merging otherwise.

samples/spring-eventsourced-shopping-cart/src/main/java/com/example/shoppingcart/ShoppingCartService.java
efgpinto: Yeah, thought the same earlier today when looking at what you've done for the customer registry sample... I think it's a better pattern. Will update. Thanks

octonato: Yep, I think is a good way forward. 

One thing that still not perfect, is that the Service extends something called `EventSourcedEntity` and that's kind of weird.

Also, when we subscribe to it from a view, we end up with:

`@Subscribe.EventSourcedEntity(ShoppingCartService.class)` 

That's also weird.

octonato (approved):
LGTM

PR #1198, opened 2022-10-20
chore: clean-up spring samples poms

efgpinto:
Just noticed these leftovers, inspired by https://github.com/lightbend/kalix-jvm-sdk/pull/1197

### DIFF
--- a/samples/spring-customer-registry-views-quickstart/pom.xml
+++ b/samples/spring-customer-registry-views-quickstart/pom.xml
@@ -25,10 +25,6 @@
 
   <build>
     <resources>
-      <!-- Add the generated protobuf descriptor to the classpath, so that source mapping works -->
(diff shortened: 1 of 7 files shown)


samples/spring-customer-registry-views-quickstart/pom.xml
```diff
@@ -57,38 +53,6 @@
         </configuration>
       </plugin>
 
-      <plugin>
-        <groupId>org.xolstice.maven.plugins</groupId>
-        <artifactId>protobuf-maven-plugin</artifactId>
```
octonato, line 62: I initially thought we needed it to generate some protobuf used inside. But that's only the SDK that needs it. The final user project doesn't.

octonato (approved):

PR #1199, opened 2022-10-21
feat: samples available for spring docs

franciscolopezsancho:
to allow loading samples to the docs on the `spring` folder

### DIFF
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -43,7 +43,7 @@ attributes:
 		>> "${java_managed_partials}/attributes.adoc"
 	echo ":console: https://console.kalix.io/" \
 		>> "${java_managed_partials}/attributes.adoc"
-	mkdir -p "${java_managed_partials}"
+	mkdir -p "${spring_managed_partials}"
 	bin/version.sh | xargs -0  printf ":kalix-jvm-sdk-version: %s" \
 		> "${spring_managed_partials}/attributes.adoc"
 	echo ":java-version: 11" \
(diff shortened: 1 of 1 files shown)


docs/Makefile
```diff
@@ -29,39 +34,49 @@ prepare:
...
+		> "${spring_managed_partials}/attributes.adoc"
 	echo ":java-version: 11" \
-		>> "${managed_partials}/attributes.adoc"
+		>> "${spring_managed_partials}/attributes.adoc"
 	echo ":java_minimum_sdk_version: 0.7.0" \
-		>> "${managed_partials}/attributes.adoc"
+		>> "${spring_managed_partials}/attributes.adoc"
 	echo ":console: https://console.kalix.io/" \
-		>> "${managed_partials}/attributes.adoc"
+		>> "${spring_managed_partials}/attributes.adoc"
```
efgpinto, line 54: Could we maybe replace this with `rsync -a ${java_managed_partials}/ ${spring_managed_partials}`? meaning, just do a copy of the java folder and contents since I'm not sure there will ever be differences..

efgpinto (approved):
LGTM to me, thanks for solving. 

left a suggestion but happy with this version as well. Up to you.

efgpinto:
Ah, another small thing.. maybe we could use a more specific message when merging? `fix: samples available for spring docs` or something alike

docs/Makefile
franciscolopezsancho: very good point

PR #1201, opened 2022-10-24
docs(spring-sdk): document event sourced entities

efgpinto:
References https://github.com/lightbend/kalix/issues/7715


docs/dev/src/modules/ROOT/partials/eventsourced-snapshots.adoc
```diff
@@ -0,0 +1,11 @@
...
+
+Snapshots are stored and handled automatically by Kalix without any specific code required. Snapshots are stored after a configured number of events:
+
+[source,conf,indent=0]
+.src/main/resources/application.conf
+----
+include::java:example$spring-eventsourced-shopping-cart/src/main/resources/application.conf[tag=snapshot-every]
+----
+
+When the Event Sourced Entity is loaded again, the snapshot will be loaded before any other events are received.
```
efgpinto, line 11: This would be duplicated between java/scala and spring so decided to move to a partial.. (we should probably agree if this is a strategy we want to follow, or if this shared partials should be somewhere else). I've tried to add it to the "developing" module but it did not work.. ended-up adding to the ROOT. but let me know what would be best here..

docs/src/modules/java/pages/event-sourced-entities.adoc
```diff
@@ -231,17 +231,7 @@ When more components are added the `KalixFactory` is regenerated and you have to
...
-Snapshots are stored and handled automatically by Kalix without any specific code required. Snapshots are stored after a configured number of events:
-
-[source,conf,indent=0]
-.src/main/resources/application.conf
-----
-include::example$java-eventsourced-shopping-cart/src/main/resources/application.conf[tag=snapshot-every]
-----
-
-When the Event Sourced Entity is loaded again, the snapshot will be loaded before any other events are received.
+include::ROOT:partial$eventsourced-snapshots.adoc[]
```
efgpinto, line 234: using the (now) partial here

docs/src/modules/spring/pages/event-sourced-entities.adoc
```diff
@@ -0,0 +1,160 @@
...
+<1> Stores the `entityId` on an internal attribute so we can use it later.
+<2> Provides initial state --overriding `emptyState()` is optional but if not doing it, be careful to deal with a currentState() with a `null` value when receiving the first command.
+<3> Marks this method as a command handler for `GET` requests.
+<4> Returns the current state as reply for the request.
+
+IMPORTANT: For simplicity purposes, we are returning the internal state directly back to the requester. In a real-world scenario, it's usually better to instead convert this internal domain model into a public model so the internal representation is free to evolve without breaking clients code.
+
+== Snapshots
+
+include::ROOT:partial$eventsourced-snapshots.adoc[]
```
efgpinto, line 126: using the (now) partial here

samples/spring-eventsourced-shopping-cart/src/main/java/com/example/shoppingcart/ShoppingCartService.java
```diff
@@ -12,20 +12,24 @@
 import java.util.Collections;
 
 // tag::class[]
-@Entity(entityKey = "cartId", entityType = "shopping-cart")
-@RequestMapping("/cart/{cartId}")
-public class ShoppingCartService extends EventSourcedEntity<ShoppingCart> {
+@Entity(entityKey = "cartId", entityType = "shopping-cart") // <2>
+@RequestMapping("/cart/{cartId}") // <3>
+public class ShoppingCartService extends EventSourcedEntity<ShoppingCart> { // <1>
```
octonato: Let's change it to `ShoppingCartEntity` as discussed the other day.

samples/spring-eventsourced-shopping-cart/src/main/resources/application.conf
```diff
@@ -0,0 +1,4 @@
+
+# //tag::snapshot-every[]
+kalix.event-sourced-entity.snapshot-every = 100
```
octonato, line 3: Which bring us to the fact that we are not yet supporting entity options programmatically. Users can define their options in code as well. 

I will create a ticket for it.

docs/src/modules/spring/pages/event-sourced-entities.adoc
```diff
@@ -0,0 +1,160 @@
...
+
+[source,java]
+.src/main/java/com/example/shoppingcart/domain/ShoppingCart.java
+----
+include::example$spring-eventsourced-shopping-cart/src/main/java/com/example/shoppingcart/domain/ShoppingCart.java[tag=domain]
+----
+<1> Our `ShoppingCart` is fairly simple, being composed only by a `cartId` and a list of line items.
+<2> A `LineItem` represents a single product and the quantity we intend to buy.
+
+NOTE: Above we are taking advantage of the Java `record` to reduce the amount of boilerplate code, but you can use regular classes so long as they can be serialized.
```
octonato, line 32: Maybe better to be explicit that it needs to be serializable to json and therefore properly annotated with Jackson annotations.

samples/spring-eventsourced-shopping-cart/src/main/java/com/example/shoppingcart/domain/ShoppingCartEvent.java
```diff
@@ -4,21 +4,20 @@
...
         @JsonSubTypes.Type(value = ShoppingCartEvent.ItemAdded.class, name = "item-added"),
         @JsonSubTypes.Type(value = ShoppingCartEvent.ItemRemoved.class, name = "item-removed"),
         @JsonSubTypes.Type(value = ShoppingCartEvent.CheckedOut.class, name = "checked-out")
     })
-public sealed interface ShoppingCartEvent {
+public sealed interface ShoppingCartEvent { // <3>
 
-  record ItemAdded(ShoppingCart.LineItem item) implements ShoppingCartEvent {
-  }
+  record ItemAdded(ShoppingCart.LineItem item) implements ShoppingCartEvent { }
```
octonato, line 17: And alternative here is to use `@JsonTypeName` to annotated each record. 

If we use it, we can remove the `name` attribute from the `@JsonSubTypes.Type` above. 

Anyway, that's just a taste thing. Both techniques works.

octonato: Hmm, now I'm having a second thought about what I just said. 

Having the name declared together with the type in the `@JsonSubTypes.Type` may be a better option. 

My reasoning is that each time that a user adds a new event type, they need to add a new `@JsonSubTypes.Type` line. As such, they have only one place to care about and they also see the other lines, so they will remember that they need to choose a name for the type hint. 

My point is, using `@JsonSubTypes.Type` like we are doing here should be the recommended technique as it increases the chances of doing the right thing.

docs/src/modules/spring/pages/event-sourced-entities.adoc
```diff
@@ -0,0 +1,160 @@
...
+Another fundamental aspect of our entity will be its domain events. For now, we will have 3 different events `ItemAdded`, `ItemRemoved` and `CheckedOut`, defined as below:
+
+[source,java]
+.src/main/java/com/example/shoppingcart/domain/ShoppingCartEvent.java
+----
+include::example$spring-eventsourced-shopping-cart/src/main/java/com/example/shoppingcart/domain/ShoppingCartEvent.java[tag=events]
+----
+<1> Include logical type name as JSON property "type."
+<2> Declare subtypes of the serializable polymorphic type `ShoppingCartEvent` associating logical names with each.
+<3> The 3 types of event all derive from the same type `ShoppingCartEvent`.
```
octonato, line 43: Maybe worth to say that we recommend the usage of an explicit name. Failing to do so, we cause jackson to use the FQCN. But this data will be persisted and if later users decide to change the package, they won't be able to deserialize the persisted payloads.

docs/src/modules/spring/pages/event-sourced-entities.adoc
```diff
@@ -0,0 +1,160 @@
...
+.src/main/java/com/example/shoppingcart/ShoppingCartService.java
+----
+include::example$spring-eventsourced-shopping-cart/src/main/java/com/example/shoppingcart/ShoppingCartService.java[tag=class]
+----
+<1> Create a class that extends `EventSourcedEntity<S>`, where `S` is the state type this entity will store (i.e. `ShoppingCart`).
+<2> Make sure to annotate such class with `@Entity` and pass in an `entityKey` and an `entityType`.
+<3> Make use of Spring's RequestMapping annotation to define the routes to your entity.
+
+
+NOTE: The entity key `cartId` must match a path parameter (i.e. `cartId`) and such value needs to be unique per entity. On the other hand, the entity type `shopping-cart` is common for all instances of this entity but should be unique across the different entity types.
```
octonato, line 66: Maybe good to mention that the entityType must be stable and can't be changed once the application is deployed in production.

efgpinto:
Updated this to include the rename from *Service to *Entity on the sample.

docs/src/modules/spring/pages/event-sourced-entities.adoc
```diff
@@ -0,0 +1,160 @@
...
+.src/main/java/com/example/shoppingcart/domain/ShoppingCart.java
+----
+include::example$spring-eventsourced-shopping-cart/src/main/java/com/example/shoppingcart/domain/ShoppingCart.java[tag=itemAdded]
+----
+<1> For an existing item, we will make sure to sum the existing quantity with the incoming one.
+<2> Returns an update list of items without the existing item.
+<3> Adds the update item to the shopping cart.
+<4> Returns a new instance of the shopping cart with the updated line items.
+
+- the `emptyState()` override is optional. If not using it, be careful to deal with a currentState() with a `null` value when receiving the first command
```
octonato, line 103: Also when applying the first event. Here users are expected to initialize the state for the first time.

docs/src/modules/spring/pages/event-sourced-entities.adoc
```diff
@@ -0,0 +1,160 @@
...
+
+To have access to the current state of the entity we can use `currentState()` as you have probably noticed from the examples above. However, what if this is the first command we are receiving for this entity? The following example shows the implementation of the read-only command handler (accessed through `GET /cart/myCarId`):
+
+[source,java,indent=0]
+.src/main/java/com/example/shoppingcart/ShoppingCartService.java
+----
+include::example$spring-eventsourced-shopping-cart/src/main/java/com/example/shoppingcart/ShoppingCartService.java[tag=getCart]
+----
+<1> Stores the `entityId` on an internal attribute so we can use it later.
+<2> Provides initial state --overriding `emptyState()` is optional but if not doing it, be careful to deal with a currentState() with a `null` value when receiving the first command.
```
octonato, line 116:
```suggestion
<2> Provides initial state -- overriding `emptyState()` is optional but if not doing it, be careful to deal with a currentState() with a `null` value when receiving the first command.
```

docs/src/modules/spring/pages/event-sourced-entities.adoc
```diff
@@ -0,0 +1,160 @@
...
+<3> Marks this method as a command handler for `GET` requests.
+<4> Returns the current state as reply for the request.
```
octonato, line 118: So here we can also mention that the state in the lambda is the updated one. After applying the event and not the same as returned by `currentState`.

docs/src/modules/spring/pages/event-sourced-entities.adoc
```diff
@@ -0,0 +1,160 @@
...
+<3> Asserts the return value is `"OK"`.
+<4> Returns the next event of type `IdemAdded` and asserts on the quantity.
+<5> Add a new item with quantity `5`.
+<6> Asserts that the total number of events should be 2.
+<7> Calls the `getCart` method and asserts that quantity should be `15`.
+
+
+*EventSourcedResult*
+
+Calling a command handler through the TestKit gives us back an link:{attachmentsdir}/testkit/kalix/javasdk/testkit/EventSourcedResult.html[`EventSourcedResult` {tab-icon}, window="new"]. This class has methods that we can use to assert of handling the command, such as:
```
octonato, line 146:
```suggestion
Calling a command handler through the TestKit gives us back an link:{attachmentsdir}/testkit/kalix/javasdk/testkit/EventSourcedResult.html[`EventSourcedResult` {tab-icon}, window="new"]. This class has methods that we can use to assert the handling of the command, such as:
```

octonato (commented):
Looking good

Left a few comments and suggestions.

samples/spring-eventsourced-shopping-cart/src/main/java/com/example/shoppingcart/ShoppingCartService.java
octonato: Oh, already obsolete comment. 😄

docs/src/modules/spring/pages/event-sourced-entities.adoc
efgpinto: This is actually a Vale rule... it throws an error if using the suggested version (which is also what I was using): ` 116:1  error  Remove the spaces around ' —'.  Lightbend.Dashes`

however, I just checked and it is ok with using "state--overriding" so I will probably update to that since seems to be the format being used elsewhere in the docs..

docs/src/modules/spring/pages/event-sourced-entities.adoc
efgpinto: Thanks, good to clarify it is "to json". I wonder if we should explicitely mention Jackson though, since although not the most likely, one could still decide to use another lib (e.g. gson) instead. But maybe I can mention Jackson as an example rather giving the impression it is required to use that one. 👍

docs/src/modules/spring/pages/event-sourced-entities.adoc
efgpinto: Definitely worth. updated with an `IMPORTANT` note about that. Thanks

docs/src/modules/spring/pages/event-sourced-entities.adoc
efgpinto: I've added this note back on the other example where we actually handle the event.. Just because on this one, there's no event produced so I think it does not fit so well.

docs/src/modules/spring/pages/event-sourced-entities.adoc
```diff
@@ -0,0 +1,158 @@
...
+
+[source,java,indent=0]
+.src/main/java/com/example/shoppingcart/ShoppingCartEntity.java
+----
+include::example$spring-eventsourced-shopping-cart/src/main/java/com/example/shoppingcart/ShoppingCartEntity.java[tag=addItem]
+----
+<1> The validation ensures the quantity of items added is greater than zero and it fails for calls with illegal values by returning an `Effect` with `effects().error`.
+<2> From the current incoming `LineItem` we create a new `ItemAdded` event representing the change of the cart.
+<3> We store the event by returning an `Effect` with `effects().emitEvent`.
+<4> The acknowledgment that the command was successfully processed is only sent if the event was successfully stored and processed, otherwise there will be an error reply. The lambda parameter `newState` gives us access to the new state returned by the processing of such event.
```
efgpinto, line 86: Updated this to address your [comment](https://github.com/lightbend/kalix-jvm-sdk/pull/1201/files#r1004239371) @octonato . Does this look fine to you?

efgpinto:
Thanks for the suggestions Renato. Updated now.

docs/src/modules/spring/pages/event-sourced-entities.adoc
```diff
@@ -0,0 +1,158 @@
...
+[source,java]
+.src/main/java/com/example/shoppingcart/domain/ShoppingCartEvent.java
+----
+include::example$spring-eventsourced-shopping-cart/src/main/java/com/example/shoppingcart/domain/ShoppingCartEvent.java[tag=events]
+----
+<1> Include logical type name as JSON property "type."
+<2> Declare subtypes of the serializable polymorphic type `ShoppingCartEvent` associating logical names with each.
+<3> The 3 types of event all derive from the same type `ShoppingCartEvent`.
+
+IMPORTANT: The use of logical names for subtypes is essential for maintainability purposes. Since that information is persisted, failing to do so might lead to the use a FQCN preventing the application from correctly deserialize the events in case of a package change.
```
octonato, line 45: Nice sentence!

docs/src/modules/spring/pages/event-sourced-entities.adoc
octonato: That's better, but I think we should say applying the event instead of processing. We apply the event on the state and produce a new state. That's how we usually explain it. I'm sure we say that in other parts of the documentation as well.

octonato (approved):
LGTM, left one more suggestion about using the word 'processing' vs. 'applying. Ok for merge otherwise.

PR #1203, opened 2022-10-25
chore: minor improvements to accomotade spring docs

octonato:
Those are some basic changes to the docs that will help us add the Spring SDK docs.

### DIFF
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -39,10 +39,15 @@ attributes:
 		> "${java_managed_partials}/attributes.adoc"
 	echo ":java-version: 11" \
 		>> "${java_managed_partials}/attributes.adoc"
+	echo ":spring-java-version: 17" \
+		>> "${java_managed_partials}/attributes.adoc"
+	echo ":minimum_docker_version: 19.03" \
+		>> "${java_managed_partials}/attributes.adoc"
 	echo ":java_minimum_sdk_version: 0.7.0" \
 		>> "${java_managed_partials}/attributes.adoc"
 	echo ":console: https://console.kalix.io/" \
 		>> "${java_managed_partials}/attributes.adoc"
+
 	mkdir -p "${spring_managed_partials}"
 	cp "${java_managed_partials}/attributes.adoc" "${spring_managed_partials}/attributes.adoc"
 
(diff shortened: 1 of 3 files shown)


docs/Makefile
```diff
@@ -39,10 +39,15 @@ attributes:
 		> "${java_managed_partials}/attributes.adoc"
 	echo ":java-version: 11" \
 		>> "${java_managed_partials}/attributes.adoc"
+	echo ":spring-java-version: 17" \
+		>> "${java_managed_partials}/attributes.adoc"
+	echo ":minimum_docker_version: 19.03" \
+		>> "${java_managed_partials}/attributes.adoc"
```
octonato, line 45: I don't know why, but this was not being added to the attributes, but load each time by the include.adoc. 

Anyway, better to manage the versions from a single place.

docs/Makefile
```diff
@@ -39,10 +39,15 @@ attributes:
 		> "${java_managed_partials}/attributes.adoc"
 	echo ":java-version: 11" \
 		>> "${java_managed_partials}/attributes.adoc"
+	echo ":spring-java-version: 17" \
```
octonato, line 42: Spring docs should recommend the usage of Java 17

docs/Makefile
```diff
@@ -61,8 +66,9 @@ apidocs:
 
 examples:
 	mkdir -p "${java_managed_examples}"
-	rsync -a --exclude-from=.examplesignore ../samples/ "${java_managed_examples}/"
-	rsync -a --exclude-from=.examplesignore ../samples/ "${spring_managed_examples}/"
+	rsync -a --exclude-from=.examplesignore ../samples/java-* "${java_managed_examples}/"
+	rsync -a --exclude-from=.examplesignore ../samples/scala-* "${java_managed_examples}/"
+	rsync -a --exclude-from=.examplesignore ../samples/spring-* "${spring_managed_examples}/"
```
octonato, line 71: Java and Scala goes together on the java managed examples.

The Spring ones are apart.

docs/dev/src/antora.yml
```diff
@@ -3,3 +3,4 @@ version: master
 
 nav:
 - modules/java/nav.adoc
+- modules/spring/nav.adoc
```
octonato, line 6: When we build it locally, we also want to access the spring sections. That may change later when we integrate with the main docs. But we can go with this approach for now.

efgpinto, line 6: We might want to not add this for now, otherwise it might be released with a new version of the sdk, right?

docs/Makefile
octonato: Btw, I don't even think we recommend docker 19.03 all over the places. I think I have seen other versions in other places. Probably hardcoded in the adoc files.

docs/Makefile
efgpinto, line 71: nice, thought about this myself as well 👍

docs/Makefile
franciscolopezsancho: I wasn't aware of this

docs/dev/src/antora.yml
efgpinto: Won't this mean the spring docs will be visible if we merge this and then release the sdk?

octonato: hmm, you are right

octonato: It's now removed

efgpinto (approved):
LGTM

PR #1205, opened 2022-10-25
feat: adding actions spring docs #7719

franciscolopezsancho:
References https://github.com/lightbend/kalix/issues/7719


docs/src/modules/spring/pages/actions.adoc
```diff
@@ -0,0 +1,79 @@
...
+include::example$spring-fibonacci-action/src/main/java/com/example/fibonacci/FibonacciAction.java[tag=implementing-action]
+----
+<1> `isFibonacci` checks if a number is a Fibonacci number
+<2> `nextFib` calculates the next number.
+<3> This `nextNumber` implementation calls the `nextNumber` implementation below.
+<4> The `nextNumber` implementation first checks if the input number belongs to the Fibonacci series. If so, it calculates the
+next number and builds a reply using `effects().reply()`.
+<5> Otherwise, if the input number doesn't belong to the Fibonacci series, it builds an `Effect` reply error.
+
+=== Multiple replies / reply streaming
```
franciscolopezsancho, line 32: please ignore this at the moment

docs/src/modules/spring/pages/actions.adoc
```diff
@@ -0,0 +1,77 @@
...
+implement an Action as a pure stateless function. We will define a `FibonacciAction` that takes a number and returns the
+next number in the Fibonacci series.
+
+== Implementing the Action
+
+To implement this action we need the following: 
+
+- Extend our class from `kalix.javasdk.action.Action`. This is generic. No matter what action we want to create we always need to extend from link:{attachmentsdir}/testkit/kalix/javasdk/action/Action.html[`Action` {tab-icon}, window="new"]. The rest is specific to this case
+- Add the Spring annotation @RequestMapping to provide a REST endpoint for the function. Here the stateless function should be reachable via HTTP.
+- Add the Spring annotations @GetMapping and @PostMapping to provide paths for GET and POST to calculate the Fibonacci of a number. Both functions do the same thing and implementation-wise the funtion exposed with GET calls the function exposed with POST.
```
octonato, line 17:
```suggestion
- Add the Spring annotations @GetMapping and @PostMapping to provide paths for GET and POST to calculate the Fibonacci of a number. Both functions do the same thing and implementation-wise the function exposed with GET calls the function exposed with POST.  
```

First time I see this bot suggesting something. Nice!

docs/src/modules/spring/pages/actions.adoc
```diff
@@ -0,0 +1,77 @@
...
+<1> `isFibonacci` checks if a number is a Fibonacci number
+<2> `nextFib` calculates the next number.
+<3> This `nextNumber` implementation calls the `nextNumber` implementation below.
+<4> The `nextNumber` implementation first checks if the input number belongs to the Fibonacci series. If so, it calculates the
+next number and builds a reply using `effects().reply()`.
+<5> Otherwise, if the input number doesn't belong to the Fibonacci series, it builds an `Effect` reply error.
+
+=== Multiple replies / reply streaming
+
+An Action may return a stream of integers. To do this you need by marking the return type as `reactor.core.publisher.Flux<Effect<Integer>>`.
```
octonato, line 34:
```suggestion
An Action may return a stream of integers. To do this you need to define the return type as `reactor.core.publisher.Flux<Effect<Integer>>`. 
```

docs/src/modules/spring/pages/actions.adoc
```diff
@@ -0,0 +1,77 @@
...
+TODO: create Main docs and add partial here
+
+== Testing the Action
+
+
+=== Unit tests
+
+The following snippet shows how the `ActionTestkit` is used to test the `FibonacciAction` implementation.
+
+With the `ActionTestkit` we can call the methods of `FibonacciAction`. Each call we execute over to the test kit returns an `ActionResult` that contains the effect produced by the underlying action method.
```
octonato, line 52:
```suggestion
With the `ActionTestkit` we can call the methods of `FibonacciAction`. Each call we pass over to the test kit returns an `ActionResult` that contains the effect produced by the underlying action method.
```

octonato (commented):
Looking good.

docs/src/modules/spring/pages/actions.adoc
```diff
@@ -0,0 +1,78 @@
+= Implementing Actions in Spring 
+:page-supergroup-java-scala: Language
```
octonato, line 2:
```suggestion
:page-supergroup-java-scala: Language
```

We should not have it. This is only useful when we have more than one language in a page. So we get the option to switch between the Java and Scala variants. For Spring, we only show Java.

I noticed that we have it also on the Event Sourced page. I will remove it on the PR that I'm preparing.

docs/src/modules/spring/pages/actions.adoc
```diff
@@ -0,0 +1,78 @@
...
+An Action may return a stream of integers. To do this you need to define the return type as `reactor.core.publisher.Flux<Effect<Integer>>`. 
+
+The stream may publish an arbitrary number of replies.
+
+ifdef::todo[TODO: add a streamed Fib series calculation]
+
+
+== Registering the Action
+
+TODO: create Main docs and add partial here
```
octonato, line 44: there is no action registration. The Spring SDK does it automatically by classpath scanning.

docs/src/modules/spring/pages/actions.adoc
franciscolopezsancho: ok

docs/src/modules/spring/pages/actions.adoc
franciscolopezsancho: Nice catch. I forgot to remove the title although I wanted to explain how to set up the Main here. How about I create a partial and we later use it in other parts of the documentation? 
Or I skip entirely this point here?

docs/src/modules/spring/pages/actions.adoc
```diff
@@ -0,0 +1,77 @@
+= Implementing Actions in Spring
```
octonato, line 1: I'm wondering what will be the best title here. 

Should we repeat `in Spring` each time? Assuming that users are already under the Spring SDK chapter? 

The event sourced one is saying "with Spring".

docs/src/modules/spring/pages/actions.adoc
efgpinto: I ended up just removing that section from the EventSourced page.. I think somewhere else we should explain that the components are auto registered...

docs/src/modules/spring/pages/actions.adoc
efgpinto: I think it helps when the user parachuts into these pages directly. And I kept it mainly to be consistent with the Java/Scala version... However, I see now that some pages have it others not, which makes it weird. Maybe we should remove it and just rely on the user to understand the chapter it is in?

franciscolopezsancho: I'm thinking what if we share the link to this page? They will have no previous navigation.  I'm fine either way. Do we have a quorum?

docs/src/modules/spring/pages/actions.adoc
franciscolopezsancho: I would rather don't explain auto registering anywhere. The user would be better off don't knowing anything about it I think. 
I thought this placeholder to explain how a `Main` is created so the user has a view of the whole application, with the entry point included.  I guess my question is how you feel about it guys? Do you think we should explain in Action, Value Entity, and Event Sourced Entity how to create the Main that starts the app? and if so and it is the same in all of them, should we create a partial for it and reuse it?

docs/src/modules/spring/pages/actions.adoc
franciscolopezsancho: I deleted it here FYI

docs/src/modules/spring/pages/actions.adoc
octonato: No, we should not explain it because users should not create it. They should start from the archetype and the archetype comes with it already correctly configured. 

I added a blank page about dependency injection. We do need to explain that components get picked by Spring Boot and that users can build components and let Spring wire dependencies into it.

docs/src/modules/spring/pages/actions.adoc
octonato: If users arrive directly on the page, they still have the Spring section selected on the left side menu. I'm ok to remove for now. 

We can gives us more time to think about it and eventually re-introduce it.

docs/src/modules/spring/pages/actions.adoc
franciscolopezsancho: ok. I'll remove

docs/src/modules/spring/pages/actions.adoc
franciscolopezsancho: 👍

docs/src/modules/spring/pages/actions.adoc
```diff
@@ -1 +1,73 @@
...
+
+Actions return effects, that is, `Action.Effect`. You have different types of effects. A reply, an error, a forward to the call to another component, and to all of those you can add side effects. Here you want only the result of the calculation or an error. Therefore you are using `.reply` and `.error`.
+
+=== Multiple replies / reply streaming
+
+An Action may return a stream of integers. To do this you need to define the return type as `reactor.core.publisher.Flux<Effect<Integer>>`. 
+
+The stream may publish an arbitrary number of replies.
+
+ifdef::todo[TODO: add a streamed Fib series calculation]
```
efgpinto, line 38: Shall we open a separate issue for this if we think it's worth? Just so we don't forget it forever?

efgpinto (approved):
Left some small suggestions but otherwise LGTM.

docs/src/modules/spring/pages/actions.adoc
```diff
@@ -1 +1,73 @@
...
+include::partial$actions.adoc[]
+
+In this first example, you will learn how to implement an Action as a pure stateless function. You will create a `FibonacciAction` that takes a number and returns the
+next number in the Fibonacci series.
+
+== Implementing the Action
+
+To implement this action you need the following:
+
+- Extend our class from `kalix.javasdk.action.Action`. This is generic. No matter what action you want to create you always need to extend from link:{attachmentsdir}/testkit/kalix/javasdk/action/Action.html[`Action` {tab-icon}, window="new"]. The rest is specific to this case
```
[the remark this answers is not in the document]
octonato: An action doesn't need to have REST annotations. You can have an Action that subscribes to the event journal and publish messages to a Topic. 

In such a case, there is now REST endpoint and the Action is only accessible internally from the Proxy.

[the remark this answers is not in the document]
octonato: But I agree that we can remove the sentence "The rest is specific to this case". 

It's not relevant to say it IMO. When we show other usages of Action, we can mention that a REST annotation is useless, for example.

docs/src/modules/spring/pages/actions.adoc
franciscolopezsancho: I'm working on this right now. I'll add it to the PR

docs/src/modules/spring/pages/actions.adoc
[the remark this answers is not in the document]
efgpinto: > An action doesn't need to have REST annotations. You can have an Action that subscribes to the event journal and publish messages to a Topic.
> 
> In such a case, there is now REST endpoint and the Action is only accessible internally from the Proxy.

Ah nice, indeed! 👍

octonato (approved):
LGTM, good to merge after addressing Eduardo's suggestions.

docs/src/modules/spring/pages/actions.adoc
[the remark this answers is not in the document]
franciscolopezsancho: I think is better to add it here. In this example so the reader knows it straight away. Instead of letting them think that this may be required. Just as @efgpinto thought it was. But as you both agreed I'll remove it.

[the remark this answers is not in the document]
efgpinto: > I think is better to add it here. In this example so the reader knows it straight away. Instead of letting them think that this may be required. Just as @efgpinto thought it was. But as you both agreed I'll remove it.

Yeah, understand why you think it's a good note. I wouldn't mind having it as a `Note:` instead maybe after ennumerating the steps saying that in this case we want the action to be exposed but if that's not the case, the first step is the only mandatory or something along those lines.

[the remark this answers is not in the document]
franciscolopezsancho: Alrighty. Let's do that in another section later.

docs/src/modules/spring/pages/actions.adoc
franciscolopezsancho: https://github.com/lightbend/kalix/issues/7787

PR #1206, opened 2022-10-25
chore: fix display for group-scala in quickstart template page

octonato:
Rendering mistake in java/scala page.

CleanShot 2022-10-25 at 15 55 57


franciscolopezsancho (approved):
Yup, that does the trick.

PR #1211, opened 2022-10-26
feat(spring-sdk): support more HTTP methods in KalixClient

efgpinto:
References https://github.com/lightbend/kalix/issues/7628

This:
- adds support for `PUT`, `PATCH` and `DELETE`

Still confirming if/how to support custom HTTP methods.

### DIFF
--- a/samples/spring-reliable-timers/docker-compose.yml
+++ b/samples/spring-reliable-timers/docker-compose.yml
@@ -2,7 +2,7 @@
 version: "3"
 services:
   kalix-proxy:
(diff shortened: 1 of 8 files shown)


efgpinto:
On custom HTTP methods: Spring MVC does not provide support for custom HTTP methods and forces you to use one of the standard methods. There are workarounds (see [this](https://www.appsloveworld.com/springmvc/100/6/custom-http-methods-in-spring-mvc) and [this](https://stackoverflow.com/a/56198556) but I'm not sure we need to consider this use case for now.. What are your thoughts?

octonato (approved):
LGTM

PR #1214, opened 2022-10-26
fix: disable ACL in dev-mode

octonato:
References https://github.com/lightbend/kalix/issues/7783

### DIFF
--- a/docs/src/modules/java/pages/access-control.adoc
+++ b/docs/src/modules/java/pages/access-control.adoc
@@ -40,7 +40,7 @@ This is the default ACL included in the quickstarts and project templates, it al
 
 Checking the ACLs are in general done for you by Kalix, however in some cases programmatic access to the principal of a call can be useful.
 
-Accessing the principal of a call inside of a service is possible through the request metadata `Metadata.principals()`. The `Metadata` for a call is available through the context (`actionContext`, `commandContext`) of the component.
+Accessing the principal of a call inside a service is possible through the request metadata `Metadata.principals()`. The `Metadata` for a call is available through the context (`actionContext`, `commandContext`) of the component.
 
 == ACLs when running unit tests
 
@@ -48,8 +48,8 @@ In the generated unit test testkits, the ACLs are ignored.
 
 == ACLs when running integration tests
 
-When running integration tests, ACLs are enabled by default but can be explicitly disabled per test by running the test with `Settings.withAclDisabled`.
+When running integration tests, ACLs are disabled by default but can be explicitly enabled per test by running the test with `Settings.withAclEnabled`.
 
 For integration tests that call other services that have ACLs limiting access to specific service names `Settings.withServiceName` allows specifying what the service identifies itself as to other services.
 
-`KalixTestKit.getGrpcClientForPrincipal` makes it possible to get a integration test client that is authenticated with specific credentials for calling a service with ACLs.
+`KalixTestKit.getGrpcClientForPrincipal` makes it possible to get an integration test client that is authenticated with specific credentials for calling a service with ACLs.
--- a/sdk/java-sdk-testkit/src/main/java/kalix/javasdk/testkit/KalixTestKit.java
+++ b/sdk/java-sdk-testkit/src/main/java/kalix/javasdk/testkit/KalixTestKit.java
@@ -71,7 +71,7 @@ public class KalixTestKit {
      */
     @Deprecated
     public Settings(final Duration stopTimeout) {
-      this(stopTimeout, "self", true);
+      this(stopTimeout, "self", false);
     }
 
     private Settings(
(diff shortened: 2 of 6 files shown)


johanandren (approved):

octonato:
Failed due to flaky test
```java
[error] Test com.example.wiring.SpringSdkWiringIntegrationTest.verifyActionIsNotSubscribedToMultiplyAndRouterIgnores failed: org.opentest4j.AssertionFailedError: expected: <1236> but was: <1237>, took 0.398s
[error]     at com.example.wiring.SpringSdkWiringIntegrationTest.verifyActionIsNotSubscribedToMultiplyAndRouterIgnores(SpringSdkWiringIntegrationTest.java:139)
[error]     at jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
[error]     at jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:77)
[error]     at jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
[error]     at java.lang.reflect.Method.invoke(Method.java:568)
[error]     ...
[info] Test com.ex
```

docs/src/modules/java/pages/access-control.adoc
```diff
@@ -40,16 +40,16 @@ This is the default ACL included in the quickstarts and project templates, it al
 
 Checking the ACLs are in general done for you by Kalix, however in some cases programmatic access to the principal of a call can be useful.
 
-Accessing the principal of a call inside of a service is possible through the request metadata `Metadata.principals()`. The `Metadata` for a call is available through the context (`actionContext`, `commandContext`) of the component.
+Accessing the principal of a call inside a service is possible through the request metadata `Metadata.principals()`. The `Metadata` for a call is available through the context (`actionContext`, `commandContext`) of the component.
```
octonato, line 43: IntelliJ grammer suggestion

sdk/java-sdk-testkit/src/main/java/kalix/javasdk/testkit/KalixTestKit.java
```diff
@@ -111,6 +111,15 @@ public Settings withServiceName(final String serviceName) {
...
     }
+
+    /**
+     * Enable ACL checking in this service.
+     *
+     * @return The updated settings.
+     */
+    public Settings withAclEnabled() {
+      return new Settings(stopTimeout, serviceName, true);
+    }
```
octonato, line 122: Adding `withAclEnabled`, but keeping the `withAclDisabled` that is now less useful since it's the default. Shall we deprecate it?

sdk/java-sdk-testkit/src/main/java/kalix/javasdk/testkit/KalixTestKit.java
```diff
@@ -71,7 +71,7 @@ public static class Settings {
      */
     @Deprecated
     public Settings(final Duration stopTimeout) {
-      this(stopTimeout, "self", true);
+      this(stopTimeout, "self", false);
```
octonato, line 74: To align with the docker-compose settings, I think this should also be disabled by default. 

All our sample tests were passing because we configured the default ACL to be `principal = INTERNET`. 

Users starting a new service and trying to write an integration tests will get blocked. Worst, there is no clear message in the logs explaining that the tests are failing because of the ACL restriction. It just doesn't work and users are left without any clue.

octonato (commented):
@johanandren, I will ask you to re-review because I changed the behavior of the integration tests testkit.

johanandren (approved):
I think changing the testkit is the right thing, but it is a potentially silent-breaking thing because a user could be integration testing verifying that a certain service _is_ accessible from the interwebs and that would now pass even if it in fact isn't. We should make sure to note this change in the release notes/and next SDK version announcement.

octonato:
> I think changing the testkit is the right thing, but it is a potentially silent-breaking thing because a user could be integration testing verifying that a certain service _is_ accessible from the interwebs and that would now pass even if it in fact isn't. We should make sure to note this change in the release notes/and next SDK version announcement.

In which places should we add such a note? GitHub release notes for sure. Any other place? 

I will do it right now to make sure we don't forget it.

efgpinto:
> > I think changing the testkit is the right thing, but it is a potentially silent-breaking thing because a user could be integration testing verifying that a certain service _is_ accessible from the interwebs and that would now pass even if it in fact isn't. We should make sure to note this change in the release notes/and next SDK version announcement.
> 
> In which places should we add such a note? GitHub release notes for sure. Any other place?
> 
> I will do it right now to make sure we don't forget it.

I think Johan means adding a note [here](https://github.com/lightbend/kalix-docs/blob/main/docs/modules/release-notes/pages/index.adoc). Although I think that will only make sense when we release a new version. (When adding that note we should probably use the tag `changed` instead of `new` so it's highlighted as a change in behavior.

PR #1217, opened 2022-10-26
chore: formatAll commands alias

octonato:
Because I'm too lazy to type `scalafmtAll;javafmtAll` over and over again.


efgpinto (approved):
LGTM

PR #1221, opened 2022-10-27
docs: Spring Boot integration page

octonato:


docs/src/modules/spring/pages/spring-boot-integration.adoc
```diff
@@ -0,0 +1,32 @@
...
+class is annotated with `@SpringBootApplication` and starts exactly the same way as in a Spring Boot application, namely
+by calling `SpringApplication.run(Main.class, args)`.
+
+For all matters, it's a Spring Boot application that runs and connects with the Kalix Proxy. As such, your application can benefit of most features available to Spring applications. However, some features are not available or are not recommended when building a Kalix application based on the Spring SDK. Those are explained in the sections below.
+
+== Dependency Injection
+
+*Actions*, *Entities* and *Views* are all Kalix components, and they get special treatment.
+
+When your application starts, it will scan the classpath for any class extending one of the Kalix components. Each time such a component is found, it's analysed in order to generate the necessary metadata that Kalix requires to be able to interact with them. The component is then registered to Kalix and made available to the Proxy through Kalix internal protocol.
```
franciscolopezsancho, line 13: Shall we add some further information or a link to explain to the users what the proxy is? Looking at the current docs I don't see a dedicated section for this.
 
I also understand that we would like to avoid the user any knowledge of the proxy if possible. What I don't know is if is it possible at the moment.

franciscolopezsancho: These ideas are obviously contradictory. I couldn't help thinking I could use an SBR.

docs/src/modules/spring/pages/spring-boot-integration.adoc
```diff
@@ -0,0 +1,32 @@
...
+
+It's important to understand that it's not possible to interact directly with a Kalix component. All interaction must happen through the Proxy. That's how Kalix manages state and distribution. That said, it's not possible to use dependency injection to inject Kalix components inside other components. Doing so will raise a runtime error.
+
+On the other hand, is it possible to inject non-Kalix classes into Kalix components. Any Spring bean defined in your application can be injected into a Kalix component. However, you must keep in mind that Kalix components can only receive other dependencies through constructor injection.
+
+
+== @Transactional
```
franciscolopezsancho, line 20: This section is so good. It gives such a great perspective of Kalix.

docs/src/modules/spring/pages/spring-boot-integration.adoc
```diff
@@ -0,0 +1,32 @@
...
+
+Kalix is a stateful serverless solution and as such it takes care of all the aspects of persisting data in a distributed environment. Using `@Transactional` annotations on methods have no effect because the database access and transactional management won't happen at the level of the Spring container. All database operations take place on the Proxy side.
+
+Moreover, and also related with the subject of transactions, each entity in Kalix lives in its own transactional boundary. Therefore, it's not possible to send two commands to two different entities and have the mutations be applied transitionally.
+
+Kalix is a distributed runtime and two different entities may be located on different JVMs, on different nodes, running on different hardware, making it, if not impossible, extremely expensive and inefficient to mutate two different entities on the same transaction.
+
+== Spring Repositories
+
+Spring developers are used to build Spring Data Repositories to persist and retrieve data. However, in Kalix this tasks is fulfilled by the Proxy, not by the Spring container. Therefore, the usage of Spring Data Repositories is not recommended.
```
franciscolopezsancho, line 30: Could you elaborate some more on this? I'm learning here and this seems a crucial point. 
A user should not use a repository because the proxy persists and retrieves data for the components? 
Maybe adding an example and showing what would be the downsides could help?

docs/src/modules/spring/pages/spring-boot-integration.adoc
octonato: We don't hide the fact that there is a proxy and the docs talk about it in a few places.

https://docs.kalix.io/reference/glossary.html
https://docs.kalix.io/services/state-model.html
https://docs.kalix.io/services/programming-model.html

Maybe you have only looked at the docs in the jvm project? 

In any case, the goal is not to explain all the internals of the proxy. That would be too much. But it's important that people have a high-level architectural view. A minimal understanding of how Kalix manages state and distributions is unavoidable.

docs/src/modules/spring/pages/spring-boot-integration.adoc
octonato: I will see how I can elaborate it more. Need to think more. We can discuss it next week.

Thanks for all your input.

docs/src/modules/spring/pages/spring-boot-integration.adoc
```diff
@@ -0,0 +1,32 @@
+= Spring Boot Integration
+
+The Kalix Spring SDK provides a familiar developer experience for Java developers that are used to Spring. Its `Main`
+class is annotated with `@SpringBootApplication` and starts exactly the same way as in a Spring Boot application, namely
```
efgpinto, line 4: Both seem correct, below one sounds better to me but might be just a matter of taste, so feel free to ignore.

```suggestion
class is annotated with `@SpringBootApplication` and starts exactly the same way as a common Spring Boot application, namely
```

docs/src/modules/spring/pages/spring-boot-integration.adoc
```diff
@@ -0,0 +1,32 @@
...
+by calling `SpringApplication.run(Main.class, args)`.
+
+For all matters, it's a Spring Boot application that runs and connects with the Kalix Proxy. As such, your application can benefit of most features available to Spring applications. However, some features are not available or are not recommended when building a Kalix application based on the Spring SDK. Those are explained in the sections below.
```
efgpinto, line 7: Maybe link "Kalix Proxy" to https://docs.kalix.io/reference/glossary.html#proxy ? Might be useful for this first reference.

docs/src/modules/spring/pages/spring-boot-integration.adoc
```diff
@@ -0,0 +1,32 @@
...
+
+== Dependency Injection
+
+*Actions*, *Entities* and *Views* are all Kalix components, and they get special treatment.
+
+When your application starts, it will scan the classpath for any class extending one of the Kalix components. Each time such a component is found, it's analysed in order to generate the necessary metadata that Kalix requires to be able to interact with them. The component is then registered to Kalix and made available to the Proxy through Kalix internal protocol.
+
+It's important to understand that it's not possible to interact directly with a Kalix component. All interaction must happen through the Proxy. That's how Kalix manages state and distribution. That said, it's not possible to use dependency injection to inject Kalix components inside other components. Doing so will raise a runtime error.
```
efgpinto, line 15: Some re-wording.. see what you think

```suggestion
It's important to understand that interacting directly with a Kalix component it's NOT possible. All interactions must happen through the Proxy since that's how Kalix manages state and distribution. Therefore, it's not possible to use dependency injection to inject Kalix components inside other components and doing so will raise a runtime error.
```

docs/src/modules/spring/pages/spring-boot-integration.adoc
```diff
@@ -0,0 +1,32 @@
...
+
+On the other hand, is it possible to inject non-Kalix classes into Kalix components. Any Spring bean defined in your application can be injected into a Kalix component. However, you must keep in mind that Kalix components can only receive other dependencies through constructor injection.
```
efgpinto, line 17:
```suggestion
On the other hand, it is possible to inject non-Kalix classes into Kalix components. Any Spring bean defined in your application can be injected into a Kalix component. However, you must keep in mind that Kalix components can only receive other dependencies through constructor injection.
```

docs/src/modules/spring/pages/spring-boot-integration.adoc
```diff
@@ -0,0 +1,32 @@
...
+
+
+== @Transactional
+
+Kalix is a stateful serverless solution and as such it takes care of all the aspects of persisting data in a distributed environment. Using `@Transactional` annotations on methods have no effect because the database access and transactional management won't happen at the level of the Spring container. All database operations take place on the Proxy side.
+
+Moreover, and also related with the subject of transactions, each entity in Kalix lives in its own transactional boundary. Therefore, it's not possible to send two commands to two different entities and have be applied in the same transaction.
```
efgpinto, line 24:
```suggestion
Moreover, and also related with the subject of transactions, each entity in Kalix lives in its own transactional boundary. Therefore, it's not possible to send two commands to two different entities and have them be applied in the same transaction.
```

docs/src/modules/spring/pages/spring-boot-integration.adoc
```diff
@@ -0,0 +1,32 @@
...
+
+Kalix is a distributed runtime and two different entities may be located on different JVMs, on different nodes, running on different hardware, making it, if not impossible, extremely expensive and inefficient to mutate two different entities on the same transaction.
+
+== Spring Repositories
+
+Spring developers are used to build Spring Data Repositories to persist and retrieve data. However, in Kalix this tasks is fulfilled by the Proxy, not by the Spring container. Therefore, the usage of Spring Data Repositories is not recommended.
+
+The equivalent of Spring repositories in Kalix are xref:spring:views.adoc[Views]. A View allows you to listen to events from Event Sourced entities or state updates from Value entities and build an indexed representation from them.
```
efgpinto, line 32: Not sure I would use "equivalent" here in the sense that the two concepts are very different.. Maybe "one way to achieve similar functionality"?

docs/src/modules/spring/pages/spring-boot-integration.adoc
efgpinto: If the phrase remains after you revisit this..

```suggestion
Spring developers are used to build Spring Data Repositories to persist and retrieve data. However, in Kalix these tasks are fulfilled by the Proxy, not by the Spring container. Therefore, the usage of Spring Data Repositories is not recommended.
```

efgpinto (commented):
Very useful section for a Spring dev I believe.. Left some grammar notes, otherwise looks good.

docs/src/modules/spring/pages/spring-boot-integration.adoc
franciscolopezsancho: I missed it, yeah. Actually, I tried to look up 'proxy' on the search box and as I see now the 'programming model' page appeared quite far in the list. It makes sense, yeah. The users will be better off if they understand the architecture. 

I really like https://docs.kalix.io/services/programming-model.html. Maybe is worth adding a link to that?

docs/src/modules/spring/pages/spring-boot-integration.adoc
franciscolopezsancho: Sure, let's discuss it next week. I think I'll learn a thing or two.

docs/src/modules/spring/pages/spring-boot-integration.adoc
octonato: Will do

docs/src/modules/spring/pages/spring-boot-integration.adoc
octonato: Thanks, good idea. 

I will the link.

efgpinto (approved):
LGTM

efgpinto:
I think all feedback was addressed here, right? Shall we merge?

PR #1222, opened 2022-10-27
feat: adds support for entity key generator

octonato:
References https://github.com/lightbend/kalix/issues/7790

### DIFF
--- a/samples/spring-valueentity-counter/src/main/java/com/example/CounterEntity.java
+++ b/samples/spring-valueentity-counter/src/main/java/com/example/CounterEntity.java
@@ -8,12 +8,12 @@ import org.springframework.web.bind.annotation.PostMapping;
 import org.springframework.web.bind.annotation.RequestBody;
 
 @EntityType("counter")                                     // <1>
-@EntityKey("counter_id")                                   // <2>
-public class CounterEntity extends ValueEntity<Integer> {  // <3>
+public class CounterEntity extends ValueEntity<Integer> {  // <2>
 
   @Override
-  public Integer emptyState() { return 0; }                  // <4>
+  public Integer emptyState() { return 0; }                  // <3>
 
+  @EntityKey("counter_id")                                   // <4>
   @PostMapping("/counter/{counter_id}/increase")             // <5>
   public Effect<Number> increaseBy(@RequestBody Number increaseBy) {
     int newCounter = currentState() + increaseBy.value();    // <6>
--- a/sdk/spring-sdk/src/main/java/kalix/springsdk/annotations/GenerateEntityKey.java
+++ b/sdk/spring-sdk/src/main/java/kalix/springsdk/annotations/GenerateEntityKey.java
@@ -1,0 +1,45 @@
+/*
+ * Copyright 2021 Lightbend Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package kalix.springsdk.annotations;
+
+import java.lang.annotation.*;
+
+/**
+ * Indicates that Kalix should generate an entity key when this method is invoked.
+
+ * The generated entity key will be accessible via the entity context supplied in the SDK.
+ * A method annotated with this annotations should not be annotated with {@link EntityKey},
+ * if it does, an error will be raised.
+ *
+ * The generated key will be a Version 4 (random) UUID. The UUID will be generated using a cryptographically secure
+ * random number generator.
+ */
+@Target(ElementType.METHOD)
+@Retention(RetentionPolicy.RUNTIME)
+@Documented
+public @interface GenerateEntityKey {
+
+  Type value() default Type.VERSION_4_UUID;
+
+  enum Type {
+    /**
+     * Generate a Version 4 (random) UUID. The UUID will be generated using a cryptographically secure random
+     * number generator.
+     */
+    VERSION_4_UUID
+  }
+}
--- a/sdk/spring-sdk/src/main/java/kalix/springsdk/annotations/GenerateEntityKey.java
+++ b/sdk/spring-sdk/src/main/java/kalix/springsdk/annotations/GenerateEntityKey.java
@@ -21,8 +21,8 @@ import java.lang.annotation.*;
 /**
  * Indicates that Kalix should generate an entity key when this method is invoked.
 
- * The generated entity key will be accessible via the entity context supplied in the SDK.
- * A method annotated with this annotations should not be annotated with {@link EntityKey},
+ * The generated entity key is accessible via the entity context supplied in the SDK.
+ * A method annotated with this annotation should not be annotated with {@link EntityKey},
  * if it does, an error will be raised.
  *
  * The generated key will be a Version 4 (random) UUID. The UUID will be generated using a cryptographically secure
--- a/sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/EntityDescriptorFactory.scala
+++ b/sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/EntityDescriptorFactory.scala
@@ -67,7 +67,7 @@ private[impl] object EntityDescriptorFactory extends ComponentDescriptorFactory 
                 restMethod.javaMethod,
                 "Invalid command method. No @EntityKey nor @GenerateEntityKey annotations found. " +
                 "A command method should be annotated with either @EntityKey or @GenerateEntityKey, or " +
-                "the an @EntityKey annotation should be present at class level.")
+                "an @EntityKey annotation should be present at class level.")
 
             KalixMethod(restMethod, entityKeys = entityKeysToUse)
           }
--- a/sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/EntityDescriptorFactory.scala
+++ b/sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/EntityDescriptorFactory.scala
@@ -61,6 +61,14 @@ private[impl] object EntityDescriptorFactory extends ComponentDescriptorFactory 
             val entityKeysToUse =
               if (entityKeyOnMethod != null) entityKeyOnMethod.value()
               else entityKeysOnType
+
+            if (entityKeysToUse.isEmpty)
+              throw ServiceIntrospectionException(
+                restMethod.javaMethod,
+                "Invalid command method. No @EntityKey nor @GenerateEntityKey annotations found. " +
+                "A command method should be annotated with either @EntityKey or @GenerateEntityKey, or " +
+                "the an @EntityKey annotation should be present at class level.")
+
             KalixMethod(restMethod, entityKeys = entityKeysToUse)
           }
 
--- a/sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/EntityDescriptorFactory.scala
+++ b/sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/EntityDescriptorFactory.scala
@@ -16,11 +16,14 @@
 
 package kalix.springsdk.impl
 
+import kalix.EntityMethodOptions.Generator
 import kalix.springsdk.annotations.EntityKey
+import kalix.springsdk.annotations.GenerateEntityKey
 import kalix.springsdk.impl.ComponentDescriptorFactory.buildJWTOptions
 import kalix.springsdk.impl.reflection.KalixMethod
 import kalix.springsdk.impl.reflection.NameGenerator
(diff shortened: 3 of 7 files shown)


samples/spring-valueentity-counter/src/main/java/com/example/CounterEntity.java
```diff
@@ -8,12 +8,12 @@
...
 @EntityType("counter")                                     // <1>
-@EntityKey("counter_id")                                   // <2>
-public class CounterEntity extends ValueEntity<Integer> {  // <3>
+public class CounterEntity extends ValueEntity<Integer> {  // <2>
 
   @Override
-  public Integer emptyState() { return 0; }                  // <4>
+  public Integer emptyState() { return 0; }                  // <3>
 
+  @EntityKey("counter_id")                                   // <4>
```
octonato, line 16: Now that EntityKey can be on a method, it makes it easier to put it next to the PostMapping. At least for the getting started guide where we don't want to be too verbose.

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/EntityDescriptorFactory.scala
```diff
@@ -16,23 +16,55 @@
...
-        KalixMethod(restMethod, entityKeys = entityKeys).withKalixOptions(buildJWTOptions(restMethod.javaMethod))
+
+        val entityKeyOnMethod = restMethod.javaMethod.getAnnotation(classOf[EntityKey])
+        val generateEntityKey = restMethod.javaMethod.getAnnotation(classOf[GenerateEntityKey])
+
+        if (entityKeyOnMethod != null && generateEntityKey != null)
+          throw ServiceIntrospectionException(
+            restMethod.javaMethod,
+            "Invalid annotation usage. Found both @EntityKey and @GenerateEntityKey annotations. " +
+            "A method can only be annotated with of them, but not both.")
```
octonato, line 51:
```suggestion
            "A method can only be annotated with one of them, but not both.")
```

sdk/spring-sdk/src/main/java/kalix/springsdk/annotations/GenerateEntityKey.java
```diff
@@ -0,0 +1,45 @@
...
+ * A method annotated with this annotations should not be annotated with {@link EntityKey},
+ * if it does, an error will be raised.
+ *
+ * The generated key will be a Version 4 (random) UUID. The UUID will be generated using a cryptographically secure
+ * random number generator.
+ */
+@Target(ElementType.METHOD)
+@Retention(RetentionPolicy.RUNTIME)
+@Documented
+public @interface GenerateEntityKey {
```
octonato, line 34: The documentation for this quite hidden in the gRPC SDK, I think we should add it on the entities pages. 

I'm planning to do it on follow up PR and then also cover Value Entities.

sdk/spring-sdk/src/test/java/kalix/springsdk/testmodels/eventsourcedentity/EventSourcedEntitiesTestModels.java
```diff
@@ -75,6 +75,46 @@ private Integer privateMethodSimilarSignature(Integer event) {
...
+    @PutMapping("/eventsourced/int/{number}")
+    public Integer getInteger(@PathVariable Integer number) {
+      return number;
+    }
+  }
+
+  @EntityType("counter")
+  public static class IllDefinedEntityWithEntityKeyGeneratorAndEntityKey extends EventSourcedEntity<Integer> {
+    @GenerateEntityKey
+    @EntityKey("id")
```
octonato, line 111: this should be illegal, method can't have both

sdk/spring-sdk/src/test/java/kalix/springsdk/testmodels/eventsourcedentity/EventSourcedEntitiesTestModels.java
```diff
@@ -75,6 +75,46 @@ private Integer privateMethodSimilarSignature(Integer event) {
...
+    @GetMapping("/eventsourced/{counter_id}/int/{number}")
+    public Integer getInteger(@PathVariable Integer number) {
+      return number;
+    }
+  }
+
+  @EntityType("counter")
+  public static class CounterEventSourcedEntityWithEntityKeyGenerator extends EventSourcedEntity<Integer> {
+    @GenerateEntityKey
+    @PutMapping("/eventsourced/int/{number}")
```
octonato, line 102: No EntityKey! Get it generated.

sdk/spring-sdk/src/test/java/kalix/springsdk/testmodels/eventsourcedentity/EventSourcedEntitiesTestModels.java
```diff
@@ -75,6 +75,46 @@ private Integer privateMethodSimilarSignature(Integer event) {
...
+    public Integer getInteger(@PathVariable Integer number) {
+      return number;
+    }
+  }
+
+  @EntityKey("id")
+  @EntityType("counter")
+  public static class CounterEventSourcedEntityWithEntityKeyMethodOverride extends EventSourcedEntity<Integer> {
+
+    @EntityKey("counter_id")
```
octonato, line 92: Question is, what should we do when we have EntityKey on type level and on method level? 

I think a sensible default it to let pick the method one and override the type one. But open for other suggestions.

sdk/spring-sdk/src/main/java/kalix/springsdk/annotations/GenerateEntityKey.java
```diff
@@ -0,0 +1,45 @@
...
+
+  Type value() default Type.VERSION_4_UUID;
+
+  enum Type {
+    /**
+     * Generate a Version 4 (random) UUID. The UUID will be generated using a cryptographically secure random
+     * number generator.
+     */
+    VERSION_4_UUID
```
octonato, line 43: In gRPC, we need to have a value. In Java, we could just have an annotation. 

Not sure if we should keep this. Looking for ideas / feedback.

samples/spring-valueentity-counter/src/main/java/com/example/CounterEntity.java
franciscolopezsancho: I like this. It separates, even more, the idea of 'service' from the 'entity'.

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/EntityDescriptorFactory.scala
```diff
@@ -16,23 +16,55 @@
...
-    val entityKeys = component.getAnnotation(classOf[EntityKey]).value()
+
+    val entityKeysOnType = {
+      val anno = component.getAnnotation(classOf[EntityKey])
+      if (anno != null) anno.value()
+      else Array.empty[String]
+    }
 
     val kalixMethods =
       RestServiceIntrospector.inspectService(component).methods.map { restMethod =>
```
franciscolopezsancho, line 42: If the user adds `EntityKey` or `GeneratedEntityKey` to more than one method the last method wins?
If this is the case, Would be better to add validation so only one method can have one annotation or the other?

sdk/spring-sdk/src/test/java/kalix/springsdk/testmodels/eventsourcedentity/EventSourcedEntitiesTestModels.java
franciscolopezsancho: Another possibility I can think of would be raising an error, not allowing two EntityKey annotations.

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/EntityDescriptorFactory.scala
franciscolopezsancho: oh sorry. I misunderstood. Never mind. There will be one annotation on the type or one per method with GET/POST. Although now this makes me think. Would make sense to add validation that there is at least one instance of 'EntityKey' or 'GeneratedEntityKey' that the rest method can use?

sdk/spring-sdk/src/test/java/kalix/springsdk/testmodels/eventsourcedentity/EventSourcedEntitiesTestModels.java
franciscolopezsancho: It makes sense to me, yup.

sdk/spring-sdk/src/main/java/kalix/springsdk/annotations/GenerateEntityKey.java
```diff
@@ -0,0 +1,45 @@
...
+
+import java.lang.annotation.*;
+
+/**
+ * Indicates that Kalix should generate an entity key when this method is invoked.
+
+ * The generated entity key will be accessible via the entity context supplied in the SDK.
+ * A method annotated with this annotations should not be annotated with {@link EntityKey},
+ * if it does, an error will be raised.
+ *
```
franciscolopezsancho, line 27:
```suggestion
 * The generated entity key is accessible via the entity context supplied in the SDK.
 * A method annotated with this annotation should not be annotated with {@link EntityKey},
 * if it does, an error will be raised.
 *
```

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/EntityDescriptorFactory.scala
```diff
@@ -16,23 +16,55 @@
...
+          if (generateEntityKey != null) {
+            val entityOptions = kalix.EntityMethodOptions.newBuilder().setKeyGenerator(Generator.VERSION_4_UUID)
+            val methodOpts = kalix.MethodOptions.newBuilder().setEntity(entityOptions)
+            KalixMethod(restMethod).withKalixOptions(methodOpts.build())
+
+          } else {
+            // keys defined on Method level get precedence
+            val entityKeysToUse =
+              if (entityKeyOnMethod != null) entityKeyOnMethod.value()
+              else entityKeysOnType
```
octonato, line 63: @franciscolopezsancho, concerning you previous comment. Here we can indeed validate it once more. If `entityKeysToUse` end up being empty that's an invalid method. 

I don't know if that's already validated down the road, but in any case, better to have the validation here. 

I will add it.

sdk/spring-sdk/src/main/java/kalix/springsdk/annotations/GenerateEntityKey.java
franciscolopezsancho: would you like to add the reference to an issue before closing this PR?

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/EntityDescriptorFactory.scala
```diff
@@ -16,23 +16,63 @@
...
+
+            if (entityKeysToUse.isEmpty)
+              throw ServiceIntrospectionException(
+                restMethod.javaMethod,
+                "Invalid command method. No @EntityKey nor @GenerateEntityKey annotations found. " +
+                "A command method should be annotated with either @EntityKey or @GenerateEntityKey, or " +
+                "the an @EntityKey annotation should be present at class level.")
```
franciscolopezsancho, line 70:
```suggestion
                "an @EntityKey annotation should be present at class level.")
```

sdk/spring-sdk/src/main/java/kalix/springsdk/annotations/GenerateEntityKey.java
franciscolopezsancho: I'm not sure I understand the problem.
`In gRPC, we need to have a value. In Java, we could just have an annotation.` I can't see any other elegant way to provide the users the ability to express that they want an auto-generated key. Another option would be adding a boolean property in the `Entity`? Such as `generateKey `?  But this doesn't match the current model of annotations in the SDK. Maybe if we were creating things with a [builder pattern](https://www.digitalocean.com/community/tutorials/builder-design-pattern-in-java) like in the java-SDK that would make sense. 
So all of that is to say that I would keep this implementation.

On top of that from what I understand this class `value` is never used and that is why you set it to a fixed value and with an enum with only one option. To inform the user this is it. There are no other options here. This makes sense to me. I quite like it.

franciscolopezsancho (approved):
LGTM

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/EntityDescriptorFactory.scala
franciscolopezsancho: I can see you did

sdk/spring-sdk/src/main/java/kalix/springsdk/annotations/GenerateEntityKey.java
octonato: A gRPC annotation can't stands on its own. It needs to have an assigned value. Even though we only have one type, UUID V4, we need to write `option (kalix.method).entity.key_generator = VERSION_4_UUID`. 

The Java annotation could have been just `@GenerateEntityKey` without any value.
Basically, the enum I introduced is not needed and is there more to align with the gRPC counterpart. That's all.

efgpinto:
@octonato did you consider having a property of `@EntityKey` instead of a new annotation (as mentioned by Cisco above)? I feel like this concept is so close to an EntityKey, that I was wondering if we really should introduce a new annotation for it.. Just concerned that we start having too fine grained Annotations.

PR #1225, opened 2022-10-29
docs(spring-sdk): add docs for timers

efgpinto:
Adding docs for timers. refs https://github.com/lightbend/kalix/issues/7721

### DIFF
--- a/docs/src/modules/java/pages/timers.adoc
+++ b/docs/src/modules/java/pages/timers.adoc
@@ -5,7 +5,7 @@ include::ROOT:partial$timers-intro.adoc[]
 
 To demonstrate its functionality, let's consider an Ordering Service composed of a xref:value-entity.adoc[Value Entity] and an xref:actions.adoc[Action]. The Action will work as a xref:actions-as-controller.adoc[Controller] receiving calls and delegating to the Order Entity.
 
-Users can place an order, but the order must be confirmed within a period of time. You can compare it with an Ordering Food application where the restaurant needs to confirm if it can accept the order. If no confirmation is sent within some pre-defined period of time, the order is automatically cancelled.
+Users can place an order, but the order must be confirmed within a period of time. You can think of it as an Ordering Food application where the restaurant needs to confirm if it can accept the order. If no confirmation is sent within some pre-defined period of time, the order is automatically cancelled.
 
 == Order Entity
 Let's have a look on how the Order Entity can be implemented.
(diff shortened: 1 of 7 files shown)


docs/src/modules/java/pages/timers.adoc
```diff
@@ -75,13 +55,13 @@ Java::
 ----
 include::example$java-reliable-timers/src/main/java/com/example/actions/OrderAction.java[tags=timers;place-order]
 ----
-<1> First we generate a random identifier for the Order. We will use it for identifying the Order, but also as unique name for the timer.
+<1> First we generate a random identifier for the Order. We will use it for identifying the Order, but also as an unique name for the timer.
```
franciscolopezsancho, line 58:
```suggestion
<1> First we generate a random identifier for the Order. We will use it for identifying the Order, but also as a unique name for the timer.
```

docs/src/modules/java/pages/timers.adoc
```diff
@@ -1,31 +1,11 @@
...
-
-**Timer limitations**: 
-
-* Timers can only be scheduled from within an Action.
-* Timers can only call other components (e.g., Actions and Entities), therefore external service calls must be wrapped by an Action in the deployed service.
-
-To demonstrate its functionality, let's consider an Ordering Service composed of a xref:value-entity.adoc[Value Entity] and an xref:actions.adoc[Action]. The Action will work as a xref:actions-as-controller.adoc[Controller] receiving calls and delegating to the Order Entity. 
-
-Users can place an order, but the order must be confirmed within a period of time. You can compare it with an Ordering Food application where the restaurant needs to confirm if it can accept the order. If no confirmation is sent within some pre-defined period of time, the order is automatically cancelled. 
+Users can place an order, but the order must be confirmed within a period of time. You can compare it with an Ordering Food application where the restaurant needs to confirm if it can accept the order. If no confirmation is sent within some pre-defined period of time, the order is automatically cancelled.
```
franciscolopezsancho, line 8: I'm not sure how to rephrase it but `compare it with an Ordering Food application` seems not quite right. Especially the use of `compare`.  Maybe it would be better something like `think of it as part of an Ordering Food application`?

docs/src/modules/spring/pages/timers.adoc
```diff
@@ -1 +1,82 @@
...
+
+== Order Entity
+Let's have a look on how the Order Entity can be implemented.
+
+.src/main/java/com/example/domain/OrderEntity.java
+[source,java]
+----
+include::example$spring-reliable-timers/src/main/java/com/example/domain/OrderEntity.java[tag=order]
+----
+<1> The first method to look at is the `placeOrder`. It's basically the creation of an order. Note that we set the `placed` field to true.
```
franciscolopezsancho, line 17: I would divide this point in two. One pointing to line 28 on OrderEntity with `The first method to look at is the `placeOrder`. It's basically the creation of an order` and the other in line 33 on OrderEntity with `Note that we set the "placed" field to true` maybe further elaborating why setting `placed` to true is important to note?

franciscolopezsancho:
Looking good. Don't mind the couple of comments I made if don't make much sense to you, please. One thing I noted is that we are using `we` in the doc most of the time. Shall we change it to `you`?

docs/src/modules/spring/pages/timers.adoc
octonato: Also good to change it in java/scala docs.

octonato (approved):
LGTM, after the minor suggestions applied

docs/src/modules/java/pages/timers.adoc
efgpinto: agreed, changing

docs/src/modules/spring/pages/timers.adoc
efgpinto: changed now in both. tks

octonato (approved):
LGTM

Shall I merge it?

PR #1223, opened 2022-10-27
feat: support for stream updates in Views

octonato:
References https://github.com/lightbend/kalix/issues/7696

PR #1226, opened 2022-10-31
7718 spring docs implementing views

franciscolopezsancho:
References https://github.com/lightbend/kalix/issues/7718


franciscolopezsancho:
It's not yet finished but I think it's ready for review.

docs/src/modules/spring/pages/views.adoc
```diff
@@ -1 +1,472 @@
...
+
+
+=== Define the View
+
+To get a View of multiple customers by their name, you need the set the following:
+
+ - Define the name of the table with `@Table`.
+ - Subscribe to the Value Entity with `@Subscribe.ValueEntity`.
+ - Extend from the class `View`.
+ - Define an endpoint for the view with `@GetMapping`. Is this optional? Can the view be called by a component?
```
octonato, line 47: No, it's not optional. The goal of a View is to expose some representation of the model. The REST annotation is mandatory. 

If user wants to block it they need to add a restrictive ACL to this method.

Actions can call Views.

docs/src/modules/spring/pages/views.adoc
```diff
@@ -1 +1,472 @@
...
+[source,java]
+----
+include::example$spring-customer-registry-views-quickstart/src/main/java/customer/view/CustomerByNameView.java[tags=class]
+----
+<1> defining table name
+<2> subscribing to `CustomerEntity`
+<3> extending from `View`
+<4> defining endpoint
+<5> defining the query
+<6> note that no return is needed
```
octonato, line 60: We have more or less the same listing just above the code snippet. We can move the bullet point to here.

docs/src/modules/spring/pages/views.adoc
```diff
@@ -1 +1,472 @@
...
+
+IMPORTANT: Note that the return value of the method is `null`. You may ask yourself,  how is it that the endpoint respond with any `Customer` at all? The fact is that this customer object is created in the proxy and not in the View. When you call this endpoint, the view calls the proxy, which in turn calls the database. When the proxy receives the response, it sends it directly to you without any further intervention from the View. Therefore, null is valid as a return value in the endpoint of a View. The choice of null is a way to signal that the response doesn't come from the return of this method. 
+
+TODO: ? how do we know there is only one Customer?
```
octonato, line 64: We don't. That's specific to the application. It will only return one customer if there is a unique match, typically when search on a unique key. 

Actually, not a very realistic use case on a system that only support one unique key (only EntityKey is guaranteed to be unique).

I think we should return a `Flux<Customer>`. That's a more sensible default of a query.

Note that the gRPC version is returning a `stream CustomerState`. So here we should return a `Flux<Cutomer>`

docs/src/modules/spring/pages/views.adoc
```diff
@@ -1 +1,472 @@
...
+<4> defining endpoint
+<5> defining the query
+<6> note that no return is needed
+
+IMPORTANT: Note that the return value of the method is `null`. You may ask yourself,  how is it that the endpoint respond with any `Customer` at all? The fact is that this customer object is created in the proxy and not in the View. When you call this endpoint, the view calls the proxy, which in turn calls the database. When the proxy receives the response, it sends it directly to you without any further intervention from the View. Therefore, null is valid as a return value in the endpoint of a View. The choice of null is a way to signal that the response doesn't come from the return of this method. 
+
+TODO: ? how do we know there is only one Customer? 
+
+
+TODO: ? When no result is found, the request fails with gRPC status code `NOT_FOUND`. A streamed call completes with an empty stream when no result is found.
```
octonato, line 67: Yeah, probably.

docs/src/modules/spring/pages/views.adoc
```diff
@@ -1 +1,472 @@
...
+
+
+[#event-sourced-entity]
+== Creating a View from an Event Sourced Entity
+
+Create a View from an Event Sourced Entity by using events that the Entity emits to build a state representation. Using a Customer Registry service example, to create a View for querying customers by name:
+
+This example assumes a Customer equal to the previous example. This example assumes an Event Sourced Entity that uses this Customer and it is in charge of producing the events that update the View. These events are defined as subtypes of the class `CustomerEvent` following standard https://github.com/FasterXML/jackson-annotations#handling-polymorphic-types[Jackson notation] like this:
```
octonato, line 75: some repetition in this paragraph

docs/src/modules/spring/pages/views.adoc
```diff
@@ -1 +1,472 @@
...
+
+.src/main/java/customer/api/CustomerEvent.java
+[source,java]
+----
+include::example$spring-eventsourced-customer-registry/src/main/java/customer/api/CustomerEvent.java[tags=class]
+----
+<1> Includes the logical type name (defined in implementation classes) as type 
+<2> Required for deserialization and to define the logical type name 
+
+Adding `@JsonTypeInfo` is needed so that Jackson can read back correctly the Object type when deserializing. By using `name`, you set its value as the fully qualified message name instead of the using the fully qualified Java class. This is called a `logical type name` and it allows you refactor the class to another package after you have serialized messages of this type. Otherwise, after such a refactor the objects serialized with the old package won't be recognized as the same type.
```
octonato, line 85:
```suggestion
Adding `@JsonTypeInfo` is needed so that Jackson can read back correctly the Object type when deserializing. By using `name`, you set its value as the fully qualified message name instead of the using the fully qualified Java class. This is called a `logical type name` and it allows you to refactor the class to another package after you have serialized messages of this type. Otherwise, after such a refactor the objects serialized with the old package won't be recognized as the same type. 
```

docs/src/modules/spring/pages/views.adoc
```diff
@@ -1 +1,472 @@
...
+<2> Required for deserialization and to define the logical type name 
+
+Adding `@JsonTypeInfo` is needed so that Jackson can read back correctly the Object type when deserializing. By using `name`, you set its value as the fully qualified message name instead of the using the fully qualified Java class. This is called a `logical type name` and it allows you refactor the class to another package after you have serialized messages of this type. Otherwise, after such a refactor the objects serialized with the old package won't be recognized as the same type. 
+
+
+=== Define the View to consume events
+
+The definition of the view for an Event Sourced Entity is the same as for a Value Entity. However, in this example, the subscription is at the method level rather than the type level. The advantage of this approach is that you can create multiple methods to handle different events.  
+
+Every time an event is processed by the view, the state of the view can be updated. You can do this with the `.updateState` method, which is available through the `effects()` of the View. Here you can see how the view is updated with a new name:
```
octonato, line 92:
```suggestion
Every time an event is processed by the view, the state of the view can be updated. You can do this with the `.updateState` method, which is available through the `effects()` API. Here you can see how the View is updated with a new name: 
```

We are not being consistent when naming the component. Sometimes `View`, other times `view`.

docs/src/modules/spring/pages/views.adoc
```diff
@@ -1 +1,472 @@
...
+ TODO: Ask Renato as he just added it to the Spring SDK
+
+This will first list the complete result for the query and then keep the response stream open, emitting new or updated
+entries matching the query as they are added to the view. The stream does not complete until the client closes it.
+
+Note: This is not intended as transport for service to service propagation of updates and does not guarantee delivery, for such use cases you
+should instead publish events to a topic, see xref:actions-publishing-subscribing.adoc[Publishing and Subscribing with Actions]
+
+[#query]
+== Query syntax reference
```
octonato, line 192: we should make a partial for the query syntax reference

docs/src/modules/spring/pages/views.adoc
```diff
@@ -1 +1,472 @@
...
+* <<#query>>
+
+IMPORTANT: Be aware that Views are not updated immediately when Entity state changes. Kalix does update Views as quickly as possible, but it is not instant and can take up to a few seconds for the changes to become visible in the query results. View updates might also take more time during failure scenarios than during normal operation.
+
+[#value-entity]
+== Creating a View from a Value Entity
+
+Consider an example of a Customer Registry service with a `customer` Value Entity. When `customer` state changes, the entire state is emitted as a value change. Value changes update any associated Views. To create a View that lists customers by their name:
+
+* <<_define_the_view>> for a service that selects customers by name and associates a table name with the View. The table is created and used by Kalix to store the View.
```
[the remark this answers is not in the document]
franciscolopezsancho: yes. Deleted

docs/src/modules/spring/pages/views.adoc
franciscolopezsancho: thanks for the clarification. Deleted

docs/src/modules/spring/pages/views.adoc
franciscolopezsancho: I'd like the reader first to read the bullet points straight. How about I explain less on the bullet points? Let me know if you like it more now.

docs/src/modules/spring/pages/views.adoc
```diff
@@ -1 +1,472 @@
...
+include::example$spring-customer-registry-views-quickstart/src/main/java/customer/view/CustomerByNameView.java[tags=class]
+----
+<1> defining table name
+<2> subscribing to `CustomerEntity`
+<3> extending from `View`
+<4> defining endpoint
+<5> defining the query
+<6> note that no return is needed
+
+IMPORTANT: Note that the return value of the method is `null`. You may ask yourself,  how is it that the endpoint respond with any `Customer` at all? The fact is that this customer object is created in the proxy and not in the View. When you call this endpoint, the view calls the proxy, which in turn calls the database. When the proxy receives the response, it sends it directly to you without any further intervention from the View. Therefore, null is valid as a return value in the endpoint of a View. The choice of null is a way to signal that the response doesn't come from the return of this method.
```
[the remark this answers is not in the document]
franciscolopezsancho: :) great. Writing these docs is helpful. 
Changed now.

docs/src/modules/spring/pages/views.adoc
franciscolopezsancho: done

docs/src/modules/spring/pages/views.adoc
franciscolopezsancho: removed

docs/src/modules/java/pages/views.adoc
```diff
@@ -262,7 +262,7 @@ NOTE: This type of update transformation is a natural fit for Events emitted by
 
 == Ignoring events
 
-When consuming events, each event must be matched by a Protobuf service method. In case your View is only interested in certain events, you may declare a method to receive all events that are not received by the other methods. If an event type is not handled, the Action will fail. Views are designed to restart, but since the handler is missing, it will fail again. Therefore, it's important to define methods for all events or define a catch-all method in case you want to discard some events.
+When consuming events, each event must be matched by a Protobuf service method. In case your View is only interested in certain events, you may declare a method to receive all events that are not received by the other methods. If an event type is not handled, the View will fail. Views are designed to restart, but since the handler is missing, it will fail again. Therefore, it's important to define methods for all events or define a catch-all method in case you want to discard some events.
```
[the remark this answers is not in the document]
franciscolopezsancho: :)

docs/src/modules/spring/pages/views.adoc
```diff
@@ -1 +1,186 @@
...
+
+The View definitions are stored and validated when a new version is deployed. There will be an error message if the changes are not compatible.
+
+== Streaming view updates
+
+A query can provide a near real time stream of results for the query, emitting new entries matching the query as they are added or updated in
+the view.
+
+
+ TODO: Ask Renato as he just added it to the Spring SDK
```
octonato, line 177: Yes, we have it already. Added recently.

https://github.com/lightbend/kalix-jvm-sdk/pull/1223

docs/src/modules/spring/pages/views.adoc
```diff
@@ -1 +1,186 @@
...
+----
+
+Such changes require you to define a new View. Kalix will then rebuild it from the source event log or value changes.
+
+WARNING: Views from topics cannot be rebuilt from the source messages, because it's not possible to consume all events from the topic again. The new View will be built from new messages published to the topic.
+
+Rebuilding a new View may take some time if there are many events that have to be processed. The recommended way when changing a View is multi-step, with two deployments:
+
+
+. Define the new View, and keep the old View intact. A new View is defined by a new View??. The `viewId` is the same as the `service` name, i.e. it will be a different `viewId` than the old View. Keep the old `register` of the old `service` in `Main`.
```
octonato, line 162: > A new View is defined by a new View??

Still draft PR? 

Maybe good to add very clear markers otherwise we risk to miss this kind of thing and end up merging it.

docs/src/modules/spring/pages/views.adoc
```diff
@@ -1 +1,186 @@
...
+. Deploy the new View, and let it rebuild. Verify that the new query works as expected. The old View can still be used.
+. Remove the old View definition and rename the new `service` to the old name if the public API is compatible, but keep the new `viewId` by defining it as shown below.
+. Deploy the second change.
+
+This is how to define a custom `viewId`: ...??? just changing the id would be enough right?
```
octonato, line 167: We are not supporting it now. We need an annotation for that. We can't call `withViewId` like in gRPC SDK. Users don't have access to the provider.

docs/src/modules/spring/pages/views.adoc
```diff
@@ -1 +1,186 @@
...
+ .src/main/java/customer/view/CustomerByNameView.class
+[source,java]
+----
+include::example$spring-eventsourced-customer-registry/src/main/java/customer/view/CustomerByNameView.java[tags=class]
+----
+<1> updating the state of the view with the new name 
+
+An Event Sourced entity can emit many types of events. You need to define a method for each event type. They return an `UpdateEffect`, which describes next processing actions, such as updating the view state.
+
+NOTE: This type of update transformation is a natural fit for Events emitted by an Event Sourced Entity, but it can also be used for Value Entities. For example, if the View representation is different from the Entity state you might want to transform it before presenting the View to the client.
```
octonato, line 97: The odd thing about the gRPC API is that we need to add `transform_updates: true` when consuming events from an event sourced entity. 

That creates the need to have to add such a paragraph. Makes sense for gRPC, but loses its weight in the Spring SDK. If there is a method, is a transformation, period. 

I think we can remove this and talk about transformation in the part that we explain how to create Views from Value Entities. Transformation simply doesn't make sense for Event Sourced entities because it's always a transformation from events to derived view state.

docs/src/modules/spring/pages/views.adoc
octonato: We can remove that sentence actually. I mean "A new View is defined by a new View". 

Users should define a new View and keep the old, that's All. The service name is the name of the class.

octonato: Which makes me realise now that users should be free to change the name of their class without breaking the view. If they do that today, they will create a new view and rebuild it from scratch. They may not even notice it.

octonato: > Keep the old `register` of the old `service` in `Main`

Not applicable

octonato (commented):
left some more comments

I think we should park this one and work on viewId issue

docs/src/modules/spring/pages/views.adoc
```diff
@@ -1 +1,186 @@
...
+
+.src/main/java/customer/api/CustomerEvent.java
+[source,java]
+----
+include::example$spring-eventsourced-customer-registry/src/main/java/customer/api/CustomerEvent.java[tags=class]
+----
+<1> Includes the logical type name (defined in implementation classes) as type 
+<2> Required for deserialization and to define the logical type name 
+
+Adding `@JsonTypeInfo` is needed so that Jackson can read back correctly the Object type when deserializing. By using `name`, you set its value as the fully qualified message name instead of the using the fully qualified Java class. This is called a `logical type name` and it allows you to refactor the class to another package after you have serialized messages of this type. Otherwise, after such a refactor the objects serialized with the old package won't be recognized as the same type.
```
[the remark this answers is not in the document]
franciscolopezsancho: deleted

docs/src/modules/spring/pages/views.adoc
franciscolopezsancho: deleted

docs/src/modules/spring/pages/views.adoc
franciscolopezsancho: removed the TODO

docs/src/modules/spring/pages/views.adoc
franciscolopezsancho: Changed to suggestion.

docs/src/modules/spring/pages/views.adoc
```diff
@@ -1 +1,186 @@
...
+include::example$spring-customer-registry-views-quickstart/src/main/java/customer/view/CustomerByNameView.java[tags=class]
+----
+<1> defining table name
+<2> subscribing to `CustomerEntity`
+<3> extending from `View`
+<4> defining endpoint
+<5> defining the query
+<6> note that no return is needed
+
+IMPORTANT: Note that the return value of the method is `null`. You may ask yourself,  how is it that the endpoint respond with any `Customer` at all?. When you call this endpoint, it first hits the proxy, which calls directly to the database. When the proxy receives the response, it sends it directly to you without any further intervention from the View. Therefore, null is valid as a return value in the endpoint of a View. The choice of null is a way to signal that the response doesn't come from the return of this method. The return could be anything that is compatible with the return type.
```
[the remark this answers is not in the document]
franciscolopezsancho: reworded it

docs/src/modules/spring/pages/views.adoc
franciscolopezsancho: You are right. It should have been a Draft PR.
removed both. 

> Which makes me realise now that users should be free to change the name of their class without breaking the view. If they do that today, they will create a new view and rebuild it from scratch. They may not even notice it.

Is there anything we should add to this regard?

docs/src/modules/spring/pages/views.adoc
franciscolopezsancho: I'm not sure I follow. Let's discuss this.

docs/src/modules/spring/pages/views.adoc
franciscolopezsancho: make sure all our view include a viewId . We highly recommend to have an ViewId

franciscolopezsancho: create an issue about nott clear transformations in view in protocol first.

octonato: As discussed on call, my memory failed me. We do have already a `ViewId` and if not defined we use the class name. Not great. We should make `ViewId` mandatory. I will create a ticket for it.

That said, the paragraph that explain how to create a new view by change the view id should only be re-written to explain the annotation.

@franciscolopezsancho, can you include a paragraph explaining that the usage of `ViewId` is highly recommended? 

We should also have it on each View that have in the samples, but can be on another PR.

franciscolopezsancho: roger that

docs/src/modules/spring/pages/views.adoc
franciscolopezsancho: after our call is clear no mention of transformations should appear in the docs

franciscolopezsancho:
removed references to transformations
added View Id
Still missing explaining how to modify a view. I couldn't make sense of it so I didn't add it yet.

docs/src/modules/spring/pages/views.adoc
```diff
@@ -1 +1,167 @@
...
+
+Adding a view ID to your View allows you to refactor the name of the class later on without the risk of losing the history of events. If you don't define a view ID the class name becomes its ID. Therefore, if you change the name of your class afterwards Kalix will not recognize this new name as the same view and will create a brand new view. This is resource consuming for a view from an entity because it will reprocess all the event of that entity to rebuild the view. While for a view built from a topic you will loose all the previous events because these views only process events from current time forwards.
+
```
octonato, line 60: It's worth adding an example showing how to transform from `Customer` to `CustomerView`. 

We can add on this PR or later on. If on follow-up, let's create an issue so we don't forget.

docs/src/modules/spring/pages/views.adoc
```diff
@@ -1 +1,167 @@
...
+When consuming events, each event must be matched by a View service method. In case your View is only interested in certain events, you may declare a catch-all method to receive the `sealed interface` and don't update the View. If an event type is not handled, the View will fail. Views are designed to restart, but since the handler is missing, it will fail again. Therefore, it's important to define methods for all events or define a catch-all method in case you want to discard some events.
+
+
+[#topic-view]
+== Creating a View from a topic
+
+The source of a View can be a topic. It works the same way as shown in <<event-sourced-entity>> or <<value-entity>>, but you define it with `@Subscribe.Topic` instead.
+
+
+=== Relational projection
```
octonato, line 111: We are missing the section "How to transform results" above "Relational projection". That's why the link at the beginning is broken.

docs/src/modules/spring/pages/views.adoc
```diff
@@ -1 +1,167 @@
...
+You may realize after a deployment that you forgot adding some parameters to the query Parameters that aren't exposed to the endpoint of the View. After adding these parameters the query is changed and therefore Kalix needs to add indexes for these new columns. For example, changing the above query to filter to add users that are active would mean that Kalix needs to build a View with the index on the `is-active` column.
+
+[source,proto,indent=0]
+----
+SELECT * FROM customers WHERE name = :customer_name AND is-active = true
+----
+
+Such changes require you to define a new View. Kalix will then rebuild it from the source event log or value changes.
+
+WARNING: Views from topics cannot be rebuilt from the source messages, because it's not possible to consume all events from the topic again. The new View is built from new messages published to the topic.
```
octonato, line 143:
```suggestion
WARNING: Views from topics cannot be rebuilt from the source messages, because it might not be possible to consume all events from the topic again. The new View is built from new messages published to the topic.
```

docs/src/modules/spring/pages/views.adoc
```diff
@@ -1 +1,167 @@
...
+
+Rebuilding a new View may take some time if there are many events that have to be processed. The recommended way when changing a View is multi-step, with two deployments:
+
+. Define the new View, and keep the old View intact.
```
octonato, line 147:
```suggestion
. Define the new View with a new `@ViewId`, and keep the old View intact. 
```

docs/src/modules/spring/pages/views.adoc
```diff
@@ -1 +1,167 @@
...
+
+Such changes require you to define a new View. Kalix will then rebuild it from the source event log or value changes.
+
+WARNING: Views from topics cannot be rebuilt from the source messages, because it's not possible to consume all events from the topic again. The new View is built from new messages published to the topic.
+
+Rebuilding a new View may take some time if there are many events that have to be processed. The recommended way when changing a View is multi-step, with two deployments:
+
+. Define the new View, and keep the old View intact. 
+. Deploy the new View, and let it rebuild. Verify that the new query works as expected. The old View can still be used.
+. Remove the old View but keep its `@ViewId` and `@GetMapping` path, and use them in the new View.
```
octonato, line 149: We can't keep the old ViewId. The goal is to use the old path on the new View.

If we do what is described here, Kalix won't deploy the view. Because the old view id is associate with another view definition (different index).

docs/src/modules/spring/pages/views.adoc
```diff
@@ -1 +1,167 @@
...
+== Streaming view updates
+
+A query can provide a near real time stream of results for the query, emitting new entries matching the query as they are added or updated in
+the view.
+
+This will first list the complete result for the query and then keep the response stream open, emitting new or updated
+entries matching the query as they are added to the view. The stream does not complete until the client closes it.
+
+Note: This is not intended as transport for service to service propagation of updates and does not guarantee delivery, for such use cases you
+should instead publish events to a topic, see xref:actions-publishing-subscribing.adoc[Publishing and Subscribing with Actions]
```
octonato, line 164: maybe worth to also link to Service to Service page?

octonato (commented):
Shaping up well!

I left a few comments that we need to address and some suggestions that can go later on future PR.

docs/src/modules/spring/pages/views.adoc
franciscolopezsancho: I will delete the above link as we agreed we won't mention transformations.

docs/src/modules/spring/pages/views.adoc
franciscolopezsancho: how is it possible to consume all events from the topic again?

docs/src/modules/spring/pages/views.adoc
```diff
@@ -1 +1,167 @@
...
+
+Views can be defined from any of the following:
+
+* xref:value-entity[Value Entities state changes]
+* xref:event-sourced-entity[Event Sourced Entity events]
+* xref:topic-view[Messages received from subscribing to topics on a broker]
+
+The remainder of this page describes:
+
+* <<transform-results>>
```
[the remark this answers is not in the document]
franciscolopezsancho: removed as we agreed not to mention transformations.

docs/src/modules/spring/pages/views.adoc
franciscolopezsancho: I see now. With Kafka could have a retention policy `earliest`.

docs/src/modules/spring/pages/views.adoc
franciscolopezsancho: https://github.com/lightbend/kalix/issues/7884

docs/src/modules/spring/pages/views.adoc
franciscolopezsancho: Thanks for the explanation. 
Removed mention to keeping the old ViewId

docs/src/modules/spring/pages/views.adoc
franciscolopezsancho: added

docs/src/modules/spring/pages/views.adoc
```diff
@@ -1 +1,167 @@
...
+
+Adding a view ID to your View allows you to refactor the name of the class later on without the risk of losing the history of events. If you don't define a view ID the class name becomes its ID. Therefore, if you change the name of your class afterwards Kalix will not recognize this new name as the same view and will create a brand new view. This is resource consuming for a view from an entity because it will reprocess all the event of that entity to rebuild the view. While for a view built from a topic you will loose all the previous events because these views only process events from current time forwards.
```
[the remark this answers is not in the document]
franciscolopezsancho: > users will lose the view, not the history of events.

Changed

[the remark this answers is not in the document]
franciscolopezsancho: > For instance, the Topic is not always the case. If PubSub, indeed, the description is correct, but it Kafka it starts from lowest available offset (depends on retention).

Added hopefully without making the text difficult to read.

docs/src/modules/spring/pages/views.adoc
```diff
@@ -1 +1,167 @@
...
+An Event Sourced entity can emit many types of events. You need to define a method for each event type. They return an `UpdateEffect`, which describes next processing actions, such as updating the view state.
+
+
+
+
+See <<#query>> for more examples of valid query syntax.
+
+== Ignoring events
+
+When consuming events, each event must be matched by a View service method. In case your View is only interested in certain events, you may declare a catch-all method to receive the `sealed interface` and don't update the View. If an event type is not handled, the View will fail. Views are designed to restart, but since the handler is missing, it will fail again. Therefore, it's important to define methods for all events or define a catch-all method in case you want to discard some events.
```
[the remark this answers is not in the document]
franciscolopezsancho: I elaborated on the `ignoreUnknown` explanation but I didn't promote nor add type level annotations explanation. Let me see if we have an example already.

docs/src/modules/spring/pages/views.adoc
octonato: The transformation in this paragraph is not about `A => B`, but about "Relational projection"

docs/src/modules/spring/pages/views.adoc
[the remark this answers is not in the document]
franciscolopezsancho: I'd like to discuss this last point if possible.

docs/src/modules/spring/pages/views.adoc
```diff
@@ -1 +1,169 @@
...
+
+
+
+See <<#query>> for more examples of valid query syntax.
+
+== Ignoring events
+
+When consuming events, each event must be matched by a View service method. In case your View is only interested in certain events:
+1. You can declare a catch-all method to receive the `sealed interface` and don't update the View.
+2. You can add `ignoreUnknown = true` to your `@Subcribe` annotation but only if it is a type level annotation.
```
octonato, line 103: This is not working as order list. That's markdown syntax, we need to use asciidoc syntax for lists.

octonato: A catch-all won't work for Spring SDK. 

The type url will be the one from the concrete type, not the parent interface. We will do the look up for the method and we won't find the catch-all method. Because the catch-all will associated with the interface type url. 

So, either the user adds event handlers for all of them and return `Effect.ignore` for those they are not interested, or they use type level annotation.

docs/src/modules/spring/pages/views.adoc
```diff
@@ -1 +1,169 @@
...
+You may realize after a deployment that you forgot adding some parameters to the query Parameters that aren't exposed to the endpoint of the View. After adding these parameters the query is changed and therefore Kalix needs to add indexes for these new columns. For example, changing the above query to filter to add users that are active would mean that Kalix needs to build a View with the index on the `is-active` column.
+
+[source,proto,indent=0]
+----
+SELECT * FROM customers WHERE name = :customer_name AND is-active = true
+----
+
+Such changes require you to define a new View. Kalix will then rebuild it from the source event log or value changes.
+
+WARNING: Views from topics cannot be rebuilt from the source messages, because it might not be possible to consume all events from the topic again. The new View is built from new messages published to the topic.
```
octonato, line 145: This is probably a problem for the other SDKs as well. True when we had only PubSub, but with Kafka it might be possible. It will depend on the retention criteria.

docs/src/modules/spring/pages/views.adoc
franciscolopezsancho: fixed now

docs/src/modules/spring/pages/views.adoc
franciscolopezsancho: I see now. Yes, of course, you are right.

franciscolopezsancho: modified

docs/src/modules/spring/pages/views.adoc
franciscolopezsancho: roger that

docs/src/modules/spring/pages/views.adoc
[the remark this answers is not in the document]
franciscolopezsancho: Elaborated on it and linked to type-level annotations in actions which has an example explaining exactly the same. I have chosen to use the link because I think is more didactic. It doesn't add more info than the reader needs to understand views and brings the idea that Actions work in the same way when ignoring events.

octonato (approved):
LGTM

Thanks for the all the efforts! 😄

PR #1230, opened 2022-11-02
feat: support for service to service streaming

octonato:
References https://github.com/lightbend/kalix/issues/7725

### DIFF
--- a/docs/src/modules/spring/pages/service-to-service.adoc
+++ b/docs/src/modules/spring/pages/service-to-service.adoc
@@ -1,1 +1,46 @@
 = Service to Service Eventing
+
+include::ROOT:partial$include.adoc[]
+
+
+Kalix provides brokerless at-least-once event delivery across Kalix services through the Service to Service eventing.
+
+The source of the events is an xref:event-sourced-entities.adoc[Event Sourced Entity]. Its events can be published as a stream and consumed by another Kalix service without the need to set up a message broker.
+
+Note:: For eventing from an entity inside the same Kalix service as the consuming component, use regular
+xref:actions-publishing-subscribing.adoc[Subscription to the entity] instead of Service to Service eventing.
+
+== Event Producer
+
+The event producer controls which entity to publish events for. Each entity published is identified by a stream `id` so
+that one Kalix service can publish more than one of the entity types it contains.
+
+[source,java,indent=0]
+.src/main/java/customer/api/CustomerEventsService.java
+----
+include::example$spring-eventsourced-customer-registry/src/main/java/customer/api/CustomerEventsService.java[tag=producer]
+----
+<1> Identifying which event sourced entity to publish events for.
+<2> Ignore any event types not handled by a method and move on with the event stream, rather than fail which is the default.
+<3> Public identifier for consumers of this stream.
+<4> Allowing access from other Kalix services, but not the public internet.
+<5> All methods on the service are transformer methods for turning internal event message types into public API message types for other services to consume
+
+== Event Consumer
+
+The consumer can be an Action or a View, annotated with `@Subscribe.Stream` with a `service` identifying the publishing service, and the `id` of the stream to subscribe to.
+
+We then define a component subscribing to the service to service publisher. In this example we do that with a View:
+
+[.tabset]
+Java::
++
+[source,java,indent=0]
+.src/main/java/customer/views/CustomersByNameView.java
+----
+include::example$spring-eventsourced-customer-registry-subscriber/src/main/java/customer/views/CustomersByNameView.java[tag=view]
+----
+<1> Annotate the component with `@Subscribe.Stream` to subscribe to an event stream from another Kalix service.
+<2> The name of the Kalix service publishing the event stream.
+<3> The public identifier of the specific stream from the publisher.
+<4> One update method per message type that the stream may contain.
--- a/samples/spring-eventsourced-customer-registry-subscriber/src/main/java/customer/Main.java
+++ b/samples/spring-eventsourced-customer-registry-subscriber/src/main/java/customer/Main.java
@@ -1,0 +1,26 @@
+package customer;
+
+import kalix.springsdk.annotations.Acl;
+import kalix.springsdk.KalixConfiguration;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.boot.SpringApplication;
+import org.springframework.boot.autoconfigure.SpringBootApplication;
+import org.springframework.context.annotation.Import;
+
+@SpringBootApplication
+@Import(KalixConfiguration.class)
+// Allow all other Kalix services deployed in the same project to access the components of this
+// Kalix service, but disallow access from the internet. This can be overridden explicitly
+// per component or method using annotations.
+// Documentation at https://docs.kalix.io/services/using-acls.html
+@Acl(allow = @Acl.Matcher(service = "*"))
+public class Main {
+
+  private static final Logger logger = LoggerFactory.getLogger(Main.class);
+
+  public static void main(String[] args) {
+    logger.info("Starting Kalix - Spring SDK");
+    SpringApplication.run(Main.class, args);
+  }
+}
--- a/samples/spring-eventsourced-customer-registry/src/main/java/customer/api/CustomerEventsService.java
+++ b/samples/spring-eventsourced-customer-registry/src/main/java/customer/api/CustomerEventsService.java
@@ -1,0 +1,28 @@
+package customer.api;
+
+import customer.domain.CustomerEvent;
+import kalix.javasdk.action.Action;
+import kalix.springsdk.annotations.Publish;
+import kalix.springsdk.annotations.Subscribe;
+import kalix.springsdk.annotations.Acl;
+
+// tag::producer[]
+@Subscribe.EventSourcedEntity(
+    value = CustomerEntity.class, // <1>
+    ignoreUnknown = true) // <2>
+@Publish.Stream(id = "customer_events") // <3>
+@Acl(allow = @Acl.Matcher(service = "*")) // <4>
+public class CustomerEventsService extends Action {
+
+  public Effect<CustomerPublicEvent.Created> onEvent( // <5>
+      CustomerEvent.CustomerCreated created) {
+    return effects().reply(
+        new CustomerPublicEvent.Created(created.email(), created.name()));
+  }
+
+  public Effect<CustomerPublicEvent.NameChanged> onEvent( // <5>
+      CustomerEvent.NameChanged nameChanged) {
+    return effects().reply(new CustomerPublicEvent.NameChanged(nameChanged.newName()));
+  }
+}
+// end::producer[]
--- a/samples/spring-eventsourced-customer-registry/src/main/java/customer/view/CustomerByEmailView.java
+++ b/samples/spring-eventsourced-customer-registry/src/main/java/customer/view/CustomerByEmailView.java
@@ -1,19 +1,23 @@
 package customer.view;
 
 import customer.api.CustomerEntity;
-import customer.api.CustomerEvent;
+import customer.domain.CustomerEvent;
 import kalix.javasdk.view.View;
 import kalix.springsdk.annotations.Query;
 import kalix.springsdk.annotations.Subscribe;
 import kalix.springsdk.annotations.Table;
+import kalix.springsdk.annotations.ViewId;
+import reactor.core.publisher.Flux;
+
 import org.springframework.web.bind.annotation.GetMapping;
 
+@ViewId("view_customers_by_email")
 @Table("customers_by_email")
 public class CustomerByEmailView extends View<CustomerView> {
 
   @GetMapping("/customer/by_email/{email}")
   @Query("SELECT * FROM customers_by_email WHERE email = :email")
-  public CustomerView getCustomer(String email) {
+  public Flux<CustomerView> getCustomer(String email) {
     return null;
   }
 
--- a/sdk/spring-sdk/src/main/java/kalix/springsdk/annotations/Publish.java
+++ b/sdk/spring-sdk/src/main/java/kalix/springsdk/annotations/Publish.java
@@ -32,4 +32,18 @@ public @interface Publish {
     /** Assign the name of the topic to be used for eventing out. */
     String value();
   }
+
+
+  /**
+   * Annotation to configure the component to publish an event stream to other Kalix services.
+   */
+  @Target(ElementType.TYPE)
+  @Retention(RetentionPolicy.RUNTIME)
+  @Documented
+  @interface Stream {
+    /**
+     * Identifier for the event stream. Must be unique inside the same Kalix service.
+     */
+    String id();
+  }
 }
--- a/sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/AclDescriptorFactory.scala
+++ b/sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/AclDescriptorFactory.scala
@@ -48,36 +48,30 @@ object AclDescriptorFactory {
 
     aclJavaAnnotation.allow.zipWithIndex.foreach { case (allow, idx) =>
       val principalMatcher = PrincipalMatcher.newBuilder()
-      if (allow.principal != Acl.Principal.UNSPECIFIED) {
-        allow.principal match {
-          case Acl.Principal.ALL =>
-            principalMatcher.setPrincipal(PrincipalMatcher.Principal.ALL)
-          case Acl.Principal.INTERNET =>
-            principalMatcher.setPrincipal(PrincipalMatcher.Principal.INTERNET)
-        }
-        aclBuilder.addAllow(idx, principalMatcher)
-
-      } else {
-        principalMatcher.setService(allow.service())
-        aclBuilder.addAllow(idx, principalMatcher)
+      allow.principal match {
+        case Acl.Principal.ALL =>
+          principalMatcher.setPrincipal(PrincipalMatcher.Principal.ALL)
+        case Acl.Principal.INTERNET =>
+          principalMatcher.setPrincipal(PrincipalMatcher.Principal.INTERNET)
+        case Acl.Principal.UNSPECIFIED =>
+          principalMatcher.setService(allow.service())
       }
+
+      aclBuilder.addAllow(idx, principalMatcher)
     }
 
     aclJavaAnnotation.deny.zipWithIndex.foreach { case (deny, idx) =>
       val principalMatcher = PrincipalMatcher.newBuilder()
-      if (deny.principal != Acl.Principal.UNSPECIFIED) {
-        deny.principal match {
-          case Acl.Principal.ALL =>
-            principalMatcher.setPrincipal(PrincipalMatcher.Principal.ALL)
-          case Acl.Principal.INTERNET =>
-            principalMatcher.setPrincipal(PrincipalMatcher.Principal.INTERNET)
-        }
-        aclBuilder.addDeny(idx, principalMatcher)
+      deny.principal match {
+        case Acl.Principal.ALL =>
+          principalMatcher.setPrincipal(PrincipalMatcher.Principal.ALL)
+        case Acl.Principal.INTERNET =>
+          principalMatcher.setPrincipal(PrincipalMatcher.Principal.INTERNET)
+        case Acl.Principal.UNSPECIFIED =>
+          principalMatcher.setService(deny.service())
 
-      } else {
-        principalMatcher.setService(deny.service())
-        aclBuilder.addDeny(idx, principalMatcher)
       }
+      aclBuilder.addDeny(idx, principalMatcher)
     }
 
     aclBuilder.setDenyCode(aclJavaAnnotation.denyCode())
--- a/sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/ActionDescriptorFactory.scala
+++ b/sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/ActionDescriptorFactory.scala
@@ -30,14 +30,15 @@ import kalix.springsdk.impl.ComponentDescriptorFactory.hasTopicPublication
 import kalix.springsdk.impl.ComponentDescriptorFactory.hasTopicSubscription
 import kalix.springsdk.impl.ComponentDescriptorFactory.hasValueEntitySubscription
 import kalix.springsdk.impl.ComponentDescriptorFactory.publishToEventStream
+import kalix.springsdk.impl.ComponentDescriptorFactory.streamSubscription
 import kalix.springsdk.impl.ComponentDescriptorFactory.subscribeToEventStream
 import kalix.springsdk.impl.ComponentDescriptorFactory.validateRestMethod
 import kalix.springsdk.impl.reflection.CombinedSubscriptionServiceMethod
 import kalix.springsdk.impl.reflection.KalixMethod
 import kalix.springsdk.impl.reflection.NameGenerator
-import kalix.springsdk.impl.reflection.SubscriptionServiceMethod
 import kalix.springsdk.impl.reflection.ReflectionUtils
 import kalix.springsdk.impl.reflection.RestServiceIntrospector
+import kalix.springsdk.impl.reflection.SubscriptionServiceMethod
 import org.slf4j.Logger
 import org.slf4j.LoggerFactory
 
@@ -113,6 +114,22 @@ private[impl] object ActionDescriptorFactory extends ComponentDescriptorFactory 
         Map(entityType -> kalixMethods)
 
       } else Map.empty
+
+    val subscriptionStreamClass: Map[String, Seq[KalixMethod]] = {
+      streamSubscription(component)
+        .map { ann =>
+          val kalixMethods =
+            component.getMethods
+              .filter(hasActionOutput)
+              .sorted // make sure we get the methods in deterministic order
+              .map { method => KalixMethod(SubscriptionServiceMethod(method)) }
+              .toSeq
+
+          val streamId = ann.id()
+          Map(streamId -> kalixMethods)
+        }
+        .getOrElse(Map.empty)
+    }
 
     // methods annotated with @Subscribe.Topic
     val subscriptionTopicMethods: IndexedSeq[KalixMethod] = component.getMethods
@@ -240,6 +257,7 @@ private[impl] object ActionDescriptorFactory extends ComponentDescriptorFactory 
       ++ subscriptionValueEntityMethods
       ++ combineByES(subscriptionEventSourcedEntityMethods, messageCodec, component)
       ++ combineByES(subscriptionEventSourcedEntityClass, messageCodec, component)
+      ++ combineByES(subscriptionStreamClass, messageCodec, component)
       ++ combineByTopic(subscriptionTopicClass)
       ++ combineByTopic(subscriptionTopicMethods)
       ++ removeDuplicates(syntheticMethods, publicationTopicMethods))
--- a/sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/ActionDescriptorFactory.scala
+++ b/sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/ActionDescriptorFactory.scala
@@ -19,38 +19,62 @@ package kalix.springsdk.impl
 import kalix.springsdk.impl.ComponentDescriptorFactory.buildJWTOptions
 import kalix.springsdk.impl.ComponentDescriptorFactory.combineByES
 import kalix.springsdk.impl.ComponentDescriptorFactory.eventingInForEventSourcedEntity
+import kalix.springsdk.impl.ComponentDescriptorFactory.eventingInForEventSourcedEntityServiceLevel
 import kalix.springsdk.impl.ComponentDescriptorFactory.eventingInForTopic
 import kalix.springsdk.impl.ComponentDescriptorFactory.eventingInForValueEntity
 import kalix.springsdk.impl.ComponentDescriptorFactory.eventingOutForTopic
+import kalix.springsdk.impl.ComponentDescriptorFactory.findEventSourcedEntityType
 import kalix.springsdk.impl.ComponentDescriptorFactory.hasActionOutput
 import kalix.springsdk.impl.ComponentDescriptorFactory.hasEventSourcedEntitySubscription
 import kalix.springsdk.impl.ComponentDescriptorFactory.hasTopicPublication
 import kalix.springsdk.impl.ComponentDescriptorFactory.hasTopicSubscription
 import kalix.springsdk.impl.ComponentDescriptorFactory.hasValueEntitySubscription
+import kalix.springsdk.impl.ComponentDescriptorFactory.publishToEventStream
+import kalix.springsdk.impl.ComponentDescriptorFactory.subscribeToEventStream
 import kalix.springsdk.impl.ComponentDescriptorFactory.validateRestMethod
 import kalix.springsdk.impl.reflection.CombinedSubscriptionServiceMethod
 import kalix.springsdk.impl.reflection.KalixMethod
 import kalix.springsdk.impl.reflection.NameGenerator
+import kalix.springsdk.impl.reflection.SubscriptionServiceMethod
 import kalix.springsdk.impl.reflection.ReflectionUtils
 import kalix.springsdk.impl.reflection.RestServiceIntrospector
-import kalix.springsdk.impl.reflection.SubscriptionServiceMethod
+import org.slf4j.Logger
+import org.slf4j.LoggerFactory
 
 private[impl] object ActionDescriptorFactory extends ComponentDescriptorFactory {
+
+  private val logger: Logger = LoggerFactory.getLogger(classOf[ActionDescriptorFactory.type])
 
   override def buildDescriptorFor(
       component: Class[_],
       messageCodec: SpringSdkMessageCodec,
       nameGenerator: NameGenerator): ComponentDescriptor = {
+
+    // TODO: we need more robust validation covering all the corner cases
+    // verify component is correctly configured
+    if (hasEventSourcedEntitySubscription(component) &&
+      component.getMethods.exists(hasEventSourcedEntitySubscription))
+      throw InvalidComponentException(
+        "You cannot use @Subscribe.EventSourcedEntity annotation in " +
+        "methods and class. You can do either one or the other.")
+
+    if (hasTopicSubscription(component) && component.getMethods.exists(hasTopicSubscription))
+      throw InvalidComponentException(
+        "You cannot use @Subscribe.Topic annotation in " +
+        "both methods and class. You can do either one or the other.")
+
     //we should merge from here
-    val syntheticMethods =
+    // methods with REST annotations
+    val syntheticMethods: Seq[KalixMethod] =
       RestServiceIntrospector.inspectService(component).methods.map { serviceMethod =>
         validateRestMethod(serviceMethod.javaMethod)
         KalixMethod(serviceMethod).withKalixOptions(buildJWTOptions(serviceMethod.javaMethod))
       }
 
     //TODO make sure no subscription should be exposed via REST.
+    // methods annotated with @Subscribe.ValueEntity
     import ReflectionUtils.methodOrdering
-    val subscriptionValueEntityMethods = component.getMethods
+    val subscriptionValueEntityMethods: IndexedSeq[KalixMethod] = component.getMethods
       .filter(hasValueEntitySubscription)
       .sorted // make sure we get the methods in deterministic order
       .map { method =>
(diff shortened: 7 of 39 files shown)


sdk/spring-sdk/src/test/java/kalix/springsdk/testmodels/subscriptions/PubSubTestModels.java
```diff
@@ -195,14 +195,58 @@ public Action.Effect<Message> messageOne(@RequestBody Message message) {
...
+  }
+
+
+  @Publish.EventStream(source = EmployeeEntity.class, name = "employee_events", ignoreUnknown = true)
+  public static class EventStreamPublishingAction extends Action {
+
+    public String transform(EmployeeCreated created) {
+      return created.toString();
+    }
+  }
```
octonato, line 224: I think this is the best place to show how it would look like if we have separated annotations. 

My original design was something similar to that.

```java
@Subscribe.EventSourcedEntity(value = EmployeeEntity.class, ignoreUnkown = true)
@Publish.EventStream("employee_events")
public static class EventStreamPublishingAction extends Action {

  public String transform(EmployeeCreated created) {
    return created.toString();
  }
}
```

What I don't like with that solution is that the use case is: "User wants to expose the event from entity A to other services". 
But to do that, they need to understand that they need to compose two annotations. I think we do a better job if we provide a single annotation that unambiguously describe the use case. 

And there are some flaws...

* It's possible to only add `@Publish.EventStream("employee_events")` without the `@Subscribe`. Only possible to validate at runtime. Arguably, codegen and annotations processors could validate at compile time, but puts the burden on us to create such validation
* It's possible to combine `@Publish.EventStream("employee_events")` with `@Subscribe.ValueEntity`. Same validation issues as previous point.

In both cases, we will need to add validation to ensure that users are using and combining the annotations in a way that makes sense. While providing one single annotation free us from the task of needing to validate and it frees the user from the task to understand how to compose annotations.

octonato (commented):
Still draft, left some comments about the annotation options/choices.

sdk/spring-sdk/src/main/java/kalix/springsdk/annotations/Publish.java
```diff
@@ -32,4 +32,26 @@
...
+  @Documented
+  @interface EventStream {
+
+    String name();
+
+    /**
+     * Assign the class type of the entity one intends to subscribe to, which must extend {@link
+     * kalix.javasdk.eventsourcedentity.EventSourcedEntity EventSourcedEntity}.
+     */
+    Class<? extends kalix.javasdk.eventsourcedentity.EventSourcedEntity<?>> source();
```
aludwiko, line 48: What about VE?

sdk/spring-sdk/src/test/java/kalix/springsdk/testmodels/subscriptions/PubSubTestModels.java
aludwiko: In case I would like to publish events to a kafka topic I still need to use 2 annotations right? Or are you planning to merge it all to sth like:
```
@Publish.EventStream(source = EmployeeEntity.class, destination = @Stream("employee_events"), ignoreUnknown = true)
@Publish.EventStream(source = EmployeeEntity.class, destination = @Topic("employee_events"), ignoreUnknown = true)
```

sdk/spring-sdk/src/main/java/kalix/springsdk/annotations/Publish.java
octonato: no supported yet
 
and then, it should not be on `@Publish.EventStream`, but something else, eg: `@Publish.ChangeStream' or a better name

sdk/spring-sdk/src/test/java/kalix/springsdk/testmodels/subscriptions/PubSubTestModels.java
octonato: I had this thought as well, I was thinking about:

```java
@Publish.Topic(source = EmployeeEntity.class, ...)
```

Source would be optional though because it should be possible to have a REST endpoint that publishes to a Topic. In such a case, the source is the endpoint.

We can explore some ideas. Your proposal looks good as well. The only thing is that will have two different ways to publish to topic (`@Publish.Topic` and `@Publish.EventStream(destination = @Topic)`).

But I absolutely agree that publishing from entity to topic should be done with a single annotation. The question is how.

johanandren: @aludwiko Eventing out to multiple destinations does not work anyway (method eventing.out is a oneof on method level, and on service level only direct is supported - mixing service and method outs leads to validation failure).

I liked it better with the composed/more aligned with the underlying, one reason is that if we diverge completely with the underlying annotations it will be really hard for us to understand/remember how things work. Not sure how strong that argument is though, when it has already diverged with the publish/subscribe vs eventing in/out.

efgpinto: > ''What I don't like with that solution is that the use case is: "User wants to expose the event from entity A to other services". But to do that, they need to understand that they need to compose two annotations. I think we do a better job if we provide a single annotation that unambiguously describe the use case.

If that's the main use case, should we also allow the user to define the stream when defining the Entity itself? Rather than needing to define a new action externally? This could be useful when the user is not interested in doing any transformation.

That said, giving the option to change your events for public consumption seems critical to me, and there we would still require the user to have a separate action. Then I think once the user needs to define a new action, it becomes "required" for them to understand this is a two-step thing, so requiring two annotations does not seem farfetched to me. 

> * It's possible to only add `@Publish.EventStream("employee_events")` without the `@Subscribe`. Only possible to validate at runtime. Arguably, codegen and annotations processors could validate at compile time, but puts the burden on us to create such validation

This would need to be prevented for Actions but could be used as a valid scenario for Entities where you don't want to transform the events.

In general, I think I would probably lean towards 2 separate annotations because they seem more composable and closer to what we have at the moment and they seem valuable as well when used separately but I think you're right that validation gets tricker.

sdk/spring-sdk/src/main/java/kalix/springsdk/annotations/Publish.java
efgpinto: Ok yeah, so I was looking at this from the second option perspective (where we have to separate annotations). In that sense, it would be similar to `@Publish.Topic` where it's not typed as well... Being all in one, it makes sense to be typed.

sdk/spring-sdk/src/test/java/kalix/springsdk/testmodels/subscriptions/PubSubTestModels.java
johanandren: Note that the publish only without any subscribe will lead to validation failure at discovery, so even if it's not a short feedback loop, it is a feedback loop.

patriknw: I'm not convinced that it's the right thing to have this higher level definition rather than composing in and out. I haven't studied the other Spring annotations for eventing.in (from topic, from event sourced entity, from value entity, from service) and eventing.out (to action, to view, to service).

If we go with these higher level definitions we should do it for all of the combinations, and that could become an explosion.

I was quite happy with that the service-to-service eventing could hook into the same eventing.in and eventing.out concepts as the other things. The code and proto is almost identical when switching between service-to-service and same-service.

octonato: There are a few reasons why the annotations are not aligned. 

It started with an attempt to use names that are more common or even correct english. There is no such a verb as "to event". The word "eventing" is a made-up concept that needs to be explained or forced into someone else brain. Sure, it's been used by others as well, but stays a strange word in any case. Certainly for newcomers. 

Another reason is that it gives us more type-safety. A generic `@Eventing.In` annotation will require us to have string fields for `topic`, `eventSourcedEntity` and `valueEntity`. 

Specific annotations bounded to some types allows us to inspect the types and pick extra information from there, for example: we can pick the `@EntityType` value and correctly configure the subscription.  

Another reason is that Java annotations, although limited in power, are much more expressive than protobuf annotations. Using protobuf as the common denominator produces a limited Java experience. We see that in other parts as well, for example the fact that event handlers can be typed. 

On the topic of using high-level annotations or not, I can see the concern. But I'm also not convinced that we should let users pick things from the shelf and compose. That means that we put the burden on then to understand what can be composed or not. And we put the burden on us to validate that what they composed is legal or not. 

The more type-safety we have the better. Less validation on our side, more guidance for the users. But true, it means that we need to do more translation on our side. But isn't that the goal of Kalix? Users do less and we do the hard lifting?

Providing high composability looks more something we want for Akka, but in Kalix we want guard-rails.

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/AclDescriptorFactory.scala
```diff
@@ -48,36 +48,30 @@ object AclDescriptorFactory {
...
-      } else {
-        principalMatcher.setService(allow.service())
-        aclBuilder.addAllow(idx, principalMatcher)
+      allow.principal match {
+        case Acl.Principal.ALL =>
+          principalMatcher.setPrincipal(PrincipalMatcher.Principal.ALL)
+        case Acl.Principal.INTERNET =>
+          principalMatcher.setPrincipal(PrincipalMatcher.Principal.INTERNET)
+        case Acl.Principal.UNSPECIFIED =>
+          principalMatcher.setService(allow.service())
```
octonato, line 57: This change in the pattern matching and the one below is just to avoid some compilation warnings.

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/ActionDescriptorFactory.scala
```diff
@@ -19,106 +19,131 @@ package kalix.springsdk.impl
...
+            .sorted // make sure we get the methods in deterministic order
+            .map { method => KalixMethod(PubSubServiceMethod(method)) }
+            .toSeq
 
-            KalixMethod(SubscriptionServiceMethod(method))
-              .withKalixOptions(kalixOptions)
-          }
-      } else Array.empty[KalixMethod]
+        val entityType = findEventSourcedEntityType(component)
+        Map(entityType -> kalixMethods)
```
octonato, line 113: There some structural changes here to make it possible to have service level annotations for subscription. That's needed when publishing to stream

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/ActionDescriptorFactory.scala
```diff
@@ -19,106 +19,131 @@ package kalix.springsdk.impl
...
-            s"You cannot use Subscribe annotation in both the methods and the class. You can do either one or the other.")
         val subscriptionOptions = eventingInForTopic(method)
         val kalixOptions =
           kalix.MethodOptions.newBuilder().setEventing(subscriptionOptions).build()
 
-        KalixMethod(SubscriptionServiceMethod(method))
+        KalixMethod(PubSubServiceMethod(method))
           .withKalixOptions(kalixOptions)
       }
+      .toIndexedSeq
```
octonato, line 129: all this conversions to IndexedSeq is to eliminate warnings from compiler

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/ActionDescriptorFactory.scala
```diff
@@ -19,106 +19,131 @@ package kalix.springsdk.impl
...
 
-    val subscriptionTopicClass: Array[KalixMethod] =
+    // type level @Subscribe.Topic, methods eligible for subscription
+    val subscriptionTopicClass: IndexedSeq[KalixMethod] =
       if (hasTopicSubscription(component)) {
-        component.getMethods.sorted // make sure we get the methods in deterministic order
+        component.getMethods
           .filter(hasActionOutput)
-          .collect { case method =>
+          .sorted // make sure we get the methods in deterministic order
```
octonato, line 136: better to sort after filtering

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/ComponentDescriptor.scala
```diff
@@ -211,7 +211,7 @@ private[impl] object ComponentDescriptor {
 
           CommandHandler(grpcMethodName, messageCodec, JavaPbAny.getDescriptor, methodInvokers)
 
-        case method: SubscriptionServiceMethod =>
+        case method: PubSubServiceMethod =>
```
octonato, line 214: better name as this ServiceMethod can be used for both subscription and publishing

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/ActionDescriptorFactory.scala
```diff
@@ -33,7 +33,7 @@ import kalix.springsdk.impl.reflection.KalixMethod
 import kalix.springsdk.impl.reflection.NameGenerator
 import kalix.springsdk.impl.reflection.ReflectionUtils
 import kalix.springsdk.impl.reflection.RestServiceIntrospector
-import kalix.springsdk.impl.reflection.SubscriptionServiceMethod
+import kalix.springsdk.impl.reflection.PubSubServiceMethod
```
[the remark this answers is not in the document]
franciscolopezsancho: Maybe creating a `PublishingServiceMethod`? Although this may be too much to add to the PR. If it is, then this name looks perfectly fine to me.

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/ActionDescriptorFactory.scala
```diff
@@ -19,106 +19,131 @@ package kalix.springsdk.impl
...
+    if (hasEventSourcedEntitySubscription(component) &&
+      component.getMethods.exists(hasEventSourcedEntitySubscription))
+      throw InvalidComponentException(
+        "You cannot use @Subscribe.EventSourcedEntity annotation in " +
+        "methods and class. You can do either one or the other.")
+
+    if (hasTopicSubscription(component) && component.getMethods.exists(hasTopicSubscription))
+      throw InvalidComponentException(
+        "You cannot use @Subscribe.Topic annotation in " +
+        "both methods and class. You can do either one or the other.")
```
franciscolopezsancho, line 64: This helps to reading 👍

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/ActionDescriptorFactory.scala
franciscolopezsancho: I was wondering. Nice.

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/ActionDescriptorFactory.scala
franciscolopezsancho: rule of 👍

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/ActionDescriptorFactory.scala
```diff
@@ -183,11 +210,32 @@ private[impl] object ActionDescriptorFactory extends ComponentDescriptorFactory
         springMethods.exists(s => p.serviceMethod.methodName.equals(s.serviceMethod.methodName)))
     }
 
+    val serviceLevelOptions = {
```
franciscolopezsancho, line 213: `typeLevelOptions` vielleicht?

franciscolopezsancho: I'm failing to understand this change. How come we were only passing ` AclDescriptorFactory.serviceLevelAclAnnotation(component)` before but now all these options are needed?

samples/spring-eventsourced-customer-registry/src/main/java/customer/view/CustomerByEmailView.java
```diff
@@ -1,19 +1,21 @@
...
 import customer.api.CustomerEntity;
-import customer.api.CustomerEvent;
+import customer.domain.CustomerEvent;
 import kalix.javasdk.view.View;
 import kalix.springsdk.annotations.Query;
 import kalix.springsdk.annotations.Subscribe;
 import kalix.springsdk.annotations.Table;
+import reactor.core.publisher.Flux;
+
 import org.springframework.web.bind.annotation.GetMapping;
```
franciscolopezsancho, line 12: maybe add `@ViewId`? and if so in the rest of the views

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/AclDescriptorFactory.scala
franciscolopezsancho: and is easier to read

docs/src/modules/spring/pages/service-to-service.adoc
```diff
@@ -1 +1,47 @@
...
+[source,java,indent=0]
+.src/main/java/customer/api/CustomerEventsService.java
+----
+include::example$spring-eventsourced-customer-registry/src/main/java/customer/api/CustomerEventsService.java[tag=producer]
+----
+<1> `@Subscribe.EventSourcedEntity` identifying which event sourced entity to publish events for
+<2> Ignore any event types not handled by a method and move on with the event stream, rather than fail which is the default
+<3> `@Publish.Stream` marks the public identifier for consumers of this stream
+<4> An ACL annotation, allowing access from other Kalix services, but not the public internet
+<5> All methods on the service are transformer methods for turning internal event message types into public API message types for other services to consume
```
franciscolopezsancho, line 28:
```suggestion
----
<1> Identifying which event sourced entity to publish events for.
<2> Ignore any event types not handled by a method and move on with the event stream, rather than fail which is the default.
<3> Public identifier for consumers of this stream.
<4> Allowing access from other Kalix services, but not the public internet.
```

docs/src/modules/spring/pages/service-to-service.adoc
```diff
@@ -1 +1,47 @@
...
++
+[source,java,indent=0]
+.src/main/java/customer/views/CustomersByNameView.java
+----
+include::example$spring-eventsourced-customer-registry-subscriber/src/main/java/customer/views/CustomersByNameView.java[tag=view]
+----
+<1> Annotate the component with `@Subscribe.Stream` to subscribe to an event stream from another Kalix service.
+<2> The name of the Kalix service publishing the event stream.
+<3> The public identified of the specific stream from the publisher
+<4> One update method per message type that the stream may contain
```
franciscolopezsancho, line 47:
```suggestion
<3> The public identifier of the specific stream from the publisher.
<4> One update method per message type that the stream may contain.
```

franciscolopezsancho:
Looks pretty good to me. I really like the cleansing, that is awesome. We should make official such a task and add it to the board once a month or something like that.

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/ActionDescriptorFactory.scala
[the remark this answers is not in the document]
octonato: A method can be both of the them. 

But in the meantime, I can to the conclusion that PubSub is also not correct. 
What this `ServiceMethod` is all about is the fact that it has one single parameter of type `Any`.

It's possible to have many `Any` methods that end-up combined. When that happen, they are all about subscription. 

You can also have a publishing method with a REST annotation. In such case, it's a `ServiceMethod` with a synthetic request and it won't become a `PubSubServiceMethod`.

sdk/spring-sdk/src/main/java/kalix/springsdk/annotations/Publish.java
octonato: I split the annotations so the discussion here is not relevant anymore. We can revisit it with we take that path in the future.

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/ActionDescriptorFactory.scala
octonato: they all become gPRC service level annotations. 

Now, we can have ACL, but also eventing.in and eventing.out.direct at service level.

docs/src/modules/spring/pages/service-to-service.adoc
octonato: Thanks for the suggestion. I think it's better so.
However, point 5 is being dropped in this suggestion. So I will create another one to apply.

docs/src/modules/spring/pages/service-to-service.adoc
```diff
@@ -1 +1,47 @@
...
+
+[source,java,indent=0]
+.src/main/java/customer/api/CustomerEventsService.java
+----
+include::example$spring-eventsourced-customer-registry/src/main/java/customer/api/CustomerEventsService.java[tag=producer]
+----
+<1> `@Subscribe.EventSourcedEntity` identifying which event sourced entity to publish events for
+<2> Ignore any event types not handled by a method and move on with the event stream, rather than fail which is the default
+<3> `@Publish.Stream` marks the public identifier for consumers of this stream
+<4> An ACL annotation, allowing access from other Kalix services, but not the public internet
```
octonato, line 27:
```suggestion
<1> Identifying which event sourced entity to publish events for.
<2> Ignore any event types not handled by a method and move on with the event stream, rather than fail which is the default.
<3> Public identifier for consumers of this stream.
<4> Allowing access from other Kalix services, but not the public internet.
```

samples/spring-eventsourced-customer-registry/src/main/java/customer/view/CustomerByEmailView.java
octonato: done

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/ActionDescriptorFactory.scala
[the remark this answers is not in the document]
octonato: I have another branch with some more clean-up on this method. For the moment, I think we can go with this naming. I don't want to renamed more stuff in this PR.

sdk/spring-sdk/src/test/java/kalix/springsdk/testmodels/subscriptions/PubSubTestModels.java
```diff
@@ -93,6 +93,7 @@ public Action.Effect<Integer> methodTwo(Integer number) {
   }
 
   @Subscribe.EventSourcedEntity(value = CounterEventSourcedEntity.class, ignoreUnknown = true)
+  @Publish.Stream(id = "")
```
franciscolopezsancho, line 96: Is this necessary?

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/ComponentDescriptorFactory.scala
```diff
@@ -20,16 +20,24 @@ import java.lang.reflect.Method
...
 import kalix.MethodOptions
+import kalix.ServiceEventing
+import kalix.ServiceEventingOut
 import kalix.javasdk.action.Action
 import kalix.javasdk.view.View
 import kalix.springsdk.annotations.EntityType
 import kalix.springsdk.annotations.JWT
 import kalix.springsdk.annotations.Publish
+import kalix.springsdk.annotations.Publish
+import kalix.springsdk.annotations.Subscribe
```
franciscolopezsancho, line 40:
```suggestion
```

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/ComponentDescriptorFactory.scala
```diff
@@ -20,16 +20,24 @@ import java.lang.reflect.Method
 import java.lang.reflect.Modifier
 import java.lang.reflect.ParameterizedType
 
+import scala.collection.immutable.Map
```
franciscolopezsancho, line 23:
```suggestion
```

sdk/spring-sdk/src/test/java/kalix/springsdk/testmodels/subscriptions/PubSubTestModels.java
franciscolopezsancho, line 96:
```suggestion
```

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/ComponentDescriptorFactory.scala
```diff
@@ -179,6 +193,28 @@ private[impl] object ComponentDescriptorFactory {
...
   }
 
+  def eventingInForEventSourcedEntityServiceLevel(clazz: Class[_]): Option[kalix.ServiceOptions] = {
+    eventSourcedEntitySubscription(clazz).map { ann =>
+
+      val entityType = findEventSourcedEntityType(clazz)
+
+      val in = EventSource
+        .newBuilder()
+        .setIgnoreUnknown(ann.ignoreUnknown())
```
franciscolopezsancho, line 203: I remember not adding `ignoredUnknown` on the descriptor when working with subscriptions. I had to do this because otherwise, the proxy ignores all messages for the Spring SDK. I tried to find the source code of this in the proxy but couldn't find it. Could you keep me honest here @johanandren?

samples/spring-eventsourced-customer-registry-subscriber/src/main/java/customer/Main.java
```diff
@@ -0,0 +1,26 @@
...
+import org.springframework.boot.SpringApplication;
+import org.springframework.boot.autoconfigure.SpringBootApplication;
+import org.springframework.context.annotation.Import;
+
+@SpringBootApplication
+@Import(KalixConfiguration.class)
+// Allow all other Kalix services deployed in the same project to access the components of this
+// Kalix service, but disallow access from the internet. This can be overridden explicitly
+// per component or method using annotations.
+// Documentation at https://docs.kalix.io/services/using-acls.html
```
franciscolopezsancho, line 16: This link will be pointing to the Java docs.

sdk/spring-sdk/src/main/java/kalix/springsdk/annotations/Publish.java
```diff
@@ -32,4 +32,19 @@
     /** Assign the name of the topic to be used for eventing out. */
     String value();
   }
+
+
+  /**
+   * Annotation for marking a method as producing information to be published on a PubSub or Kafka
```
aludwiko, line 38: Direct stream, not PubSub or Kafka.

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/ActionDescriptorFactory.scala
```diff
@@ -183,11 +210,31 @@ private[impl] object ActionDescriptorFactory extends ComponentDescriptorFactory
...
+
+      val allOptions =
+        AclDescriptorFactory.serviceLevelAclAnnotation(component) ::
+        eventingInForEventSourcedEntityServiceLevel(component) ::
```
franciscolopezsancho, line 217: If I understood correctly now when annotating the class this has to be signaled in the gRPC annotation at the service level. Is this the case for all annotations? Because I'm missing here subscription to a topic.

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/ComponentDescriptorFactory.scala
```diff
@@ -205,6 +241,53 @@ private[impl] object ComponentDescriptorFactory {
...
+    Option(component.getAnnotation(classOf[Subscribe.Stream])).map { streamAnn =>
+      val direct = DirectSource
+        .newBuilder()
+        .setEventStreamId(streamAnn.id())
+        .setService(streamAnn.service())
+
+      val in = EventSource
+        .newBuilder()
+        .setDirect(direct)
+        .setIgnoreUnknown(streamAnn.ignoreUnknown())
```
franciscolopezsancho, line 254: Same comment as above about the  `ignoreUnknown`.

sdk/spring-sdk/src/test/scala/kalix/springsdk/impl/ViewDescriptorFactorySpec.scala
```diff
@@ -374,19 +375,38 @@ class ViewDescriptorFactorySpec extends AnyWordSpec with ComponentDescriptorSuit
...
         methodOptions.getEventing.getIn.getIgnore shouldBe false // we don't set the property so the proxy won't ignore. Ignore is only internal to the SDK
 
       }
     }
+
+    "generate mappings for service to service subscription " in {
+      assertDescriptor[EventStreamSubscriptionView] { desc =>
+
+        val serviceOptions = findKalixServiceOptions(desc)
+        val eventingInDirect = serviceOptions.getEventing.getIn.getDirect
```
franciscolopezsancho, line 397: Maybe adding `serviceOptions.getEventing.getIn.getIgnore shouldBe false`? This maybe not necessary but could be quite useful if it turns out we have to avoid passing `ignore = true` to the proxy. As I pointed out above. If that is the case could be nice to set the Publish.Stream to ignore true and verify here that is still false.

franciscolopezsancho:
In general looks to me like this is ready to merge. Except for minor details I have only two outstanding questions you may have already read in the comments. 
1. Can we pass ignore = true to the proxy?
2. Do we need to add to `serviceLevelOptions` the `ServiceOptions` for subscribing to topics as well?

If both answers are no, then LGTM right away.

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/ActionDescriptorFactory.scala
octonato: I didn't touch the topic part because not relevant for this feature. 

We should indeed ensure that a type level subscription to a topic becomes a gRPC service level annotation, but I prefer to do it later since not relevant for now.

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/ComponentDescriptorFactory.scala
octonato: Good catch! You are right. 

The ignore should be propagated to the proxy, but kept and handled on the SDK side. I will change it.

sdk/spring-sdk/src/test/java/kalix/springsdk/testmodels/subscriptions/PubSubTestModels.java
octonato: That was intentional. The goal was to validate that empty ID is not accepted, but that's probably something odd here. The class name is not exposing it. Either this is not being used or the validation was removed. 

I will check.

sdk/spring-sdk/src/main/java/kalix/springsdk/annotations/Publish.java
octonato:
```suggestion
   * Annotation to configure the component to publish an event stream to other Kalix services.
```

octonato: Just realised that it should not be possible to use this annotation in anything other than an Action and it should be combined with a subscription to an Event Sourced entity. 

Btw, the text here comes directly from the proto annotation in the proxy. I copied it blindly. 

That's also not correct. This is an annotation that is only valid at service level. We should not mention `method`.

sdk/spring-sdk/src/main/java/kalix/springsdk/annotations/Publish.java
```diff
@@ -32,4 +32,19 @@
...
+   * topic.
+   */
+  @Target(ElementType.TYPE)
+  @Retention(RetentionPolicy.RUNTIME)
+  @Documented
+  @interface Stream {
+    /**
+     * Identifier for the direct event stream. Must be unique inside the same Kalix service.
```
octonato, line 46:
```suggestion
     * Identifier for the event stream. Must be unique inside the same Kalix service.
```

efgpinto (commented):
I was mostly looking to check the main changes so not a very detailed review here but looks good overall.

samples/spring-eventsourced-customer-registry-subscriber/src/main/java/customer/Main.java
franciscolopezsancho: fine by me

franciscolopezsancho (approved):
LGTM

samples/spring-eventsourced-customer-registry/src/main/java/customer/api/CustomerEventsService.java
```diff
@@ -0,0 +1,29 @@
...
+@Subscribe.EventSourcedEntity(
+    value = CustomerEntity.class, // <1>
+    ignoreUnknown = true) // <2>
+@Publish.Stream(id = "customer_events") // <3>
+@Acl(allow = @Acl.Matcher(service = "*")) // <4>
+public class CustomerEventsService extends Action {
+
+  public Effect<CustomerPublicEvent.Created> onEvent( // <5>
+      CustomerEvent.CustomerCreated created) {
+//    var id = actionContext().eventSubject().get();
```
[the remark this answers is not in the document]
octonato: no, that was a test, I will remove

sdk/spring-sdk/src/main/java/kalix/springsdk/annotations/Publish.java
```diff
@@ -32,4 +32,19 @@
     /** Assign the name of the topic to be used for eventing out. */
     String value();
   }
+
+
+  /**
+   * Annotation for marking a method as producing information to be published on a PubSub or Kafka
+   * topic.
```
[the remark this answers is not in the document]
octonato: I thought I had updated this.

docs/src/modules/spring/pages/service-to-service.adoc
```diff
@@ -1 +1,47 @@
 = Service to Service Eventing
+
+include::ROOT:partial$include.adoc[]
+
+
+Kalix provides brokerless at-least-once event delivery across Kalix services through the Service to Service eventing.
+
+The source of the events is an xref:event-sourced-entities.adoc[Event Sourced Entity], that the service provides
+a Service to Service publisher for.
```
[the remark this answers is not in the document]
octonato: I think this is from the original Java docs.

Would it be more clear if we say?

> The source of the events is an xref:event-sourced-entities.adoc[Event Sourced Entity]. Its events are then published to another service through Service to Service eventing.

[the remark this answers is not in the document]
octonato: Reworked it here: [c6d0c19](https://github.com/lightbend/kalix-jvm-sdk/pull/1230/commits/c6d0c19717e85cd03ed32de31bd081c82173b847)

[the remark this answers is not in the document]
efgpinto: Nice 👍

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/ActionDescriptorFactory.scala
```diff
@@ -61,64 +86,81 @@ private[impl] object ActionDescriptorFactory extends ComponentDescriptorFactory
...
+      streamSubscription(component)
+        .map { ann =>
+          val kalixMethods =
+            component.getMethods
+              .filter(hasActionOutput)
+              .sorted // make sure we get the methods in deterministic order
+              .map { method => KalixMethod(SubscriptionServiceMethod(method)) }
+              .toSeq
+
+          val streamId = ann.id()
```
octonato, line 128: Here we will use the streamId as differentiator.

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/ActionDescriptorFactory.scala
```diff
@@ -61,64 +86,81 @@ private[impl] object ActionDescriptorFactory extends ComponentDescriptorFactory
...
-          }
-      } else Array.empty[KalixMethod]
+        val entityType = findEventSourcedEntityType(component)
+        Map(entityType -> kalixMethods)
+
+      } else Map.empty
 
-    val subscriptionTopicMethods = component.getMethods
+    val subscriptionStreamClass: Map[String, Seq[KalixMethod]] = {
+      streamSubscription(component)
```
octonato, line 119: This was missing. I mentioned today that subscribing with an action was not working. 

Well, the reason was that when use `@Subscribe.Stream`, the `ActionDescriptorFactory` won't generate any method in gRPC. It was only generating methods for `@Subscribe.EventSourcedEntity`, `@Subscribe.ValueEntity` and `@Subscribe.Topic`.

And the tests was not picking it because of a mistake and lack of some extra checks.

sdk/spring-sdk/src/test/scala/kalix/springsdk/impl/ActionDescriptorFactorySpec.scala
```diff
@@ -437,6 +441,46 @@ class ActionDescriptorFactorySpec extends AnyWordSpec with ComponentDescriptorSu
...
+        eventingInDirect.getService shouldBe "employee_service"
+        eventingInDirect.getEventStreamId shouldBe "employee_events"
+
+        // we don't set the property so the proxy won't ignore. Ignore is only internal to the SDK
+        eventingIn.getIgnore shouldBe false
+        eventingIn.getIgnoreUnknown shouldBe false
+
+        val methodDescriptor = findMethodByName(desc, "KalixSyntheticMethodOnESEmployee_events")
+        methodDescriptor.isServerStreaming shouldBe false
+        methodDescriptor.isClientStreaming shouldBe false
```
octonato, line 481: This was missing and we were not detecting that a gPRC method was not being generated. 

Note the name of the methods. Ends with `Employee_events'. Not very java style, but this name comes from the name of the stream. Chosen by the use, so not under our control. We could try to remove some symbols, but I don't think it's worth. That's hidden from users.

sdk/spring-sdk/src/test/java/kalix/springsdk/testmodels/subscriptions/PubSubTestModels.java
```diff
@@ -232,14 +232,16 @@ public Effect<String> transform(EmployeeEmailUpdated emailUpdated) {
 
     }
 
-    @Subscribe.EventSourcedEntity(EmployeeEntity.class)
```
octonato: This test model had two subscriptions. That's currently not legal, but we don't have proper validation yet. 

I have been experimenting with some validation, but will PR next week (hopefully).

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/ActionDescriptorFactory.scala
franciscolopezsancho: Nice

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/ComponentDescriptorFactory.scala
```diff
@@ -247,15 +338,14 @@ private[impl] object ComponentDescriptorFactory {
             (typeUrl, k.serviceMethod.javaMethodOpt.get)
           }.toMap
 
-        val subscriptions = kMethods.map(_.serviceMethod.asInstanceOf[SubscriptionServiceMethod])
         KalixMethod(
           CombinedSubscriptionServiceMethod(
             component.getName,
             "KalixSyntheticMethodOnES" + eventSourcedEntity.capitalize,
```
franciscolopezsancho, line 344: https://github.com/lightbend/kalix/issues/7911

sdk/spring-sdk/src/test/java/kalix/springsdk/testmodels/subscriptions/PubSubTestModels.java
franciscolopezsancho: added https://github.com/lightbend/kalix/issues/7912 for visibility

sdk/spring-sdk/src/test/scala/kalix/springsdk/impl/ActionDescriptorFactorySpec.scala
franciscolopezsancho: Not doing gives room to hyphens. Just added https://github.com/lightbend/kalix/issues/7911

PR #1233, opened 2022-11-03
docs: spring docs handling serialization

franciscolopezsancho:
References https://github.com/lightbend/kalix/issues/7727


docs/src/modules/spring/pages/serialization.adoc
```diff
@@ -1 +1,38 @@
 = Handling Serialization
+
+You do not need to handle serialization for the messages of an Event Source Entity or a Value Entity. On the other hand, Kalix exposes endpoints with JSON messages and you need to handle serialization of their inputs and outputs: if you are using Java classes you need to annotate them with the https://github.com/FasterXML/jackson-annotations#usage-general[proper Jackson annotation] but if you are using Java https://openjdk.org/jeps/395[record] no annotation is needed. 
+
+== When consuming Messages from Topics
+
+When consuming events from a topic deserialization works out of the box. By default Kalix serialize Java objects to JSON when publish to a topic, or when emitting events from an entity, whatever they are value changes or snapshots. Therefore it can deserialize them as JSON when consuming them.
```
octonato, line 7:
```suggestion
When consuming events from a topic deserialization works out of the box. By default Kalix serialize Java objects to JSON when publishing to a topic, or when emitting events from an entity, whatever they are value changes or snapshots. Therefore it can deserialize them as JSON when consuming them. 
```

I find this confusing. The title is about consuming message from topics, but we talk about emitting events, value changes and snapshots.

docs/src/modules/spring/pages/serialization.adoc
```diff
@@ -1 +1,38 @@
 = Handling Serialization
+
+You do not need to handle serialization for the messages of an Event Source Entity or a Value Entity. On the other hand, Kalix exposes endpoints with JSON messages and you need to handle serialization of their inputs and outputs: if you are using Java classes you need to annotate them with the https://github.com/FasterXML/jackson-annotations#usage-general[proper Jackson annotation] but if you are using Java https://openjdk.org/jeps/395[record] no annotation is needed.
```
octonato, line 3:
```suggestion
You do not need to handle serialization for the messages of an Event Source Entity or a Value Entity. On the other hand, Kalix exposes endpoints with JSON messages and you need to handle serialization of their inputs and outputs: if you are using Java classes you need to annotate them with the https://github.com/FasterXML/jackson-annotations#usage-general[proper Jackson annotation] but if you are using Java https://openjdk.org/jeps/395[record] no annotation is needed. 
```

Here we give the impression that only input and output matters. But events and state are also serializable. Users don't need to serialize and deserialize them themselves, but they do need to create classes that can be serializable. Either by using records or by annotating their domains with Jackson.

docs/src/modules/spring/pages/serialization.adoc
```diff
@@ -1 +1,38 @@
...
+include::example$spring-eventsourced-counter/src/main/java/com/example/actions/CounterTopicSubscriptionAction.java[tags=class]
+----
+<1> subscribing to the `ValueIncreased` from topic 'counter-events'. 
+<2> setting the method input type to the events produced by the counter.
+<3> any return is valid.
+<4> subscribing to the `ValueMultiplied` from topic 'counter-events'. 
+<5> setting the method input type to the events produced by the counter.
+<6> any return is valid.
+
+As we mentioned it works out of the box. No
```
octonato, line 23: No?

octonato (commented):
Left some comments. 

I think this page can talk a bit more about Jackson. Users need to be aware of it even if using java records make it super simple.

docs/src/modules/spring/pages/serialization.adoc
franciscolopezsancho: changed now

docs/src/modules/spring/pages/serialization.adoc
franciscolopezsancho: :) deleted

octonato (approved):
LGTM, with a small typo fix that I will apply myself.

PR #1235, opened 2022-11-03
docs(spring-sdk): add generating entity keys section

efgpinto:
Closes https://github.com/lightbend/kalix/issues/7825


docs/src/modules/spring/pages/value-entity.adoc
```diff
@@ -44,6 +44,18 @@ different existing entities within a Kalix application.
 <3> The `CounterEntity` class should extend `kalix.javasdk.valueentity.ValueEntity`.
 <4> The initial state of each counter is defined as 0.
 
+include::partial$entities-key.adoc[tag=intro]
```
octonato, line 47: It will be easier for us to find back this file if we call it `generate-entity-keys.adoc` instead.

octonato (approved):
LGTM, with one small change request.

docs/src/modules/spring/pages/value-entity.adoc
efgpinto: indeed, tks

PR #1236, opened 2022-11-03
feat: handle deletes in views

aludwiko:
References https://github.com/lightbend/kalix/issues/7773

### DIFF
--- a/sdk/spring-sdk/src/main/java/kalix/springsdk/annotations/Subscribe.java
+++ b/sdk/spring-sdk/src/main/java/kalix/springsdk/annotations/Subscribe.java
@@ -28,14 +28,7 @@ public @interface Subscribe {
    * Annotation for subscribing to updates from a Value Entity. It can be used both at type and
    * method levels. When used at type level, it means the `View` will not be transforming state.
    * When used at method level, it gives the ability to transform the updates into a different state
-   * for the view and the underlying method must be declared to receive one or two parameters:
-   *
-   * <ul>
-   *   <li>when one parameter is passed, the single parameter will be considered the event type such
-   *       method will handle;
-   *   <li>when two parameters are passed, the first one will be considered the view state and the
-   *       second one the event type.
-   * </ul>
+   * for the view or delete the view state (handleDeletes flag).
    */
   @Target({ElementType.TYPE, ElementType.METHOD})
   @Retention(RetentionPolicy.RUNTIME)
@@ -46,6 +39,23 @@ public @interface Subscribe {
      * kalix.javasdk.valueentity.ValueEntity ValueEntity}.
      */
     Class<? extends kalix.javasdk.valueentity.ValueEntity<?>> value();
+
+    /**
+     * When true at type level the subscription will automatically delete the view state based on ValueEntity deletion fact.
+     * When true at method level it allows to create a special handler for deletes (must be declared to receive zero parameters):
+     * <pre>{@code
+     * @Subscribe.ValueEntity(MyValueEntity.class)
+     * public UpdateEffect<MyView> onChange(ValueEntity valueEntity) {
+     *   return effects().updateState(...);
+     * }
+     *
+     * @Subscribe.ValueEntity(value = MyValueEntity.class, handleDeletes = true)
+     * public UpdateEffect<MyView> onDelete() {
+     *   return effects().deleteState();
+     * }
+     * </pre>
+     */
+    boolean handleDeletes() default false;
   }
 
   /**
--- a/sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/ComponentDescriptor.scala
+++ b/sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/ComponentDescriptor.scala
@@ -17,7 +17,9 @@
 package kalix.springsdk.impl
 
 import java.lang.reflect.Type
+
 import scala.jdk.CollectionConverters.CollectionHasAsScala
+
 import com.google.api.AnnotationsProto
 import com.google.api.CustomHttpPattern
 import com.google.api.HttpRule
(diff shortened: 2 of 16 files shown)


sdk/spring-sdk/src/it/http/test-requests.http
```diff
@@ -0,0 +1,14 @@
+POST {{url}}/user/{{userId}}/email/andrzej
+Content-Type: application/json
```
aludwiko, line 2: I can remove it, but it's pretty convenient to have some predefined requests directly in IDE.

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/ComponentDescriptor.scala
```diff
@@ -58,6 +62,8 @@ import org.springframework.web.bind.annotation.RequestMethod
  */
 private[impl] object ComponentDescriptor {
 
+  val ProtobufEmptyTypeUrl = s"type.googleapis.com/google.protobuf.Empty"
```
aludwiko, line 65: No idea where to put this constant :)

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/ViewDescriptorFactory.scala
```diff
@@ -255,19 +265,45 @@ private[impl] object ViewDescriptorFactory extends ComponentDescriptorFactory {
...
+          case _ => // happy days, dev did good with the signature
+        }
+
+        val methodOptionsBuilder = kalix.MethodOptions.newBuilder()
+        methodOptionsBuilder.setEventing(eventingInForValueEntityHandleDeletes(method))
+        addTableOptionsToUpdateMethod(tableName, tableProtoMessageName, methodOptionsBuilder, transform = true)
+
+        KalixMethod(HandleDeletesServiceMethod(method))
+          .withKalixOptions(methodOptionsBuilder.build())
+          .withKalixOptions(buildJWTOptions(method)) //TODO JWT needed?
```
aludwiko, line 303: Not sure about JWT options.

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/ComponentDescriptor.scala
aludwiko: 👍

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/ViewDescriptorFactory.scala
johanandren: Not entirely sure the return of the method would be hooked up to view updates, probably not, if not, i am not sure what would happen when opening up the ACL and calling from the outside. The view update effect can't become a response automagically.

sdk/spring-sdk/src/it/http/test-requests.http
efgpinto: You can also use a readme with curl commands inside `shell` code blocks and it's similarly convenient (you need to repeat host and port I guess). We have been using that I think, [example](https://github.com/lightbend/kalix-jvm-sdk/blob/main/samples/java-valueentity-shopping-cart/README.md).

sdk/spring-sdk/src/test/java/kalix/springsdk/testmodels/view/ViewTestModels.java
```diff
@@ -149,6 +171,122 @@ public TransformedUser getUser(@RequestBody ByEmail byEmail) {
...
+
+    @Query("SELECT * FROM users_view WHERE email = :email")
+    @PostMapping("/users/by-email")
+    public TransformedUser getUser(@RequestBody ByEmail byEmail) {
+      return null;
+    }
+  }
+
+  @Table("users_view")
+  public static class ViewWithMissingSubscriptionForHandleDeletes extends View<TransformedUser> {
```
octonato, line 263: We can drop this case for now.

When we allow multiple subscriptions, then we will need a new set of rules. 

* All subscriptions must be at method level when subscribing to more than one entity. 
* The number of allowed `HandleDeletes` should equal the number of unique value entity types that are being subscribed to.
* There must be no more than one `HandleDelete` per entity (that's a relevant check already)

octonato (commented):
Looking good. 

I left some comments.

sdk/spring-sdk/src/it/http/test-requests.http
aludwiko: Interesting, I wasn't aware of that, thanks!

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/ViewDescriptorFactory.scala
aludwiko: Yeah, personally I wouldn't expose subs methods at all, because it could lead to a very nasty problems with consistency.  At the same time, I know that it might be handy to use them "for a good purpose". Tricky.

sdk/spring-sdk/src/main/java/kalix/springsdk/annotations/Subscribe.java
```diff
@@ -28,14 +28,8 @@
...
-   * for the view and the underlying method must be declared to receive one or two parameters:
-   *
-   * <ul>
-   *   <li>when one parameter is passed, the single parameter will be considered the event type such
-   *       method will handle;
-   *   <li>when two parameters are passed, the first one will be considered the view state and the
-   *       second one the event type.
-   * </ul>
+   * for the view or delete the view state (handleDeletes flag).
+   * The underlying method must be declared to receive zero (handle deletes) or one (handle update) parameters.
```
octonato, line 32: I feel that the last sentence is not needed. It's already explained on the `handleDeletes` field.

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/reflection/KalixMethod.scala
```diff
@@ -128,6 +119,47 @@ case class SubscriptionServiceMethod(javaMethod: Method) extends AnyJsonRequestS
...
+  def buildPathTemplate(componentName: String, methodName: String): String = {
+    val cls = componentName.replace("$", ".")
+    s"/$cls/${methodName.capitalize}" // FIXME: must pass through NameGenerator?
+  }
+}
+
+/**
+ * Additional trait to simplify pattern matching for actual and virtual delete service method
+ */
+trait DeleteServiceMethod extends ServiceMethod {}
```
octonato, line 132:
```suggestion
trait DeleteServiceMethod extends ServiceMethod
```

octonato (approved):
LGTM, so suggestions and minor change request.

Docs are missing, but Cisco is working on the View docs. Can be added after his PR gets merged

PR #1241, opened 2022-11-04
docs(spring-sdk): add actions as controllers page

efgpinto:
closes https://github.com/lightbend/kalix/issues/7720


samples/spring-valueentity-shopping-cart/src/main/java/com/example/ShoppingCartController.java
```diff
@@ -0,0 +1,106 @@
...
+import org.springframework.web.bind.annotation.PostMapping;
+import org.springframework.web.bind.annotation.RequestBody;
+import org.springframework.web.bind.annotation.RequestMapping;
+
+import java.util.List;
+import java.util.UUID;
+import java.util.concurrent.CompletionStage;
+
+@RequestMapping("/carts")
+public class ShoppingCartController extends Action {
```
aludwiko, line 21: Shouldn't this be a part of the api package?

samples/spring-valueentity-shopping-cart/src/main/java/com/example/ShoppingCartEntity.java
```diff
@@ -0,0 +1,102 @@
...
+
+import java.time.Instant;
+
+/**
+ * A value entity.
+ */
+@EntityKey("cartId")
+@EntityType("shopping-cart")
+@RequestMapping("/cart/{cartId}")
+public class ShoppingCartEntity extends ValueEntity<ShoppingCart> {
```
aludwiko, line 34: Same here, api?

samples/spring-valueentity-shopping-cart/src/main/java/com/example/ShoppingCartController.java
efgpinto: it should, tks 🙇

samples/spring-valueentity-shopping-cart/src/main/java/com/example/ShoppingCartEntity.java
efgpinto: tks

efgpinto:
Ready for review now. Includes new spring valueentity shopping cart sample used in docs.

docs/src/modules/spring/pages/actions-as-controller.adoc
```diff
@@ -1 +1,112 @@
...
+forwarded operation must have the same response type as the original command.
+
+In this example we accept requests with the same structure as the entity, by receiving a `LineItem`, but add some additional verification of the
+request and only conditionally forward the request to the entity if the verification is successful:
+
+
+.src/main/java/com/example/api/ShoppingCartController.java
+[source,java]
+----
+include::example$spring-valueentity-shopping-cart/src/main/java/com/example/api/ShoppingCartController.java[tag=forward]
```
franciscolopezsancho, line 22: I think it would be nice to explain what is the `ActionCreationContext creationContext` that appears on the constructor.  Especially because `kalixContext` maps to a variable but the `creationContext` doesn't.

docs/src/modules/spring/pages/actions-as-controller.adoc
```diff
@@ -1 +1,112 @@
...
+----
+<1> `KalixClient` is injected on the constructor. It will be used to build calls to the underlining Entity.
+<2> Expose the command handler as a `POST` endpoint at specified path.
+<3> Check if the added item is carrots.
+<4> If it is "carrots" immediately return an error, disallowing adding the item.
+<5> For allowed requests, use `kalixClient` to get a `DeferredCall`.
+<6> The `DeferredCall` is used with `effects().forward()` to forward the request to the entity.
```
franciscolopezsancho, line 29: Would it be better to give some context of what a `DeferredCall` is?

docs/src/modules/spring/pages/actions-as-controller.adoc
```diff
@@ -1 +1,112 @@
...
+acting as the external interface of a service, receiving requests, operating over the requests values and forwarding the call
+to other components in the same service.
+
+To illustrate how you can use an Action as a Controller, we will build on top of the xref:structuring-your-application.adoc[Value Entity Shopping Cart] example, adding a new Action to the existing shopping cart service.
+
+== Forwarding Commands
+
+The `forward` effect allows us to transform or further validate an incoming request before passing it on to another
+component and have the response message directly passed back to the client making the request. The response from the
+forwarded operation must have the same response type as the original command.
```
franciscolopezsancho, line 13: I understand this is the same text as the original. I struggled a bit to understand it so here is a suggestion. 
`The `forward` effect allows you to call another
component and have the response message passed back to the method forwarding.  The response from the
forwarded operation must have the same response type as the original method. You can transform or further validate an incoming request before passing it on to another component`

I'm not sure we should have different info than the java docs but at the same time I think is a good opportunity to improve the docs. You may want to check with Renato if you are interested in doing such a change.

docs/src/modules/spring/pages/actions-as-controller.adoc
```diff
@@ -1 +1,112 @@
...
+<5> For allowed requests, use `kalixClient` to get a `DeferredCall`.
+<6> The `DeferredCall` is used with `effects().forward()` to forward the request to the entity.
+
+NOTE: You might be wondering what the `kalixClient` is about. For now, think of it as a lightweight HTTP client allowing you to reach out to other Kalix services. All details can be found at xref:call-another-service.adoc[] chapter.
+
+== Transform Request and Response to Another Component
+
+The `asyncReply` and `asyncEffect` effects allow us to process and transform a request before calling another component and then also transform the response.
+
+As an example, let us look at the problem of creating a new entity with an id generated by the service itself.
```
franciscolopezsancho, line 37:
```suggestion
As an example, let us look at the problem of creating a new entity with an id generated by another component.
```

docs/src/modules/spring/pages/actions-as-controller.adoc
```diff
@@ -1 +1,112 @@
...
+<7> `effects().asyncEffect()` allows us to reply with a `CompletionStage<Effect<String>>`.
+
+The action generates a link:https://en.wikipedia.org/wiki/Universally_unique_identifier[UUID] to use as entity id for the shopping cart. UUIDs are extremely unlikely to lead to the same id
+being generated, but to completely guarantee two calls can never be assigned the same shopping cart we
+make use of the "boundary of consistency" provided by the entity - the entity will only process a single command at a time
+and can safely make decisions based on its state - for example to only allow creation once by storing something in its state
+signifying that it has been created.
+
+In this case we mark that the entity has been created using a creation timestamp in the shopping cart state stored on first
+`create` call (when the timestamp has the default value of `0`). If the cart has already been stored with a timestamp we return an error effect:
```
franciscolopezsancho, line 62:
```suggestion
`create` call - the timestamp has the default value of `0`. If the cart has already been stored with a timestamp we return an error effect:
```

franciscolopezsancho:
It looks pretty good to me. I left some sparsed comments you may find useful.

Two general comments apart from the previous comments. 
- There are 'wes' instead of 'yous' :) 
- I think it would be beneficial to add the imports on top of the classes. Especially if they are from Kalix basic components.

docs/src/modules/spring/pages/actions-as-controller.adoc
efgpinto: I think it should actually work without it, I'll try removing

franciscolopezsancho: good idea

docs/src/modules/spring/pages/actions-as-controller.adoc
franciscolopezsancho: yeah, I think it helps

docs/src/modules/spring/pages/actions-as-controller.adoc
franciscolopezsancho: > > return the response message to the forwarding method before passing it back to the client making the request
> 
> Don't think this is accurate actually.. because the forward doesn't allow you to transform the response, so effectively the response is passed directly to the client making the request and not to the forwarding method.

Good point. I guess that depends on what before means here. Before in space or in time? (we are doing philosophy now:) Meaning space as in the code, and time as in runtime. I meant in time but I see now it can be interpreted as in the code. In any case, it was a mild suggestion, I guess the original text is just okay the way it is.

docs/src/modules/spring/pages/actions-as-controller.adoc
efgpinto: for the record, I did this change but also for the java counterpart

docs/src/modules/spring/pages/actions-as-controller.adoc
efgpinto: for the record, I did this change but also for the java counterpart

efgpinto:
> * There are 'wes' instead of 'yous' :)
> 
> * I think it would be beneficial to add the imports on top of the classes. Especially if they are from Kalix basic components.

Missed this previously. Will have a look now. :)

efgpinto:
Updated now @franciscolopezsancho, let me know if there's anything else you would prefer to be different. :)

docs/src/modules/spring/pages/actions-as-controller.adoc
efgpinto: Oh I see, makes sense. :) 

Yeah, I couldn't improve it much further I think.. Maybe we can revisit in the future (famous last words).

franciscolopezsancho: hehehe okay, let's leave it like it is then.

docs/src/modules/spring/pages/actions-as-controller.adoc
```diff
@@ -1 +1,112 @@
...
+<7> `effects().asyncEffect()` allows us to reply with a `CompletionStage<Effect<String>>`.
+
+The action generates a link:https://en.wikipedia.org/wiki/Universally_unique_identifier[UUID] to use as entity id for the shopping cart. UUIDs are extremely unlikely to lead to the same id
+being generated, but to completely guarantee two calls can never be assigned the same shopping cart we
+make use of the "boundary of consistency" provided by the entity - the entity will only process a single command at a time
+and can safely make decisions based on its state - for example to only allow creation once by storing something in its state
+signifying that it has been created.
+
+In this case you mark that the entity has been created using a creation timestamp in the shopping cart state stored on first
+`create` call - when the timestamp has the default value of `0`). If the cart has already been stored with a timestamp it returns an error effect:
```
franciscolopezsancho, line 62:
```suggestion
`create` call - when the timestamp has the default value of `0`. If the cart has already been stored with a timestamp it returns an error effect:
```

Also question. Where is this timestamp with the default value of 0? Couldn't find it.

samples/spring-valueentity-shopping-cart/src/main/java/com/example/api/ShoppingCartController.java
```diff
@@ -0,0 +1,114 @@
+package com.example.api;
+
+import com.example.api.ShoppingCartDTO.LineItemDTO;
+// tag::forward[]
+import kalix.javasdk.action.Action;
+import kalix.springsdk.KalixClient;
```
franciscolopezsancho, line 6:
```suggestion
import kalix.springsdk.KalixClient;

```

franciscolopezsancho (commented):
Except for that minor question LGTM

docs/src/modules/spring/pages/actions-as-controller.adoc
efgpinto: Nice catch!

> Also question. Where is this timestamp with the default value of 0? Couldn't find it.

On the domain model, one of the last lines.. we use that static builder on the `emptyState` override.

efgpinto:
Updated now with the added context for the value entity shopping cart we had agreed on @franciscolopezsancho . Also, addressed your suggestions.

franciscolopezsancho:
It looks pretty good to me. I'm not sure you saw the question I posted. I should have included it in its own comment. https://github.com/lightbend/kalix-jvm-sdk/pull/1241#discussion_r1019957541

franciscolopezsancho:
Oh sorry. I should have refreshed my page. Then is all good. I think it is happening the same to you that happened to me a minute ago. You need to bump the SDK to 1.0.11.

franciscolopezsancho (approved):
LVGTM

PR #1245, opened 2022-11-08
feat: codegen Scala support for handle deletes

aludwiko:


sdk/scala-sdk/src/main/scala/kalix/scalasdk/valueentity/ValueEntity.scala
```diff
@@ -34,7 +34,7 @@ object ValueEntity {
 
       def updateState(newState: S): OnSuccessBuilder[S]
 
-      def deleteState: OnSuccessBuilder[S]
+      def deleteState(): OnSuccessBuilder[S]
```
aludwiko, line 37: not sure how this will affect existing users, but it's a side effect function, so `()`

johanandren: Compiler warning (or possibly error) saying don't do non-param call to method with empty params, I think that is ok. 

(I'm fine with this change but It's not a side effect function though, it returns an effect saying "delete it", it doesn't actually do the delete)

octonato: They will get a warning only, unless they are using a strict compiler configuration.

Not big issue, let's just do it.

project/ReflectiveCodeGen.scala
```diff
@@ -38,6 +38,9 @@ object ReflectiveCodeGen extends AutoPlugin {
             gen(
               akkaGrpcCodeGeneratorSettings.value ++ Seq("enable-debug", "flat_package"),
               "kalix.codegen.scalasdk.KalixGenerator$") -> (Compile / sourceManaged).value,
+            gen(
+              akkaGrpcCodeGeneratorSettings.value ++ Seq("enable-debug", "flat_package"),
+              "kalix.codegen.scalasdk.KalixTestGenerator$") -> (Test / sourceManaged).value,
```
johanandren, line 43: 👍

octonato (approved):

project/Dependencies.scala
```diff
@@ -132,7 +132,7 @@ object Dependencies {
   // FIXME
   val sdkScala = deps ++= coreDeps ++ Seq(jacksonScala)
 
-  val sdkScalaTestKit = deps ++= Seq(testContainers, logback % "test;provided", scalaTest % Test)
+  val sdkScalaTestKit = deps ++= Seq(testContainers, logback % "test;provided", scalaTest)
```
johanandren, line 135: Do we require scalatest for testing with the testkit? Isn't it just that we _use_ it ourselves (so not a transitive dependency by default, as before).

johanandren (commented):
Change itself looks good, some comments on the semi-related changes.

project/Dependencies.scala
aludwiko: Actually, this change is for `codegenScalaCompilationTest` which depends on `sdkScalaTestKit` where we want to compile the test code together with the main code. At least, it's how I understand it. A similar story was with the junit dependency in `Dependencies#sdkJavaTestKit`.

johanandren: Shouldn't we add the non test scoped ScalaTest dependency to that project instead in that case?

octonato: Yeah, indeed. 

The java testkit has junit in provided scope, but it does have a dependency on teskit. The scala testkit production code has not dependency on scalatest at all.

johanandren (approved):

project/Dependencies.scala
aludwiko: Actually, I don't need to change the scope of this dependency. Somehow it is there with the correct scope already.

PR #1246, opened 2022-11-08
docs: spring docs acls #7731

franciscolopezsancho:
References https://github.com/lightbend/kalix/issues/7731


docs/src/modules/spring/pages/using-acls.adoc
```diff
@@ -0,0 +1,192 @@
...
+
+Access Control Lists (ACLs) allow you to control which services and sources of requests may access your services.
+
+Within a Kalix project, all communication between services uses Mutual TLS (mTLS). This is injected transparently by Kalix. The clients you use to make calls from one service to another do NOT need to be configured to do this. Kalix transparently captures outgoing requests to other services and wraps them in an mTLS connection.
+
+Based on this mTLS support, Kalix is able to read and apply policies based on where the request was made from.
+
+== Principals
+
+A principal in Kalix is an abstract concept that represents anything that can make or be the source of a request. Principals that are currently supported by Kalix include other services, and the internet. Kalix uses the above described mTLS support to associate requests with one or more principals.
```
franciscolopezsancho, line 13: But then we have `service` and `principal` as settings. So it seems that the category Principal is divided in two. One is called `principal` and the other `service`. It seems it's difficult to know when one talks about the principal to which one is referring.

docs/src/modules/spring/pages/using-acls.adoc
```diff
@@ -0,0 +1,192 @@
...
+
+Note that requests that have the internet principal are requests that Kalix has identified as coming through the Kalix ingress, according to a configured route. This is identified by mTLS, however it does not imply that mTLS has been used to connect to the ingress from the client in the internet. These are separate hops. To configure mTLS from internet clients, see https://docs.kalix.io/services/using-client-certificates.html[Client certificates]
```
franciscolopezsancho, line 15: what does it mean by Kalix ingress?

docs/src/modules/spring/pages/using-acls.adoc
```diff
@@ -0,0 +1,192 @@
...
+      return ...;
+    }
+}
+```
+
+The `MyMethod` rpc method will allow calls by `service-b`, but not by `service-a`.
+
+=== Configuring the default policy
+
+The default policy can be configured by specifying a project level annotation in the `Main`, for example, to set a default policy of allowing all local services:
```
franciscolopezsancho, line 87: what means project and service here? I think this issue has been raised before. The distinction between, 
1. project as deployable, 
2. project as project id in Kalix
3. service as project deployable 
4. service as component

I'm quite sure this list is wrong. Shall we try to standardize it and maybe add it to the glossary?

franciscolopezsancho: I mean. Not right now, but add an issue maybe?

docs/src/modules/spring/pages/using-acls.adoc
```diff
@@ -0,0 +1,192 @@
...
+```java
+@Acl(allow = @Acl.Matcher(service = "*"))
+public class Main { ... } 
+```
+
+An ACL declared at the project level is used as the default for all services that don't declare their own explicit ACL.
+
+=== Customizing the deny code
+
+When a request is denied, by default, a gRPC error code of 7, `PERMISSION_DENIED`, is sent. This gets transcoded to an HTTP status code of 403, forbidden. The code that is returned when a request is denied can be customised using the `deny_code` property. The deny code must be a valid gRPC error code, which is any integer from 1 to 16, described https://grpc.github.io/grpc/core/md_doc_statuscodes.html[here]. Each one maps as follows to an https://doc.akka.io/japi/akka-http/current/akka/http/scaladsl/model/StatusCodes.html[Akka HTTP status code].
```
franciscolopezsancho, line 98: The user will be better off not knowing about gRPC codes. I put some thoughts here https://github.com/lightbend/kalix/issues/7901

docs/src/modules/spring/pages/access-control.adoc
```diff
@@ -1 +1,36 @@
...
+
+Accessing the principal of a call inside a service is possible through the request metadata `Metadata.principals()`. The `Metadata` for a call is available through the context (`actionContext`, `commandContext`) of the component.
+
+== ACLs when running unit tests
+
+In the generated unit test testkits, the ACLs are ignored.
+
+== ACLs when running integration tests
+
+When running integration tests, ACLs are disabled by default but can be explicitly enabled per test by running the test with `Settings.withAclEnabled`.
```
octonato, line 34: I just realised that the testkit for the Spring SDK won't allow us to tweak this setting.

octonato: We will need to change the testkit to allow users to enable it and also to allow users to set the `serviceName`. So, for now, we can only state that ACLs are disabled for integration tests. 

On follow-up PR we can change the testkit and then add these sentences back.

docs/src/modules/spring/pages/access-control.adoc
```diff
@@ -1 +1,36 @@
 = Access Control Lists (ACLs)
+
+This section describes functionality in Spring SDK related to ACLs (Access Control Lists), for information about how the ACLs are defined and control access see xref:spring:using-acls.adoc[Using ACLs]
```
octonato, line 3: I would merge `Using ACLs` inside this page. We don't need to have two pages.

docs/src/modules/spring/pages/using-acls.adoc
octonato: That comes from the java docs, no? Let's leave it as is.

docs/src/modules/spring/pages/using-acls.adoc
octonato: I agree, but that was defined as such in gRPC. At the end, 'service' is also a principal. Confusing, but we will have to go with it for now.

docs/src/modules/spring/pages/using-acls.adoc
octonato: Indeed, here is project as deployable.

docs/src/modules/spring/pages/using-acls.adoc
octonato: We should go with that for now. There is already a tickets for that:
https://github.com/lightbend/kalix/issues/7743

octonato (commented):
Looking good. 

Left some comments. Some minor adjustments only.

docs/src/modules/spring/pages/using-acls.adoc
franciscolopezsancho: yes. Okay.

docs/src/modules/spring/pages/using-acls.adoc
franciscolopezsancho: no need :)

docs/src/modules/spring/pages/access-control.adoc
```diff
@@ -1 +1,212 @@
...
+
+```java
+    @PostMapping
+    @Acl(
+       allow = @Acl.Matcher(service = "*"), 
+       deny =  @Acl.Matcher(service = "my-service"))
+    public Effect<String> createUser(@RequestBody CreateEmployee create) {
+     ...
+    }
+```
```
octonato, line 29: We should have all those snippets in a sample so we can prove that they are compiling correctly.

Isn't the gRPC one using snippets?

franciscolopezsancho: gRPC is not using snippets.

franciscolopezsancho: I'll add them np

.circleci/config.yml
```diff
@@ -1184,6 +1184,23 @@ jobs:
...
+      image: ubuntu-2004:202201-02
+    steps:
+      - checkout-and-merge-to-main
+      - setup_sbt
+      - restore_deps_cache
+      # note: `copy-from-workspace` requires `publish-local` to be run first as declared in workflow section below
+      - copy-from-workspace
+      - set-sdk-version
+      - run:
+          name: Spring SDK Reliable Timers
```
efgpinto, line 1198:
```suggestion
          name: Spring SDK ACL Examples
```

.circleci/config.yml
```diff
@@ -1184,6 +1184,23 @@ jobs:
...
+          command: |
+            cd samples/spring-acl-examples
+            echo "Running mvn with SDK version: '$SDK_VERSION'"
+            mvn -Dkalix-sdk.version=$SDK_VERSION compile 
+
```
efgpinto, line 1203: I think you're missing a change to make this job run... somewhere closer to the end of this file.

samples/spring-acl-examples/src/main/java/com/example/MyAction.java
```diff
@@ -0,0 +1,63 @@
...
+    // tag::acl6[]
+    @PostMapping
+    @Acl(allow = @Acl.Matcher(service = "*"), denyCode = 5)
+    public Effect<String> updateUser(@RequestBody CreateUser create) {
+        //...
+        // end::acl6[]
+        return null;
+        // tag::acl6[]
+    }
+    // end::acl6[]
```
efgpinto, line 58: nitpicking here so feel free to ignore.. could we maybe use names for tags that describe the sample? e.g. `acl-deny-code`

docs/src/modules/spring/pages/access-control.adoc
```diff
@@ -1 +1,201 @@
...
+
+[source, java]
+----
+include::example$spring-acl-examples/src/main/java/com/example/UserAction.java[tag=acl]
+----
+
+== ACLs on eventing methods
+
+
+Any method with an `@Subscribe` annotation on it will not automatically inherit either the default or its component's ACL, rather, all outside communication will be blocked, since it's assumed that a method that subscribes to an event stream must only be intended to be invoked in response to events on that stream. This can be overridden by explicitly defining an ACL on that method:
```
efgpinto, line 136: Missing example here or wrong ponctuation? since it says: `by explicitly defining an ACL on that method:`

docs/src/modules/spring/pages/access-control.adoc
```diff
@@ -1 +1,201 @@
...
+    SERVICE_NAME: my-service-name
+```
+
+Note that in local development, the services don't actually authenticate with each other, they only pass their identity in a header. It is assumed in local development that a client can be trusted to set that header correctly.
+
+== Programmatically accessing principals
+
+The current principal associated with a request can be accessed by reading metadata headers. If the request came from another service, the `_kalix-src-svc` header will be set to the name of the service that made the request. Kalix guarantees that this header will only be present from an authenticated principal, it can't be spoofed.
+
+For internet, self and backoffice requests, the `_kalix-src` header will be set to `internet`, `self` and `backoffice` respectively. Backoffice requests are requests that have been made using the `kalix service proxy` command, they are authenticated and authorized to ensure only developers of your project can make them.
```
efgpinto, line 185: I think all of this could be moved to a partial and shared with the java/scala sdk right? Seems exactly the same and can't see it changing any time...

docs/src/modules/spring/pages/access-control.adoc
efgpinto: Maybe create a issue for this so we don't forget about it?

efgpinto (commented):
Looking good overall, left some comments.. I noticed we don't seem to have validation of ACLs for the spring-sdk.. So definitely support adding a IT tests suite to the sample and run them on CI once we support enabling the ACLs.

docs/src/modules/spring/pages/access-control.adoc
franciscolopezsancho: done

docs/src/modules/spring/pages/access-control.adoc
franciscolopezsancho: https://github.com/lightbend/kalix/issues/7906

samples/spring-acl-examples/src/main/java/com/example/MyAction.java
franciscolopezsancho: done

docs/src/modules/spring/pages/access-control.adoc
franciscolopezsancho: Good catch. Added example.

docs/src/modules/spring/pages/access-control.adoc
franciscolopezsancho: As a principle is a good idea the problem is that this content reappears in the `kalix-docs`, not in the `kalix-jvm-sdk/docs/java`. The Spring ACL doc is a mix of those two. Making a partial here wouldn't have any benefit. There is no other place to share it.

efgpinto: Oh I see.. But in that case, this is an example of a kalix-docs section that will need to be moved down.. I mean it doesn't make sense anymore to have the top-level ["Using ACLs"](https://docs.kalix.io/services/using-acls.html) which mentions gRPC... it needs to become one for Java/Scala and Javascript and another for Spring I guess. And in that case, this "Local development with ACLs" can still be shared. We'll probably need to sync with @beritou on that maybe

efgpinto: Fair enough 👍

efgpinto (approved):
LGTM as long as CI passes. 😁

franciscolopezsancho:
heheh cool! CI is convinced that the pom for ACLs project has the `1.0.10` SDK version in there. I'll try to convince it otherwise.

PR #1253, opened 2022-11-09
docs: handle deletes in scala and java proto

aludwiko:


octonato (commented):
Didn't review yet, but we can cover this functionality on the main docs (views.adoc) and leave it out form the quickstart. 

Quickstart should stay minimalist.

aludwiko:
Well, I can revert the code but, from my perspective, it stays minimalistic with deletes. VE = CRUD entity and the letter D was missing. We could argue if we should mention deletes in Views in the quickstart examples. The problem is that the first place where you can read about `transform_updates: true` in the context of Views based on VE is here https://docs.kalix.io/java/quickstart/cr-value-entity-views-java.html#_define_the_customerbyname_view so my plan was to extend a bit this part. Unfortunately, in the "main" docs about Views, `transform_updates: true` is in the section with Views based on ES entities. I think that now I understand your concerns about docs in general and why they should be rewritten.

efgpinto (approved):
Probably good to wait for Renato anyway but this LGTM.

johanandren (approved):

PR #1255, opened 2022-11-10
feat(spring-sdk): support return of collection on view queries

efgpinto:
Refs https://github.com/lightbend/kalix/issues/7463

The docs and samples for this will be added under https://github.com/lightbend/kalix/issues/7884 since we will need a new sample. Using this feature requires the proxy version currently in main but unreleased, but since we are not adding docs yet, I think this can be merged anyway.

Example of usage:

```java
public record CustomerViewCollection(Collection<CustomerView> customers, int total)
```

```java
@GetMapping("/customer/by_name/{customer_name}")
@Query(value = "SELECT * AS customers FROM customers_by_name, count(*) as total WHERE name = :customer_name")
public CustomerViewCollection getCustomer(@PathVariable String customer_name) {
  return null;
}
```


efgpinto:
Here's an example of a generated descriptor:
```proto
message CustomerViewCollection {
  repeated google.protobuf.Any results = 1;
}

service CustomerByNameView {
  rpc GetCustomer(GetCustomerKalixSyntheticRequest) returns (CustomerViewCollection) {
    option (kalix.method) = {
      view {
        query {
             query: "SELECT * AS results FROM customers_by_name WHERE name = :customer_name"
        }
        json_schema {
           output: "CustomerView"
        }
      }
    };
    option (google.api.http) = {
      get: "/customer/by_name/{customer_name}"
    };
  }
//...
```

This is currently not being accepted by the proxy:
```shell
2022-11-10 17:05:08,223 ERROR kalix.javasdk.impl.DiscoveryImpl - Error reported from Kalix system: KLX-00122 Invalid JSON annotations in customer.view.CustomerByNameView.GetCustomer: JSON descriptor for query result must match the stored entry JSON descriptor. View cannot be projected into result type:
  * Projected field [results] with type [message Untyped] not found in result type [customer.view.CustomerView]

Update the view so that the SELECT clause in the view query and the result type are compatible.

At CustomerByNameView.proto
```

Having a look at the proxy at the moment.

pvlugter:
I haven't looked at the JSON support in detail, but I would think that the descriptor should be different in a few ways: I expect the rpc method output should still be Any for JSON (if that's what it usually does), the `json_schema.output` should refer to the output type (`CustomerViewCollection`), which should describe the schema for each result type. So more along the lines of:

```protobuf
message CustomerViewCollection {
  repeated CustomerView results = 1;
}

message CustomerView {
  // customer fields
}

service CustomerByNameView {
  rpc GetCustomer(GetCustomerKalixSyntheticRequest) returns (google.protobuf.Any) {
    option (kalix.method) = {
      view {
        query {
          query: "SELECT * AS results FROM customers_by_name WHERE name = :customer_name"
        }
        json_schema {
          output: "CustomerViewCollection"
        }
      }
    };
    option (google.api.http) = {
      get: "/customer/by_name/{customer_name}"
    };
  }
//...
```
 
The error message says that `results` field can't be projected into `CustomerView` — the JSON schema output type, pointing out that this type doesn't have the expected shape.

The `results` field name is hardcoded currently, probably want to allow the user to specify what this field should be named. For more complex structured projections into the result type, then may need to start parsing the view query to check the result structure. The use case for putting results into a repeated field is usually for paging, so the paging functions would also need to be looked at, and fields added to the generated output type.

efgpinto:
Thanks for having a look @pvlugter. Indeed, my first approach was similar to what you're suggesting (with `google.protobuf.Any` on the output and the json schema as `CustomerViewCollection`. Although in that case, the proxy rejected it because it expects all the json_schema to be the same `CustomerView`. It might be the case of a validation that needs to be changed so I will give it another shot and confirm if that's what should happen or not.


> The `results` field name is hardcoded currently, probably want to allow the user to specify what this field should be named. For more complex structured projections into the result type, then may need to start parsing the view query to check the result structure. The use case for putting results into a repeated field is usually for paging, so the paging functions would also need to be looked at, and fields added to the generated output type.

Definitely, I will try to have the basic working and then follow up with those. Thanks. 🙇

pvlugter:
> Indeed, my first approach was similar to what you're suggesting (with `google.protobuf.Any` on the output and the json schema as `CustomerViewCollection`). However, in that case, the proxy rejected it because it expects all the json_schema to be the same `CustomerView`. It might be the case of a validation that needs to be changed so I will give it another shot and confirm if that's what should happen or not.

Could be that the old validation (current proxy release) wasn't taking the json schema output type into account properly. I think it did have a check that the type stored in the view and the output type were an exact match. Validation has since been improved, so that it instead checks the projected results have the same shape as the output type, and should cover the json schema properly as well.

pvlugter:
@efgpinto tweaked things to try it out too. Pushed an update. Still fails validation on current proxy, but works fine for latest proxy on main.

efgpinto:
Thanks for the help Peter, your suggestion works indeed. 👍

pvlugter:
Thinking on this some more, and given that the projected types can be more complex that just wrapped results, also supporting something more like the gRPC approach could make sense — where the result type includes the results field, so it's a direct mapping of the types. That would probably make it easier to extend to paging functions, and the nested aggregations from the new joins stuff, and other things.

So instead of the automatic mapping to a wrapped repeated field, it's explicit, along the lines of:

```java
public record CustomerViewCollection(Collection<CustomerView> customers)
```

```java
@GetMapping("/customer/by_name/{customer_name}")
@Query(value = "SELECT * AS customers FROM customers_by_name WHERE name = :customer_name")
public CustomerViewCollection getCustomer(@PathVariable String customer_name) {
  return null;
}
```

Allows naming this field easily. And can add the paging functions (`has_more`, `next_page_token`) into the result type. Or work with other aggregated results.

sdk/spring-sdk/src/test/scala/kalix/springsdk/impl/ViewDescriptorFactorySpec.scala
```diff
@@ -20,23 +20,26 @@ import com.google.protobuf.Descriptors.FieldDescriptor.JavaType
...
-import kalix.springsdk.testmodels.view.ViewTestModels.UserByEmailWithPostRequestBodyOnly
-import kalix.springsdk.testmodels.view.ViewTestModels.UserByEmailWithStreamUpdates
-import kalix.springsdk.testmodels.view.ViewTestModels.UserByNameEmailWithPost
-import kalix.springsdk.testmodels.view.ViewTestModels.UserByNameStreamed
-import kalix.springsdk.testmodels.view.ViewTestModels.ViewWithMethodLevelAcl
-import kalix.springsdk.testmodels.view.ViewTestModels.ViewWithNoQuery
-import kalix.springsdk.testmodels.view.ViewTestModels.ViewWithServiceLevelAcl
-import kalix.springsdk.testmodels.view.ViewTestModels.ViewWithSubscriptionsInMixedLevels
-import kalix.springsdk.testmodels.view.ViewTestModels.ViewWithTwoQueries
+import kalix.springsdk.testmodels.view.ViewTestModels.{
```
aludwiko, line 23: not sure about the convention, but it looks like this will overridden in other PRs, in my case it's flat: https://github.com/lightbend/kalix-jvm-sdk/pull/1236/files#diff-51a4487b6b60e90723ded89d0ecafd7a6296804c23d3015ce2bd691080191b04, probably some intellij formatting setting

efgpinto: Thanks. yeah, I'll see if I find the setting and disable but I also think we should probably add this to scalafmt rules if we want to enforce the convention.

efgpinto (commented):
Turns out this ended up much simpler than expected thanks to Peter.

I've followed his suggestion to use an approach closer to what we have for gRPC, expecting the user to declare a jsonable class to handle the results.

I think we might consider adding a simplified version in the future where the user can use a `Collection` directly but don't think it's worth adding right way (it's more complex to implement and would allow only for basic usage).

The docs and samples for this will be added under https://github.com/lightbend/kalix/issues/7884 since we will need a new sample. Using this feature requires the proxy version currently in main but unreleased, but since we are not adding docs yet, I think this can be merged anyway.

pvlugter (approved):
👍🏼 LGTM

docs/src/modules/spring/pages/views.adoc
```diff
@@ -132,6 +132,7 @@ In a similar way, you can include values from the request message in the respons
 SELECT :request_id, customer_id as id, name FROM customers WHERE name = :customer_name
 ----
 
+ifdef::todo[TODO: complete with example for transformed results github.com/lightbend/kalix/issues/7884]
```
octonato, line 135: The transformation example that are talking about in https://github.com/lightbend/kalix/issues/7884 is about taking a `User` as input but indexing a different representation, like `UserView`. 

While here we should have a more concrete example of relational projection. 

The original docs, have a paragraph about returning a wrapper for collections:
https://docs.kalix.io/java/views.html#_response_message_including_the_result

octonato (commented):
Looking good, but I think we have a few things to address before merging. See my comments below.

docs/src/modules/spring/pages/views.adoc
```diff
@@ -172,8 +173,10 @@ the view.
...
 entries matching the query as they are added to the view. The stream does not complete until the client closes it.
 
-Note: This is not intended as transport for xref:service-to-service.adoc[service to service] propagation of updates and does not guarantee delivery, for such use cases you
+NOTE: This is not intended as transport for xref:service-to-service.adoc[service to service] propagation of updates and it does not guarantee delivery. For such use cases you
 should instead publish events to a topic, see xref:actions-publishing-subscribing.adoc[Publishing and Subscribing with Actions]
 
 [#query]
 include::ROOT:partial$query-syntax-reference.adoc[]
+
+include::ROOT:partial$query-syntax-paging.adoc[]
```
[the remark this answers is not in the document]
efgpinto: Damn, you're right, I missed that.

docs/src/modules/spring/pages/views.adoc
octonato: Yep, no problem. We can extend the docs when we add the transformation example.

sdk/spring-sdk/src/test/java/kalix/springsdk/testmodels/view/ViewTestModels.java
```diff
@@ -259,6 +261,17 @@ public Flux<User> getUser(@RequestBody ByEmail byEmail) {
     }
   }
 
+  @Table(value = "users_view_collection")
+  @Subscribe.ValueEntity(UserEntity.class)
+  public static class UserByEmailWithCollectionReturn extends View<User> {
+
+    @Query(value = "SELECT * AS users FROM users_view WHERE name = :name")
+    @PostMapping("/users/by-name/{name}")
+    public UserCollection getUser() {
```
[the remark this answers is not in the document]
efgpinto: Agreeded offline to have a follow-up [issue](https://github.com/lightbend/kalix-jvm-sdk/issues/1259) for this.

efgpinto:
Updated now so that we show a specific return type example for grpc vs spring. Ended up not copying the whole page but using tags instead. I think it's a good tradeoff for now but let me know if you disagree. cc @octonato

linux-root:
```scala
println("Hello")
```

aludwiko (approved):

PR #1262, opened 2022-11-15
feat: handle deletes in spring Action

aludwiko:


aludwiko:
@octonato in a separate PR, I'm waiting for some other PR about views with a transformed model to place parts about deletes.

PR #1263, opened 2022-11-15
docs(spring-sdk): add examples for transforming results in views

efgpinto:
Closes https://github.com/lightbend/kalix/issues/7884

### DIFF
--- a/docs/src/modules/spring/pages/views.adoc
+++ b/docs/src/modules/spring/pages/views.adoc
@@ -178,14 +178,14 @@ include::example$spring-valueentity-customer-registry/src/main/java/com/example/
 
 Kalix creates indexes for the View based on the query. For example, the following query will result in a View with an index on the `name` column:
 
-[source,proto,indent=0]
+[source,sql,indent=0]
 ----
 SELECT * FROM customers WHERE name = :customer_name
 ----
 
 You may realize after a deployment that you forgot adding some parameters to the query Parameters that aren't exposed to the endpoint of the View. After adding these parameters the query is changed and therefore Kalix needs to add indexes for these new columns. For example, changing the above query to filter to add users that are active would mean that Kalix needs to build a View with the index on the `is-active` column.
 
-[source,proto,indent=0]
+[source,sql,indent=0]
 ----
 SELECT * FROM customers WHERE name = :customer_name AND is-active = true
 ----
--- a/docs/src/modules/spring/pages/views.adoc
+++ b/docs/src/modules/spring/pages/views.adoc
@@ -26,7 +26,7 @@ Consider an example of a Customer Registry service with a `customer` Value Entit
 <<_define_the_view>> for a service that selects customers by name and associates a table name with the View. The table is created and used by Kalix to store the View.
 
 
-This example assumes the following `customer` exists
+This example assumes the following `Customer` exists:
 
 [source,java,indent=0]
 .src/main/java/customer/api/Customer.java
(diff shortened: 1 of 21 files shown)


samples/spring-valueentity-customer-registry/src/main/java/com/example/api/CustomerSummaryByName.java
```diff
@@ -0,0 +1,26 @@
...
+@Subscribe.ValueEntity(CustomerEntity.class)
+public class CustomerSummaryByName extends View<Customer> { // <1>
+
+  @GetMapping("/summary/by_name/{customerName}")   // <2>
+  @Query(
+      value = "SELECT customerId AS id, name FROM customers WHERE name = :customerName", // <3>
+      streamUpdates = true) // <4>
+  public Flux<CustomerSummary> getCustomer() { // <5>
+    return null;
+  }
```
octonato, line 24: This is a good example to illustrate relational projection, but the original ticket was about doing a transformation and indexing the transformed model. 

The effect is slightly different. 
Here, we are persisting `Customer` to the table with a index on name. When we retrieve the results, we map it to `CustomerSummary`. 

The other alternative is to have a `onUpdate` method that receives `Customer` and returns `CustomerSummary`. 

For that, we need an example with `@Subscribe.ValueEntity(CustomerEntity.class)` at method level, not at type level.

efgpinto: yeah, I see now.. I took over the issue but thought it was about that section 😂 

Anyway, I will see where to fit that new example.. this is one we didn't have in Java for same reason, right?

docs/src/modules/spring/pages/views.adoc
```diff
@@ -55,16 +55,41 @@ include::example$spring-customer-registry-views-quickstart/src/main/java/custome
...
+IMPORTANT: Adding a view ID to your View allows you to refactor the name of the class later on without the risk of losing the view. If you don't define a view ID the class name becomes its ID. Therefore, if you change the name of your class afterwards Kalix will not recognize this new name as the same view and will create a brand-new view. This is resource consuming for a view from an Event Sourced Entity because it will reprocess all the events of that entity to rebuild it. While for a view built from a topic you can lose all the previous events because depending on the topic configuration you may only process events from current time forwards. Last but not least, it's also a problem for Value Entities because it will need to index again them when grouping them by some value.
+
+=== Using a transformed model
+
+Often times, you will want to transform the entity model to which the view is subscribing into a different representation. To do that, let's have a look at the example in which we store a summary of the `Customer` used in the previous section instead of the original one:
```
aludwiko, line 64: I think it should be `Oftentimes`, or simply `often`.

docs/src/modules/spring/pages/views.adoc
```diff
@@ -26,7 +26,7 @@ Consider an example of a Customer Registry service with a `customer` Value Entit
 <<_define_the_view>> for a service that selects customers by name and associates a table name with the View. The table is created and used by Kalix to store the View.
 
 
-This example assumes the following `customer` exists
+This example assumes the following `Customer` exists:
```
aludwiko, line 29: Maybe a link to sample/doc on how to create such an Entity?

octonato (approved):
LGTM

Andrzej left some relevant comments. Good for merge after addressed.

samples/spring-valueentity-customer-registry/src/main/java/com/example/api/CustomersResponseByName.java
```diff
@@ -0,0 +1,23 @@
...
+import kalix.springsdk.annotations.Subscribe;
+import kalix.springsdk.annotations.Table;
+import kalix.springsdk.annotations.ViewId;
+import org.springframework.web.bind.annotation.GetMapping;
+
+@ViewId("view_response_customers_by_name")
+@Table("customers_by_name")
+// tag::class[]
+@Subscribe.ValueEntity(CustomerEntity.class)
+public class CustomersResponseByName extends View<Customer> { // <1>
```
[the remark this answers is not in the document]
efgpinto: moved to views package for consistency. tks

docs/src/modules/spring/pages/views.adoc
efgpinto: thanks. added now.

samples/spring-valueentity-customer-registry/src/main/java/com/example/api/CustomersResponseByName.java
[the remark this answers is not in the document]
efgpinto: Yup, it's very hard to draw the line. I think I will leave it within`views` for consistency but it will be good to get some common understanding of how to approach that and if we want to change all the samples to be consistent with whatever we decide.

To be honest, at times I think it can be harder if you have 6, 7, 8 files all under the same package rather than having them separated. So I'm not sure I agree with the extra brain cycle usage but that's one for later.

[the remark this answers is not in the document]
aludwiko: Being opinionated is not always a bad thing, because I find it quite useful if someone gives me a direction about something, in this case, packaging. If I prefer to go with a different packing I will do so. And personally, I think that it requires more brain cycles to filter classes that are not relevant for understanding a given concept (like @efgpinto said). 

What is an API and what's not is a separate discussion. Agree with that.

aludwiko (approved):
LGTM

PR #1264, opened 2022-11-15
docs: Spring docs side effect #7723

franciscolopezsancho:
References https://github.com/lightbend/kalix/issues/7723

### DIFF
--- a/docs/src/modules/spring/pages/side-effects.adoc
+++ b/docs/src/modules/spring/pages/side-effects.adoc
@@ -1,1 +1,43 @@
-= Running Side Effects
+= Running Side Effects in Java
+
+include::ROOT:partial$include.adoc[]
+
+== Emitting effects on another component
+
+An Entity or an Action may also emit one or more side effects. A side effect is something whose result has no impact on the result of the current command--if it fails, the current command still succeeds. The result of the side effect is therefore ignored. When used from inside an Entity, side effects are only performed after the successful completion of any state actions requested by the command handler.
+
+There is no guarantee that a side effect will be executed successfully. If a failure occurs after the command is fully handled, effects might not be executed. Side effects are not retried in case of failures.
+
+Side effects may be declared as synchronous or asynchronous. Asynchronous commands run in a "fire and forget" fashion. The code flow of the caller (the command handler of the entity which emitted the asynchronous command) continues while the command is being asynchronously processed. Meanwhile, synchronous commands run sequentially, that is, the commands are processed in order, one at a time. The final result of the command handler, either a reply or a forward, is not sent until all synchronous commands are completed.
+
+=== Use case: mobile notification
+
+You might want to run side effects to notify interested parties of a change in state. For example, after a withdrawal is made from a bank account, an account entity could send a notification to the account owner's mobile phone.
+
+
+== Emitting a side effect
+
+To illustrate how you can emit a side effect, you can build on top of the xref:spring:actions-as-controller.adoc[Action as a Controller] example. In that previous example, you  build a controller around the xref:value-entity.adoc[Value Entity Counter] and forwarded the incoming request
+after modifying it.
+
+This time, instead of using a `forward`, you will call the entity using a side effect.
+
+== Implementing the Action
+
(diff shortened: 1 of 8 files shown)


sdk/spring-sdk-testkit/src/main/java/kalix/springsdk/testkit/ActionTestkit.java
```diff
@@ -75,6 +75,23 @@ public <R> ActionResult<R> call(Function<A, Action.Effect<R>> func) {
...
+   * The {@code call} method can be used to simulate a unary call to the Action. The passed java lambda should
+   * return an Action.Effect. The Effect is interpreted into an ActionResult that can be used in
+   * test assertions.
+   *
+   * @param func A function from Action to Action.Effect
+   * @param metadata Medatada accessible to the Action
+   * @return an ActionResult
+   * @param <R> The type of reply that is expected from invoking a command handler
+   */
+  public <R> ActionResult<R> call(Function<A, Action.Effect<R>> func, Metadata metadata) {
```
franciscolopezsancho, line 88: I found we will need to create such a call to pass the subject, at least. To allow retrieval of the value [here](https://github.com/lightbend/kalix-jvm-sdk/pull/1264/files#diff-00139a7ecdb8f7994ec8b144ed5fb581be8f45060a59c7f8bfd2e30a02c78f61R24)

franciscolopezsancho:
Unable to test Action side effects. More about it, here https://github.com/lightbend/kalix/issues/7935

docs/src/modules/spring/pages/side-effects.adoc
```diff
@@ -1 +1,43 @@
...
+[source,java,indent=0]
+.src/main/java/com/example/actions/DoubleCounterAction.java
+----
+include::example$spring-valueentity-counter/src/main/java/com/example/action/DoubleCounterAction.java[tag=controller-side-effect]
+----
+<1> Retrieving the id of the counter 
+<2> On incoming request, doubling the value of `increase`, 
+<3> Building a reply using `Confirmed.getDefaultInstance()`. And attaching a side effect, i.e. calling to the `Counter` to increase double the previous amount.
+
+Please note that, the result of a side effect is ignored by the command meaning that even if the call to
```
octonato, line 38: I know that this is coming from the original documentation, but maybe better to say that the "result is ignored". As it is now it may give the impression that the side-effect is not even run.

samples/spring-valueentity-counter/src/main/java/com/example/action/Confirmed.java
```diff
@@ -0,0 +1,16 @@
+package com.example.action;
+
+public class Confirmed {
+
+    private static Confirmed instance = null;
```
octonato, line 5:
```suggestion
    final public static Confirmed instance = new Confirmed();
```

And then we can remove the more verbose `getDefaultInstance`

octonato: I see that we have the same `Confirmed` type in other samples. We could do the same there as well. I mean, have a `final public static instance`.

Btw, it seems that the other variants are using tab instead of space for indentation.

octonato (approved):
LGTM, with a few minor suggestions

docs/src/modules/spring/pages/side-effects.adoc
franciscolopezsancho: totally

franciscolopezsancho: I think it may be better `the result of a side effect is ignored` for `the response of a side effect is ignored` @Renato I think without the reference to side effect the reader may not be sure of what result are we talking about.

samples/spring-valueentity-counter/src/main/java/com/example/action/Confirmed.java
franciscolopezsancho: Done

docs/src/modules/spring/pages/side-effects.adoc
octonato: @franciscolopezsancho, you tagged another Renato here.

PR #1266, opened 2022-11-16
fix: missing partials in docs

efgpinto:
Moved the partials to java folder and make spring ones depend on that one. I'm not sure this is the final setup we want to have but it's better than duplicating everything for now, I think.


docs/src/modules/spring/pages/timers.adoc
```diff
@@ -1,6 +1,6 @@
 = Timers
 
-include::ROOT:partial$timers-intro.adoc[]
+include::java:partial$timers-intro.adoc[]
```
franciscolopezsancho, line 3:
```suggestion
include::spring:partial$timers-intro.adoc[]
```

octonato (approved):

docs/src/modules/java/pages/event-sourced-entities.adoc
```diff
@@ -3,7 +3,7 @@
 :page-supergroup-java-scala: Language
 
 include::ROOT:partial$include.adoc[]
-include::ROOT:partial$eventsourced.adoc[]
+include:::java:partial$eventsourced.adoc[]
```
franciscolopezsancho, line 6:
```suggestion
include::java:partial$eventsourced.adoc[]
```

PR #1273, opened 2022-11-17
feat: support for inter-services calls

octonato:
References https://github.com/lightbend/kalix/issues/7729


sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/KalixServer.scala
```diff
@@ -142,48 +146,47 @@ object KalixServer {
     }
   }
 
-  class ActionCreationContextFactoryBean(loco: ThreadLocal[ActionCreationContext])
-      extends FactoryBean[ActionCreationContext] {
+  abstract class ThreadLocalFactoryBean[T: ClassTag] extends FactoryBean[T] {
```
octonato, line 149: All in one!

Introducing to avoid having to define each time a thread local and a factory bean.

sdk/java-sdk/src/main/scala/kalix/javasdk/impl/ProxyInfoHolder.scala
```diff
@@ -0,0 +1,101 @@
...
+
+  def setProxyInfo(proxyInfo: ProxyInfo): Unit = {
+
+    val chosenProxyName =
+      if (proxyInfo.internalProxyHostname.isEmpty) {
+        // for backward compatibility with proxy 1.0.14 or older
+        proxyInfo.proxyHostname
+      } else {
+        proxyInfo.internalProxyHostname
+      }
```
octonato, line 57: Keeping this logic, because people can still have docker-compose files using very old versions of the Proxy.

octonato:
CI not triggering, I will close and re-open it.

docs/src/modules/spring/pages/call-another-service.adoc
```diff
@@ -1 +1,31 @@
 = Calling other services
+
+
+== Kalix services
+
+Calling other Kalix services in the same project from an Action is done by invoking them using a https://docs.spring.io/spring-framework/docs/5.0.13.RELEASE/spring-framework-reference/web-reactive.html#webflux-client[`Spring WebFlux WebClient`]. The service is identified by the name it has been deployed. Kalix takes care of routing requests to the service and keeping the data safe by encrypting the connection for you.
+
+In this sample we will make an action that does a call to the xref:value-entity.adoc[Value Entity Counter] service, deployed with the service name "counter."
+
+The Kalix Spring SDK provides a utility class calls `WebClientProvider` that can provide already configured `WebClients` for other Kalix services deployed on the same Kalix project.
```
efgpinto, line 10: "already configured" sounded a bit weird..

```suggestion
The Kalix Spring SDK provides a utility class `WebClientProvider` that can provide previously configured `WebClient`s to reach other Kalix services deployed on the same Kalix project.
```

efgpinto (approved):
Nitpicking.. LGTM.

efgpinto:
Something that just crossed my mind.. maybe we need something here to align the max size with the changes done in  https://github.com/lightbend/kalix-jvm-sdk/pull/1261 ?

octonato:
> Something that just crossed my mind.. maybe we need something here to align the max size with the changes done in #1261 ?

We need to see how we can do this with the WebClient. They have different underlying clients. The WebClient is just a wrapper. Not sure if we will be able do do some fine tuning like that.

octonato:
Not ready to merge this one. I search a little further and it seems that we should cache the WebClient. Initially I thought it was not necessary because the connections are managed by Spring, but it seems that is still a good idea to cache it. See comment here: https://stackoverflow.com/questions/51999432/limit-http-response-size-on-spring-webflux

sdk/spring-sdk/src/test/scala/kalix/springsdk/impl/RestKalixClientImplSpec.scala
```diff
@@ -62,7 +62,6 @@ class RestKalixClientImplSpec extends AnyWordSpec with Matchers with BeforeAndAf
         restDefCall.fullServiceName shouldBe targetMethod.getService.getFullName
         restDefCall.methodName shouldBe targetMethod.getName
         assertMethodParamsMatch(targetMethod, restDefCall.message)
-        restDefCall.metadata.get("testServiceHeader").get() shouldBe "testName"
```
octonato, line 65: @efgpinto, this verification is not needed anymore because the calls from the Rest client doesn't need to have any extra header.

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/WebClientProviderHolder.scala
```diff
@@ -0,0 +1,107 @@
...
+    }
+  }
+
+  private def buildClient(host: String, port: Int, identificationHeader: Option[(String, String)]) = {
+
+    val builder =
+      WebClient.builder
+        .baseUrl(s"http://$host:$port")
+        .defaultHeader(HttpHeaders.CONTENT_TYPE, MediaType.APPLICATION_JSON_VALUE)
+        .filter(ExchangeFilterFunctions.limitResponseSize(MaxCrossServiceResponseContentLength))
```
octonato, line 98: Found this here: https://stackoverflow.com/questions/51999432/limit-http-response-size-on-spring-webflux

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/WebClientProviderHolder.scala
```diff
@@ -0,0 +1,107 @@
...
+
+  override def get(system: ClassicActorSystemProvider): WebClientProviderHolder = super.get(system)
+
+  override def createExtension(system: ExtendedActorSystem): WebClientProviderHolder =
+    new WebClientProviderHolder(system)
+  override def lookup: ExtensionId[_ <: Extension] = this
+}
+
+class WebClientProviderHolder(system: ExtendedActorSystem) extends Extension {
+  val webClientProvider = new WebClientProviderImpl(system)
```
octonato, line 54: Needed to introduce one more Akka extension, for two reasons:

1. we need the actor system to pass the the provider impl
2. we should had only one instance because we are not caching the clients
3. we must control where the `WebClientProvider` can be inject (see KalixServer)

Point 3 is actually about having a safe way of building a singleton provider and inject it into Actions.

sdk/java-sdk/src/main/scala/kalix/javasdk/impl/ProxyInfoHolder.scala
```diff
@@ -58,15 +57,10 @@ class ProxyInfoHolder(system: ExtendedActorSystem) extends Extension {
...
-    // if portOverride is filled, we choose it.
-    // Otherwise we use the announced one.
-    // Note: with old proxy versions `proxyInfo.proxyPort` will default to 0
-    this._chosenPort = _portOverride.getOrElse(proxyInfo.proxyPort)
-
     this._identificationInfo = proxyInfo.identificationInfo
 
     log.debug("Proxy hostname: [{}]", chosenProxyName)
-    log.debug("Proxy port to: [{}]", _chosenPort)
+    log.debug("Proxy port to: [{}]", proxyInfo)
```
octonato, line 63: When running tests, this log will print the announced proxy port, even if later it can be overwritten.

sdk/java-sdk/src/main/scala/kalix/javasdk/impl/ProxyInfoHolder.scala
```diff
@@ -75,10 +69,14 @@ class ProxyInfoHolder(system: ExtendedActorSystem) extends Extension {
   def identificationInfo: Option[IdentificationInfo] = _identificationInfo
 
   def proxyPort: Option[Int] = {
+    // If portOverride is filled, we choose it. Otherwise we use the announced one.
+    // Note: with old proxy versions `proxyInfo.proxyPort` will default to 0
+    val chosenPort = _portOverride.getOrElse(_proxyAnnouncedPort)
```
octonato, line 74: Previously, we were fixing the `_chosenPort` when we were receive the `ProxyInfo`. We can't do this for the integration tests as the port is only available afterwards. Therefore, each time we call `proxyPort`, we must calculate the `chosenPort`.

octonato (commented):
Integration tests are not fixed. 

When running the tests we need to pass the testcontainer port to the `ProxyInfoHolder` using the `overridePort`. 

We were doing so, but by the time we did, the port was already frozen. The last commit ensure that `proxyPort` will return the override if defined.

efgpinto (approved):
Reviewed the latest changes, LGTM. 👍

PR #1275, opened 2022-11-18
feat: introduce @TypeName for logical type names

octonato:
References https://github.com/lightbend/kalix/issues/7949


docs/src/modules/spring/pages/views.adoc
```diff
@@ -106,15 +106,14 @@ You can create a View from an Event Sourced Entity by using events that the Enti
...
 ----
-<1> Includes the logical type name (defined in implementation classes) as type 
-<2> Required for deserialization and to define the logical type name 
+<1> Includes the logical type name using `@TypeName` annotation.
 
-Adding `@JsonTypeInfo` is needed so that Jackson can read back correctly the Object type when deserializing. By using `name`, you set its value as the fully qualified message. This is called a `logical type name` and it allows you to refactor the class to another package after you have serialized messages of this type. Otherwise, after such a refactor the objects serialized with the old package won't be recognized as the same type. 
+IMPORTANT: It's highly recommended to add a `@TypeName` to your persisted events. Kalix needs to identify each event in order to deliver them to the right event handlers. If a logical type name isn't specified, Kalix will use the fully qualified class name. However, once a fully qualified class name is persisted in Kalix, you won't be able to rename your class or to change its package, as Kalix won't be able to recognize previously persisted event. Therefore, we recommend all event classes to use a logical type name.
```
octonato, line 116: The question is, should we highly recommend or make it compulsory?

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/SpringSdkMessageCodec.scala
```diff
@@ -54,59 +61,3 @@ private[springsdk] class SpringSdkMessageCodec extends MessageCodec {
...
-
-private[springsdk] object SpringSdkMessageCodec {
-
-  /**
-   * Used in to compute cache value if absent. This method will try to scan the type hierarchy from the passed
-   * `messageClass` to for either an explicit JsonTypeName annotation or a JsonSubTypes.
-   *
-   * In the absence of any annotation from the JsonTypeInfo family, it will fallback to use the FQCN as a type hint.
-   */
-  private def findTypeHint(messageClass: Class[_]): String = {
```
octonato, line 66: There is no reason for us to keep this complexity. If we keep it we'll need to document that users can choose between our own `@TypeName` or Jackson's `@JsonSubTypes` and `@JsonTypeName`. 

Previously it was important to have `@JsonSubTypes`, but since https://github.com/lightbend/kalix-jvm-sdk/pull/1123, it won't help the users. They do need to define individual handler per event.

docs/src/modules/spring/pages/views.adoc
franciscolopezsancho: Could we through some warnings at compilation?

franciscolopezsancho, line 116:
```suggestion
IMPORTANT: It's highly recommended to add a `@TypeName` to your persisted events. Kalix needs to identify each event in order to deliver them to the right event handlers. If a logical type name isn't specified, Kalix will use the fully qualified class name. However, once a fully qualified class name is persisted in Kalix, you won't be able to rename your class or change its package, as Kalix won't be able to recognize previously persisted events. Therefore, we recommend all event classes to use a logical type name.
```

franciscolopezsancho:
I think this is a positive change for the UX. Again, less is more. And from what I see in `SpringSdkMessageCodec` they still can use `JsonTypeName` in case they want to. So we are simplifying and at the same time maintaining flexibility for the user. I'd say this is a win win.

franciscolopezsancho:
Oh. I was looking at the first `SpringSdkMessageCodec` modification not the last. I see now that Jackson is gone. I understand you are removing complexity for us. That makes sense. Let's see when we get feedback from the users if anybody prefers to have Jackson's annotations here.

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/SpringSdkMessageCodec.scala
```diff
@@ -25,6 +25,7 @@ import com.google.protobuf.any.{ Any => ScalaPbAny }
 import com.google.protobuf.{ Any => JavaPbAny }
 import kalix.javasdk.JsonSupport
 import kalix.javasdk.impl.MessageCodec
+import kalix.springsdk.annotations.TypeName
```
franciscolopezsancho, line 28: there is a couple of import unused
```
import com.fasterxml.jackson.annotation.JsonSubTypes
import com.fasterxml.jackson.annotation.JsonTypeName
```
I would add a github `suggestion` if I could.

docs/src/modules/spring/pages/views.adoc
octonato: If we have a warning, it will be at runtime, when we scan the components to build the descriptor. 

A warning will annoy the user. If one day they decide to add the annotation, they will break the system. Unless they use the FQCN as value, because a running system will have already events persisted with the FQCN.

franciscolopezsancho: I see. Too late then. Okay

octonato: So, I change this to use the Simple Class Name in absence of a TypeName. That allows users to still change the package and it follows what Jackson does as well. 

Users are not forced to add it, so for someone that is starting with Kalix, that's one less thing to take into account. One less thing to demo and explain for us as well. 

Eventually, users will need to add the TypeName later. Also, we could solve it when we introduce schema migration, as discussed in: https://github.com/lightbend/kalix/issues/7953

franciscolopezsancho: Agreed. Great to have the issue created so we don't forget.

docs/src/modules/spring/pages/views.adoc
```diff
@@ -106,15 +106,14 @@ You can create a View from an Event Sourced Entity by using events that the Enti
...
 ----
-<1> Includes the logical type name (defined in implementation classes) as type 
-<2> Required for deserialization and to define the logical type name 
+<1> Includes the logical type name using `@TypeName` annotation.
 
-Adding `@JsonTypeInfo` is needed so that Jackson can read back correctly the Object type when deserializing. By using `name`, you set its value as the fully qualified message. This is called a `logical type name` and it allows you to refactor the class to another package after you have serialized messages of this type. Otherwise, after such a refactor the objects serialized with the old package won't be recognized as the same type. 
+IMPORTANT: It's highly recommended to add a `@TypeName` to your persisted events. Kalix needs to identify each event in order to deliver them to the right event handlers. If no logical type name is specified, Kalix uses the non-qualified class name. However, once a class name is persisted in Kalix, you won't be able to rename your class, as Kalix won't be able to recognize previously persisted events. Therefore, we recommend all event classes to use a logical type name.
```
franciscolopezsancho, line 116:
```suggestion
IMPORTANT: It's highly recommended to add a `@TypeName` to your persisted events. Kalix needs to identify each event in order to deliver them to the right event handlers. If no logical type name is specified, Kalix uses the non-qualified class name. Once an event is persisted in Kalix, you won't be able to rename your class if no logical type name has been specified, as Kalix won't be able to recognize previously persisted events. Therefore, we recommend all event classes use a logical type name.
```

sdk/spring-sdk/src/main/java/kalix/springsdk/annotations/TypeName.java
```diff
@@ -0,0 +1,43 @@
...
+/**
+ * Annotation to assign a logical type name to persisted events.
+ *
+ * Kalix needs to identify each event in order to deliver them to the right event handlers.
+ * If a logical type name isn't specified, Kalix will use the non-qualified class name.
+ *
+ * However, once a class name is persisted in Kalix, you won't be able to rename your class,
+ * as Kalix won't be able to recognize previously persisted event.
+ *
+ * Therefore, we recommend all event classes to use a logical type name.
```
franciscolopezsancho, line 33:
```suggestion
 * Once an event is persisted in Kalix, you won't be able to rename your class if no logical type name 
 * has been specified, as Kalix won't be able to recognize previously persisted events. 
 * 
 * Therefore, we recommend all event classes use a logical type name
```

franciscolopezsancho (approved):
LGTM
I left a suggestion (same repeated twice) in case you find it useful. Writing is tricky.

aludwiko:
BTW, are planning to support java pattern matching on the sealed interface in the future. So that I can have a single handler:
```
public State onEvent(CounterEvent event)
```
Dispatching is on me, but the compiler will check if all cases are covered.

octonato:
> BTW, are planning to support java pattern matching on the sealed interface in the future. So that I can have a single handler:
> 
> ```
> public State onEvent(CounterEvent event)
> ```
> 
> Dispatching is on me, but the compiler will check if all cases are covered.

It will be challenging to support it, but not impossible. 

The SDK will need to be aware that the user defined only one method and not try to dispatch to methods based on the `typeUrl`.

PR #1278, opened 2022-11-21
fix: (codegen,views) missing validation when in/out params are different

aludwiko:
References #xxxx


johanandren (approved):

codegen/core/src/main/scala/kalix/codegen/ModelBuilder.scala
```diff
@@ -289,6 +289,7 @@ object ModelBuilder {
 
     updates.toSeq
       .diff(transformedUpdates.toSeq)
+      .filterNot(_.handleDeletes)
```
johanandren, line 292: ah ofc

PR #1280, opened 2022-11-22
feat!: support views with multiple tables

pvlugter:
Refs https://github.com/lightbend/kalix/issues/7323

WIP for supporting views with multiple tables. Remove the state type parameter for views. Effects builder works at a method level now. Source compatible for protocol-first views with code generation. Breaking change for code-first and Spring SDK.

Still to do:

- [x] figure out what will happen with `emptyState` and whether there's a multi-table version of this
- [ ] switch Spring SDK back to a state parameter on update methods (currently an untyped `viewState`)
- [ ] finish off and check the internal changes for Spring SDK
- [ ] validate multiple tables work, while still compatible with existing single table views for protocol-first

### DIFF
--- a/codegen/java-gen/src/main/scala/kalix/codegen/java/ViewServiceSourceGenerator.scala
+++ b/codegen/java-gen/src/main/scala/kalix/codegen/java/ViewServiceSourceGenerator.scala
@@ -178,30 +178,24 @@ object ViewServiceSourceGenerator {
   private[codegen] def viewSource(view: ModelBuilder.ViewService, packageName: PackageNaming): String = {
     import Types.View._
 
-    val updateTypes = view.transformedUpdates.map(_.outputType).toSeq.distinct
+    val updateTypes =
+      view.transformedUpdates.map(command => command.viewTable -> command.outputType).toSeq.distinct
 
     val emptyState =
       if (updateTypes.isEmpty)
         c""
       else if (updateTypes.size > 1) {
-        val tableNames = view.transformedUpdates.map(_.viewTable).toSeq.distinct
-        val emptyStateCases = tableNames.map { tableName =>
-          c"""|case "$tableName":
+        val emptyStateMethods = updateTypes.map { case (tableName, updateType) =>
+          c"""|@Override
+              |public $updateType ${emptyStateMethodName(updateType)}() {
               |  throw new UnsupportedOperationException("Not implemented yet, replace with your empty view state for '$tableName'");
+              |}
               |"""
         }
         c"""|
-            |@Override
-            |public Object emptyState() {
-            |  switch (updateContext().viewTable()) {
-            |    $emptyStateCases
-            |    default:
-            |      return null;
-            |  }
-            |}
-            |"""
+            |$emptyStateMethods"""
       } else {
-        val stateType = updateTypes.head
+        val (_, stateType) = updateTypes.head
         c"""|
             |@Override
             |public $stateType emptyState() {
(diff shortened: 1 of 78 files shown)


pvlugter:
Updated to generate the typed empty state methods when there are multiple view tables. The protocol-first support for multi-table views should be covered now, while still source compatible for single table views.

johanandren:
One idea for more typesafety would be to generate an enum for Java/ast for Scala with the tables for the switch/pattern match (maybe tricky because the context table return type not capturing it).

pvlugter:
> Another idea would be just `emptyState()` when a single table for backwards compatibilty, but enumerate with names and the right return type when there are multiple: `emptyArtistState(): Artist`, `emptyAlbumState(): Album` (and do the untyped switch behind the scenes in the SDK/generated code)

That’s what it does currently. See the latest commits.

codegen/java-gen/src/main/scala/kalix/codegen/java/ViewServiceSourceGenerator.scala
```diff
@@ -229,12 +293,18 @@ object ViewServiceSourceGenerator {
...
          |}
          |""",
       Nil)
   }
 
+  private def emptyStateMethodName(updateType: ProtoMessageType): String = {
+    val name = updateType.name
+    val suffix = if (name.endsWith("State")) "" else "State"
+    s"empty$name$suffix"
+  }
```
johanandren, line 308: Ah, not sure if I was looking at older commits or just missed it, this is good 👍

johanandren:
I think this is a good solution for Scala/Java protocol first and then we can possibly have a separate and different abstraction more suitable for the springy reflection API.

octonato:
I believe that the Spring SDK won't require any change to support it. We can pretty much work with inner classes, for example: 


```java
@ViewId("customers_orders_view")
public class CustomerOrdersView {
  
  @Query("SELECT * " +
             "FROM customers " +
             "JOIN orders ON orders.customer_id = customers.customer_id " +
             "JOIN products ON products.product_id = orders.product_id " +
             "WHERE customers.customer_id = :customer_id")
  public Flux<CustomerOrder> get(/* ... */) { return null; )


  @Table("customers")
  public static class Customers extends View<Customer> {
    @Override
    public View.UpdateEffect<Customer> updateCustomer(
      Customer state, CustomerDomain.CustomerCreated customerCreated) {
        return effects().updateState(new Customer(/* ... */));
    };
  }

  @Table("products")
  public static class Products extends View<Product> {
    @Override
    public View.UpdateEffect<Product> updateProduct(
      Product state, ProductDomain.ProductCreated productCreated) {
        return effects().updateState(new Product(/* ... */));
    }
  }

  @Table("orders")
  public static class Orders extends View<Order> {
    @Override
    public View.UpdateEffect<Order> updateOrder(
      Order state, OrderDomain.OrderState orderState) {
        return effects().updateState(new Order(/* ... */));
    }
  }
}
```

Note that the outer class doesn't need to extend anything as long as it has a `@ViewId`. Today this annotation is not mandatory, but we could change it.

The other day, we were saying that we should move the `@Table` to the methods. That won't be necessary to repeat the table name over and over again if we work with inner classes. So this is another plus point. Less repetition, less verbose API. 

But if we decide to change the `emptyState` method, we will kind of degrade the DX for the Spring devs. But in that case, I think we will have one more reason to start working on a native Spring components. We can then step out of `View` and have a `Table` type.

octonato:
With respect to the current change proposal, I want to drop some thoughts. 

The fact we have different `emptyState` methods named after their types indicates to me that we do have different concerns being mixed in one big class. We are putting all those methods together, but they can certainly be grouped by type. Each group is responsible for updating one specific index and they have nothing to do with each other.

I understand the reasons for creating a structure that is similar to the gRPC service definition, but I think it's a bad choice to try to follow it. By doing so we limit the quality of the generated code to what gRPC can offer. 

What I mean is that the fact that we use gRPC services to create models in Java and Scala is imposing its structure. In gRPC, we can't have nested Service (it also doesn't make sense) and we can't compose services either, or express any dependency between them. So we have a flat structure and we propagate that flat structure further to Java and Scala.

I wouldn't find it bad if joins would generate more modular Java and Scala classes instead of a flat structure. 

Anyway, I just want to drop these thoughts here. I'm also not convinced that generating a non-flat structure is better. It will certainly better to work with more convenient code, but it will also introduce some cognitive load.

pvlugter:
> I understand the reasons for creating a structure that is similar to the gRPC service definition, but I think it's a bad choice to try to follow it. By doing so we limit the quality of the generated code to what gRPC can offer.

The exploration here was more about having something that was source compatible with the current protocol-first views, and then extending that approach to multiple tables. But yes, limited by what was already there to keep compatibility.

> I believe that the Spring SDK won't require any change to support it.

I'd still expect changes to the internals, otherwise I think it would see 3 views with static inner classes — these do still need to be combined into one view component for the proxy.

I can park this PR for now, and look at the inner class approach — along the lines of the first suggestion in https://github.com/lightbend/kalix/issues/7323#issuecomment-1321281453, with abstract inner classes for generated code.

octonato:
> > I believe that the Spring SDK won't require any change to support it.
> 
> I'd still expect changes to the internals, otherwise I think it would see 3 views with static inner classes — these do still need to be combined into one view component for the proxy.

Correct. What I mean is that in principle we don't even need to come up with new types.

In my proposal above, we will need to identify classes annotated with `@ViewId`. Then look-up for a method annotated with `@Query` and then inner classes. That's quite a change on how we read the View classes today. 

The `ViewDescriptorFactory` we have now is the less obvious of all factories and it's deserving some extra love. 

My move to introduce a validation framework this week was already in preparation for the upcoming joins. I feel that if the validation is extracted from the factories and run beforehand, we will make that code easier to manage.

pvlugter:
Replaced by #1299, with nested classes — just for protocol-first so far.

Issue #1281, opened 2022-11-22
Testkit support for setting accessible proxy hostname


johanandren:
Cross calls in integration tests use the mapped port from test containers, but the hostname is the default `localhost` so in environments where the proxy container is not reachable there it is not possible to run integration tests with cross component calls.

We should pick the name up from testcontainers, if possible, or allow the user to specify a hostname to use that is then passed to the proxy on startup via `kalix.proxy.internal-proxy-hostname`

PR #1282, opened 2022-11-22
fix: Use the testcontainer host name for cross component calls

johanandren:
References #1281

### DIFF
--- a/sdk/java-sdk/src/main/scala/kalix/javasdk/impl/ProxyInfoHolder.scala
+++ b/sdk/java-sdk/src/main/scala/kalix/javasdk/impl/ProxyInfoHolder.scala
@@ -93,7 +93,15 @@ class ProxyInfoHolder(system: ExtendedActorSystem) extends Extension {
   /**
    * Change port disregarding what is announced in ProxyInfo This is required for the testkit because the port to use is
    * defined by testcontainers
+   *
+   * INTERNAL API
    */
   private[kalix] def overridePort(port: Int): Unit =
     _portOverride = Some(port)
+
+  /**
+   * INTERNAL API
+   */
+  private[kalix] def overrideProxyHost(host: String): Unit =
+    _proxyHostname = Some(host)
 }
--- a/sdk/java-sdk/src/main/scala/kalix/javasdk/impl/ProxyInfoHolder.scala
+++ b/sdk/java-sdk/src/main/scala/kalix/javasdk/impl/ProxyInfoHolder.scala
@@ -55,7 +55,7 @@ class ProxyInfoHolder(system: ExtendedActorSystem) extends Extension {
         proxyInfo.internalProxyHostname
       }
 
-    // don't set if already overriden (by testkit)
+    // don't set if already overridden (by testkit)
     if (this._proxyHostname.isEmpty) {
       this._proxyHostname = Some(chosenProxyName)
     }
--- a/sdk/java-sdk/src/main/scala/kalix/javasdk/impl/ProxyInfoHolder.scala
+++ b/sdk/java-sdk/src/main/scala/kalix/javasdk/impl/ProxyInfoHolder.scala
@@ -26,6 +26,9 @@ import kalix.protocol.discovery.IdentificationInfo
 import kalix.protocol.discovery.ProxyInfo
 import org.slf4j.LoggerFactory
 
+import java.util.concurrent.atomic.AtomicReference
+import java.util.concurrent.atomic.AtomicReference
+
 object ProxyInfoHolder extends ExtensionId[ProxyInfoHolder] with ExtensionIdProvider {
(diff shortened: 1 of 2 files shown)


johanandren:
Tried this out, and it works locally (but it did before as well), not sure I can reproduce the issue where the proxy is not actually mapped to localhost. One of the reported problem envs was gitlab ci (but hard to try that out without a released SDK).

efgpinto (approved):
LGTM, can't think of an easy way to confirm either.

sdk/java-sdk/src/main/scala/kalix/javasdk/impl/ProxyInfoHolder.scala
```diff
@@ -55,7 +55,10 @@ class ProxyInfoHolder(system: ExtendedActorSystem) extends Extension {
         proxyInfo.internalProxyHostname
       }
 
-    this._proxyHostname = Some(chosenProxyName)
+    // don't set if already overridden (by testkit)
+    if (this._proxyHostname.isEmpty) {
+      this._proxyHostname = Some(chosenProxyName)
+    }
```
octonato, line 61: What about using the `def proxyHostName` to resolve it. 

So, here we do: 

```scala
this._announcedProxyHostname = 
  if (proxyInfo.internalProxyHostname.isEmpty) {
    // for backward compatibility with proxy 1.0.14 or older
    proxyInfo.proxyHostname
  } else {
    proxyInfo.internalProxyHostname
  }
```

and then 

```scala
def proxyHostname: Option[String] = 
   _proxyHostnameOverride.orElse(_announcedProxyHostname)
```

That's similar to what I did for the proxy port.

Anyway, I think both approaches are valid, but better to align it. Either like in this PR or like the proxyPort.

johanandren: I considered simplifying port to a single volatile, always feels dodgy to juggle multiple volatiles even if it probably is safe here. I'll change port to follow the same pattern as this instead.

johanandren: Done, in a safer way, that guarantee there is no race between override and discovery, in [a0026ef](https://github.com/lightbend/kalix-jvm-sdk/pull/1282/commits/a0026ef9b309d7f4ba992e120a0eef17df74d5fc)

johanandren: Heh, also, just noticed that the `_proxyAnnouncedPort` actually never was assigned in the original logic, so that can never have worked?

sdk/java-sdk/src/main/scala/kalix/javasdk/impl/ProxyInfoHolder.scala
```diff
@@ -55,23 +56,24 @@ class ProxyInfoHolder(system: ExtendedActorSystem) extends Extension {
         proxyInfo.internalProxyHostname
       }
 
-    this._proxyHostname = Some(chosenProxyName)
-
-    this._identificationInfo = proxyInfo.identificationInfo
+    // don't set if already overridden (by testkit)
+    _proxyHostname.compareAndSet(null, chosenProxyName)
+    _proxyPort.compareAndSet(-1, proxyInfo.proxyPort)
+    _identificationInfo = proxyInfo.identificationInfo
```
octonato, line 62: check here below, line 65

We are loning `proxyInfo`, we should log `proxyPort`

octonato (approved):
LGTM, but check the logging before merging. Need a small fix.

sdk/java-sdk/src/main/scala/kalix/javasdk/impl/ProxyInfoHolder.scala
octonato: > Heh, also, just noticed that the `_proxyAnnouncedPort` actually never was assigned in the original logic, so that can never have worked?

Mistake introduced by your truly here https://github.com/lightbend/kalix-jvm-sdk/pull/1273

That wasn't release yet, so we are lucky. 😄

sdk/java-sdk/src/main/scala/kalix/javasdk/impl/ProxyInfoHolder.scala
johanandren: Done in [56db155](https://github.com/lightbend/kalix-jvm-sdk/pull/1282/commits/56db15519bb3eda42b05cb69ca1040c485552056)

hithran:
> Tried this out, and it works locally (but it did before as well), not sure I can reproduce the issue where the proxy is not actually mapped to localhost. One of the reported problem envs was gitlab ci (but hard to try that out without a released SDK).

Can confirm that the fixed worked, thanks!

PR #1284, opened 2022-11-22
chore: validate before registering

octonato:
References https://github.com/lightbend/kalix/issues/7955

This adds a small validation 'framework' to allow us to scan the components before we registered them. Currently only covering Action and almost all View validations. 

Make it easier to see what are the expected rules.


sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/reflection/KalixMethod.scala
```diff
@@ -42,6 +42,7 @@ object ServiceMethod {
   // this is more for early validation. We don't support stream-in over http,
   // we block it before deploying anything
   def isStreamIn(method: Method): Boolean = {
+    // TODO: remove this once is validations completely moved to Validations object
```
franciscolopezsancho, line 45:
```suggestion
```
In case all validations have been moved to Validations

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/Validations.scala
```diff
@@ -0,0 +1,362 @@
...
+          "You cannot use @Subscribe.Topic annotation in both methods and class. You can do either one or the other.")
+      }
+    }
+  }
+
+  private def publishStreamIdMustBeFilled(component: Class[_]): Validation = {
+    Option(component.getAnnotation(classOf[Publish.Stream]))
+      .map { ann =>
+        when(ann.id().trim.isEmpty) {
+          Seq("Publish.Stream id can not be an empty string")
```
franciscolopezsancho, line 169:
```suggestion
          Seq("@Publish.Stream id can not be an empty string")
```

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/Validations.scala
```diff
@@ -0,0 +1,362 @@
...
+
+  private def noSubscriptionMethodWithAcl(component: Class[_]): Validation = {
+
+    val hasSubscriptionAndAcl = (method: Method) => hasAcl(method) && hasSubscription(method)
+
+    val messages =
+      component.getMethods.toIndexedSeq.filter(hasSubscriptionAndAcl).map { method =>
+        errorMessage(
+          method,
+          "Subscription methods are for internal use only and cannot be combined with ACL annotations.")
```
franciscolopezsancho, line 183:
```suggestion
          "Methods annotated with Kalix @Subscription annotations are for internal use only and cannot be annotated with ACL annotations.")
```
For consistency with the message below

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/Validations.scala
```diff
@@ -0,0 +1,362 @@
...
+
+  private def noSubscriptionWithRestAnnotations(component: Class[_]): Validation = {
+
+    val hasSubscriptionAndRest = (method: Method) => hasRestAnnotation(method) && hasSubscription(method)
+
+    val messages =
+      component.getMethods.toIndexedSeq.filter(hasSubscriptionAndRest).map { method =>
+        errorMessage(
+          method,
+          "Methods annotated with Kalix @Subscription annotations can not be annotated with REST annotations.")
```
franciscolopezsancho, line 197:
```suggestion
          "Methods annotated with Kalix @Subscription annotations cannot be annotated with REST annotations.")
```

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/Validations.scala
```diff
@@ -0,0 +1,362 @@
...
+          case param if param.getAnnotation(classOf[RequestBody]) != null => param
+        }
+      paramWithRequestBody.exists(_.getType == classOf[Flux[_]])
+    }
+
+    val hasRestWithStreamIn = (method: Method) => hasRestAnnotation(method) && isStreamIn(method)
+
+    val messages =
+      component.getMethods.filter(hasRestWithStreamIn).map { method =>
+        errorMessage(method, "Stream in calls are not supported.")
```
franciscolopezsancho, line 219:
```suggestion
        errorMessage(method, "@Subscribe.Stream in REST calls is not supported.")
```
Not sure I've got this one right. If not, could you rephrase it?

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/Validations.scala
```diff
@@ -0,0 +1,362 @@
...
+    val offendingMethods =
+      component.getMethods
+        .filter(hasRestAnnotation)
+        .filter(_.hasAnnotation[Query])
+        .filter { method =>
+          method.getAnnotation(classOf[Query]).streamUpdates() && !ServiceMethod.isStreamOut(method)
+        }
+
+    val messages = offendingMethods.map { method =>
+      errorMessage(method, "Query.streamUpdates can only be enabled in stream methods returning Flux")
```
franciscolopezsancho, line 272:
```suggestion
      errorMessage(method, "@Query.streamUpdates can only be enabled in stream methods returning Flux")
```

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/Validations.scala
```diff
@@ -0,0 +1,362 @@
...
+          }
+        Validation(messages)
+      } else Valid
+    }
+
+    val standaloneMethodLevelHandleDeletesIsNotAllowed = {
+      if (handleDeleteMethods.nonEmpty && updatedMethods.isEmpty) {
+        val messages =
+          handleDeleteMethods.map { method =>
+            errorMessage(method, "Method annotated with handleDeletes=true don't have matching update methods. ")
```
franciscolopezsancho, line 349:
```suggestion
            errorMessage(method, "Method annotated with handleDeletes=true don't have matching update methods.")
```

franciscolopezsancho:
A couple of thoughts:
1. Do you think it will be worth adding the name of the method or the component in the message? To help to find the source of the error more quickly?
2. Looking at this PR I thought a was being enlightened :) We have many sparsed ideas now concentrated in one place, ready to consume. Not only from the program, but for us. As you mentioned already, this was part of the idea of the PR. Thinking about this I found this could be immensely helpful to the user. What do you think about adding documentation about this? Not for all the validations, but for a good bunch of them. This way the user can avoid mistakes beforehand and keep in mind the limitations of the annotations. By looking at them in one go.

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/Validations.scala
octonato, line 197:
```suggestion
          "Methods annotated with Kalix @Subscription annotations are for internal use only and cannot be annotated with REST annotations.")
```

@franciscolopezsancho, I will align this message to the one you suggested for the ACL.

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/Validations.scala
octonato: yeah, here the concept is `stream in` calls. It's not about `@Subscribe.Stream`. 

Stream in calls are calls where the client sends an stream as input. This is not supported for REST.

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/reflection/KalixMethod.scala
octonato: That will probably be done in follow up PRs

octonato:
> A couple of thoughts:
> 
> 1. Do you think it will be worth adding the name of the method or the component in the message? To help to find the source of the error more quickly?

This is already done. When reporting a message we use `errorMessage` that accepts a message and a `AnnotatedElement` which can be a `class` or a `method`. Still possible to only pass a message (just String) though. 

> 2. Looking at this PR I thought a was being enlightened :) We have many sparsed ideas now concentrated in one place, ready to consume. Not only from the program, but for us. As you mentioned already, this was part of the idea of the PR. Thinking about this I found this could be immensely helpful to the user. What do you think about adding documentation about this? Not for all the validations, but for a good bunch of them. This way the user can avoid mistakes beforehand and keep in mind the limitations of the annotations. By looking at them in one go.

I agree. I was thinking about that as well. We can add more javadoc to the annotations, but some annotations are not under our control, like Spring Rest ones. We could also have a documentation page listing all the annotations and explain each usage in case in detail.

franciscolopezsancho:
> 
> I agree. I was thinking about that as well. We can add more javadoc to the annotations, but some annotations are not under our control, like Spring Rest ones. We could also have a documentation page listing all the annotations and explain each usage in case in detail.
If possible I would go for the latter. Are you planning to add it to this PR or you prefer to create another? If the latter this LGTM

aludwiko:
A couple of thoughts :) 

It will be nice to have such concentrated validation in proxy as well. In fact, it will be nice to have it also on the user function side (Scala and Java). When I'm working with IT tests it is pretty annoying to wait 20 seconds for testcontainers kalix-proxy just to get an error that my protobuf's are not correct. But then it will be good to avoid duplication of such validation code.

octonato:
> > I agree. I was thinking about that as well. We can add more javadoc to the annotations, but some annotations are not under our control, like Spring Rest ones. We could also have a documentation page listing all the annotations and explain each usage in case in detail.
> 
> If possible I would go for the latter. Are you planning to add it to this PR? If you prefer to leave it for another PR then this LGTM

I think this would be out of scope for this PR. And it will be a quite some work as well.

octonato:
> A couple of thoughts :)
> 
> It will be nice to have such concentrated validation in proxy as well. In fact, it will be nice to have it also on the user function side (Scala and Java). When I'm working with IT tests it is pretty annoying to wait 20 seconds for testcontainers kalix-proxy just to get an error that my protobuf's are not correct. But then it will be good to avoid duplication of such validation code.

It will be hard to have a generic validation for gRPC and Spring. If we validate the generated gRPC, we will produce error messages that won't make much sense to Spring SDK users. We will make things worst. 

I agree that gRPC is lacking proper validation. It will be better to have it in some linting library that we include in each SDK (also Javascript/Typescript).

pvlugter:
> It will be nice to have such concentrated validation in proxy as well. In fact, it will be nice to have it also on the user function side (Scala and Java). When I'm working with IT tests it is pretty annoying to wait 20 seconds for testcontainers kalix-proxy just to get an error that my protobuf's are not correct. But then it will be good to avoid duplication of such validation code.

Earlier issue on this: https://github.com/lightbend/kalix-proxy/issues/1122

franciscolopezsancho (approved):
LGTM

PR #1288, opened 2022-11-23
fix(spring-sdk): fails to handle persisted events for ES entity

efgpinto:
Fix https://github.com/lightbend/kalix/issues/7944

This is currently preventing the use of ES entities within SpringSDK as persisted events cannot be replayed when delivered to the user function, failing with `java.lang.RuntimeException: no matching handler for 'class com.google.protobuf.any.Any'`

A similar issue occurs when recovering from snapshot and accessing `currentState`: `class com.google.protobuf.any.Any cannot be cast to class customer.domain.Customer`.

~I would like to add some test coverage on this, keeping as draft for now.~

PR #1291, opened 2022-11-24
fix(spring-sdk): avoid exception when using currentState in VE after persistence

efgpinto:
Fix https://github.com/lightbend/kalix/issues/8002

Similar to #1288 but now for the state in VE

### DIFF
--- a/sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/valueentity/ReflectiveValueEntityRouter.scala
+++ b/sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/valueentity/ReflectiveValueEntityRouter.scala
@@ -17,11 +17,14 @@
 package kalix.springsdk.impl.valueentity
 
 import com.google.protobuf.any.{ Any => ScalaPbAny }
+import kalix.javasdk.JsonSupport
 import kalix.javasdk.impl.valueentity.ValueEntityRouter
 import kalix.javasdk.valueentity.CommandContext
 import kalix.javasdk.valueentity.ValueEntity
 import kalix.springsdk.impl.CommandHandler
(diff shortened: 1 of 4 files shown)


sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/valueentity/ReflectiveValueEntityRouter.scala
```diff
@@ -54,4 +56,26 @@ class ReflectiveValueEntityRouter[S, E <: ValueEntity[S]](
...
+      case s if s == null || state.getClass == entityStateType =>
+        // note that we set the state even if null, this is needed in order to
+        // be able to call currentState() later
+        entity._internalSetCurrentState(s)
+      case s =>
+        val deserializedState =
+          JsonSupport.decodeJson(entityStateType, ScalaPbAny.toJavaProto(s.asInstanceOf[ScalaPbAny]))
+        entity._internalSetCurrentState(deserializedState)
+    }
+  }
```
efgpinto, line 80: we now have this code in 3 different places (ES, VE and views router) but I'm not sure we want to extract it to a common place tbh... might have a look again later or let me know if you have an opinion.

octonato (approved):
LGTM

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/valueentity/ReflectiveValueEntityRouter.scala
octonato: Yep, we can extract the transformation to some utility class, but also fine to keep it as such for accelerate the release. We can move the code later.

PR #1293, opened 2022-11-24
chore: decrease CI pipeline run time

efgpinto:
This actually reverts something I did some months ago I think.

Current CI pipeline takes about ~11mins which I think is a bit too much. Making `tests` and `IT tests` run in parallel  should reduce that to about ~8mins.


efgpinto:
Confirmed... 
Screenshot 2022-11-24 at 14 48 04

octonato (approved):

PR #1297, opened 2022-11-28
feat: adding setting to testkit through properties #7906

franciscolopezsancho:
This a possible implementation to provide `Settings` to the Spring test kit. The idea is to set these values via the  `.properties` file in each project.
Please bear in mind that the key part here is the User interface, that is, using properties to configure. The underlying implementation is secondary. Instead of using `org.springframework.core.env.Environment.getProperty` we may prefer to use `@ConfigurationProperties` like [this](https://www.baeldung.com/configuration-properties-in-spring-boot).

References https://github.com/lightbend/kalix/issues/7906


samples/spring-eventsourced-counter/src/main/java/com/example/actions/Confirmed.java
```diff
@@ -1,5 +1,8 @@
 package com.example.actions;
 
+import com.fasterxml.jackson.annotation.JsonTypeInfo;
+
+@JsonTypeInfo(use = JsonTypeInfo.Id.NAME, property = "type")
```
franciscolopezsancho, line 5: This has little to do with this PR. It was failing the serialization when running the integration test in this sample. Although the tests were passing.

samples/spring-valueentity-counter/src/main/java/com/example/action/Confirmed.java
```diff
@@ -1,5 +1,8 @@
 package com.example.action;
 
+import com.fasterxml.jackson.annotation.JsonTypeInfo;
+
+@JsonTypeInfo(use = JsonTypeInfo.Id.NAME, property = "type")
 public class Confirmed {
```
franciscolopezsancho, line 6: This has little to do with this PR. It was failing the serialization when running the integration test in this sample. Although the tests were passing.

franciscolopezsancho:
I would have liked to add a test to assert over the Forbidden 403 exception by enabling ACL and denying to ALL but we can't catch exceptions yet. This incapability is referenced in this older issue https://github.com/lightbend/kalix/issues/7756

samples/spring-eventsourced-counter/src/it/java/com/example/TestConfig.java
```diff
@@ -0,0 +1,14 @@
...
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+
+@Configuration
+public class TestConfig {
+    @Bean
+    public KalixTestKit.Settings settings() {
+        return KalixTestKit.Settings.DEFAULT.withAclEnabled(); // <1>
+    }
+}
```
octonato, line 13: I'm wondering which other name we can use here.

`TestConfig` is not bad because we are providing extra Spring `@Configuration` to the testkit. 

On the other hand, the only thing that we can provide are the `Settings`. So maybe, `TestkitSettings`. 

Or maybe, `TestkitConfiguration` or `TestkitConfig` to enforce that this `@Configuration` is for tweaking the testkit.

octonato (approved):
Looking good. 

Left some comments. Small suggestions and clean-ups. 

Otherwise, good for merge.

samples/spring-eventsourced-counter/src/main/resources/application.properties
```diff
@@ -1,4 +1,5 @@
 # Don't remove this. 
-# this configuration is required in order to  block Spring Boot to start it's own http server
+# this configuration is required in order to block Spring Boot to start it's own http server
 # Kalix brings in its own http server and that's the one that will be starter.
-spring.main.web-application-type=none
\ No newline at end of file
+spring.main.web-application-type=none
+kalix.settings.aclEnabled=true
```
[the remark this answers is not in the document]
franciscolopezsancho: sorry, this shouldn't be here. It's a leftover from the initial attempt to provide the capabilities through properties.

samples/spring-eventsourced-counter/src/it/java/com/example/TestConfig.java
franciscolopezsancho: Let's use `TestkitConfig`

samples/spring-eventsourced-counter/src/main/java/com/example/actions/Confirmed.java
octonato: you can have a static instance 

```java
record Confirm() {
  public static Confirm instance = new Confirm();
}
```

A little bit more code, but no other annotation.

franciscolopezsancho: :) that's simplicity 👍

PR #1299, opened 2022-11-30
feat: support views with multiple tables

pvlugter:
Refs https://github.com/lightbend/kalix/issues/7323 and replaces #1280

Support multi-table views for protocol-first and code generation, using a nested class approach. Single-table views remain unchanged, while multi-table views sit on top of this with an inner class for each view table and an extra level of routing for updates.

Samples and docs for advanced views in #1268 rebased on these changes (temporarily) and updated to use both event sourced entity with transform updates and value entity, to demonstrate what it looks like in an actual example. Leaving the samples and tests for multi-table views in that docs PR.

Spring SDK changes will be done separately.


pvlugter:
Taking @johanandren's review from https://github.com/lightbend/kalix-jvm-sdk/pull/1268#issuecomment-1332331841

Issue #1164, opened 2022-10-11
Spring-sdk: conflict between path parameters and request param


efgpinto:
There seems to be some kinf of conflict between using a path param with a request param. See example below with counter sample by trying to use a `debug` request parameter (calling would be `GET /counter/hello?debug=test`):
```
@Entity(entityKey = "id", entityType = "counter")
@RequestMapping("/counter/{id}")
public class Counter extends EventSourcedEntity<Integer> {
    @GetMapping
    public Effect<String> get(@RequestParam String debug) {
        return effects().reply(currentState().toString());
    }
}
```

It fails with:
```
Caused by: com.google.protobuf.Descriptors$DescriptorValidationException: com.example.GetKalixSyntheticRequest.debug: Field number 2 has already been used in "com.example.GetKalixSyntheticRequest" by field "id".
	at com.google.protobuf.Descriptors$Descriptor.validateNoDuplicateFieldNumbers(Descriptors.java:1022)
	at com.google.protobuf.Descriptors$Descriptor.crossLink(Descriptors.java:1003)
	at com.google.protobuf.Descriptors$Descriptor.access$1200(Descriptors.java:648)
	at com.google.protobuf.Descriptors$FileDescriptor.crossLink(Descriptors.java:600)
	at com.google.protobuf.Descriptors$FileDescriptor.buildFrom(Descriptors.java:321)
	at com.google.protobuf.Descriptors$FileDescriptor.buildFrom(Descriptors.java:290)
	at kalix.springsdk.impl.ProtoDescriptorGenerator$.genFileDescriptor(ProtoDescriptorGenerator.scala:52)
	at kalix.springsdk.impl.ComponentDescriptor$.apply(ComponentDescriptor.scala:127)
	at kalix.springsdk.impl.EntityDescriptorFactory$.buildDescriptorFor(EntityDescriptorFactory.scala:35)
	at kalix.springsdk.impl.ComponentDescriptor$.descriptorFor(ComponentDescriptor.scala:69)
	at kalix.springsdk.impl.ComponentDescriptor.descriptorFor(ComponentDescriptor.scala)
	at kalix.springsdk.eventsourced.ReflectiveEventSourcedEntityProvider.<init>(ReflectiveEventSourcedEntityProvider.java:82)
```

I don't see a reason for this to not be allowed. Btw,  I've tested with a simple action and it works fine when there are no path parameters (`id` in this case).

PR #1300, opened 2022-11-30
fix: using query params should not collide with path params

aludwiko:
References #1164

### DIFF
--- a/sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/ComponentDescriptor.scala
+++ b/sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/ComponentDescriptor.scala
@@ -16,45 +16,45 @@
 
 package kalix.springsdk.impl
 
-import java.lang.reflect.Type
-
-import scala.jdk.CollectionConverters.CollectionHasAsScala
-
-import com.google.api.AnnotationsProto
-import com.google.api.CustomHttpPattern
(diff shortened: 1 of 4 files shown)


sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/ComponentDescriptor.scala
```diff
@@ -16,46 +16,46 @@
...
+  ExtractorCreator,
+  KalixMethod,
+  NameGenerator,
+  ParameterExtractor,
+  ParameterExtractors,
+  RestServiceIntrospector,
+  ServiceMethod,
+  SubscriptionServiceMethod,
+  SyntheticRequestServiceMethod
+}
```
octonato, line 53: We usually avoid this kind of import. I don't have preferences, is just for consistency with the rest of the code base.

octonato (approved):
LGTM, with a small nipick about imports.

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/ComponentDescriptor.scala
aludwiko: yeah, something is not right with my intellij preferences, fixing that

johanandren (approved):

Issue #1289, opened 2022-11-24
ClassCastException when using ReplicatedCounterMap in Scala SDK


octonato:
This was reported by a user on Discuss forum. 
https://discuss.kalix.io/t/weird-error-when-retrieving-replicated-entity/159

I was able to reproduce it using the counter map we have in our [samples](https://github.com/lightbend/kalix-jvm-sdk/tree/main/samples/scala-replicatedentity-examples) which is similar to what the user was doing. 

The problem is that I was able to reproduce it only once. After restarting the service the issue went away. I was running it locally with sbt and I suspect that there are some classpath issues. The exception contains the text `... are in unnamed module of loader 'app'`. 

What doesn't make sense to me is that first command succeeds (Increase counter), while calls to Get fails.

efgpinto:
Was able to reproduce now, this issue actually happens for Get and Increase (and probably any other), we just need to wait for passivation to come in. As the default is 30s, doing something like the below (i.e. sleeping 35 secs after the 1st command) will trigger the issue on the second attempt:

```shell
 ➜  while :; do grpcurl -plaintext -d '{"counter_map_id":"c11", "key":"hello", "value":10}' localhost:9000 com.example.replicated.countermap.CounterMapService/Increase; sleep 35; done
{

}
ERROR:
  Code: Unknown
  Message: Entity terminating
```

I'm not too acquainted with this RE piece but now with this info, shouldn't be too hard to fix I hope.

pvlugter:
The TCK should already cover reactivation of replicated entities, but looks like the Scala SDK TCK implementation doesn't include replicated entities — could be useful to add these for test coverage.

pvlugter:
Unimplemented TCK for Scala SDK tracked by #683.

maristi:
Hi, any idea when this fix might go out? afraid holidays are coming up and there might not be much progress until Jan, and It's going to start blocking me from releasing a beta of my service soon.

Alternatively, is there any workaround? like using a bunch of of separate ReplicatedCounters instead of a single ReplicatedCounterMap?

maristi:
> Hey @maristi,
> 
> I expect to have the fix released until the end of this week (16/12).

Great to hear that! thanks @efgpinto

efgpinto:
@maristi the fix is released now in https://github.com/lightbend/kalix-jvm-sdk/releases/tag/v1.1.4 .

maristi:
[...]

PR #1303, opened 2022-12-05
fix(scala-sdk): replicated entities CCE after entity passivation

efgpinto:
Resolves https://github.com/lightbend/kalix-jvm-sdk/issues/1289

### DIFF
--- a/samples/scala-replicatedentity-examples/src/test/scala/com/example/replicated/counter/SomeCounterIntegrationSpec.scala
+++ b/samples/scala-replicatedentity-examples/src/test/scala/com/example/replicated/counter/SomeCounterIntegrationSpec.scala
@@ -41,6 +41,16 @@ class SomeCounterIntegrationSpec extends AnyWordSpec with Matchers with BeforeAn
       getResult.futureValue.value shouldBe (42 - 32)
     }
 
+    "Test value after passivation" in {
+      val counterId = "1"
+
+      counter.increase(IncreaseValue(counterId, 1)).futureValue
+      Thread.sleep(5000)
+
+      val getResult = counter.get(GetValue(counterId))
+      getResult.futureValue.value shouldBe 1
+    }
+
   }
 
   override def afterAll(): Unit = {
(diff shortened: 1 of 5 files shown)


samples/scala-replicatedentity-examples/src/test/scala/com/example/replicated/counter/SomeCounterIntegrationSpec.scala
```diff
@@ -41,6 +41,16 @@ class SomeCounterIntegrationSpec extends AnyWordSpec with Matchers with BeforeAn
...
 
+    "Test value after passivation" in {
+      val counterId = "1"
+
+      counter.increase(IncreaseValue(counterId, 1)).futureValue
+      Thread.sleep(5000)
+
+      val getResult = counter.get(GetValue(counterId))
+      getResult.futureValue.value shouldBe 1
+    }
```
efgpinto, line 52: I just wanted to have a test somewhere confirming behaviour. I'm not sure we should add one of these for each entity, or just wait for the proper TCK tests to be updated.

efgpinto (commented):
So, I think it is better to decouple this from the ongoing TCK work. So opening for review now.

samples/scala-replicatedentity-examples/src/test/scala/com/example/replicated/counter/SomeCounterIntegrationSpec.scala
pvlugter: Just this for now seems good, remove again once TCK tests added.

pvlugter (approved):
LGTM. And then follow up with the TCK tests.

samples/scala-replicatedentity-examples/src/test/scala/com/example/replicated/counter/SomeCounterIntegrationSpec.scala
efgpinto: tks, added a comment to remove after tck tests added.

PR #1310, opened 2022-12-06
fix: support for java primitives in path variables

aludwiko:
References https://github.com/lightbend/kalix/issues/8053

### DIFF
--- a/sdk/spring-sdk/src/it/java/com/example/wiring/actions/echo/ActionWithPrimitives.java
+++ b/sdk/spring-sdk/src/it/java/com/example/wiring/actions/echo/ActionWithPrimitives.java
@@ -1,3 +1,19 @@
+/*
+ * Copyright 2021 Lightbend Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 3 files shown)


sdk/spring-sdk/src/it/java/com/example/wiring/actions/echo/ActionWithPrimitives.java
```diff
@@ -0,0 +1,46 @@
...
+package com.example.wiring.actions.echo;
+
+import kalix.javasdk.action.Action;
+import org.springframework.web.bind.annotation.GetMapping;
+import org.springframework.web.bind.annotation.PathVariable;
+import org.springframework.web.bind.annotation.RequestParam;
+
+public class ActionWithPrimitives extends Action {
+
+  @GetMapping("/action/{doubleValue}/{floatValue}/{intValue}/{longValue}")
```
octonato, line 26: can we have a boolean in the path just for completeness?

octonato (approved):
LGTM, with a small request

sdk/spring-sdk/src/it/java/com/example/wiring/actions/echo/ActionWithPrimitives.java
aludwiko: there is a `booleanValue` as a query param, should I add it also for path param?

octonato: No, not needed. I see that `pathParamExtractors` also calls the new `toExtractor`. So we are covered.

I will merge it.

PR #1315, opened 2022-12-06
docs: more detailed text about EntityKey

octonato:
References https://github.com/lightbend/kalix-docs/issues/1607

This PR remove the generate keys partial and adds partial about how EntityKeys are extracted or generated in the Spring SDK. 

The partial is then used in both ES and VE pages.


samples/spring-valueentity-counter/src/main/java/com/example/CounterEntity.java
```diff
@@ -57,7 +46,7 @@ public Effect<Number> plusOne() {
   // end::behaviour[]
 
   // tag::delete[]
-  @DeleteMapping("/counter/{counter_id}/delete")
+  @DeleteMapping("/counter/{counter_id}")
```
octonato, line 49: We don't need `/delete` in the path, http method is already expressing it.

docs/src/modules/spring/partials/entity-keys.adoc
```diff
@@ -0,0 +1,27 @@
+
+== Identifying the Entity
```
octonato, line 2: In this partial, I tried to be generic enough. It's used in two different pages that use different examples (ShoppingCart for ES and Counter for VE). 

Because of that, the partial does not have any snippet and tries to explain everything inline (in text).

docs/src/modules/spring/partials/entity-keys.adoc
```diff
@@ -0,0 +1,27 @@
...
+The entity key can be defined in different ways.
+
+=== Single keys
+
+The most common use is to annotate the class with `@EntityKey` and assign one path variable name to it.
+For instance, `@EntityKey("id")` will instruct Kalix to look up a matching path variable. For an endpoint defined with `@RequestMapping("/users/\{id}")`, Kalix will extract whatever path segment is used to replace `\{id}` and treat it as the Entity unique identifier.
+
+=== Composite keys
+
+It's also possible to have composite keys. For example, `@EntityKey(\{"groupId", "id"})` defines a composite key made of `groupId` and `id`. In such a case, the endpoints for this entity will need to have both path variables, eg:  `@RequestMapping("/users/\{groupId}/\{id}")`.
```
octonato, line 19:
```suggestion
It's also possible to have composite keys. For example, `@EntityKey(\{"groupId", "id"})` defines a composite key made of `groupId` and `id`. In such a case, the endpoints for this entity will need to have both path variables, e.g.:  `@RequestMapping("/users/\{groupId}/\{id}")`.
```

samples/spring-eventsourced-shopping-cart/src/it/java/com/example/IntegrationTest.java
```diff
@@ -40,60 +40,65 @@ public class IntegrationTest extends KalixIntegrationTestKitSupport {
...
-            .uri("/cart/create")
-            .retrieve()
-            .bodyToMono(String.class)
-            .block(timeout);
+
+    String cartId = "card-abc";
+
+    ResponseEntity<String> created =
+      webClient.post()
+        .uri("/cart/" + cartId + "/create")
```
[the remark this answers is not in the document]
efgpinto: maybe we can add a formating rule to be part of the javafmt? otherwise, we end up changing this back and forth depending on who is the next person touching the file and what settings they have on intellij...

docs/src/modules/spring/partials/entity-keys.adoc
```diff
@@ -0,0 +1,27 @@
+
+== Identifying the Entity
+
+In order to interact with an Entity in Kalix, we need to assign an entity type and one or more entity keys. The entity type is a unique identifier for all entities of a given type. The entity key, on the other hand, is unique per instance.
+
+In most cases, the entity key is passed as a path parameter of a REST request. The exception to the rule is when we request Kalix to auto-generate a key for us. In such a case, Kalix won't try to extract the key from the endpoint path.
+
+To define the entity type, the entity class must be annotated with `@EntityType` and have a unique and stable name assigned.
+
+The entity key can be defined in different ways.
```
efgpinto, line 10: I think in this paragraph we jump between entity key and entity type several times which can make it a bit more difficult to process and refer to.. What do you think of aggregating the info into 2 bullet points? e.g:

```suggestion
In order to interact with an Entity in Kalix, we need to assign an *entity type* and one or more *entity keys*: 
- *entity type* is a unique identifier for all entities of a given type. To define the entity type, the entity class must be annotated with `@EntityType` and have a unique and stable name assigned.
- *entity key*, on the other hand, is unique per instance. In most cases, the entity key is passed as a path parameter of a REST request. The exception to the rule is when we request Kalix to auto-generate a key for us. In such a case, Kalix won't try to extract the key from the endpoint path.

The entity key can be defined in different ways, as detailed below.
```

efgpinto (approved):
LGTM with a small suggestion.

samples/spring-eventsourced-shopping-cart/src/it/java/com/example/IntegrationTest.java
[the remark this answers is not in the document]
octonato: The problem is that samples are not part of the build. In that specific case, it's not even a sbt project. 

I'm wondering if can create some virtual sbt module that is only used to format code in the samples. As such, it doesn't need to run compilations and be aware of sample dependencies, but only run the formatting on it.

docs/src/modules/spring/partials/entity-keys.adoc
octonato: Bullet points are not working. Wrong syntax. I will fix it later. Let's not merge it for now.

octonato: fixed now

Issue #1314, opened 2022-12-06
API docs mixing javadoc and scaladoc


octonato:
I noticed that in https://docs.kalix.io/java/event-sourced-entities.html#_testing_the_entity, there is a link to the API docs for `EventSourcedResult`. 

The java one is pointing to a scaladoc. That's probably some wrong settings in sbt. 

The scala one is broken. It points to https://docs.kalix.io/java/_attachments/testkit-scala/kalix/scalasdk/testkit/EventSourcedResult.html, it should be https://docs.kalix.io/java/_attachments/scala-testkit-api/kalix/scalasdk/testkit/EventSourcedResult.html

We might have other broken links like that. Strange that link validation is not picking it. 

And with the latest change in https://github.com/lightbend/kalix-jvm-sdk/pull/1312, the Spring SDK is correctly published, but the its index.html is a scaladoc one. It certainly overwrote the original javadoc index.html because the other javadoc files are published.

franciscolopezsancho:
As per Sbt documentation, it seems that using `scaladoc` is unavoidable. See [here](https://www.scala-sbt.org/1.x/docs/Howto-Scaladoc.html#Select+javadoc+or+scaladoc)

PR #1319, opened 2022-12-08
cleansing

franciscolopezsancho:
I found the interface depending on the implementation when working on https://github.com/lightbend/kalix-jvm-sdk/issues/1314 and I needed to break the dependency build only the java docs. WDYT?


franciscolopezsancho:
Forget about it. It's a bit of a stretch.

PR #1360, opened 2022-12-16
feat: multi-table views for Spring SDK

pvlugter:
Refs https://github.com/lightbend/kalix/issues/7323

~Introduces Spring SDK specific components: `MultiTableView` and `ViewTable`~


sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/Validations.scala
```diff
@@ -121,8 +126,13 @@ object Validations {
       commonValidation(component) ++
       commonSubscriptionValidation(component) ++
       viewMustHaveTableName(component) ++
-      viewMustHaveOneQueryMethod(component) ++
       streamUpdatesQueryMustReturnFlux(component)
+    } ++
+    whenExcluding[View[_], ViewTable[_]](component) {
```
octonato, line 131: I didn't find so clear which is the typed that needs to match and which is the type that must not match. Maybe an if and else is better, for instance...

```scala
    val commonViewValidations =
      commonValidation(component) ++
      commonSubscriptionValidation(component) ++
      viewMustHaveTableName(component) ++
      streamUpdatesQueryMustReturnFlux(component)

    if (assignable[ViewTable[_]](component)) {
      commonViewValidations
    } else if (assignable[View[_]](component)) {
      commonViewValidations ++ viewMustHaveOneQueryMethod(component)
    } else if (assignable[MultiTableView](component)) {
      viewMustHaveOneQueryMethod(component)
    } else {
      Valid
    }
```

octonato: Or maybe this...

```scala
private def whenNot[T: ClassTag](component: Class[_])(block: => Validation): Validation = 
  if (!assignable[T](component)) block else Valid
```

```scala
    when[View[_]](component) {
      commonValidation(component) ++
      commonSubscriptionValidation(component) ++
      viewMustHaveTableName(component) ++
      streamUpdatesQueryMustReturnFlux(component) ++
      whenNot[ViewTable[_]](component) {
        viewMustHaveOneQueryMethod(component)
      }
    } ++
    when[MultiTableView](component) {
      viewMustHaveOneQueryMethod(component)
    }
```

octonato (commented):
Looks very nice. 

I left a comment about the validations and some extra thoughts.

We could have the following...

```java
@ViewId("joined-customer-orders")
public class JoinedCustomerOrdersView {

  @GetMapping("/joined-customer-orders/{customerId}")
  @Query( 
      """
      SELECT *
      FROM customers
      WHERE customers.customerId = :customerId
      """)
  public Flux<CustomerOrder> get(String customerId) {
    return null;
  }

  @Table("customers")
  @Subscribe.EventSourcedEntity(CustomerEntity.class)
  public static class Customers extends View<Customer> {
    ...
  }
}
```

A class annotated with `@ViewId` qualifies to be a View. 
It can have one or more static inner `View` and we use this pattern for all kind of views, not necessarily views with joins. 

I'm a little concerned with `springsdk.ViewTable` extending `javasdk.View`. We would like to have native spring components. If we create this dependency now, we will have to come up with another name later. 

Maybe a strategic approach could be to stick with `@ViewId` and `javasdk.View` for now. And later we move to `@View` with one or more inner `springsdk.ViewTable`. WDYT?

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/Validations.scala
octonato: And if make `component` an implicit? 😄 

```scala
when[View[_]] {
  whenNot[ViewTable[_]] {
    viewMustHaveOneQueryMethod
  } ++
  commonValidation ++
  commonSubscriptionValidation ++
  viewMustHaveTableName ++
  streamUpdatesQueryMustReturnFlux
} ++
when[MultiTableView] {
  viewMustHaveOneQueryMethod
}
```

pvlugter:
> Maybe a strategic approach could be to stick with `@ViewId` and `javasdk.View` for now. And later we move to `@View` with one or more inner `springsdk.ViewTable`. WDYT?

Sure. Since all the other components are class-based, seemed clearer to have the additional classes. An alternative would be moving these to `javasdk` to keep the `springsdk` package clean. But no strong opinion on this.

Have updated to the `@ViewId` + nested `View` approach in https://github.com/lightbend/kalix-jvm-sdk/pull/1360/commits/b23c2006558f0e911dc427a06cc43e374de19472.

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/Validations.scala
pvlugter: With the `ViewTable` and `MultiTableView` classes removed, this needed to be different anyway.

samples/spring-customer-registry-quickstart/src/main/resources/application.properties
```diff
@@ -1,4 +1,4 @@
 # Don't remove this. 
-# this configuration is required in order to block Spring Boot to start it's own http server
-# Kalix brings in its own http server and that's the one that will be starter.
-spring.main.web-application-type=none
\ No newline at end of file
+# This configuration is required in order to block Spring Boot starting its http server.
+# Kalix brings in its own http server and that's the one that will be started.
+spring.main.web-application-type=none
```
octonato, line 4: Thanks for spotting this silly typos of mine. 😄

octonato (approved):
LGTM

Nice that we could do it without introducing new types. Later, when we introduce Spring native components, we take that same approach for all kind of Views. Also for Views using a single table.

Issue #1327, opened 2022-10-25
Spring SDK: support for components options


octonato:
gRPC SDKs have components options that can be passed to providers. Spring SDK is not supporting it yet. Main problem is that users don't have access to the providers so we need to have annotations at component level that we then translated to provider options.

aludwiko:
Heh, I almost finished with adding the passivation strategy but then I've realized that:
```
// Note: passivation strategy settings for value entities and event sourced entities are ignored.
``` so I'm skipping this option to focus only on `forwardHeaders`.

aludwiko:
I still think that it will be good to (at least) discuss passivation strategy in terms of workflows. https://github.com/lightbend/kalix-proxy/issues/1810

aludwiko:
A question about deferred calls. I've noticed that for deferred calls we are not passing metadata for the call. It's an empty list for grpc or http calls. I would expect that a deferred call will have the same metadata information as the origin request.

johanandren:
I don't think that can be automatic in general, many of the metadata attributes for the incoming request is specific to that request (think of content-type or ce-subject) and not at all suitable to pass along, I think that forwarding may be better off as something the user does explicitly. 

I think we may be missing a `withMetadata` for that on the `DeferredCall`s though (and if we want to support it for timers as well it would need adding support in the proxy, actually persisting the metadata)

aludwiko:
`withMetadata` sounds good

PR #1375, opened 2022-12-27
feat: forward headers option support

aludwiko:
References #1327

### DIFF
--- a/sdk/java-sdk/src/main/java/kalix/javasdk/eventsourcedentity/EventSourcedEntityOptions.java
+++ b/sdk/java-sdk/src/main/java/kalix/javasdk/eventsourcedentity/EventSourcedEntityOptions.java
@@ -39,7 +39,11 @@ public interface EventSourcedEntityOptions extends EntityOptions {
    */
   EventSourcedEntityOptions withSnapshotEvery(int numberOfEvents);
 
+  /**
+   * @deprecated passivation strategy is ignored
+   */
   @Override
+  @Deprecated(since = "1.1.4", forRemoval = true)
   EventSourcedEntityOptions withPassivationStrategy(PassivationStrategy strategy);
 
   @Override
--- a/sdk/java-sdk/src/main/java/kalix/javasdk/eventsourcedentity/EventSourcedEntityOptions.java
+++ b/sdk/java-sdk/src/main/java/kalix/javasdk/eventsourcedentity/EventSourcedEntityOptions.java
@@ -19,9 +19,9 @@ package kalix.javasdk.eventsourcedentity;
 import kalix.javasdk.EntityOptions;
 import kalix.javasdk.PassivationStrategy;
 import kalix.javasdk.impl.eventsourcedentity.EventSourcedEntityOptionsImpl;
-import scala.collection.immutable.Set;
 
 import java.util.Collections;
+import java.util.Set;
 
 /** Root entity options for all event sourcing entities. */
 public interface EventSourcedEntityOptions extends EntityOptions {
@@ -39,7 +39,11 @@ public interface EventSourcedEntityOptions extends EntityOptions {
    */
   EventSourcedEntityOptions withSnapshotEvery(int numberOfEvents);
 
+  @Override
   EventSourcedEntityOptions withPassivationStrategy(PassivationStrategy strategy);
+
+  @Override
+  EventSourcedEntityOptions withForwardHeaders(Set<String> headers);
 
   /**
    * Create a default entity option for an event sourced entity.
--- a/sdk/spring-sdk/src/main/java/kalix/springsdk/view/ReflectiveViewProvider.java
+++ b/sdk/spring-sdk/src/main/java/kalix/springsdk/view/ReflectiveViewProvider.java
@@ -24,7 +24,6 @@ import kalix.javasdk.view.ViewCreationContext;
 import kalix.javasdk.view.ViewOptions;
 import kalix.javasdk.view.ViewProvider;
 import kalix.springsdk.annotations.ViewId;
-import kalix.springsdk.common.ForwardHeadersExtractor;
 import kalix.springsdk.impl.ComponentDescriptor;
 import kalix.springsdk.impl.ComponentDescriptorFactory;
 import kalix.springsdk.impl.SpringSdkMessageCodec;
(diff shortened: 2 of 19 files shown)


sdk/spring-sdk/src/main/java/kalix/springsdk/view/ReflectiveViewProvider.java
```diff
@@ -62,7 +63,7 @@ private ReflectiveViewProvider(
       Function<ViewCreationContext, V> factory,
       ViewOptions options) {
     this.factory = factory;
-    this.options = options;
+    this.options = options.withForwardHeaders(ForwardHeadersExtractor.extractFrom(cls));
```
aludwiko, line 66: I added forward headers support for views, but I'm not sure if there is an actual use case for that. GET - methods should be without implementation and subscription methods shouldn't be called from the outside.

sdk/java-sdk/src/main/java/kalix/javasdk/eventsourcedentity/EventSourcedEntityOptions.java
```diff
@@ -39,8 +39,12 @@ public interface EventSourcedEntityOptions extends EntityOptions {
    */
   EventSourcedEntityOptions withSnapshotEvery(int numberOfEvents);
 
+  @Override
   EventSourcedEntityOptions withPassivationStrategy(PassivationStrategy strategy);
```
aludwiko, line 43: This option is ignored, should I make it deprecated in java and scala sdks?

octonato: yep, we should deprecate them.

sdk/spring-sdk/src/main/java/kalix/springsdk/view/ReflectiveViewProvider.java
octonato: Yep, I agree. I don't see I reason to have support for forward headers in views, but maybe @johanandren has some extra insight about it? Maybe we added it for some future use cases?

octonato (approved):
LGTM

Fine for me to deprecate the passivation options in this PR. I also think we should deprecate the forward headers for Views. I don't see a reason to have it, but maybe others can shed a light on this topic.

sdk/java-sdk/src/main/java/kalix/javasdk/eventsourcedentity/EventSourcedEntityOptions.java
aludwiko: added, the thing is that I can't add this on an EntityOptions class, because the passivation strategy is not ignored for replicated entities, only for VE and ES.

octonato: That's annoying. 

Can we do the following...
We move this method to each of the variants of `EntityOptions` and we deprecate it only in ES and VE. True, someone depending on `EntityOptions` directly may be impacted, but that's unlike to happen, IMO. 

Maybe better to do it in another PR in that case.

octonato: Wait, that's exactly what you did. I don't see the issue then.

aludwiko: Yeah, it was only a comment about why I did it this way.

sdk/spring-sdk/src/main/java/kalix/springsdk/view/ReflectiveViewProvider.java
johanandren: Will not be useful until/if we add view result transformations in user code. No strong opinion whether adding is worth it or not.

sdk/java-sdk/src/main/java/kalix/javasdk/eventsourcedentity/EventSourcedEntityOptions.java
johanandren: What does passivation mean for replicated/crdt entities? Do we have some special GC:y cleanup for them somehow?

sdk/spring-sdk/src/main/java/kalix/springsdk/view/ReflectiveViewProvider.java
octonato: But even if we have result transformation done on the UF side, does it makes sense to have forward headers? The proxy will be calling that transformer method. We won't have an external client that could potentially send some headers that users could be interested.

johanandren: Only for the query method(s). When/if we add query request/result transform, headers the user logic could use the request headers to transform the request parameters, result rows or even completely deny a request based on some headers.

octonato: I don't see how it would be possible to have query methods that also run transformations. Query methods have query input as parameter. How would we pass the row content for transformation? 

Anyway, we can discuss that in detail the if/when we decide about it. 

For now, I would simply not support forward headers in views because we simply can't support it.

sdk/java-sdk/src/main/java/kalix/javasdk/eventsourcedentity/EventSourcedEntityOptions.java
aludwiko: it is used in `context.setReceiveTimeout(serviceSettings.passivationTimeout.duration)` and
```
    case ReceiveTimeout =>
      context.parent ! ReplicatedEntityManager.Passivate
```

sdk/spring-sdk/src/main/java/kalix/springsdk/view/ReflectiveViewProvider.java
aludwiko: Removed

sdk/java-sdk/src/main/java/kalix/javasdk/eventsourcedentity/EventSourcedEntityOptions.java
johanandren: I see, looking into how they are implemented I definitely did not have the whole picture.

Issue #1374, opened 2022-12-22
Spring SDK: support for Spring Boot 3


octonato:
Spring Boot 3 is out and we should preferably depend on this new vesion.

octonato:
For the record, @mckeeh3 tried it out and run into a ClassNotFoundException. We are possibly dependending on something that is now requiring an extra jar.

aludwiko:
Ok, after bumping to Spring 3 one test is flaky:
```
[info] Test com.example.wiring.EventSourcedEntityIntegrationTest#verifyCounterEventSourcedAfterRestartFromSnapshot() started
[info] 2022-12-27 12:01:10,042 INFO  com.example.wiring.eventsourcedentities.counter.CounterEntity - Restarting counter with commandId=12 commandName=Restart seqNr=10 current=com.example.wiring.eventsourcedentities.counter.Counter@4d1986a6
[info] 2022-12-27 12:01:10,043 ERROR com.example.wiring.eventsourcedentities.counter.CounterEntity - Terminating entity [restartFromSnapshot] due to unexpected failure for command [Restart]
[info] kalix.javasdk.impl.EntityExceptions$EntityException: Unexpected failure: java.lang.reflect.InvocationTargetException
[info] 	at kalix.javasdk.impl.EntityExceptions$EntityException$.apply(EntityExceptions.scala:46)
[info] 	at kalix.javasdk.impl.eventsourcedentity.EventSourcedEntitiesImpl.liftedTree1$1(EventSourcedEntitiesImpl.scala:196)
[info] 	at kalix.javasdk.impl.eventsourcedentity.EventSourcedEntitiesImpl.$anonfun$runEntity$6(EventSourcedEntitiesImpl.scala:186)
[info] 	at akka.stream.impl.fusing.Scan$$anon$11.onPush(Ops.scala:444)
[info] 	at akka.stream.impl.fusing.GraphInterpreter.processPush(GraphInterpreter.scala:542)
[info] 	at akka.stream.impl.fusing.GraphInterpreter.execute(GraphInterpreter.scala:423)
[info] 	at akka.stream.impl.fusing.GraphInterpreterShell.runBatch(ActorGraphInterpreter.scala:650)
[info] 	at akka.stream.impl.fusing.ActorGraphInterpreter$SimpleBoundaryEvent.execute(ActorGraphInterpreter.scala:61)
[...]

PR #1376, opened 2022-12-27
feat: update to spring 3

aludwiko:
References #1374

### DIFF
--- a/maven-java/kalix-spring-boot-archetype/src/main/resources/archetype-resources/pom.xml
+++ b/maven-java/kalix-spring-boot-archetype/src/main/resources/archetype-resources/pom.xml
@@ -282,7 +282,7 @@
     <dependency>
       <groupId>org.springframework.boot</groupId>
       <artifactId>spring-boot-starter-test</artifactId>
-      <version>3.0.1</version>
+      <version>3.0.2</version>
       <scope>test</scope>
     </dependency>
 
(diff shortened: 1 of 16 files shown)


octonato:
Hmm, we need to rebase this PR with `main` and update the kotlin archetype as well.

maven-java/kalix-spring-boot-archetype/src/main/resources/archetype-resources/pom.xml
```diff
@@ -282,7 +282,7 @@
     <dependency>
       <groupId>org.springframework.boot</groupId>
       <artifactId>spring-boot-starter-test</artifactId>
-      <version>2.7.0</version>
+      <version>3.0.1</version>
```
efgpinto, line 285: I suppose there's a reason for not using 3.0.2 here as well?

efgpinto (approved):
LGTM with a question.

maven-java/kalix-spring-boot-archetype/src/main/resources/archetype-resources/pom.xml
octonato: Should be 3.0.2

octonato: Actually, I had already noticed that 3.0.2 was out and was planning to update it. I will do it now.

Issue #1358, opened 2022-12-15
Improve error message for Views using ValueEntity


octonato:
In the Spring SDK, if we create a View from an ValueEntity, using transformation, but with the `@Subscribe.ValueEntity` annotation at class level we get the following error. 

```java
Caused by: kalix.springsdk.impl.InvalidComponentException: View subscribes to ValueEntity 
[customer.CustomerEntity] and subscribes to state changes which will be of type 
[customer.CustomerEntity$Customer] but view type parameter is 
[customer.CustomersByCity$CustomerView] which does not match, the types of the entity and the 
subscribing must be the same.
```

Here is the View code

```java
@Table("customer_by_city")
@Subscribe.ValueEntity(CustomerEntity.class)
public class CustomersByCity extends View<CustomersByCity.CustomerView> {

  record CustomerView(String name, String city){}

  @GetMapping("/customers/city/{city}")
  @Query("SELECT * FROM customer_by_city WHERE city = :city")
  public Flux<CustomerView> byCity(String city) {
    return null;
  }

  public UpdateEffect<CustomerView> onEvent(CustomerEntity.Customer state) {
    return effects().updateState(new CustomerView(evt.name(), evt.city()));
  }

}
```

I run into it during my webinar yesterday. Unfortunately, I didn't spot it immediately and decide to index `Customer` instead of `CustomerView`.

The error happens because annotation at type level means 'no transformation', but the intention is to transform and that's why the View is typed to `CustomerView`. 

We should give more hints in this error message. For example, we could say, if you want to index a different format, make sure you a method with `public UpdateEffect<CustomerView> onEvent(CustomerEntity.Customer state)` and move the `@Subscribe` annotation to it.

PR #1377, opened 2022-12-28
fix: better error message for value entity subscription

aludwiko:
References #1358

### DIFF
--- a/sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/Validations.scala
+++ b/sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/Validations.scala
@@ -250,11 +250,16 @@ object Validations {
         component.getAnnotation(classOf[Subscribe.ValueEntity]).value().asInstanceOf[Class[_]]
       val entityStateClass = valueEntityStateClassOf(valueEntityClass)
 
+      /**
+       * You are using a type level annotation in this View and that requires the View type to match the ValueEntity type [${entityStateClass.getName}]. If your intention is to transform the type, you should instead add a method like UpdateEffect<${tableType.getName}> onChange(${entityStateClass.getName} state) and move the @Subscribe.ValueEntity to it.
+       */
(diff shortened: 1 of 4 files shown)


sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/Validations.scala
```diff
@@ -241,6 +243,25 @@ object Validations {
...
+      val tableType: Class[_] = tableTypeOf(component)
+      val valueEntityClass: Class[_] =
+        component.getAnnotation(classOf[Subscribe.ValueEntity]).value().asInstanceOf[Class[_]]
+      val entityStateClass = valueEntityStateClassOf(valueEntityClass)
+
+      when(entityStateClass != tableType) {
+        val message = s"View subscribes to ValueEntity [${valueEntityClass.getName}] and subscribes to state changes " +
+          s"which will be of type [${entityStateClass.getName}] but view type parameter is [${tableType.getName}] which does not match, " +
+          "the types of the entity and the subscribing must be the same. Make sure that @Subscribe.ValueEntity annotation is used on a method like " +
+          s"`UpdateEffect<${tableType.getName}> onChange(${entityStateClass.getName} state)`."
```
octonato, line 257: I'm not sure users will understand it. I think we can be more direct. Something like: 

"You are using a type level annotation in this View and that requires the View type to match the ValueEntity type [${entityStateClass.getName}]. If your intention is to transform the type, you should instead add a method like `UpdateEffect<${tableType.getName}> onChange(${entityStateClass.getName} state)` and move the `@Subscribe.ValueEntity` to it."

Something on those lines. WDYT?

aludwiko: Updated.

octonato (approved):
Great!

I wish we had this fix before my webinar the other day. I made that mistake and couldn't figure out what was the issue.

Issue #1379, opened 2023-01-03
Support for passing metadata in deferred calls


johanandren:
Noticed in https://github.com/lightbend/kalix-jvm-sdk/issues/1327#issuecomment-1369096730

We have a `Metadata` field on deferred call but there is no way to set it, it is always empty. 

A `DeferredCall#withMetadata` would be good.

PR #1380, opened 2023-01-09
feat: adding metadata to deferred calls

aludwiko:
References #1379


sdk/java-sdk/src/main/java/kalix/javasdk/DeferredCall.java
```diff
@@ -41,4 +42,10 @@
    * effects().asyncReply} and {@code effects().asyncEffect}
    */
   CompletionStage<O> execute();
+
+  /**
+   * Add given entries to metadata
+   * @return DeferredCall with updated metadata
+   */
+  DeferredCall<I, O> withMetadata(List<Metadata.MetadataEntry> entries);
```
aludwiko, line 50: Not sure about the naming: `withMetadata` vs `addMetadata`. Current semantics is about adding.

sdk/java-sdk/src/main/scala/kalix/javasdk/impl/MetadataImpl.scala
```diff
@@ -45,6 +44,16 @@ private[kalix] class MetadataImpl(val entries: Seq[MetadataEntry]) extends Metad
         value
     }
 
+  private[kalix] def addAll(entries: util.List[Metadata.MetadataEntry]): MetadataImpl = {
+    entries.asScala.foldLeft(this) { (metadata, entry) =>
+      if (entry.isText) {
+        metadata.add(entry.getKey, entry.getValue)
```
aludwiko, line 50: Should I add, or replace current entries?

sdk/scala-sdk/src/main/scala/kalix/scalasdk/impl/DeferredCallAdapters.scala
```diff
@@ -48,6 +50,27 @@ private[scalasdk] final case class ScalaDeferredCallAdapter[I, O](javaSdkDeferre
     new MetadataImpl(javaSdkDeferredCall.metadata.asInstanceOf[kalix.javasdk.impl.MetadataImpl])
 
   def execute(): Future[O] = javaSdkDeferredCall.execute().asScala
+
+  override def withMetadata(entries: Seq[MetadataEntry]): ScalaDeferredCallAdapter[I, O] = {
+    val updatedDefCall = javaSdkDeferredCall.withMetadata(
+      entries
+        .map(entry =>
+          new kalix.javasdk.Metadata.MetadataEntry {
```
aludwiko, line 58: Maybe we should replace `MetadataEntry` interface with a class because, from a user perspective, it's quite tedious to always provide the same MetadataEntry implementation.

sdk/spring-sdk/src/it/java/com/example/wiring/SpringSdkIntegrationTest.java
```diff
@@ -490,6 +490,19 @@ public void verifyForwardHeaders() throws InterruptedException {
     Assertions.assertEquals(esHeaderValue, esResponse);
   }
 
+  @Test
+  public void shouldPropagateMetadataWithHttpDeferredCall() {
+    String value = "someValue";
+
+    String actionResponse = webClient.get().uri("/action-with-meta/myKey/" + value)
```
aludwiko, line 497: I wonder how I could write a similar test for java/scala sdk? Ubiquitous app?

sdk/java-sdk/src/main/java/kalix/javasdk/DeferredCall.java
johanandren: I think it should be `withMetadata(Metadata)` rather than a list of entries. I think we provide APIs for manipulating that metadata elsewhere. (In general across the Akka and Kalix APIs we use `withX` to replace and `addX` for update/add)

I expect the main use scenario will be "do this deferred call with this set of metadata", rather than iteratively adding metadata entries to one deferred call, so optimize DX for that:

sdk/scala-sdk/src/main/scala/kalix/scalasdk/impl/DeferredCallAdapters.scala
johanandren: See previous comment, the JVM SDK Metadata APIs already provides a smoother API with `add(String, String)` etc.

sdk/java-sdk/src/main/java/kalix/javasdk/DeferredCall.java
aludwiko: That was my first idea, but the implementation is hidden and I cannot use it `private[kalix] class MetadataImpl`. Implementing `Metadata.MetadataEntry` is faster than `Metadata` interface, but maybe we should simply expose some classes, as suggested in one of my comments.

johanandren: I don't understand this comment. `kalix.scalasdk.Metadata` and `kalix.javasdk.Metadata` are public APIs for dealing with metadata. A user would do `deferredCall.withMetadata(Metadata.empty.add("some-key", "some value"))` in Scala and `deferredCall.withMetadata(Metadata.EMPTY.add("some-key", "some value"))` in Java.

The internal impl would be done by dealing with the underlying MetadataImpl turning it into gRPC/protocol metadata just like for responses in for example `ActionsImpl#toProtocol`?

aludwiko: 🤦 of course, fixed

sdk/spring-sdk/src/it/java/com/example/wiring/SpringSdkIntegrationTest.java
franciscolopezsancho: Are you waiting for this or do you consider the PR done? The PR looks pretty solid to me.

johanandren (commented):
Looks good but I don't think the added conversion is needed.

sdk/java-sdk/src/main/scala/kalix/javasdk/impl/MetadataImpl.scala
```diff
@@ -250,6 +249,19 @@ object MetadataImpl {
...
+      .iterator()
+      .asScala
+      .foldLeft(MetadataImpl.Empty)((meta, entry) => {
+        if (entry.isText) {
+          meta.add(entry.getKey, entry.getValue)
+        } else {
+          meta.addBinary(entry.getKey, entry.getBinaryValue)
+        }
+      })
+  }
```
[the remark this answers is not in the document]
aludwiko: It is needed only to avoid duplication of code in `GrpcDeferredCall` and `HttpDeferredCall`, unless I will be brutal and do sth like:
```
override def withMetadata(metadata: Metadata): GrpcDeferredCall[I, O] = {
    this.copy(metadata = metadata.asInstanceOf[MetadataImpl])
  }
```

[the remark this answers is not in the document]
johanandren: I think brutal would be ok here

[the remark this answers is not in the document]
johanandren: (since the interface is not really an extension point but rather a way to hide impl details)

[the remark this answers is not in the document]
aludwiko: done

johanandren (approved):
LGTM

PR #1386, opened 2023-01-10
fix: CPU arch detection for docker build fix

johanandren:
MacOS intel CPU (sometimes?) reports arch as x86_64 and not amd64 so check for both and cross build only if neither matches.


pvlugter (approved):
Yes, set to `x86_64` on my machine.

pvlugter:
Reauthorize issue on CI again. Does that just keep happening regularly?

efgpinto:
I think this was due to circleci evicting all configured keys due to the security bridge.. I updated those now again. Closing and reopening didn't seem to kick the circleci pipeline.. Maybe push something on to the branch will?

efgpinto (approved):

Issue #1336, opened 2022-11-21
Spring SDK: supported types for Views


octonato:
Views page is mentioning some protobuf types. The Spring variant should talk about Java types instead. 

On top of that, we will probably need to add support for timestamp. As of today, we are only supporting `google.protobuf.Timestamp`

efgpinto:
I've created a different issue to track the problem I found while trying to support timestamps and will be using this one only for updating the docs.

PR #1392, opened 2023-01-12
docs(spring-sdk): update supported types for views

efgpinto:
Closes https://github.com/lightbend/kalix-jvm-sdk/issues/1336

### DIFF
--- a/docs/src/modules/java/partials/query-syntax-reference.adoc
+++ b/docs/src/modules/java/partials/query-syntax-reference.adoc
@@ -1,3 +1,5 @@
+// tag::grpc[]
+// tag::spring[]
 == Query syntax reference
 
 Define View queries in a language that is similar to SQL. The following examples illustrate the syntax. To retrieve:
@@ -118,6 +120,7 @@ SELECT * FROM customers WHERE text_search(profile, :search_words, 'english')
 NOTE: Text search is currently only available for deployed services, and can't be used in local testing.
 
 ==== Data types
+// end::spring[]
 
 The following data types are supported, for their corresponding Protobuf types. Arrays are created for a repeated field in a Protobuf message. Timestamps can be stored and compared using the `google.protobuf.Timestamp` message type.
 
@@ -166,6 +169,60 @@ In some use cases it is important to explicitly express that a value is missing,
 
 Optional fields with values present can be queried just like regular view fields:
 
+// end::grpc[]
+
+
+// tag::spring[]
+
+When modeling your queries, the following data types are supported:
+
+[cols="1,1"]
+|===
+| Data type | Java type
+
+| Text
+| `String`
+
+| Integer
+| `int` / `Integer`
+
+| Long (Big Integer)
+| `long` / `Long`
+
+| Float (Real)
+| `float` / `Float`
+
+| Double
+| `double` / `Double`
+
+| Boolean
+| `boolean` / `Boolean`
+
+| Byte String
+| `ByteString`
+
+| Array
+| `Collection<T>` and derived
+
+// not supported at the moment, see https://github.com/lightbend/kalix-jvm-sdk/issues/1391
+//| Timestamp
+//| `Instant`
+
+|===
+
+==== Optional fields
+
+Fields in a view type that were not given a value are handled as the default value for primitive Java data types.
+
+However, in some use cases it is important to explicitly express that a value is missing, doing that in a view column can be done in two ways:
+
+* use one of the Java non-primitive types for the field (e.g. use `Integer` instead of `int`)
+* make the field a part of another class and leave it uninitialized (i.e. `null`), for example `address.street` where the lack of an `address` message implies there is no `street` field.
+
+Optional fields with values present can be queried just like regular view fields:
+
+// tag::grpc[]
+
 ----
 SELECT * FROM customers WHERE phone_number = :number
 ----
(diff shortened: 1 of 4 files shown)


docs/src/modules/java/partials/query-syntax-reference.adoc
```diff
@@ -194,4 +251,7 @@ If no explicit ordering is specified in a view query, results will be returned i
 SELECT * FROM customers WHERE name = :name AND age > :min_age ORDER BY age DESC
 ----
 
-NOTE: Some orderings may be rejected, if the view index cannot be efficiently ordered. Generally, to order by a column it should also appear in the `WHERE` conditions.
\ No newline at end of file
+NOTE: Some orderings may be rejected, if the view index cannot be efficiently ordered. Generally, to order by a column it should also appear in the `WHERE` conditions.
```
franciscolopezsancho, line 254: Tangent. I understand you didn't write this. Do you know what it means `efficiently ordered` and why `Generally` to order by a column, it should also appear in the `WHERE` condition?

docs/src/modules/java/partials/query-syntax-reference.adoc
```diff
@@ -166,6 +169,60 @@ In some use cases it is important to explicitly express that a value is missing,
...
+| Boolean
+| `boolean` / `Boolean`
+
+| Byte String
+| `ByteString`
+
+| Array
+| `Collection<T>` and derived
+
+// not supported at the moment, see https://github.com/lightbend/kalix-jvm-sdk/issues/1391
```
franciscolopezsancho, line 207: Nice to keep a reference. Somehow unorthodox but I think it is useful.

franciscolopezsancho (approved):
LGTM

docs/src/modules/java/partials/query-syntax-reference.adoc
efgpinto: As for the first part, one example seems to be when streaming updates, but could also when trying to ORDER BY columns that are not indexed (see Indexes.scala#483 on the proxy).. For the second part, I think it's just saying that we probably don't want to sort every single row in the table, so having some sort of selection of rows is generally advisable.

docs/src/modules/java/partials/query-syntax-reference.adoc
efgpinto: Yeah, I was not sure.. but seemed okay just because we have it for GRPC counter-part.

docs/src/modules/java/partials/query-syntax-reference.adoc
franciscolopezsancho: Thanks for the pointer 👍 👍  . I would advise rewriting this `Note` to let the user know what are those validations/limitations. 
Regarding the second I think the word `generally` implies that in some cases you can order by column even though it doesn't appear in the WHERE condition. [Indexes.scala#486](https://github.com/lightbend/kalix-proxy/blob/f3b42718793077b7033dcad8269db3a414bf8363/proxy/core/src/main/scala/kalix/proxy/views/Indexes.scala#L486) seems to indicate that is never the case.

efgpinto: > I would advise rewriting this `Note` to let the user know what are those validations/limitations.

Let me see if I can do something about that..

> Regarding the second I think the word generally implies that in some cases you can order by column even if it doesn't appear in the WHERE condition. [Indexes.scala#486](https://github.com/lightbend/kalix-proxy/blob/f3b42718793077b7033dcad8269db3a414bf8363/proxy/core/src/main/scala/kalix/proxy/views/Indexes.scala#L486) seems to indicate that is never the case.

That code indeed seems to indicate such restriction.. however, yesterday I happen to have tried a query such `@Query("SELECT * FROM customers ORDER BY createdAt")` and it works.. I've just retried it now to make sure and is accepted. So maybe there's more to that code than what it looks at first sight.

franciscolopezsancho: Nice catch! `createdAt` is a field that belongs to customer? or we are creating it?

octonato (approved):

docs/src/modules/java/partials/query-syntax-reference.adoc
efgpinto: In that case it was part of a `CustomerSummary`, the view transformed type.

pvlugter: Yes, there's more to it than always requiring these columns to be in the WHERE clause. Extra sort columns will also be indexed if possible — when it's possible to create an index in the expected order.

https://github.com/lightbend/kalix-proxy/blob/f3b42718793077b7033dcad8269db3a414bf8363/proxy/core/src/main/scala/kalix/proxy/views/Indexes.scala#L121

pvlugter: We could expand the note to try explaining this in more detail. Or we could suggest people contact us if they run into the restrictions — as we can enable extra sort operations for some customers.

So by default, it only allows ORDER BY columns where an extra sort operation isn't needed — the index scan will return already ordered results, based on the indexing. So that's the 'efficiently' part. But if possible, it will add the sort columns to the index to create the expected order — it can do this when the columns indexed for the WHERE conditions are all exact matches (same for when there are no WHERE conditions).

PR #1394, opened 2023-01-12
feat: workflows support in Spring SDK

aludwiko:
References #1383 
Some tests are missing so a draft only.


aludwiko:
there are some additional commits on this branch (from main) I need to fix it

PR #1395, opened 2023-01-12
feat: workflows support in Spring SDK

aludwiko:
References #1383


sdk/spring-sdk/src/it/java/com/example/wiring/SpringSdkIntegrationTest.java
```diff
@@ -490,6 +494,70 @@ public void verifyForwardHeaders() throws InterruptedException {
...
+    createWallet(walletId2);
+    var transfer = new Transfer(walletId1, walletId2, -10);
+
+    ResponseEntity<Void> response = webClient.put().uri("/transfer/123")
+        .bodyValue(transfer)
+        .retrieve()
+        .toBodilessEntity()
+        .block(timeout);
+
+    assertThat(response.getStatusCode()).isEqualTo(HttpStatus.BAD_REQUEST);
```
aludwiko, line 512: If we are ok with using asserj then I can fix assertions in other tests, or we can we use both in parallel.

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/workflow/ReflectiveWorkflowRouter.scala
```diff
@@ -0,0 +1,95 @@
...
+        // be able to call currentState() later
+        workflow._internalSetCurrentState(s)
+      case s =>
+        val deserializedState =
+          JsonSupport.decodeJson(workflowStateType, ScalaPbAny.toJavaProto(s.asInstanceOf[ScalaPbAny]))
+        workflow._internalSetCurrentState(deserializedState)
+    }
+  }
+
+  override protected def decodeToAny(input: ScalaPbAny, messageCodec: MessageCodec): Any = {
```
aludwiko, line 87: temporal solution before we find something better

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/SpringSdkMessageCodec.scala
```diff
@@ -28,34 +27,73 @@ import kalix.springsdk.annotations.TypeName
...
+  }
+
+  def decodeToJson(value: ScalaPbAny): Any = {
+    if (value.typeUrl.startsWith(JsonSupport.KALIX_JSON)) {
+      val any =
+        JavaPbAny
+          .newBuilder()
+          .setTypeUrl(value.typeUrl)
+          .setValue(value.value)
+          .build()
```
octonato, line 90:
```suggestion
      val any = ScalaPbAny.toJavaProto(value)
```

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/SpringSdkMessageCodec.scala
```diff
@@ -28,34 +27,73 @@ import kalix.springsdk.annotations.TypeName
...
 
-  /**
-   * In the Spring SDK, input data are kept as proto Any and delivered as such to the router
-   */
-  override def decodeMessage(value: ScalaPbAny): Any = value
+  override def decodeMessage(value: ScalaPbAny): Any = {
+    value
+  }
+
+  def decodeToJson(value: ScalaPbAny): Any = {
```
octonato, line 83: Why we have this method? Shouldn't we just use `decodeMessage` above?

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/workflow/ReflectiveWorkflowRouter.scala
```diff
@@ -0,0 +1,95 @@
...
+    state match {
+      case s if s == null || state.getClass == workflowStateType =>
+        // note that we set the state even if null, this is needed in order to
+        // be able to call currentState() later
+        workflow._internalSetCurrentState(s)
+      case s =>
+        val deserializedState =
+          JsonSupport.decodeJson(workflowStateType, ScalaPbAny.toJavaProto(s.asInstanceOf[ScalaPbAny]))
+        workflow._internalSetCurrentState(deserializedState)
+    }
```
octonato, line 84: Do we still need this dance? I thought this would be solved in the codec, no?

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/workflow/ReflectiveWorkflowRouter.scala
```diff
@@ -0,0 +1,95 @@
...
+  }
+
+  override protected def decodeToAny(input: ScalaPbAny, messageCodec: MessageCodec): Any = {
+    messageCodec.asInstanceOf[SpringSdkMessageCodec].decodeToJson(input)
```
octonato, line 88: I may be missing something, I don't see why this specialization.  Calling `decodeMessage` should be enough if its impl is smart enough to use the types in the cache.

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/workflow/ReflectiveWorkflowRouter.scala
```diff
@@ -0,0 +1,95 @@
...
+import kalix.springsdk.impl.CommandHandler
+import kalix.springsdk.impl.InvocationContext
+import kalix.springsdk.impl.SpringSdkMessageCodec
+
+import java.lang.reflect.ParameterizedType
+
+class ReflectiveWorkflowRouter[S, W <: Workflow[S]](
+    override protected val workflow: W,
+    commandHandlers: Map[String, CommandHandler],
+    messageCodec: SpringSdkMessageCodec)
```
octonato, line 34: This `messageCodec` is not being used and we pass the codec. 

Maybe we don't need to pass it to `decodeToAny` because already in scope?

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/KalixServer.scala
```diff
@@ -371,6 +386,26 @@ case class KalixServer(applicationContext: ApplicationContext, config: Config) {
...
+        if (hasContextConstructor(clz, classOf[WorkflowContext])) {
+          WorkflowContextFactoryBean.set(context)
+        }
+
+        val webClientProviderHolder = WebClientProviderHolder(context.materializer().system)
+
+        if (hasContextConstructor(clz, classOf[KalixClient])) {
+          kalixClient.setWebClient(webClientProviderHolder.webClientProvider.localWebClient)
+          // we only have one KalixClient, but we only set it to the ThreadLocalFactoryBean
+          // when building actions, because it's only allowed to inject it in Actions
```
octonato, line 403: time to update this commend. We only allowed it to be inject in Actions and Workflows

sdk/spring-sdk/src/test/scala/kalix/springsdk/impl/SpringSdkMessageCodecSpec.scala
```diff
@@ -67,11 +79,56 @@ class SpringSdkMessageCodecSpec extends AnyWordSpec with Matchers {
       encoded.getTypeUrl shouldBe jsonTypeUrlWith("SimpleClass")
     }
 
+    "not encode java twice" in {
+      val encoded = messageCodec.encodeJava(SimpleClass("abc", 10))
+      val reEncoded = messageCodec.encodeJava(encoded)
+      reEncoded shouldBe encoded
+    }
+
+    "not encode scala to java" in {
```
octonato, line 88: Not so clear from the test title what we are trying to test. 

"not re-code (wrap) ScalaPbAny" ?

octonato: Maybe "not encode when already PbAny"? 
Isn't that what we are looking for?

sdk/spring-sdk/src/test/scala/kalix/springsdk/impl/WorkflowDescriptorFactorySpec.scala
```diff
@@ -0,0 +1,114 @@
...
+      assertDescriptor[WorkflowTestModels.WorkflowWithTypeLevelKey] { desc =>
+        val method = desc.commandHandlers("StartTransfer")
+        val fieldKey = "transferId"
+        assertRequestFieldJavaType(method, fieldKey, JavaType.STRING)
+        assertEntityKeyField(method, fieldKey)
+        assertRequestFieldJavaType(method, "json_body", JavaType.MESSAGE)
+      }
+    }
+
+    "generate mappings for a Workflow with  keys in path and EntityKey on method" in {
```
octonato, line 40: nitpick
```suggestion
    "generate mappings for a Workflow with keys in path and EntityKey on method" in {
```

octonato (commented):
Looking good. 

I have a few comments and questions about the codec (as expected 😄).

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/KalixServer.scala
aludwiko: updated

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/workflow/ReflectiveWorkflowRouter.scala
aludwiko: Unfortunately, I don't see a way to simplify this code, since we need to decode to a specific class and `decodeToJson` returns Any.

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/workflow/ReflectiveWorkflowRouter.scala
aludwiko: As discussed, we can't change the current behavior of `decodeMessage` without changes in other components.

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/workflow/ReflectiveWorkflowRouter.scala
aludwiko: it is used now

sdk/spring-sdk/src/test/scala/kalix/springsdk/impl/SpringSdkMessageCodecSpec.scala
aludwiko: yeah, wrap is a good world

octonato (approved):

Issue #1372, opened 2022-12-22
Spring SDK: path and query params should respect defaults


efgpinto:
When defining an endpoint with spring mvc and using `@PathVariable` or `@RequestParam`, the behaviour should respect the defaults from spring. The default is those are required fields unless using i.e. `@PathVariable(required = false)`. Thus, in the  example below, it shoud mean a request such as `curl "localhost:9000/wrapped/by_city"` should be rejected:

```java
  @GetMapping("/wrapped/by_city")
  public CustomersResponse getCustomers(@RequestParam List<String> cities) {
    return null;
  }
```

However, that is not the case currently because the fields are marked as optional by default.

efgpinto:
Just to note, this is a breaking change but I think it's a needed one. So probably better to do it sooner than later.

octonato:
One concern I have with this is that today it's not required to have a `@PathVariable` for the entity key. 

That's because the entity key is already available in the context and it's not necessarily something that we need all the time. 

Users that want to have access to the id can either access it using a `@PathVariable` or using `context.entityId`. If we change this, users will be required to always add an extra method param that they might not even use.

efgpinto:
Maybe I miss-understood you or do not have full context but I don't see how this affects the entityId.

In cases where the user does not declare any param on the method, it should work as of today. The change would be when a user adds a parameter, we should consider the `required` prop of the annotation being used to mark the field as optional or required. The user would still have the option to either read the entityId from context or declare as `@PathVariable`.

One thing I'm not sure actually is if it needs to apply both for `@PathVariable` and `@RequestParam`.. I need to confirm but I think we might already make the path variable required by default.

octonato:
Ok, it's on me. I miss-understood the issue. 

My point was, that a method defined as:

```java
@GetMapping("/customers/{customerId}")
public Customer get() {
  return null;
}
```

should not require a method param annotated with `@PathVariable`. Users can access it from a method param if they want, but should not be forced to.

PR #1401, opened 2023-01-16
fix(spring-sdk)!: mark params as required according to annotation

efgpinto:
Closes https://github.com/lightbend/kalix-jvm-sdk/issues/1372


octonato (approved):
LGTM

PR #1403, opened 2023-01-16
fix(spring-sdk)!: mark params as required according to annotations

efgpinto:
Closes https://github.com/lightbend/kalix-jvm-sdk/issues/1372


efgpinto:
Duplicated PR. Closing

Issue #1348, opened 2022-10-20
Spring ACL annotations should use http status code 


octonato:
This is something that I didn't realise before. The deny code are gRPC codes. I think we can leave it as it for now, but we may want to change it to http status code and then convert to proto under the hood.

_Originally posted by @octonato in https://github.com/lightbend/kalix-jvm-sdk/pull/1195#discussion_r1000582808_

octonato:
also related to https://github.com/lightbend/kalix-jvm-sdk/issues/1323

franciscolopezsancho:
Done in https://github.com/lightbend/kalix-jvm-sdk/pull/1406

PR #1406, opened 2023-01-19
chore: using HTTP status codes for ACL denyCode #1348

franciscolopezsancho:
References #1348

### DIFF
--- a/samples/spring-doc-snippets/src/main/java/com/example/acl/MyAction.java
+++ b/samples/spring-doc-snippets/src/main/java/com/example/acl/MyAction.java
@@ -51,7 +51,7 @@ public class MyAction extends Action {
 
     // tag::deny-code[]
     @PostMapping
-    @Acl(allow = @Acl.Matcher(service = "*"), denyCode = 5)
+    @Acl(allow = @Acl.Matcher(service = "*"), denyCode = Acl.DenyStatusCode.NOT_FOUND_404)
     public Effect<String> updateUser(@RequestBody CreateUser create) {
         //...
         // end::deny-code[]
--- a/samples/spring-doc-snippets/src/main/java/com/example/acl/MyAction.java
+++ b/samples/spring-doc-snippets/src/main/java/com/example/acl/MyAction.java
@@ -51,7 +51,7 @@ public class MyAction extends Action {
 
     // tag::deny-code[]
     @PostMapping
-    @Acl(allow = @Acl.Matcher(service = "*"), denyCode = Acl.DenyStatusCode.NOT_FOUND_404)
+    @Acl(allow = @Acl.Matcher(service = "*"), denyCode = Acl.DenyStatusCode.NOT_FOUND)
     public Effect<String> updateUser(@RequestBody CreateUser create) {
         //...
         // end::deny-code[]
--- a/samples/spring-doc-snippets/src/main/java/com/example/acl/UserAction.java
+++ b/samples/spring-doc-snippets/src/main/java/com/example/acl/UserAction.java
@@ -3,7 +3,7 @@ package com.example.acl;
 import kalix.javasdk.action.Action;
 import kalix.springsdk.annotations.Acl;
 // tag::acl[]
-@Acl(denyCode = 5)
+@Acl(denyCode = Acl.DenyStatusCode.NOT_FOUND_404)
 public class UserAction extends Action {
     //...
 }
--- a/samples/spring-doc-snippets/src/main/java/com/example/acl/UserAction.java
+++ b/samples/spring-doc-snippets/src/main/java/com/example/acl/UserAction.java
@@ -3,7 +3,7 @@ package com.example.acl;
 import kalix.javasdk.action.Action;
 import kalix.springsdk.annotations.Acl;
 // tag::acl[]
-@Acl(denyCode = Acl.DenyStatusCode.NOT_FOUND_404)
+@Acl(denyCode = Acl.DenyStatusCode.NOT_FOUND)
 public class UserAction extends Action {
     //...
 }
--- a/sdk/spring-sdk/src/main/java/kalix/springsdk/annotations/Acl.java
+++ b/sdk/spring-sdk/src/main/java/kalix/springsdk/annotations/Acl.java
@@ -55,7 +55,7 @@ public @interface Acl {
    * If set to `true` in the top most parent - like the `Main` class - then it will be equivalent to set {@code denyCode} to 'FORBIDDEN'
    * @return
    */
-  boolean inherit() default false;
+  boolean inheritDenyCode() default false;
 
   enum DenyStatusCode {
     BAD_REQUEST(3),
--- a/sdk/spring-sdk/src/main/java/kalix/springsdk/annotations/Acl.java
+++ b/sdk/spring-sdk/src/main/java/kalix/springsdk/annotations/Acl.java
@@ -48,19 +48,18 @@ public @interface Acl {
    * Found'. If 'Inherited', indicates that the code should be inherited from the parent (regardless of the inherit field).
    *
    */
-  DenyStatusCode denyCode() default  DenyStatusCode.FORBIDDEN_403;
+  DenyStatusCode denyCode() default DenyStatusCode.FORBIDDEN;
 
   enum DenyStatusCode {
     INHERITED(0),
-    BAD_REQUEST_400(3),
-    FORBIDDEN_403(7),
-    NOT_FOUND_404(5),
-
-    AUTHENTICATION_REQUIRED_407(16),
-    CONFLICT_409(6),
-    INTERNAL_SERVER_ERROR_500(13),
-    SERVICE_UNAVAILABLE_503(14),
-    GATEWAY_TIMEOUT_504(4);
+    BAD_REQUEST(3),
+    FORBIDDEN(7),
+    NOT_FOUND(5),
+    AUTHENTICATION_REQUIRED(16),
+    CONFLICT(6),
+    INTERNAL_SERVER_ERROR(13),
+    SERVICE_UNAVAILABLE(14),
+    GATEWAY_TIMEOUT(4);
 
 
 
--- a/sdk/spring-sdk/src/main/java/kalix/springsdk/annotations/Acl.java
+++ b/sdk/spring-sdk/src/main/java/kalix/springsdk/annotations/Acl.java
@@ -42,23 +42,27 @@ public @interface Acl {
 
 
   /**
-   * The gRPC status code to respond with when access is denied.
+   * The status code to respond with when access is denied.
    *
-   * By default, this will be 7 (permission denied), but alternatives might include 16 (unauthenticated) or 5 (not
-   * found). If 0, indicates that the code should be inherited from the parent (regardless of the inherit field).
+   * By default, this will be 'Forbidden', but alternatives might include 'Authentication required' or 'Not
+   * Found'. If 'Inherited', indicates that the code should be inherited from the parent (regardless of the inherit field).
    *
-   * When HTTP transcoding is in use, this code will be translated to an equivalent HTTP status code.
    */
   DenyStatusCode denyCode() default  DenyStatusCode.FORBIDDEN_403;
 
   enum DenyStatusCode {
-    BAD_REQUEST_400(4),
-    FORBIDDEN_403(8),
-    NOT_FOUND_404(6),
-    CONFLICT_409(7),
-    INTERNAL_SERVER_ERROR_500(2),
-    SERVICE_UNAVAILABLE_503(9),
-    GATEWAY_TIMEOUT_504(5);
+    INHERITED(0),
+    BAD_REQUEST_400(3),
+    FORBIDDEN_403(7),
+    NOT_FOUND_404(5),
+
+    AUTHENTICATION_REQUIRED_407(16),
+    CONFLICT_409(6),
+    INTERNAL_SERVER_ERROR_500(13),
+    SERVICE_UNAVAILABLE_503(14),
+    GATEWAY_TIMEOUT_504(4);
+
+
 
     public final int value;
     DenyStatusCode(int value){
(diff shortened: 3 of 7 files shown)


efgpinto (commented):
Looking good, left some questions

sdk/spring-sdk/src/main/java/kalix/springsdk/annotations/Acl.java
```diff
@@ -42,14 +42,34 @@
...
+    INHERITED(0),
+    BAD_REQUEST_400(3),
+    FORBIDDEN_403(7),
+    NOT_FOUND_404(5),
+
+    AUTHENTICATION_REQUIRED_407(16),
+    CONFLICT_409(6),
+    INTERNAL_SERVER_ERROR_500(13),
+    SERVICE_UNAVAILABLE_503(14),
+    GATEWAY_TIMEOUT_504(4);
```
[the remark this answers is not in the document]
franciscolopezsancho: yep, that's why. 
I thought about it removing it but I wanted to see what you guys think about it. I like to have the reference namely because is an indicator that it is an HTTP status code but I agree it doesn't look great.

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/AclDescriptorFactory.scala
```diff
@@ -74,11 +77,33 @@ object AclDescriptorFactory {
       aclBuilder.addDeny(idx, principalMatcher)
     }
 
-    aclBuilder.setDenyCode(aclJavaAnnotation.denyCode())
+    aclBuilder.setDenyCode(denyCodeHTTPtogRPC(aclJavaAnnotation.denyCode()))
```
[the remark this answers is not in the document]
franciscolopezsancho: You are not. Silly me :)

sdk/spring-sdk/src/main/java/kalix/springsdk/annotations/Acl.java
```diff
@@ -42,14 +42,34 @@
...
+   * By default, this will be 'Forbidden', but alternatives might include 'Authentication required' or 'Not
+   * Found'. If 'Inherited', indicates that the code should be inherited from the parent (regardless of the inherit field).
    *
-   * When HTTP transcoding is in use, this code will be translated to an equivalent HTTP status code.
    */
-  int denyCode() default 0;
+  DenyStatusCode denyCode() default  DenyStatusCode.FORBIDDEN_403;
+
+  enum DenyStatusCode {
+    INHERITED(0),
```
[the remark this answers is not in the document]
franciscolopezsancho: There was an OK before but receiving an OK from a call that has been denied I find it counterintuitive for the user. Could you maybe elaborate?

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/AclDescriptorFactory.scala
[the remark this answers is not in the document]
franciscolopezsancho: removed

sdk/spring-sdk/src/main/java/kalix/springsdk/annotations/Acl.java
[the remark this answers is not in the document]
efgpinto: yeah, I mean.. it was certainly more explicit but also felt weird.. maybe we can have the input from someone else? since this is user facing could be good to have another pair of eyes maybe? @octonato WDYT?

sdk/spring-sdk/src/main/java/kalix/springsdk/annotations/Acl.java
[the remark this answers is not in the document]
efgpinto: Oh yeah, it's counterintuitive for sure. I just think that since we are letting the user configure it, maybe we should provide that option as well. But it's also something we can easily add later so I'm happy to go with this version as well.

samples/spring-doc-snippets/src/main/java/com/example/acl/MyAction.java
```diff
@@ -51,7 +51,7 @@ public void example5() {
 
     // tag::deny-code[]
     @PostMapping
-    @Acl(allow = @Acl.Matcher(service = "*"), denyCode = 5)
+    @Acl(allow = @Acl.Matcher(service = "*"), denyCode = Acl.DenyStatusCode.NOT_FOUND_404)
```
efgpinto, line 54: refactor missed this one?

samples/spring-doc-snippets/src/main/java/com/example/acl/UserAction.java
```diff
@@ -3,7 +3,7 @@
 import kalix.javasdk.action.Action;
 import kalix.springsdk.annotations.Acl;
 // tag::acl[]
-@Acl(denyCode = 5)
+@Acl(denyCode = Acl.DenyStatusCode.NOT_FOUND_404)
```
efgpinto, line 6: same?

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/AclDescriptorFactory.scala
```diff
@@ -16,12 +16,15 @@
 
 package kalix.springsdk.impl
 
-import java.lang.reflect.Method
+import akka.http.javadsl.model.StatusCode
```
efgpinto, line 19: unused import?

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/AclDescriptorFactory.scala
```diff
@@ -16,12 +16,15 @@
...
 
+import java.lang.reflect.Method
 import com.google.protobuf.DescriptorProtos
 import com.google.protobuf.Descriptors
+import io.grpc.Status
```
efgpinto, line 24: unused import?

sdk/spring-sdk/src/test/scala/kalix/springsdk/impl/AclDescriptorFactorySpec.scala
```diff
@@ -54,7 +55,7 @@ class AclDescriptorFactorySpec extends AnyWordSpec with Matchers {
     "generate a default ACL file descriptor with deny code" in {
       val extension = lookupExtension[MainDenyWithCode].get
       val denyCode = extension.getAcl.getDenyCode
-      denyCode shouldBe 7
+      denyCode shouldBe Status.Code.ALREADY_EXISTS.value()
```
efgpinto, line 58: Wrong import? better to use the acl one instead of the grpc one?

efgpinto (approved):
LGTM, with small notes on imports

samples/spring-doc-snippets/src/main/java/com/example/acl/MyAction.java
franciscolopezsancho: yes. done now

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/AclDescriptorFactory.scala
franciscolopezsancho: and a few more. Removed now.

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/AclDescriptorFactory.scala
franciscolopezsancho: removed

sdk/spring-sdk/src/test/scala/kalix/springsdk/impl/AclDescriptorFactorySpec.scala
franciscolopezsancho: yes. changed

samples/spring-doc-snippets/src/main/java/com/example/acl/UserAction.java
franciscolopezsancho: changed

sdk/spring-sdk/src/main/java/kalix/springsdk/annotations/Acl.java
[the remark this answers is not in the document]
octonato: What we probably can do, is to extract this enums to some other place. Not part of the ACL annotation. 

My reasoning is that with native components, calls to `effect.error("msg", code)` should receive our own http code. 

We should not allow users to pass raw integers, and we may want to offer a limited set of status codes to have better control of the conversion to gRPC codes. 

In addition to that, we may want to also offer (later) a limited set of 2xx codes. What I have in mind, is that a native code-first API may have effects that can be tweaked to returned different status codes (also for successful calls, ie: 2xx).

octonato (commented):
Looking good, but I think we should have the status code enum/mapping outside the annotation. See my comment below.

sdk/spring-sdk/src/main/java/kalix/springsdk/annotations/Acl.java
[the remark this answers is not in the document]
franciscolopezsancho: I like that. One caveat I see. What do we do with `Acl.DenyStatusCode.INHERITED`, that only has meaning for the ACL? As per documentation  `If '0' [me: that is inherited], indicates that the code should be inherited from the parent (regardless of the inherit field).`
Maybe we could set the ACL default as `0`, INHERITED, and delete it from the list.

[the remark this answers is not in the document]
franciscolopezsancho: Nope, I won't be able to set it to `0`. It must be an enum.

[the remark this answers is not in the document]
octonato: Umm, that's indeed an issue.

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/AclDescriptorFactory.scala
```diff
@@ -74,7 +73,7 @@ object AclDescriptorFactory {
       aclBuilder.addDeny(idx, principalMatcher)
     }
 
-    aclBuilder.setDenyCode(aclJavaAnnotation.denyCode())
+    aclBuilder.setDenyCode(aclJavaAnnotation.denyCode().value)
```
octonato, line 76:
```scala
if (aclJavaAnnotation.denyCode() == null)
  aclBuilder.setDenyCode(0)
else 
  aclBuilder.setDenyCode(aclJavaAnnotation.denyCode().value)
```

sdk/spring-sdk/src/main/java/kalix/springsdk/annotations/Acl.java
```diff
@@ -42,14 +42,33 @@
...
    *
-   * By default, this will be 7 (permission denied), but alternatives might include 16 (unauthenticated) or 5 (not
-   * found). If 0, indicates that the code should be inherited from the parent (regardless of the inherit field).
+   * By default, this will be 'Forbidden', but alternatives might include 'Authentication required' or 'Not
+   * Found'. If 'Inherited', indicates that the code should be inherited from the parent (regardless of the inherit field).
    *
-   * When HTTP transcoding is in use, this code will be translated to an equivalent HTTP status code.
    */
-  int denyCode() default 0;
+  DenyStatusCode denyCode() default DenyStatusCode.FORBIDDEN;
```
octonato, line 51: So, here instead of `FORBIDEN`, we should not have a default. The default should be `null` and then we translate to `0`.

octonato: I realise not that there is some confusion here. We say that the default is `7` (permission denied), but then the default was `0`. 

That was wrong, the default should be indeed `DenyStatusCode.FORBIDDEN`, mapped to `7`

sdk/spring-sdk/src/main/java/kalix/springsdk/annotations/Acl.java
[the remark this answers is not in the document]
octonato: We have a long talk to discuss it. Current decision is to keep the status code inside the ACL annotation. As such we can ensure that only a subset of status code can be used for a deny code. 

And we plan to introduce a boolean (`inheritDenyCode`) to make it more explicit that the intention is to inherit from the parent. Also, because `0` maps to http status `Ok`.  

And we need to check how the proxy is dealing with `0` as well. No deny code, means default protobuf value, which is `0` for integer. We believe that the proxy is will ultimately convert `0` to `7` (permission denied) when it reaches the top parent ACL. However, that's just a guess.

franciscolopezsancho:
related to https://github.com/lightbend/kalix-jvm-sdk/issues/1409

franciscolopezsancho:
For completeness when the `denyCode` reaches the proxy if `0` then becomes `Status.PERMISSION_DENIED`.
https://github.com/lightbend/kalix-proxy/blob/56dd38682a7013f183f57f4ea87e544ffc016397/proxy/core/src/main/scala/kalix/proxy/acl/AclRouter.scala#L186-L187
https://github.com/lightbend/kalix-proxy/blob/56dd38682a7013f183f57f4ea87e544ffc016397/proxy/core/src/main/scala/kalix/proxy/acl/AclRouter.scala#L113

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/AclDescriptorFactory.scala
```diff
@@ -74,7 +73,11 @@ object AclDescriptorFactory {
       aclBuilder.addDeny(idx, principalMatcher)
     }
 
-    aclBuilder.setDenyCode(aclJavaAnnotation.denyCode())
+    if (aclJavaAnnotation.inheritDenyCode()) {
+      aclBuilder.setDenyCode(0)
+    } else {
+      aclBuilder.setDenyCode(aclJavaAnnotation.denyCode().value)
+    }
```
octonato, line 80: I guess we don't need to force it to `FORBIDDEN` if used on the `Main` class and `inheritDenyCode` is set to true.

At the end, we will send `0` to proxy, but proxy will convert to `7`, correct?

octonato (approved):
LGTM

Issue #1047, opened 2022-07-07
Allow override of proxy docker image version used for testkit integration tests


efgpinto:
Needing to build a local SDK as well, just to inject the local proxy snapshot version, adds some complexity. Maybe we should just have the proxy docker image tag used by the SDK testkit overridable with a system property or config? Then can use a released SDK and set the local snapshot proxy version for the integration tests. Would also make it easier to run locally like this when needed.

_Originally posted by @pvlugter in https://github.com/lightbend/kalix-proxy/pull/1539#pullrequestreview-1030963156_

Note that we already provide a way to override the proxy version introduced at #1037 for when building / releasing the SDK but not at runtime from the end-user app perspective.

aludwiko:
For local development, it will be also nice to simply set the proxy address for IT test and launch the proxy from Main, without testcontainers.

johanandren:
Leaving this open as a opt out of testcontainers to manually run a proxy also sounds very useful.

PR #1411, opened 2023-01-20
feat: KALIX_TESTKIT_USE_TEST_CONTAINERS env to disable testcontainers

aludwiko:
… for it tests

References #1047

### DIFF
--- a/sdk/java-sdk-testkit/src/main/java/kalix/javasdk/testkit/KalixTestKit.java
+++ b/sdk/java-sdk-testkit/src/main/java/kalix/javasdk/testkit/KalixTestKit.java
@@ -44,8 +44,6 @@ import java.util.concurrent.CompletionStage;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeUnit;
 
-import static java.util.Optional.ofNullable;
-
 /**
  * Testkit for running Kalix services locally.
  *
@@ -198,7 +196,7 @@ public class KalixTestKit {
   public KalixTestKit start(final Config config) {
     if (started) {throw new IllegalStateException("KalixTestkit already started");}
 
-    Boolean useTestContainers = ofNullable(System.getenv("KALIX_TESTKIT_USE_TEST_CONTAINERS")).map(Boolean::valueOf).orElse(true);
+    Boolean useTestContainers = Optional.ofNullable(System.getenv("KALIX_TESTKIT_USE_TEST_CONTAINERS")).map(Boolean::valueOf).orElse(true);
     int port = userFunctionPort(useTestContainers);
     Map<String, Object> conf = new HashMap<>();
     conf.put("kalix.user-function-port", port);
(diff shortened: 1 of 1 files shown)


sdk/java-sdk-testkit/src/main/java/kalix/javasdk/testkit/KalixTestKit.java
```diff
@@ -145,7 +154,9 @@ public Settings withAdvancedViews() {
   private final Settings settings;
 
   private boolean started = false;
-  private KalixProxyContainer proxyContainer;
+  private String proxyHost = "localhost";
+  private int proxyPort = 9000;
```
johanandren, line 158: More clear these defaults are only set in the local-proxy-scenario, since they are always overwritten with the values from test containers for the default case?

johanandren (commented):
Docs mentioning the env var we put in some proxy README?

sdk/java-sdk-testkit/src/main/java/kalix/javasdk/testkit/KalixTestKit.java
```diff
@@ -185,9 +196,10 @@ public KalixTestKit start() {
    * @return this KalixTestkit
    */
   public KalixTestKit start(final Config config) {
-    if (started) throw new IllegalStateException("KalixTestkit already started");
+    if (started) {throw new IllegalStateException("KalixTestkit already started");}
 
-    int port = availableLocalPort();
+    Boolean useTestContainers = ofNullable(System.getenv("KALIX_TESTKIT_USE_TEST_CONTAINERS")).map(Boolean::valueOf).orElse(true);
```
pvlugter, line 201: Use `Optional.ofNullable` explicitly rather than static import?

pvlugter (commented):
👍🏼 useful to have this

sdk/java-sdk-testkit/src/main/java/kalix/javasdk/testkit/KalixTestKit.java
aludwiko: Yes. Do you think it will be more clear not to set them with defaults?

sdk/java-sdk-testkit/src/main/java/kalix/javasdk/testkit/KalixTestKit.java
aludwiko: fixed

sdk/java-sdk-testkit/src/main/java/kalix/javasdk/testkit/KalixTestKit.java
johanandren: Yeah, that's what I meant, since it isn't really defaults for the "normal" scenario.

aludwiko: done

PR #1417, opened 2023-01-25
removing reference to roadmap

franciscolopezsancho:
As we don't have a link to our roadmap I think is better to avoid having to update this constantly.


efgpinto (commented):
I think we should just remove most of the contents of that README and just point to the official docs. (or just remove the file altogether since we don't have it for the others sdk)

There's probably more stuff outdated there as this was meant to be temporary while we didn't have official docs.

franciscolopezsancho:
Good idea. Let's point to the docs.

franciscolopezsancho:
I finally decided to remove it altogether. For reference https://github.com/lightbend/kalix-jvm-sdk/issues/1424

efgpinto (approved):

Issue #1323, opened 2022-10-05
Native Spring SDK components


octonato:
The Spring SDK relies on the Java SDK for the components. That means that when implementing a component, users need to import classes from `kalix.javasdk`. 

That's not bad in itself, but we are constraining the Spring users to the limitations of the Java SDK.

For example, the Java SDK was designed to be used with gRPC. In gRPC, we can't have type hierarchy and therefore we can't define typed event handlers. In the Java SDK, event handlers are all of type `Object`. 

If we had a native Spring SDK EventSourcedEntity, we could have two type parameters, like in `EventSourcedEntity<Event, State>`. This would allow us to check that events handlers are all complying with `Event`. We could even do it at compilation type if we use an annotation processor and sealed classes from Java 17.
(arguably, the above could be done directly in Java by typing the Event to Object)

A native Spring SDK components would also allow us to have `asyncReply` based on Spring Reactor (see https://github.com/lightbend/kalix-jvm-sdk/issues/997).

efgpinto:
One thought that occured to me while working on this... I'm not sure if we should be implementing "native SpringSDK components" or "native codefirst components" instead. The difference being, say in a year we want to do Java/JaxRS, in theory we should not need to reimplement another set of components (but we have if those are under `springsdk`).. we should be able to reuse the "spring" ones ideally. In summary, I'm wondering if the new codefirst components can/should be placed under a different package in javasdk so they can be reused for other codefirst approaches on the JVM. (I'm not entirely aware of all the implications of such scenario, happy to hear your thoughts on it.)

PR #1426, opened 2023-01-30
feat: add support for mono in async replies

efgpinto:
References #1323 

I'm not sure if we want to merge this in or wait for the remaining small bits around this topic. Leaving as draft for now.

### DIFF
--- a/sdk/java-sdk/src/main/java/kalix/javasdk/action/Action.java
+++ b/sdk/java-sdk/src/main/java/kalix/javasdk/action/Action.java
@@ -24,7 +24,7 @@ import kalix.javasdk.impl.action.ActionContextImpl;
 import kalix.javasdk.impl.action.ActionEffectImpl;
 import kalix.javasdk.timer.TimerScheduler;
 import kalix.javasdk.impl.timer.TimerSchedulerImpl;
-import reactor.core.publisher.Mono;
+import org.reactivestreams.Publisher;
 
 import java.util.Collection;
 import java.util.Optional;
@@ -151,22 +151,25 @@ public abstract class Action {
       <S> Effect<S> asyncEffect(CompletionStage<Effect<S>> futureEffect);
 
       /**
-       * Create a message reply from an async operation result.
+       * Create a message reply from a {@link org.reactivestreams.Publisher}.
+       * Note: only the first element will be requested and used to create the reply,
+       * further elements being ignored.
        *
-       * @param message The future payload of the reply.
+       * @param publisher The publisher from which the message will be consumed to create the reply.
        * @return A message reply.
        * @param <S> The type of the message that must be returned by this call.
        */
-      <S> Effect<S> asyncReply(Mono<S> message);
+      <S> Effect<S> asyncReply(Publisher<S> publisher);
 
       /**
        * Create a reply from an async operation result returning an effect.
-       *
-       * @param futureEffect The future effect to reply with.
+       * Note: only the first element will be requested and used to create the effect,
+       * further elements being ignored.
+       * @param futureEffect The publisher from which the effect will be consumed to create the reply.
        * @return A reply, the actual type depends on the nested Effect.
        * @param <S> The type of the message that must be returned by this call.
        */
-      <S> Effect<S> asyncEffect(Mono<Effect<S>> futureEffect);
+      <S> Effect<S> asyncEffect(Publisher<Effect<S>> futureEffect);
 
       /**
        * Ignore the current element and proceed with processing the next element if returned for an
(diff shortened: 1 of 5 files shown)


sdk/java-sdk/src/main/java/kalix/javasdk/action/Action.java
```diff
@@ -24,6 +24,7 @@
 import kalix.javasdk.impl.action.ActionEffectImpl;
 import kalix.javasdk.timer.TimerScheduler;
 import kalix.javasdk.impl.timer.TimerSchedulerImpl;
+import reactor.core.publisher.Mono;
```
octonato, line 27: Mono implements `org.reactivestream.Publisher` so we could instead on Reactive Streams. 

There is a potential issue though, `Flux` also implements `Publisher`. So one could pass a `Flux` to `asyncReply`. We need to check how that will work. 

What we will happen if we have `Flux[T] => CompletionState[T]`? I guess it will only be able to emit the first element and then it will close the stream.

octonato: Actually, letting the javasdk depend on Reactor is not good. 
We either depend on Reactive Streams or we only offer CompletionStage.

I think we can accept `Publisher` and document that only the first element will be consumed.

franciscolopezsancho: I would agree with accepting `Publisher` and documenting it. The way I'm thinking about it is that you get back a lazy-List-like instance (the `Publisher`). I think it's intuitive to find out you can get one or more elements only at runtime. Nobody has a problem reasoning about lists.

efgpinto: Thanks for the feedback, updated now using reactive streams. I wonder if we should also support the [Flow.Publisher](https://docs.oracle.com/javase/9/docs/api/java/util/concurrent/Flow.Publisher.html) from java..

> The way I'm thinking about it is that you get back a lazy-List-like instance (the `Publisher`).
Not sure if I followed your suggestion.. current approach only takes the first element and ignores the rest. Let me know if your suggestion is different.

franciscolopezsancho: Looking at the implementation now I think I understand better what @octonato was saying. Which makes me think exactly what you are pointing at Eduardo.

> I wonder if we should also support the [Flow.Publisher](https://docs.oracle.com/javase/9/docs/api/java/util/concurrent/Flow.Publisher.html) from java.

From the beginning, I thought we were going to support `Flow` and `Mono` (through `Publisher`), document it and let the user deal with one or multiple elements.  I see now that I misinterpreted the previous [comment](https://github.com/lightbend/kalix-jvm-sdk/pull/1426#discussion_r1091217323).

efgpinto: > I thought we were going to support `Flow` and `Mono` (through `Publisher`)

well, we do support because both implement Publisher. So user can actually pass in a Flow. But in that case we'll only pick the first element to build the reply (I will add a test btw).

What I was mentioning as well is support for `Flow.Publisher` which is the semantically-equivalent native implementation of the reactive streams directly on the jdk, since eventually all libraries should move to use those.

johanandren: I think accepting Publisher is a bad idea, the SPI interface is to allow a stream of async messages but in this case what we really want to support is specifically a Mono, not arbitrary streaming publishers. Better keep the API tight and explicit about what it supports IMO.

octonato: Yeah, I think you are right. The initial idea was to avoid depending on Reactor directly and to avoid writing wrappers around the Java SDK. 

But your point that `Publisher` semantics is about stream and not about a single value is solid.  

I don't think it's worth to build a wrapper around the java sdk just for that. Reactor users can still build a `CompletionStage` from a `Mono`. And I don't think we should add a direct dependency on Project Reactor either. 

If we have other reasons to build a wrapper, then we can re-evaluate. 

Today, the reasons for the Spring wrapper are:

1. support for async calls accepting `Mono` 
2. error codes that don't expose gprc libraries. 
3. typed events - `EventSourcedEntity<State, Event>`

To avoid the wrapper, we can:

1. not support `Mono` directly, users will need to convert to `CompletionStage` 
2. introduce Kalix StatusCodes so code-first users don't need to deal with gRPC codes
3. expand EventSourcedEntity to have another typed parameter. Codegen can set it to `Object` or we can introduced a new type like in `EventSourcedEntity<State> extends TypedEventSourcedEntity<State, Object>`. 

So, in retrospect, we have acceptable solution for all three reasons and therefore I'm less inclined to build a wrapper.

efgpinto:
I'm closing this one and opening a cleaner PR for the remaining additions.

Issue #1391, opened 2023-01-12
Spring SDK: support timestamps in Views


efgpinto:
Split from https://github.com/lightbend/kalix-jvm-sdk/issues/1336

efgpinto:
I've spent some time trying to address this, so here are some notes:

The current main problem is when creating the descriptors. Assuming we have a Java POJO such as:
```java
public record Test(String value, Instant ts) {
}
```
We do 2 steps to get those into PB descriptors:
1. we use `ProtobufMapper` from [FasterXML/jackson-dataformats-protobuf](https://github.com/FasterXML/jackson-dataformats-binary/tree/2.15/protobuf) to map those POJOs into a ProtobufSchema
2. we manually convert that ProtobfSchema to PB descriptors

However, as reported [here](https://github.com/FasterXML/jackson-dataformats-binary/issues/349), on the first step the lib transforms any `Instance` into `double` and for that reason, on step 2 we cannot transform the field into a `google.protobuf.Timestamp` which would be the end goal. (I've tried matching by name on the original POJO and if `type == Timestamp` then harcode. The problem there is that it's a rabbit hole because you can have nested classes as well and it quickly turns into a nightmare.)


Also worth noting that the current behavior will happily take a `Instance` attribute on the POJO but if you then you will not be able to consume that on a view under the same format. It will be automatically converted to a string format and delivered to the view who will be expecting a `double` instead.

octonato:
How hard it will be to add support for in the original library? I think that at the end, it's our best option. Of course, we will need to wait for a release, but I don't see how we could fix it otherwise.

efgpinto:
> How hard it will be to add support for in the original library? I think that at the end, it's our best option. Of course, we will need to wait for a release, but I don't see how we could fix it otherwise.

Yeah, I'm not too sure what effort is involved. I can try to have a look at that. But according to the maintainer, one might need to address something else first:
> One practical blocker might be that the underlying protoc parser this module uses should probably replaced first (see https://github.com/FasterXML/jackson-dataformats-binary/issues/121). I suspect this might be necessary to find relevant metadata, unless this is actually something specified by Protobuf spec.

octonato:
Yeah, I saw it but didn't understand why the parser could be a problem.

PR #1427, opened 2023-01-30
fix: convert Java Instant fields to proto Timestamp

octonato:
Quite hacky solution for issue with views for types containing `Instant`. I still need to confirm that this will work. The current tests is only patching the proto message to use proto Timestamp instead. 

References #1391

### DIFF
--- a/docs/src/modules/java/partials/query-syntax-reference.adoc
+++ b/docs/src/modules/java/partials/query-syntax-reference.adoc
@@ -204,11 +204,11 @@ When modeling your queries, the following data types are supported:
 | Array
 | `Collection<T>` and derived
 
-// not supported at the moment, see https://github.com/lightbend/kalix-jvm-sdk/issues/1391
-//| Timestamp
-//| `Instant`
-
+| Timestamp
+| `java.time.Instant`
 |===
+
+NOTE: `Instant` are only supported for queries with a `@RequestBody`. It's currently not possible to pass an `Instant` as a `@PathVariable` or `@RequestParam`.
 
 ==== Optional fields
 
--- a/sdk/spring-sdk-testkit/src/main/java/kalix/springsdk/testkit/KalixConfigurationTest.java
+++ b/sdk/spring-sdk-testkit/src/main/java/kalix/springsdk/testkit/KalixConfigurationTest.java
@@ -16,6 +16,7 @@
 
 package kalix.springsdk.testkit;
 
+import kalix.javasdk.JsonSupport;
 import kalix.javasdk.testkit.KalixTestKit;
 import kalix.springsdk.KalixConfiguration;
 import kalix.springsdk.impl.KalixServer;
@@ -26,9 +27,9 @@ import org.springframework.boot.test.context.TestConfiguration;
 import org.springframework.context.ApplicationContext;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Import;
-import org.springframework.core.env.Environment;
 import org.springframework.http.HttpHeaders;
 import org.springframework.http.MediaType;
+import org.springframework.http.codec.json.Jackson2JsonEncoder;
 import org.springframework.web.reactive.function.client.WebClient;
 
 /** Spring test configuration for Kalix integration tests. */
@@ -52,6 +53,10 @@ public class KalixConfigurationTest {
     return WebClient.builder()
         .baseUrl("http://localhost:" + kalixTestKit.getPort())
         .defaultHeader(HttpHeaders.CONTENT_TYPE, MediaType.APPLICATION_JSON_VALUE)
+        .codecs(configurer -> {
+          configurer.defaultCodecs().jackson2JsonEncoder(
+            new Jackson2JsonEncoder(JsonSupport.getObjectMapper(), MediaType.APPLICATION_JSON));
+        })
         .build();
   }
 
--- a/sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/ProtoDescriptorGenerator.scala
+++ b/sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/ProtoDescriptorGenerator.scala
@@ -21,6 +21,7 @@ import com.google.protobuf.AnyProto
 import com.google.protobuf.DescriptorProtos
 import com.google.protobuf.Descriptors
 import com.google.protobuf.EmptyProto
+import com.google.protobuf.TimestampProto
 import kalix.{ Annotations => KalixAnnotations }
 import org.slf4j.LoggerFactory
 
@@ -33,6 +34,7 @@ private[impl] object ProtoDescriptorGenerator {
     Array(
       AnyProto.getDescriptor,
       EmptyProto.getDescriptor,
+      TimestampProto.getDescriptor,
       HttpAnnotationsProto.getDescriptor,
       KalixAnnotations.getDescriptor)
 
@@ -51,6 +53,7 @@ private[impl] object ProtoDescriptorGenerator {
 
     protoBuilder.addDependency("google/protobuf/any.proto")
     protoBuilder.addDependency("google/protobuf/empty.proto")
+    protoBuilder.addDependency("google/protobuf/timestamp.proto")
     protoBuilder.addService(service)
     messages.foreach(protoBuilder.addMessageType)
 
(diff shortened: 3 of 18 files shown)


sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/ProtoDescriptorGenerator.scala
```diff
@@ -33,6 +34,7 @@ private[impl] object ProtoDescriptorGenerator {
     Array(
       AnyProto.getDescriptor,
       EmptyProto.getDescriptor,
+      Timestamp.getDescriptor.getFile,
```
efgpinto, line 37: why not `TimestampProto.getDescriptor` ? Is there a difference actually?

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/ProtoMessageDescriptors.scala
```diff
@@ -16,34 +16,110 @@
...
+      javaClass: Class[_]): Boolean = {
+
+    def lookupOriginalJavaType(jacksonType: ProtobufMessage, javaClass: Class[_]): Option[Class[_]] = {
+
+      def isClassMatch(cls: Class[_]): Boolean = cls.getSimpleName == jacksonType.getName
+      def isFieldMatch(field: Field): Boolean = isClassMatch(field.getType)
+
+      if (isClassMatch(javaClass)) Some(javaClass)
+      else {
+        javaClass.getFields
```
efgpinto, line 67: will this `.getFields` work if javaClass is a record? I think it will always return empty.. I think you'll need to use `getDeclaredFields` or `getRecordComponents`

efgpinto (commented):
just a quick first review, will give it a 2nd look tomorrow... looking as good as possible. glad we can make this work as long as we cover it with a good amount of tests.

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/ProtoDescriptorGenerator.scala
octonato: We need a `FileDescriptor` and `Timestamp.getDescriptor` returns `Descriptors.Descriptor`. I don't know the reason. Probably related with how the proto was defined?

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/ProtoMessageDescriptors.scala
octonato: Yeah, I will check this. 

Btw, I converted one of the types to java class, namely the `TimeEnum` because IntelliJ was complaining. I will revert it back to record and may also convert some of the types we have in the tests to record.

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/ProtoDescriptorGenerator.scala
efgpinto: Yeah, but you can use `TimestampProto.getDescriptor` instead of `Timestamp.getDescriptor.getFile` I believe.. at least it worked when I was trying this a few weeks back

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/ProtoMessageDescriptors.scala
efgpinto: > I converted one of the types to java class, namely the `TimeEnum` because IntelliJ was complaining.

yeah, my intellij also compains about that one.. I wonder if we have any missing config

efgpinto:
One other thing we might need to consider is what to do with other related types? The JavaTimeModule adds serializers for related types such as OffsetDateTime, ZonedDateTime and others. I assume we would have a similar issue there.

octonato:
> One other thing we might need to consider is what to do with other related types? The JavaTimeModule adds serializers for related types such as OffsetDateTime, ZonedDateTime and others. I assume we would have a similar issue there.

This might be a little more tricker. We can only support formats with string representations that are compatible to proto Timestamp because that's what proxy will use to run the query when a request comes in. We need to check it out. May work for some of the times.

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/ProtoMessageDescriptors.scala
octonato: I got it working by:

* configuring sbt to always use 17 for all the Spring SDK artifacts
* manually configure IntelliJ to use 17 for the Spring SDK modules

build.sbt
```diff
@@ -133,9 +133,9 @@ lazy val sdkSpring = project
     buildInfoPackage := "kalix.springsdk",
     Compile / akkaGrpcGeneratedSources := Seq(AkkaGrpc.Server),
     Compile / akkaGrpcGeneratedLanguages := Seq(AkkaGrpc.Scala),
-    Test / javacOptions ++= Seq("-parameters", "--release", "17"), // for Jackson
-    Test / scalacOptions ++= Seq("-release", "17"),
-    IntegrationTest / javacOptions += "-parameters", // for Jackson
+    Compile / javacOptions ++= Seq("-parameters" /* for Jackson */, "--release", "17"),
+    Compile / scalacOptions ++= Seq("-release", "17"),
```
octonato, line 137: Actually we should revert this. We only want the tests to be on 17, not the production artifacts.

octonato (commented):
Finally got tests passing (locally).

Last run in CI failed with a timeout. Let's see.

sdk/spring-sdk-testkit/src/main/java/kalix/springsdk/testkit/KalixConfigurationTest.java
```diff
@@ -52,6 +53,10 @@ public WebClient createWebClient(KalixTestKit kalixTestKit) {
     return WebClient.builder()
         .baseUrl("http://localhost:" + kalixTestKit.getPort())
         .defaultHeader(HttpHeaders.CONTENT_TYPE, MediaType.APPLICATION_JSON_VALUE)
+        .codecs(configurer -> {
+          configurer.defaultCodecs().jackson2JsonEncoder(
+            new Jackson2JsonEncoder(JsonSupport.getObjectMapper(), MediaType.APPLICATION_JSON));
```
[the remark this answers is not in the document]
octonato: I just realised that the Jackson Protobuf lib has a limitation. 

A type with two complex fields that have the same simple class name will generate an error. For example:

```java

interface A {
  record Foo(int i)
}

interface B {
  record Foo(String s)
}

class Bar {
  public A.Foo fooA;
  public B.Foo fooB;
}
```

This will generate a message for `Bar` and two messages for `Foo`. 
The `Foo` messages will identical. 

They will have either a `int32` field or a `string` field. Basically, it will consider that they are the same.

That's because protobuf messages doesn't have namespace.

We can't do much about it. We can validate and throw an exception instructing users how to fix though.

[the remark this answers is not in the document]
octonato: I created an issue to track this. https://github.com/lightbend/kalix-jvm-sdk/issues/1432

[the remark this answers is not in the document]
efgpinto: 2 nice finds! 👍

docs/src/modules/java/partials/query-syntax-reference.adoc
```diff
@@ -204,12 +204,12 @@ When modeling your queries, the following data types are supported:
...
 
-// not supported at the moment, see https://github.com/lightbend/kalix-jvm-sdk/issues/1391
-//| Timestamp
-//| `Instant`
-
+| Timestamp
+| `java.time.Instant`
 |===
 
+NOTE: `Instant` are only supported for queries with a `@RequestBody`. It's currently not possible to pass an `Instant` as a `@PathVariable` or `@RequestParam`.
```
efgpinto, line 211: Let's create an issue for this?

efgpinto (approved):
LGTM.

I think we should open another issue for this:
> This might be a little more tricker. We can only support formats with string representations that are compatible to proto Timestamp because that's what proxy will use to run the query when a request comes in. We need to check it out. May work for some of the times.

docs/src/modules/java/partials/query-syntax-reference.adoc
octonato: Yep, I will

PR #1430, opened 2023-02-01
chore: release 11 and 17 flags

patriknw:
* Spring SDK requires JDK 17
* Others JDK 11

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1333,7 +1333,7 @@ jobs:
 
   publish:
     docker:
-      - image: circleci/openjdk:11
+      - image: cimg/openjdk:17.0
     steps:
       - checkout
       - setup_sbt
--- a/build.sbt
+++ b/build.sbt
@@ -21,8 +21,8 @@ lazy val `kalix-jvm-sdk` = project
 
 def common: Seq[Setting[_]] =
   Seq(
-    Compile / javacOptions ++= Seq("-encoding", "UTF-8", "--release", "11"),
-    Compile / scalacOptions ++= Seq("-encoding", "UTF-8", "-release", "11", "-deprecation"))
+    Compile / javacOptions ++= Seq("-encoding", "UTF-8"),
+    Compile / scalacOptions ++= Seq("-encoding", "UTF-8", "-deprecation"))
 
 lazy val sdkCore = project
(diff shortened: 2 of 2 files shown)


.circleci/config.yml
```diff
@@ -1333,7 +1333,7 @@ jobs:
 
   publish:
     docker:
-      - image: circleci/openjdk:11
+      - image: cimg/openjdk:17.0
```
patriknw, line 1336: this is for ci-release

build.sbt
```diff
@@ -133,8 +141,7 @@ lazy val sdkSpring = project
     buildInfoPackage := "kalix.springsdk",
     Compile / akkaGrpcGeneratedSources := Seq(AkkaGrpc.Server),
     Compile / akkaGrpcGeneratedLanguages := Seq(AkkaGrpc.Scala),
-    Test / javacOptions ++= Seq("-parameters", "--release", "17"), // for Jackson
-    Test / scalacOptions ++= Seq("-release", "17"),
```
patriknw, line 137: I guess these were misplaced in Test scope

efgpinto: Renato had the same thought 2 days ago but then said "We only want the tests to be on 17, not the production artifacts." [here](https://github.com/lightbend/kalix-jvm-sdk/pull/1427#discussion_r1092041066). I'm not sure what's the rational though.

.circleci/config.yml
efgpinto: the interesting part is that the "setup_sbt" already installs java 17, but from the logs sbt is still using the 11 anyway.. so I think we might be able to do some clean up but I will have a look in a later PR.

efgpinto (commented):
This LGTM but would probably be good to have @octonato input because I know he considered the change from Test to Compile scope before.

build.sbt
octonato: I changed it and then reverted in the PR for the Instant. Now, if SpringBoot 3.0 requires 17, then we can simply say that Spring SDK requires 17. 

For us that will be perfect, because we will be able to use records in our code base.

octonato (approved):
LGTM

Indeed, SpringBoot3 requires JDK 17. Fine to move the release SpringSDK for 17 then. Let's do it.

PR #1436, opened 2023-02-01
feat: add Kalix error status code

efgpinto:
References #1323 

I have some questions about this one, so leaving them here and as draft for now:
- I had discussed briefly with @octonato about maybe having some sealed class and having a type for gRPC and another for HTTP:
  - while implementing this, I was not so sure of what the advantages of this are.. since we support all gRPC possible codes anyway, I'm not sure we need to move those under this new sealed interface.
- implemented this version using a java enum (can't use sealed interfaces)  but I would like some input if this is a good option overall or if there are enough good reasons to move this to scala
  - in this case, we want to reduce the set of HTTP codes we accept (because some of them will be lost in translation HTTP -> gRPC -> HTTP), otherwise I would probably advocate for letting the user pass in the HTTP code directly.

Task breakdown:

- [x] implement new error method for all java component types
- [x] ~implement new error method for all scala component types~
- [x] ~deprecate old method~
- [x] support for deferred calls
- [x] update samples to use new `ErrorCode`
- [x] review/update docs if needed


efgpinto:
Will redo after chatting with Renato:
- we want this ErrorStatusCode to replace also the existing one
- it should be agnostic to both gRPC and HTTP but we document which code that will be turned into
- the main interface should allow not only for error codes but also success ones if we need to have them in the future
- since we are deprecating the old API, scala sdk needs the new version as well

efgpinto:
I was starting to update the samples so they don't use the deprecated gRPC StatusCode and use instead our internal representation and now I'm not sure the benefits are worth the added cognitive load, specifically for the protocol-first SDKs. See example:
- with the new API, client will do e.g. `effects().error("Error", ErrorCode.BAD_REQUEST)`
- now, when it comes to do integration tests one will have to still do: `assertThat(suppressed.getStatus().getCode(), is(Status.Code.INVALID_ARGUMENT));`
- so, one will need to know that the internal "ErrorCode.BAD_REQUEST" will be mapped to the `Status.Code.INVALID_ARGUMENT`

If I'm an experienced gRPC-minded developer I'm very likely used to deal with the gRPC codes directly, so I'm not sure the benefit of hidding those is worth the added trouble of having to double-check which internal code maps to which gRPC code. 

The same happens to some extent on the HTTP side with new API but there I think the cognitive load is much smaller simply because the names we are using match the HTTP error codes, e.g. the use of `ErrorCode.BAD_REQUEST` to reject a request will lead to a `400` code on the client side as one would expect and guess.

TLDR: I'm not so sure we need to deprecated the API for passing gRPC StatusCode directly. We would support both and each dev can choose the one they see fit.

efgpinto:
The other concerned raised has to do with `DeferredCall`s and the fact that they exposed the gRPC Status Code to the developer. In that regard, I think we can catch the exception and wrap it under a friendly exception while converting the status code to the internal error code. I will test this approach now.

efgpinto:
> The other concerned raised has to do with `DeferredCall`s and the fact that they exposed the gRPC Status Code to the developer. In that regard, I think we can catch the exception and wrap it under a friendly exception while converting the status code to the internal error code. I will test this approach now.

@octonato added a solution for the above in https://github.com/lightbend/kalix-jvm-sdk/pull/1436/commits/98c771292b0363fb603d52f0e99ab56304f92354 ... I think it's fine actually but let me know your thoughts.

samples/spring-reliable-timers/src/main/java/com/example/actions/OrderAction.java
```diff
@@ -75,20 +78,15 @@ public Effect<Order> placeOrder(@RequestBody OrderRequest orderRequest) {
...
-
     var cancelRequest = kalixClient.post("/order/"+orderId+"/cancel", "", String.class);
 
     CompletionStage<String> reply =
         cancelRequest
             .execute() // <1>
             .thenApply(cancelled -> "Ok") // <2>
             .exceptionally(e -> { // <3>
-                  if (e instanceof StatusRuntimeException && validateErrorCodes.test((StatusRuntimeException) e)) {
+                  if (e.getCause() instanceof DeferredCallResponseException ere &&
```
octonato, line 88: ere?

sdk/java-sdk/src/main/scala/kalix/javasdk/impl/RestDeferredCall.scala
```diff
@@ -31,9 +33,19 @@ final case class RestDeferredCall[I, O](
...
     asyncCall: () => CompletionStage[O])
     extends DeferredCall[I, O] {
-  override def execute(): CompletionStage[O] = asyncCall()
+  override def execute(): CompletionStage[O] = asyncCall().exceptionally {
+    case responseException: WebClientResponseException =>
+      throw DeferredCallResponseException(
+        responseException.getMessage,
+        StatusCodeConverter.fromWebClientResponse(responseException),
+        responseException)
+    case other: Throwable => throw other
```
octonato, line 42: That would also work for gRPC calls, no? 

I mean, if I'm executing a `GrpcDeferredCall` we could also convert it to a Kalix ErrorCode so the types that I'm exposed to will be the same as when I call `effect.error()`. 

The only problem is that such a change is backward incompatible. Code that is already handling this kind of exceptions is already expecting `io.grpc.Status`. So we can't change it. 

With the current solution, code-first SDK using http, will be exposed to `ErrorCode` on both ends, but Protobuf SDK will have `ErrorCode` when defining errors, but `io.grpc.Status` when handling errors.

efgpinto: Yeah, that is also one reason I would argue for not deprecating error() with Status.Code. I think I would just keep both APIs.

samples/spring-reliable-timers/src/main/java/com/example/actions/OrderAction.java
efgpinto: the exception had a different name before.. 😂 will update. tks

samples/spring-reliable-timers/src/it/java/com/example/OrderActionIntegrationTest.java
```diff
@@ -5,6 +5,7 @@
 import com.example.domain.OrderStatus;
 import kalix.springsdk.testkit.KalixIntegrationTestKitSupport;
 
+import org.jetbrains.annotations.Nullable;
```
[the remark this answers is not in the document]
efgpinto: oh yeah, it was an automatic import from a refactor, will remove

efgpinto:
After some back and forth, this is my current view for this PR:
- clients can choose between passing a grpc or an http error code (I've changed the parameter name to be more explicit, see below)
Screenshot 2023-02-08 at 11 26 42
- protocol-first and code-first users get a consistent experience as they can always use grpc or http on all places (i.e. error(), handling deferredcalls or integration testing)
- no breaking change

We can still decide as a follow-up to deprecate the old `error()` method with grpc status code, although that will be a breaking change if we want to make it consistent for the deferred calls as well. But that is due to the change itself, nothing in this PR makes that part harder.

What do you think @octonato ?

octonato (commented):
Indeed, we will have to accept the fact that both codes will have to live next to each other. I don't see how we could solve it without breaking compatibility and without having wrappers. 

As discussed on call, let's go the minimal change. Adding http codes for the Java SDK, keeping the gRPC codes and not changing the Scala SDK since it's only used with gRPC for now.

efgpinto:
> As discussed on call, let's go the minimal change. Adding http codes for the Java SDK, keeping the gRPC codes and not changing the Scala SDK since it's only used with gRPC for now.

Done now! Ready for final review

sdk/java-sdk/src/main/java/kalix/javasdk/valueentity/ValueEntity.java
```diff
@@ -171,7 +173,19 @@ protected final Effect.Builder<S> effects() {
...
+       * Create an error reply with a custom status code.
+       * This status code will be translated to an HTTP or gRPC code
+       * depending on the type of service being exposed.
+       *
+       * @param description The description of the error.
+       * @param errorCode A custom Kalix status code to represent the error.
+       * @return An error reply.
+       * @param <T> The type of the message that must be returned by this call.
+       */
+      <T> Effect<T> error(String description, StatusCode.ErrorCode httpErrorCode);
```
octonato, line 188: The javadoc tries to be more generic by saying that it will be converted to Http or gRPC, but the param name is `httpErrorCode`. And the names are aligned with http error codes. 

I think we can just refer to it as Http Error code instead. 

Eventually, a gRPC user will use this method and handle exceptions of type `StatusRuntimeException`. But that's fine. Such a user will be already passing `ErrorCode.BAD_REQUEST` on one side and handling `io.grpc.Status.Code.INVALID_ARGUMENT` on the other side.

sdk/java-sdk/src/main/java/kalix/javasdk/StatusCode.java
```diff
@@ -0,0 +1,41 @@
...
+    UNAUTHORIZED,
+    FORBIDDEN,
+    NOT_FOUND,
+    CONFLICT,
+    TOO_MANY_REQUESTS,
+    INTERNAL_SERVER_ERROR,
+    SERVICE_UNAVAILABLE,
+    GATEWAY_TIMEOUT
+  }
+}
```
octonato, line 41: What about this...

```java
public interface StatusCode {

  interface ErrorCode extends StatusCode {}
  /**
   * The supported error codes that can be used when replying from the Kalix user function.
   *
   * <p>Note: the error code will be transcoded to a gRPC code depending on the
   * type of service being exposed.
   */
  enum HttpErrorCode implements ErrorCode {
    BAD_REQUEST,
    UNAUTHORIZED,
    FORBIDDEN,
    NOT_FOUND,
    CONFLICT,
    TOO_MANY_REQUESTS,
    INTERNAL_SERVER_ERROR,
    SERVICE_UNAVAILABLE,
    GATEWAY_TIMEOUT
  }
}
```

Eventually, in the future we can introduce:

```java
enum GrpcErrorCode implements ErrorCode {
  ... 
}
```

I would keep the `error()` method accepting `ErrorCode` in case we want to have `GrpcErrorCode`.

octonato: There is a catch on my suggestion above. 

The javasdk is compiling with jdk11 and we can't use sealed interfaces.

The current impl, prevents new error codes because it's an enum. While my suggestion would make `ErrorCode' and open interface.

octonato (commented):
Looks good. 

I know we have had some many iterations on this, but I raised some minor suggestions. Just to make sure that we covered everything.

samples/spring-reliable-timers/src/main/java/com/example/actions/OrderAction.java
```diff
@@ -75,25 +74,20 @@ public Effect<Order> placeOrder(@RequestBody OrderRequest orderRequest) {
...
-
     var cancelRequest = kalixClient.post("/order/"+orderId+"/cancel", "", String.class);
 
     CompletionStage<String> reply =
         cancelRequest
             .execute() // <1>
             .thenApply(cancelled -> "Ok") // <2>
             .exceptionally(e -> { // <3>
-                  if (e instanceof StatusRuntimeException && validateErrorCodes.test((StatusRuntimeException) e)) {
+                  if (e.getCause() instanceof DeferredCallResponseException dcre &&
```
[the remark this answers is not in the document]
efgpinto: I think you might have seen it working on the scala sample though.. in this case, both the spring and the java one behave similarly.. the wrapping is on the `exceptionally` implementation.

sdk/java-sdk/src/main/java/kalix/javasdk/StatusCode.java
efgpinto: > There is a catch on my suggestion above.
> 
> The javasdk is compiling with jdk11 and we can't use sealed interfaces.
> 
> The current impl, prevents new error codes because it's an enum. While my suggestion would make `ErrorCode' and open interface.

Yup, been there. :D 

That was exactly the rational. Since we don't have a sealed interface I decided to use make it closed with the Enum.

That said, I think given the current strategy, it might make sense to follow your suggestion and have it explicitely as `HttpErrorCode`. Will update.

efgpinto: Oh, there was another reason for the current approach.. my thinking was also: if we change our minds and want to be able to have agnostic internal error codes after all, with the current approach that would still be possible. We would just need to rename the param on `error()` to drop the `http` but it would not need any breaking change for the API, it would still be source-compatible. If we tie the type to Http then I think we have a stronger commitment to have this dual approach supporting HTTP and gRPC.

octonato: Ok, I see. Let's keep it as is then. 

There is a merge conflict now. I will fix it myself and merge this PR.

octonato (approved):

project/Dependencies.scala
```diff
@@ -104,6 +105,7 @@ object Dependencies {
     jacksonJdk8,
     jacksonJsr310,
     jacksonParameterNames,
+    webflux,
```
octonato, line 108: I realise now that we introduced a spring dependency to the java-sdk. We should find another way to solve that. 

We need to convert the exception at the level of the `KalixClientImpl` instead.

efgpinto: fixed [here](https://github.com/lightbend/kalix-jvm-sdk/pull/1473) along with a few bits remaining from your review

PR #1437, opened 2023-02-02
chore: Update nexus-staging-maven-plugin for JDK 17, #1435

### LINKED ISSUE #1435: Can't release maven-java with JDK 17
Updated to JDK 17 in https://github.com/lightbend/kalix-jvm-sdk/pull/1431

```
[INFO] Kalix Java Maven parent pom ........................ FAILURE [  2.296 s]
...
[ERROR] import: Entry[import  from realm ClassRealm[maven.api, parent: null]]
[ERROR] 
[ERROR] -----------------------------------------------------: Unable to make field private final java.util.Comparator java.util.TreeMap.comparator accessible: module java.base does not "opens java.util" to unnamed module @25d3cfc8
[ERROR]
```

It's in maven-java: `mvn -P release -B deploy`

I tried building locally but then some other problem with the gpg plugin.

Maybe we are just using old versions, I haven't looked or investigated further.

patriknw:
I was able to reproduce the "Unable to make field private final..." error locally. Seems fixed in later nexus-staging-maven-plugin version.

Fixes #1435

### DIFF
--- a/maven-java/pom.xml
+++ b/maven-java/pom.xml
@@ -58,7 +58,7 @@
         <plugin>
           <groupId>org.sonatype.plugins</groupId>
           <artifactId>nexus-staging-maven-plugin</artifactId>
-          <version>1.6.7</version>
+          <version>1.6.13</version>
           <extensions>true</extensions>
           <configuration>
             <serverId>ossrh</serverId>


patriknw:
here is a confirmation that this should fix it https://stackoverflow.com/questions/70153962/nexus-staging-maven-plugin-maven-deploy-failed-an-api-incompatibility-was-enco

efgpinto (approved):
LGTM (just re-kicked the pipeline)

PR #1440, opened 2023-02-02
fix: create user using 'now' time instead

octonato:


sdk/spring-sdk/src/it/java/com/example/wiring/SpringSdkIntegrationTest.java
```diff
@@ -533,20 +533,18 @@ public void shouldPropagateMetadataWithHttpDeferredCall() {
...
-
+    createCustomer(new CustomerEntity.Customer("customer1", now));
 
     // the view is eventually updated
     await()
       .ignoreExceptions()
       .atMost(20, TimeUnit.SECONDS)
       .until(() -> getCustomersByCreationDate(now).size(), new IsEqual(1));
 
+    var later = now.plusSeconds(60 * 5);
```
octonato, line 547: I think this tests still have a flaw. I'm trying to understand how that passed previously. 

The first assertion will pass because we are searching with `now` and customer was created with `later` and now, in this PR it's create with `now`.

The second assertion is now 'correct'. It searches with `later` while Customer was created with `now`. So, indeed it show return nothing. 

I'm trying to think how that passed before. My wilde assumption is that when take a `flux.stream.toList` we may produce an empty list if there is not yet an element in the stream. 

That the only thing I can image to explain how that call returned zero elements while it should have return 1 and fail.

That all said, the test is not robust enough because it may return 0 not because the query returns 0, but because the element was not yet in the stream.

octonato: Ok, I will refactor this to use a relational projection and return a wrapper with a list instead of Flux. Then we are sure about the size of elements when we get the response.

efgpinto (approved):
LGTM

Issue #1439, opened 2023-02-02
fail: EventSourcedEntityIntegrationTest.verifyCounterEventSourcedAfterRestartFromSnapshot


patriknw:
https://app.circleci.com/pipelines/github/lightbend/kalix-jvm-sdk/5513/workflows/b1c1ec20-cbe7-4e37-ad93-e48e6d3155ac/jobs/104242?invite=true#step-109-489

```
[error] Test com.example.wiring.EventSourcedEntityIntegrationTest.verifyCounterEventSourcedAfterRestartFromSnapshot failed: java.lang.IllegalStateException: Timeout on blocking read for 10000000000 NANOSECONDS, took 10.595s
[error]     at reactor.core.publisher.BlockingSingleSubscriber.blockingGet(BlockingSingleSubscriber.java:123)
[error]     at reactor.core.publisher.Mono.block(Mono.java:1734)
[error]     at com.example.wiring.EventSourcedEntityIntegrationTest.getCounter(EventSourcedEntityIntegrationTest.java:147)
[error]     at com.example.wiring.EventSourcedEntityIntegrationTest.verifyCounterEventSourcedAfterRestartFromSnapshot(EventSourcedEntityIntegrationTest.java:113)
[error]     at jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
[error]     at jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:77)
[error]     at jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
[error]     at java.lang.reflect.Method.invoke(Method.java:568)
```

PR #1441, opened 2023-02-02
fix: use await after restarting an entity

octonato:
References https://github.com/lightbend/kalix-jvm-sdk/issues/1439

### DIFF
--- a/sdk/spring-sdk/src/it/java/com/example/wiring/EventSourcedEntityIntegrationTest.java
+++ b/sdk/spring-sdk/src/it/java/com/example/wiring/EventSourcedEntityIntegrationTest.java
@@ -110,7 +110,12 @@ public class EventSourcedEntityIntegrationTest {
     restartCounterEntity(counterId);
 
     // current state is based on snapshot and should be the same as previously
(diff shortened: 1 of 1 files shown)


sdk/spring-sdk/src/it/java/com/example/wiring/EventSourcedEntityIntegrationTest.java
```diff
@@ -110,7 +110,12 @@ public void verifyCounterEventSourcedAfterRestartFromSnapshot() {
     restartCounterEntity(counterId);
 
     // current state is based on snapshot and should be the same as previously
-    Assertions.assertEquals(10, getCounter(counterId));
+    await()
+      .ignoreExceptions()
+      .atMost(10, TimeUnit.of(SECONDS))
+      .until(
+        () ->getCounter(counterId),
+        new IsEqual(10));
```
octonato, line 118: Here we call it after restarting the entity. I'm assuming here that it simply took too long to restart it so using `await.util` to retry it.

efgpinto (approved):
Makes sense.

PR #1449, opened 2023-02-03
chore: Use regular text logs for dev mode instead of JSON

johanandren:
For the user function itself, proxy logs are a separate thing


johanandren:
IT tests in samples looks like they already have a logback-text.xml present

efgpinto:
So, for the spring ones I already have the logs not formatted in JSON without your changes.. which got me wondering what am I missing? or is the json format miss-configured for the spring ones actually and we never noticed?

Btw, this is a great improvement for DX I think. 👍

octonato:
> For grpc sdks this works as expected but for the spring ones I already have the logs not formatted in JSON without your changes.. which got me wondering what am I missing? or is the json format miss-configured for the spring ones actually and we never noticed?
> 
> Btw, this is a great improvement for DX I think. 👍

Spring ones are using STDOUT, instead of JSON-STDOUT. Although the config is there.

octonato (approved):
LGTM

I will open another one to fix the Spring samples

PR #1450, opened 2023-02-03
feat: More error details about what service when a cross call fails

johanandren:
JVM SDK version of the more general https://github.com/akka/akka-grpc/pull/1748


efgpinto (approved):
This seems very useful!

PR #1451, opened 2023-02-04
chore: fix logger format in spring samples and archetypes

octonato:
Follow up of https://github.com/lightbend/kalix-jvm-sdk/pull/1449

Spring requires different settings when configuring logback through maven. 

PR updating the archetypes as well.

### DIFF
--- a/maven-java/kalix-spring-boot-archetype/src/main/resources/archetype-resources/pom.xml
+++ b/maven-java/kalix-spring-boot-archetype/src/main/resources/archetype-resources/pom.xml
@@ -146,6 +146,7 @@
           <arguments>
             <argument>-classpath</argument>
             <classpath/>
+            <argument>-Dlogging.config=src/main/resources/logback-dev-mode.xml</argument>
             <argument>${mainClass}</argument>
           </arguments>
           <environmentVariables>
@@ -223,6 +224,11 @@
        <groupId>org.springframework.boot</groupId>
        <artifactId>spring-boot-maven-plugin</artifactId>
        <version>3.0.2</version>
+        <configuration>
+            <systemPropertyVariables>
+                <logging.config>src/main/resources/logback-dev-mode.xml</logging.config>
+            </systemPropertyVariables>
+          </configuration>
       </plugin>
       <plugin>
         <groupId>org.apache.maven.plugins</groupId>
--- a/samples/spring-customer-registry-views-quickstart/pom.xml
+++ b/samples/spring-customer-registry-views-quickstart/pom.xml
@@ -146,7 +146,7 @@
           <arguments>
             <argument>-classpath</argument>
             <classpath/>
-            <argument>-Dlogback.configurationFile=logback-dev-mode.xml</argument>
+            <argument>-Dlogging.config=src/main/resources/logback-dev-mode.xml</argument>
             <argument>${mainClass}</argument>
           </arguments>
           <environmentVariables>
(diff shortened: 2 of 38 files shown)


maven-java/kalix-spring-boot-archetype/src/main/resources/archetype-resources/pom.xml
```diff
@@ -146,6 +146,7 @@
           <arguments>
             <argument>-classpath</argument>
             <classpath/>
+            <argument>-Dlogging.config=src/main/resources/logback-dev-mode.xml</argument>
```
octonato, line 149: Spring takes control of the logging infra and tries to have a single property for all log frameworks. Instead of passing `logback.configurationFile`, we should pass `logging.config` and with the full path to the file.

maven-java/kalix-spring-boot-archetype/src/main/resources/archetype-resources/pom.xml
```diff
@@ -223,6 +224,11 @@
        <groupId>org.springframework.boot</groupId>
        <artifactId>spring-boot-maven-plugin</artifactId>
        <version>3.0.2</version>
+        <configuration>
+            <systemPropertyVariables>
+                <logging.config>src/main/resources/logback-dev-mode.xml</logging.config>
+            </systemPropertyVariables>
```
octonato, line 230: `systemPropertyVariables` is how we pass `-D` properties through the spring-boot-maven-plugin

samples/spring-customer-registry-views-quickstart/pom.xml
```diff
@@ -146,7 +146,7 @@
           <arguments>
             <argument>-classpath</argument>
             <classpath/>
-            <argument>-Dlogback.configurationFile=logback-dev-mode.xml</argument>
+            <argument>-Dlogging.config=src/main/resources/logback-dev-mode.xml</argument>
```
aludwiko, line 149: I think this file `logback-dev-mode.xml` is missing in all spring samples.

efgpinto: It was added on https://github.com/lightbend/kalix-jvm-sdk/pull/1449 .. in main now I believe.

maven-java/kalix-spring-boot-archetype/src/main/resources/archetype-resources/pom.xml
efgpinto: And this config value gets overriden somewhere with `resources/logback-dev-mode.xml` when the app is deployed?

efgpinto (commented):
Looking good, left a question

Also, wouldn't this mean no logs for spring sdk deployed apps? As the logs would need to be JSON formatted to be picked up? I _think_ I've seen logs for spring-sdk deployed apps though.

johanandren:
Very important that actually deployed apps get the json layout or else Kalix does not understand what is one entry, log levels etc.

maven-java/kalix-spring-boot-archetype/src/main/resources/archetype-resources/pom.xml
octonato: This one certainly not because it is the one from Spring Boot and is not used in the docker image. But maybe the one we are passing to maven exec may be problematic. 

What I expect is that nothing will be passed to the jvm in docker unless explicitly configured in the docker plugin. But we will need to check. I will be surprised if the docker image is build taking into consideration the configurations that we pass to `mvn exec`. 

But let's confirm it first.

octonato: Ok, I can confirm that none of the configs passed to `spring-boot-maven-plugin` or `exec-maven-plugin` are picked by `docker-maven-plugin`. 

The entry point in the image is: 

```
"Entrypoint": [
    "java",
    "-jar",
    "/maven/spring-fibonacci-action-1.0-SNAPSHOT.jar"
]
```

Which is exactly what we declared in `docker-maven-plugin`. So not hidden magic.

aludwiko (approved):

PR #1453, opened 2023-02-06
fix: decode messages from proxy after workflow restart

aludwiko:
References https://github.com/lightbend/kalix-proxy/issues/1882

### DIFF
--- a/sdk/java-sdk/src/main/java/kalix/javasdk/workflowentity/WorkflowEntity.java
+++ b/sdk/java-sdk/src/main/java/kalix/javasdk/workflowentity/WorkflowEntity.java
@@ -21,9 +21,15 @@ import io.grpc.Status;
 import kalix.javasdk.DeferredCall;
 import kalix.javasdk.Metadata;
 import kalix.javasdk.impl.workflowentity.WorkflowEntityEffectImpl;
+import net.jodah.typetools.TypeResolver;
 
-import java.util.*;
+import java.util.ArrayList;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Optional;
+import java.util.Set;
 import java.util.concurrent.CompletionStage;
+import java.util.function.Consumer;
 import java.util.function.Function;
 
 /** @param <S> The type of the state for this entity. */
@@ -263,6 +269,10 @@ public abstract class WorkflowEntity<S> {
       this.uniqueNames.add(step.name());
       return this;
     }
+
+    public void forEachStep(Consumer<Step> stepConsumer){
+      steps.forEach(stepConsumer);
+    }
   }
 
 
@@ -280,6 +290,8 @@ public abstract class WorkflowEntity<S> {
     final private String _name;
     final public Function<CallInput, DeferredCall<DefCallInput, DefCallOutput>> callFunc;
     final public Function<DefCallOutput, Effect.TransitionalEffect<Void>> transitionFunc;
+    final public Class<CallInput> callInputClass;
+    final public Class<DefCallOutput> transitionInputClass;
 
     public CallStep(String name,
                     Function<CallInput, DeferredCall<DefCallInput, DefCallOutput>> callFunc,
@@ -287,6 +299,10 @@ public abstract class WorkflowEntity<S> {
       _name = name;
       this.callFunc = callFunc;
       this.transitionFunc = transitionFunc;
+      Class<?>[] callClasses = TypeResolver.resolveRawArguments(Function.class, callFunc.getClass());
+      Class<?>[] transitionClasses = TypeResolver.resolveRawArguments(Function.class, transitionFunc.getClass());
+      this.callInputClass = (Class<CallInput>) callClasses[0];
+      this.transitionInputClass = (Class<DefCallOutput>) transitionClasses[0];
     }
 
     @Override
@@ -300,6 +316,8 @@ public abstract class WorkflowEntity<S> {
     final private String _name;
     final public Function<I, CompletionStage<O>> callFunc;
     final public Function<O, Effect.TransitionalEffect<Void>> transitionFunc;
+    final public Class<I> callInputClass;
+    final public Class<O> transitionInputClass;
 
     public AsyncCallStep(String name,
                          Function<I, CompletionStage<O>> callFunc,
@@ -307,6 +325,10 @@ public abstract class WorkflowEntity<S> {
       _name = name;
       this.callFunc = callFunc;
       this.transitionFunc = transitionFunc;
+      Class<?>[] callClasses = TypeResolver.resolveRawArguments(Function.class, callFunc.getClass());
+      Class<?>[] transitionClasses = TypeResolver.resolveRawArguments(Function.class, transitionFunc.getClass());
+      this.callInputClass = (Class<I>) callClasses[0];
+      this.transitionInputClass = (Class<O>) transitionClasses[0];
     }
 
     @Override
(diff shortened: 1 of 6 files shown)


sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/workflow/ReflectiveWorkflowEntityRouter.scala
```diff
@@ -58,28 +58,6 @@ class ReflectiveWorkflowEntityRouter[S, W <: WorkflowEntity[S]](
       .invoke(workflow, invocationContext)
       .asInstanceOf[WorkflowEntity.Effect[_]]
   }
-
-  private def _extractAndSetCurrentState(state: S): Unit = {
```
aludwiko, line 62: this is no longer needed, since we always decode state to correct object

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/KalixServer.scala
```diff
@@ -403,7 +405,29 @@ case class KalixServer(applicationContext: ApplicationContext, config: Config) {
...
+        val workflowEntity = kalixBeanFactory.getBean(clz)
+
+        val workflowStateType: Class[S] =
+          workflowEntity.getClass.getGenericSuperclass
+            .asInstanceOf[ParameterizedType]
+            .getActualTypeArguments
+            .head
+            .asInstanceOf[Class[S]]
+
+        messageCodec.lookupTypeHint(workflowStateType)
```
aludwiko, line 417: we need to register state type as well, for commands that was handled with by `_extractAndSetCurrentState` method, but this method is not invoked for a step invocation

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/SpringSdkMessageCodec.scala
```diff
@@ -54,7 +54,7 @@ private[springsdk] class SpringSdkMessageCodec extends MessageCodec {
   private def lookupTypeHint(value: Any): String =
     lookupTypeHint(value.getClass)
 
-  private def lookupTypeHint(clz: Class[_]): String = {
+  private[kalix] def lookupTypeHint(clz: Class[_]): String = {
```
franciscolopezsancho, line 57: Would it be more expressive to return `Option[String]`? Although I can imagine that changing the signature here must be a bit expensive, coding-wise.
BTW great catch finding the `null` case.

franciscolopezsancho:
I would add a couple of suggestions but since there has not been a change in the imports of the `ReflectiveWorkflowEntityRouter` I can't. 
`import java.lang.reflect.ParameterizedType` and `import kalix.javasdk.JsonSupport` I think can be removed.

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/KalixServer.scala
```diff
@@ -403,7 +405,29 @@ case class KalixServer(applicationContext: ApplicationContext, config: Config) {
...
         setKalixClient(clz, webClientProviderHolder)
 
-        kalixBeanFactory.getBean(clz)
+        val workflowEntity = kalixBeanFactory.getBean(clz)
+
+        val workflowStateType: Class[S] =
+          workflowEntity.getClass.getGenericSuperclass
+            .asInstanceOf[ParameterizedType]
+            .getActualTypeArguments
+            .head
```
franciscolopezsancho, line 414: Could it be the case that no `getActualTypeArguments` is empty?

aludwiko: no, since we managed that class

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/SpringSdkMessageCodec.scala
aludwiko: In this case, it will never be null, because we will either register a new entry or get one from the cache

sdk/java-sdk/src/main/java/kalix/javasdk/workflowentity/WorkflowEntity.java
```diff
@@ -280,13 +290,19 @@ public interface Step {
...
 
     public CallStep(String name,
                     Function<CallInput, DeferredCall<DefCallInput, DefCallOutput>> callFunc,
                     Function<DefCallOutput, Effect.TransitionalEffect<Void>> transitionFunc) {
       _name = name;
       this.callFunc = callFunc;
       this.transitionFunc = transitionFunc;
+      Class<?>[] callClasses = TypeResolver.resolveRawArguments(Function.class, callFunc.getClass());
+      Class<?>[] transitionClasses = TypeResolver.resolveRawArguments(Function.class, transitionFunc.getClass());
+      this.callInputClass = (Class<CallInput>) callClasses[0];
```
franciscolopezsancho, line 304: Again. I think this may be my ignorance of implementation details. How do you know `callClasses` are not empty and if there are multiple results you only need the first?

aludwiko: It is quite simple, you have a `Function` which is parametrized with 2 types, so you know it won't be empty, and you need the first one (they are ordered).

franciscolopezsancho: That makes sense 👍

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/KalixServer.scala
franciscolopezsancho: I see. You mean it has to be a `public static class Workflow<S> ` and therefore it has one parameter.

aludwiko: correct

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/SpringSdkMessageCodec.scala
franciscolopezsancho: Would then make sense to remove the `null` check?

aludwiko: the null check is about something different :)

franciscolopezsancho: :) it seems we are not on the same page hehe. Let me give it a last try I don't want to derail your effort. I'm thinking of the signature as `Option[String]` because of the `null` check. When `typeClass` is `null` the method raises an Exception. I thought it could be better to return a `None` instead of the exception. Hence the change of the signature.

octonato (approved):
Nice that we have found a solution for finding out the type param without increasing the API surface.

We can re-evaluate later if we want to depend on typetools or not. Great finding!

sdk/spring-sdk/src/main/scala/kalix/springsdk/impl/SpringSdkMessageCodec.scala
aludwiko: as discussed on the call, a new signature will now help much, since this this is not expected situation

PR #1455, opened 2023-02-07
fix: Instant test - use lower precision

octonato:
For now, we should truncate the Instant to lower the precision. Seems to be an issue in H2 and in Postgres.

### DIFF
--- a/sdk/spring-sdk/src/it/java/com/example/wiring/SpringSdkIntegrationTest.java
+++ b/sdk/spring-sdk/src/it/java/com/example/wiring/SpringSdkIntegrationTest.java
@@ -31,7 +31,6 @@ import org.hamcrest.core.IsEqual;
 import org.hamcrest.core.IsNull;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
-import org.junit.Ignore;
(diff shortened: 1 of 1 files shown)


johanandren (approved):

sdk/spring-sdk/src/it/java/com/example/wiring/SpringSdkIntegrationTest.java
```diff
@@ -533,10 +533,10 @@ public void shouldPropagateMetadataWithHttpDeferredCall() {
     Assertions.assertEquals(value, actionResponse);
   }
 
-  @Ignore
+  @Test
   public void searchWithInstant() {
 
-    var now = Instant.now();
+    var now = Instant.now().truncatedTo(SECONDS);
```
aludwiko, line 539: millis/macros would be just fine, but it doesn't matter

aludwiko (approved):

sdk/spring-sdk/src/it/java/com/example/wiring/SpringSdkIntegrationTest.java
octonato: I feel safer with seconds. 🤣

Issue #1345, opened 2022-09-21
Create a Kalix Spring Boot Starter


octonato:
SpringBoot has a framework to allow integrators to provide '[starters](https://docs.spring.io/spring-boot/docs/current/reference/html/features.html#features.developing-auto-configuration.custom-starter)'. 

> A typical Spring Boot starter contains code to auto-configure and customize the infrastructure of a given technology

For example, through starters one can decide if the http backend should be tomcat or netty. 

So var, we have hacked around a solution to not let Spring start it's own web server and then force the Kalix one (akka-http) to start. 

It will be better to revisit our code and instead build a `kalix-spring-boot-starter` that plug-ins into Spring Boot and replaces netty by akka-http + plus all the necessary code to start Kalix.

franciscolopezsancho:
By doing this 
- we may move @KalixConfiguration to be passed through params when starting the Spring server 
- we can start the application in one go when running in local by starting testcontainers directly

octonato:
One other advantage of having our own starter using testcontainers, is that we can easily support [auto-reload](https://docs.spring.io/spring-boot/docs/1.5.16.RELEASE/reference/html/using-boot-devtools.html#using-boot-devtools-restart).

Auto reload isn't working today for two reasons: 

1. We need to hook the akka http lifecycle to it. Currently, it will try to start the http server again and we get a "Address already in use" error.
2. Even if we solve the above, the proxy won't do a discovery call again.

octonato:
This is implemented in #1459. 

Auto-reload is more complicated and will be implement in follow-up PRs. I will open ticket about it.

PR #1459, opened 2023-02-08
feat!: spring boot starter

octonato:
References https://github.com/lightbend/kalix-jvm-sdk/issues/1345


project/Dependencies.scala
```diff
@@ -119,21 +120,28 @@ object Dependencies {
...
     jacksonDataFormatProto,
-    "org.springframework.boot" % "spring-boot" % SpringVersion,
-    "org.springframework.boot" % "spring-boot-starter-webflux" % SpringVersion)
+    "org.springframework" % "spring-web" % SpringFrameworkVersion,
+    "org.springframework" % "spring-webflux" % SpringFrameworkVersion,
+    "org.springframework" % "spring-webmvc" % SpringFrameworkVersion,
+    "org.springframework.boot" % "spring-boot" % SpringBootVersion,
+    "org.springframework.boot" % "spring-boot-starter" % SpringBootVersion,
+    "org.springframework.boot" % "spring-boot-starter-json" % SpringBootVersion,
+    "jakarta.websocket" % "jakarta.websocket-api" % "2.1.0")
```
efgpinto, line 129: just curious, where is this one needed?

build.sbt
```diff
@@ -200,6 +202,79 @@ lazy val sdkSpringTestKit = project
       "java.lang"))
   .settings(Dependencies.sdkSpringTestKit)
 
+lazy val sdkSpringBootStarter = project
+  .in(file("sdk/spring-boot-starter"))
+  .dependsOn(sdkSpring)
+  .enablePlugins(BuildInfoPlugin, PublishSonatype)
+  .settings(common)
+  .settings(
+    name := "kalix-spring-boot-starter",
```
efgpinto, line 211: I'm a bit skeptical if we should tie this to springboot rather than keeping it as spring... if there is any remote chance that we might need to include anything spring-related and not springboot specific in here, I would probably just go with spring. E.g. I'm not sure where the parsing of the spring mvc annotations will end up, but those are still spring concepts, not spring boot ones.

efgpinto (commented):
Looking good, left some questions to brainstorm

build.sbt
octonato: A starter is a Spring Boot concept and won't work with just Spring. 

The parsing of annotations is indeed generic and is not be in `kalix-spring-boot-starter`.

project/Dependencies.scala
octonato: Previously, we were including `spring-boot-starter-webflux` that was bringing in a few dependencies including this one. 

Without this jar, we get `ClassNotFoundException: jakarta.websocket.Endpoint`. 

But I still need to drill it. I'm not getting this error again on a sample that I'm testing. So not yet clear why (yet).

octonato: Something very odd. 

I added 2.1.0 as a dependency here, but in a project that I was using to test, I have it explicitly set to 2.0.0. 

It seems that 2.0.0 works, but 2.1.0 not. When a remove the deps to 2.0.0 and let it use the one it gets transitively, it starts to fail.

sdk/spring-sdk/src/main/scala/kalix/springboot/KalixReactiveWebServerFactory.scala
```diff
@@ -0,0 +1,62 @@
...
+
+      override def stop(): Unit = {
+        if (started) {
+          logger.info("Stopping Kalix ReactiveWebServer...")
+          try {
+            Await.ready(kalixApp.stop(), 10.seconds)
+          } finally {
+            started = false
+          }
+        } else ()
```
efgpinto, line 55: leftover `else ()`?

build.sbt
efgpinto: So, what your thinking is that we'll eventually get ride of the spring-sdk module, and we just have the `kalix-spring-boot-starter` that will depend directly on the `java-sdk`.. is this correct?

efgpinto (approved):
Would like to have a better understanding of what modules we'll have in the near future so might in your vision.. 

Anyway, this LGTM. Well done!

build.sbt
octonato: More or less. I still think that we will need to have something in between, but maybe something called `kalix-spring`. This module will contain the Spring annotations scanner and some other support classes. 

Starters are more like more similar to configuration package. It provides transitive dependencies and a `@AutoConfiguration` class to be loaded. 

For the moment, `KalixConfiguration` didn't move to the starter because we use it in our tests, but we can re-think it later as well.

efgpinto: Ok, cool. One thing I'm thinking about is package imports on the end-user kalix apps..

Do we foresee having something like this?
```java
import kalix.javasdk.Something;
import kalix.spring.AnotherOne;
import kalix.springboot.YetAnotherOne;
```

If this is a possibility, maybe having the below would be better:
```java
import kalix.javasdk.Something;
import kalix.spring.AnotherOne;
import kalix.spring.boot.YetAnotherOne;
```

Note `springboot` vs `spring.boot`.. This would also be consistent with the springboot packages themselves that do `springframework.boot`.

Anyway, this is nitpicking but I know you also care about this. :D

octonato: I have been thinking about that as well. `springboot` vs. `spring.boot`. 
I looked at other starters and there is also a mix. 

Maybe indeed better for us to have `kalix.spring` and `kalix.spring.boot`.

sdk/spring-sdk/src/main/scala/kalix/springboot/KalixReactiveWebServerFactory.scala
octonato: well, if not started, we just return. True that not needed in scala. 😄

build.sbt
octonato: renamed to `kalix.spring.boot`

sdk/spring-sdk/src/main/scala/kalix/spring/boot/KalixConfiguration.scala
```diff
@@ -32,25 +34,32 @@ object KalixConfiguration {
...
 }
 
-@Configuration
+@AutoConfiguration(
+  // ReactiveWebServerFactoryAutoConfiguration is responsible for auto-configure the ReactiveWebServer
+  // it has different pre-defined configs, ie: Tomcat, Netty, Jetty. And it will load one of them if its server is in
+  // the classpath. It turns out that we need reactor-netty dependency for the WebClient, but we don't want Spring Boot
+  // to configure it as the main ReactiveWebServer. Therefore, we must require our AutoConfigure to be loaded
+  // before NettyReactiveWebServer is loaded
+  before = Array(classOf[ReactiveWebServerFactoryAutoConfiguration]))
```
octonato, line 43: That was the biggest brain fkcu ever. 

The most surprising is that the tests was failing quite often in CI since I removed the `webflux-starter` (that brings in `reactor-netty`). 
Also surprising, when it failed, it was failing one the same tests. 

It turns out that the WebClient works (most of the times) without this dependency. And certainly when running locally.

We need reactor-netty in the production code because the WebClient is also used by the `KalixClient`, but if we leave `webflux-starter` in the classpath, Spring Boot will find two `ReactiveWebServers` configured and it will fail.

As soon as I brought `webflux-starter` back, all tests passed in CI.

octonato: Ok, forgot to comment on important trick. 

If we load our configuration before `ReactiveWebServerFactoryAutoConfiguration`, the Netty one won't be loaded because there is a condition to prevent a second one to be loaded. But we must load ours first. 🤦

build.sbt
```diff
@@ -200,6 +202,79 @@ lazy val sdkSpringTestKit = project
...
+    crossPaths := false,
+    Compile / javacOptions ++= Seq("--release", "17"),
+    Compile / scalacOptions ++= Seq("-release", "17"),
+    buildInfoKeys := Seq[BuildInfoKey](
+      name,
+      version,
+      "protocolMajorVersion" -> Kalix.ProtocolVersionMajor,
+      "protocolMinorVersion" -> Kalix.ProtocolVersionMinor,
+      "scalaVersion" -> scalaVersion.value),
+    buildInfoPackage := "kalix.springboot",
```
efgpinto, line 221: I'm not too well-versed in sbt.. so mostly curious.. doesn't this need to be updated to?

```suggestion
    buildInfoPackage := "kalix.spring.boot",
```

build.sbt
```diff
@@ -200,6 +202,79 @@ lazy val sdkSpringTestKit = project
...
+    crossPaths := false,
+    Compile / javacOptions ++= Seq("--release", "17"),
+    Compile / scalacOptions ++= Seq("-release", "17"),
+    buildInfoKeys := Seq[BuildInfoKey](
+      name,
+      version,
+      "proxyImage" -> "gcr.io/kalix-public/kalix-proxy",
+      "proxyVersion" -> Kalix.ProxyVersion,
+      "scalaVersion" -> scalaVersion.value),
+    buildInfoPackage := "kalix.springboot.test",
```
efgpinto, line 258: same?
```suggestion
    buildInfoPackage := "kalix.spring.boot.test",
```

efgpinto (approved):
LGTM again. 

Also, small suggestion to change this to `feat!: spring boot starter` to denote its a breaking change.

build.sbt
octonato: Oh, good catch! 

And thanks for `feat!` suggestion.

PR #1466, opened 2023-02-09
Merge with main

aludwiko:


sdk/spring-sdk/src/it/java/com/example/wiring/workflowentities/SpringWorkflowIntegrationTest.java
```diff
@@ -26,6 +26,7 @@
 import org.springframework.boot.test.context.SpringBootTest;
 import org.springframework.context.annotation.Import;
 import org.springframework.http.HttpStatus;
+import org.springframework.http.HttpStatusCode;
```
octonato, line 29: why that changed? 

probably `org.springframework.http.HttpStatus` is not in use (line above).

octonato: Is that because of Spring 3.0.2?

aludwiko: yes

octonato (approved):
LGTM

but we may have now an unused import.

PR #1468, opened 2023-02-09
feat!: typed events in eventsourced entity

efgpinto:
References #1323 

~Opening for visibility and early feedback.~

Task breakdown:
- [x] update ES entity signature and had a new type param
- [x] update testkits
  - [x] Ended-up not typing testkit ESResult, otherwise it's a breaking change for the grpc sdks. Note: the only drawback here is the getAllEvents returning a `List<Object>` instead of `List<E>`. Still, there is the `getNextEventOfType` which is typed anyway.
- [x] update spring ES samples to new format
- [x] update codegen for grpc sdks

### DIFF
--- a/sdk/java-sdk-testkit/src/main/java/kalix/javasdk/testkit/EventSourcedResult.java
+++ b/sdk/java-sdk-testkit/src/main/java/kalix/javasdk/testkit/EventSourcedResult.java
@@ -27,9 +27,8 @@ import java.util.List;
  * <p>Not for user extension, returned by the testkit.
  *
  * @param <R> The type of reply that is expected from invoking a command handler
- * @param <E> The parent type of events produced by the ES entity
  */
-public interface EventSourcedResult<R, E> {
+public interface EventSourcedResult<R> {
 
   /** @return true if the call had an effect with a reply, false if not */
   boolean isReply();
@@ -69,7 +68,7 @@ public interface EventSourcedResult<R, E> {
   boolean didEmitEvents();
 
   /** @return All the events that were emitted by handling this command. */
-  List<E> getAllEvents();
+  List<Object> getAllEvents();
 
   /**
    * Look at the next event and verify that it is of type E or fail if not or if there is no next
--- a/sdk/java-sdk-testkit/src/main/java/kalix/javasdk/testkit/EventSourcedResult.java
+++ b/sdk/java-sdk-testkit/src/main/java/kalix/javasdk/testkit/EventSourcedResult.java
@@ -27,8 +27,9 @@ import java.util.List;
  * <p>Not for user extension, returned by the testkit.
  *
  * @param <R> The type of reply that is expected from invoking a command handler
+ * @param <E> The parent type of events produced by the ES entity
  */
-public interface EventSourcedResult<R> {
+public interface EventSourcedResult<R, E> {
 
   /** @return true if the call had an effect with a reply, false if not */
   boolean isReply();
@@ -68,7 +69,7 @@ public interface EventSourcedResult<R> {
   boolean didEmitEvents();
 
   /** @return All the events that were emitted by handling this command. */
-  List<Object> getAllEvents();
+  List<E> getAllEvents();
 
   /**
    * Look at the next event and verify that it is of type E or fail if not or if there is no next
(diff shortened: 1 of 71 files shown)


sdk/java-sdk/src/main/scala/kalix/javasdk/impl/eventsourcedentity/EventSourcedEntityRouter.scala
```diff
@@ -17,14 +17,14 @@
...
 import java.util.Optional
-
 import kalix.javasdk.eventsourcedentity.CommandContext
 import kalix.javasdk.eventsourcedentity.EventContext
 import kalix.javasdk.eventsourcedentity.EventSourcedEntity
 import kalix.javasdk.impl.EntityExceptions
 import kalix.javasdk.impl.effect.SecondaryEffectImpl
 import kalix.javasdk.impl.eventsourcedentity.EventSourcedEntityEffectImpl.EmitEvents
 import kalix.javasdk.impl.eventsourcedentity.EventSourcedEntityEffectImpl.NoPrimaryEffect
+import org.checkerframework.checker.units.qual.K
```
aludwiko, line 27: leftover?

aludwiko (commented):
Nice! Is single handler for events planned in some incoming changes?

sdk/java-sdk/src/main/scala/kalix/javasdk/impl/eventsourcedentity/EventSourcedEntityRouter.scala
efgpinto: For sure, tks :)

efgpinto:
> Nice! Is single handler for events planned in some incoming changes?

I didn't remember that idea anymore.. 👍 

don't think I will include it in this PR, but definitely something to address as follow up. I thought there was an issue for that but couldn't find one. Will open one later.

sdk/java-sdk-testkit/src/main/java/kalix/javasdk/testkit/EventSourcedResult.java
```diff
@@ -77,7 +77,7 @@
    *
    * @return The next event if it is of type E, for additional assertions.
    */
-  <E> E getNextEventOfType(Class<E> expectedClass);
+  <T> T getNextEventOfType(Class<T> expectedClass);
```
octonato, line 80: I agree that we should not type the events in the result because it will be a breaking change at users face.

octonato (approved):
LGTM

As far as I can see, the only breaking change is the type returned by `effects()`. 

Although someone could have written `Effect.Builder<State> effects = effect();`, it's unlikely. 

Given the added value for code-first, I think the breaking change is justifiable. We can document it in the release notes. 

Alternatively, we can preserve the untyped builder and introduce a typed one and have a `effectsTyped` method.

sdk/java-sdk-testkit/src/main/java/kalix/javasdk/testkit/impl/EventSourcedEntityEffectsRunner.java
```diff
@@ -27,19 +27,19 @@
 import scala.jdk.javaapi.CollectionConverters;
 
 /** Extended by generated code, not meant for user extension */
-public abstract class EventSourcedEntityEffectsRunner<S> {
+public abstract class EventSourcedEntityEffectsRunner<S, E> {
```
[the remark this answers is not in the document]
efgpinto: indeed, fixed. thanks

sdk/java-sdk-testkit/src/main/java/kalix/javasdk/testkit/EventSourcedResult.java
efgpinto: Btw, just noticed this change was a leftover from when I tried to type this class as well. I reverted this as well now (so it's unchanged and consistent with the comment).

efgpinto:
> As far as I can see, the only breaking change is the type returned by `effects()`.

That's a good point, I actually missed this one.

As agreed, let's hear from @patriknw before merging.

patriknw:
I agree that we can make this change. I'm not concerned at all about breaking the Spring api, since that is the purpose of the beta phase. Clever to almost be fully compatible for code first via the the base class.

Perhaps we should make the next release that includes this a 1.2.0 so that we live up to the expectation of no breaking changes in patches at least. If we would be strict about semantic versioning it should be 2.0.0, but that probably signals too much for this small breakage.

octonato:
Completely agree that next one should be 1.2.0.

octonato:
Let's merge this. I think we have consensus about this change.

PR #1473, opened 2023-02-09
chore: refactor to hide status code impl details

efgpinto:
Some improvements after #1436 

Some leftovers from code-review: moving things to `impl` package and actually remove dependency on webflux on java sdk.


sdk/java-sdk/src/main/java/kalix/javasdk/DeferredCallResponseException.java
```diff
@@ -0,0 +1,41 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package kalix.javasdk;
+
+/**
+ * Exception used when a DeferredCall fails to wrap the origin error, plus the error code associated.
+ */
+public interface DeferredCallResponseException {
```
octonato, line 22: Not sure we need an interface. We could have a case class exception directly here.

octonato (approved):
Looks good. 

I was already working on this and then I saw your PR. 

I have one small comment. I don't think it's worth to have an interface for `DeferredCallResponseException`. Otherwise, good to merge.

sdk/java-sdk/src/main/java/kalix/javasdk/DeferredCallResponseException.java
efgpinto: update now

Issue #1456, opened 2023-02-08
Remove pub/sub images when not in use


efgpinto:
I think we talked about this, it would be good to remove or at least comment out all the google pubsub images on the docker-compose of our samples for any of those in which it is not required (which is the majority I believe).

octonato:
Should we also comment it out in the templates? I would prefer to do so. My feeling is that this is not used in most cases or its usage is delayed until later in the development cycle. 

Ideally, if someone start to use topics in dev-mode, we should present them with a clear message and a link to some documentation explaining how to enable it for local development. 

Another option is to provide a few docker-compose variants. 

One with only the proxy using the default name (ie: `docker-compose.yml`) 
And then one `docker-compose-pubsub.yml` and one `docker-compose-kafka.yml`. The user can then choose what to do. Eventually dropping all the others they don't use.

efgpinto:
> Should we also comment it out in the templates? I would prefer to do so. My feeling is that this is not used in most cases or its usage is delayed until later in the development cycle.

I think so. I would probably have it commented out on the templates as well.

octonato:
what about having three docker-compose files and user pick the one they prefer?

efgpinto:
I don't have a strong opinion tbh, I just thought that with the different docker files you would maybe need to explain somewhere that to run a specific non-standard file one needs to do `docker-compose -f file.yml up` (or delete and rename of course).. thus, having it commented with a note on the readme seemed like a simpler approach. But again, no strong feelings about it.

octonato:
Correct, that would require more explanation. I would add it to the README to start with. 

We can add comments to the docker file, but then there is two places that need to be commented.  

I'm not sure what will be the best. We experiment and see which option feels easier.

johanandren:
I'd vote for one file, comment out the pubsub container and parameter, and keep it as default compose file.

ennru:
Fixed with #1476

PR #1476, opened 2023-02-13
chore: remove unused pub/sub from docker compose

efgpinto:
Refs #1456 

~Opening as draft: Just looking to confirm the setup we want before doing the changes on all samples. Would this be ok?~

### DIFF
--- a/samples/java-customer-registry-quickstart/docker-compose.yml
+++ b/samples/java-customer-registry-quickstart/docker-compose.yml
@@ -1,8 +1,11 @@
+# If you're looking to use eventing with Google PubSub, to get an emulator running:
+# - uncomment property "-Dkalix.proxy.eventing.support=google-pubsub-emulator"
+# - uncomment the env var PUBSUB_EMULATOR_HOST and the section below for gcloud-pubsub-emulator service
 version: "3"
 services:
   kalix-proxy:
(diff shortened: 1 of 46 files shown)


samples/java-customer-registry-quickstart/docker-compose.yml
```diff
@@ -1,18 +1,21 @@
+# If you're looking to use eventing with Google PubSub, to get an emulator running:
+# - uncomment property "-Dkalix.proxy.eventing.support=google-pubsub-emulator"
+# - uncomment the env var PUBSUB_EMULATOR_HOST and the section below for gcloud-pubsub-emulator service
```
octonato, line 3: This is looking good. I think it's all that we need.

johanandren (approved):
LGTM (did not in any way cross-check which samples may use eventing)

efgpinto:
> LGTM (did not in any way cross-check which samples may use eventing)

Ahm, actually, I think I need to double check that.. I might have overlooked eventing.in.

octonato:
> > LGTM (did not in any way cross-check which samples may use eventing)
> 
> Ahm, actually, I think I need to double check that.. I might have overlooked eventing.in.

Anyone with a proto using the topic annotations. Should be easy to find it out.

PR #1480, opened 2023-02-13
feat!: java code-first classes moved to javasdk package

octonato:
~On top of #1468 because of potential conflicts.~ 

Most classes in `springsdk` package moved to `javasdk`. 

Classes that are Spring specific and only make sense when using a Spring Boot project are now moved to `kalix.spring` or `kalix.spring.boot`. 

Some of the classes that moved to `javasdk` still have Spring dependencies that we will need to abstract away. 

All classes remain in `springsdk` sbt module. They are intended for code-first java, but at present we only offset code-first java together with Spring. Later, we can move them to `javasdk` sbt module. But this can be done in different steps.


octonato:
Maybe some more clarifications of my plan here. 

First I want to move to `javasdk` everything that is strictly speaking code-first java. 
Some of those classes still have dependencies on Spring or Reactor, but this can be later abstracted away if we decide so. 

Nothing is moving to `javasdk` sbt module for now. They all stay in `sprindsdk` module. 
As of today, code-first java only works with Spring, so no need to move them right now. Later, if we move those classes or some of them to the `javasdk` module, users won't be impacted.

The `springsdk` and `springsdk-testkit` modules should be renamed and in principle, not accessible directly to the users. Users will get them from the `starters`.

As we progress and abstract away all spring and reactor dependencies, we can move the classes to `javasdk` sbt module, but that's not something we can postpone for now.

efgpinto (approved):
Boring one.. but looks good to me!

Strategy wise, I think it should be fine as well. I will leave it up to you if you want to collect anyone else's feedback before merging.

sdk/spring-sdk/src/main/scala/kalix/javasdk/impl/ComponentDescriptor.scala
```diff
@@ -537,7 +538,7 @@ private[impl] object ComponentDescriptor {
 
 }
 
-private[springsdk] final case class ComponentDescriptor private (
+private[kalix] final case class ComponentDescriptor private (
```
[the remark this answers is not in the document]
octonato: Yep. That's correct. 

Most of the annotation parsing is java only, except for the REST annotations. The reflective providers depend on it.

sdk/spring-sdk/src/test/scala/kalix/javasdk/impl/reflection/ReflectionUtilsSpec.scala
```diff
@@ -33,7 +33,7 @@ class ReflectionUtilsSpec extends AnyWordSpec with Matchers {
 
   "The reflection utils" must {
     "deterministically sort methods of the same class" in {
-      import ReflectionUtils.methodOrdering
+      import kalix.javasdk.impl.reflection.ReflectionUtils.methodOrdering
```
octonato, line 36:
```suggestion
      import ReflectionUtils.methodOrdering
```

sdk/spring-sdk-testkit/src/test/java/spring/testmodels/actions/EchoAction.java
```diff
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-package springsdk.testmodels.actions;
+package spring.testmodels.actions;
```
franciscolopezsancho, line 17: I don't have a strong opinion about this but 'spring' here seems to me like it may not be precise. Meaning, it's not actually a `spring` component but a `kalix` one used in the 'springsdk` testkit. How about leaving them as they were or maybe `kalix.spring`?. or none of it and just merging it with Eduardo's blessings (and mine :)

octonato: The notion of `springsdk` is being removed in favor of `javasdk` with a thin layer of Spring Boot integration. 

The choice to use `kalix.spring` and `kalix.spring.boot` is more related with a convention that we have seen in other projects. 

For that particular case, I think we can pick `kalix.spring.boot`. I will change it.

octonato: @franciscolopezsancho, I see now that those tests are from the kalix-spring-testkit. The package should just be `kalix.javasdk.testmodels`.

Although the module still have `spring` in its name, the classes themselves are not Spring related.

PR #1483, opened 2023-02-13
chore: rename sbt modules for code-first java

octonato:
On top of #1480. 

Sending as apart PR so we can discuss modules names apart.


build.sbt
```diff
@@ -122,25 +122,24 @@ lazy val sdkJavaTestKit = project
...
-  .in(file("sdk/spring-sdk"))
+lazy val sdkJavaSpring = project
+  .in(file("sdk/java-sdk-spring"))
   .dependsOn(sdkJava)
   .dependsOn(sdkJavaTestKit % IntegrationTest)
   .enablePlugins(AkkaGrpcPlugin, BuildInfoPlugin, PublishSonatype, IntegrationTests)
   .settings(common)
   .settings(
-    name := "kalix-spring-sdk",
+    name := "kalix-java-sdk-spring",
```
octonato, line 132: This is not the Spring SDK, but the java-sdk with a thin layer of Spring sauce.

aludwiko: I understand that `spring` is not an sdk, but for the sake of consistency maybe we could go with:
```
kalix-java-sdk
kalix-java-spring-sdk
kalix-scala-sdk
kalix-java-protobuf-sdk
```?

File: build.sbt [hunk 2]
```diff
@@ -122,25 +122,24 @@ lazy val sdkJavaTestKit = project
       "java.lang"))
   .settings(Dependencies.sdkJavaTestKit)
 
-lazy val sdkSpring = project
-  .in(file("sdk/spring-sdk"))
+lazy val sdkJavaSpring = project
+  .in(file("sdk/java-sdk-spring"))
```
aludwiko, line 126: can we adjust the val name with the folder name?

octonato: Yep

build.sbt
octonato: I don't think we should call it `java-spring-sdk` because that's the `java-sdk` with Spring integration. Hence, `java-sdk-spring`. 

So I would prefer to have: 

* `kalix-java-sdk-spring`
* `kalix-java-sdk-protobuf`
* `kalix-scala-sdk-protobuf`

`kalix-java-sdk-protobuf` should read: the java sdk using protobuf for modeling.

efgpinto: This last proposal looks good to me.. this will also define the packages names users will depend on I guess.. right?

So `<artifactId>kalix-java-sdk</artifactId>` will become `<artifactId>kalix-java-sdk-protobuf</artifactId>`.. for code-first they will depend on the starter `<artifactId>kalix-spring-boot-starter</artifactId>`.

build.sbt
```diff
@@ -122,25 +122,24 @@ lazy val sdkJavaTestKit = project
...
   .dependsOn(sdkJavaTestKit % IntegrationTest)
   .enablePlugins(AkkaGrpcPlugin, BuildInfoPlugin, PublishSonatype, IntegrationTests)
   .settings(common)
   .settings(
-    name := "kalix-spring-sdk",
+    name := "kalix-java-sdk-spring",
     crossPaths := false,
     Compile / javacOptions ++= Seq("--release", "17"),
     Compile / scalacOptions ++= Seq("-release", "17"),
-    buildInfoObject := "SpringSdkBuildInfo",
```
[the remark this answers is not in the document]
efgpinto: indeed 👍

build.sbt
aludwiko: Yeap, the last proposal is nice.

franciscolopezsancho: I join the agreement. I think the 'shoud read' it's intuitive.

octonato: Hmm, now I realise that `kalix-java-sdk-protobuf` will be a breaking change. 
But something that we can solve though. 

We change it to `kalix-java-sdk-protobuf` and we create a new one called `kalix-java-sdk`. 

`kalix-java-sdk` brings in `kalix-java-sdk-protobuf` as a transitive dependency and an extra class that we load with the sole purpose to print a warning telling users that they should move to `kalix-java-sdk-protobuf`. Kind of deprecation cycle for jar packages. 

Btw, I don't think we should do it in this PR. Let's move on with the Spring stuff. We can then discuss what to do with `kalix-java-sdk` and `kalix-scala-sdk`.

@efgpinto, you are right. Spring users won't see any of this directly, only the starter. The starters work as a kind of dependency management (bom file) + bootstrap configuration logic.

build.sbt
octonato: @aludwiko, I changed the sbt modules names to follow the package name. I think it looks better now, but I realise that the main reason for it was to have them all grouped when running `sbt projects`. 

We can revert it if we find it inconvenient. 

btw, I also change the tckJava/tckScala to javaTck/scalaTck

aludwiko: grouped in the sense to start typing `sdk...` and wait for auto-completion?

octonato: yep, but I think this is more of habit. We can learn the other way around.

efgpinto (approved):
LGTM

PR #1484, opened 2023-02-13
fix: bind to 0.0.0.0 when running springboot:run

### LINKED ISSUE #1393: Spring SDK: mvn spring-boot:run binding
We are binding the http server to `0.0.0.0` if the app is started with `mvn exec:exec`. However, When using `mvn spring-boot:run` (which is what we recommend on the READMEs) it binds to `127.0.0.1` and this does not work on all platforms (e.g. Linux) making the UF unreachable to a containerized proxy.

octonato:
Fixes #1393

### DIFF
--- a/maven-java/kalix-spring-boot-archetype/src/main/resources/archetype-resources/pom.xml
+++ b/maven-java/kalix-spring-boot-archetype/src/main/resources/archetype-resources/pom.xml
@@ -227,6 +227,7 @@
         <configuration>
             <systemPropertyVariables>
                 <logging.config>src/main/resources/logback-dev-mode.xml</logging.config>
+                <kalix.user-function-interface>0.0.0.0</kalix.user-function-interface>
             </systemPropertyVariables>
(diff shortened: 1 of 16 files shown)


maven-java/kalix-spring-boot-archetype/src/main/resources/archetype-resources/pom.xml
```diff
@@ -227,6 +227,7 @@
         <configuration>
             <systemPropertyVariables>
                 <logging.config>src/main/resources/logback-dev-mode.xml</logging.config>
+                <kalix.user-function-interface>0.0.0.0</kalix.user-function-interface>
```
octonato, line 230: The `mvn exec:exec` is explicitly bind it to 0.0.0.0. For Spring, we should just do the same and this is how we pass `-D` properties to Spring Boot.

efgpinto (approved):
LGTM

PR #1486, opened 2023-02-14
fix: use the same spring context environment for KalixComponentProvider

aludwiko:
with this change we can disable components with existing Spring annotation:
```@ConditionalOnProperty```


octonato (approved):
👍

Issue #1489, opened 2023-02-15
add support for spring-boot-devtools


octonato:
[Spring Boot Devtools](https://docs.spring.io/spring-boot/docs/1.5.16.RELEASE/reference/html/using-boot-devtools.html) allows for auto-reload on when project is recompiled. 

It works similar to Play auto-reload. All projects classes are added to a `RestartClassLoader` that are thrown away (after some running some shutdown procedure) and reloaded. Classes from jar packages are not reloaded. 

I have done some experiments with it and notice a few issues:

First, the user components classes are loaded from `KalixSpringApplication` which is in the sdk jar. Therefore, the components classes are loaded in the normal classloard, not the `RestartClassLoader`. We have a fix for this and I will PR. 

The next problem is when the Spring shutdown hook is called, it stops the ReactiveWebServer, which stops the actor system, which in turn kills the JVM. 

To make it work, we should set `kalix.system.akka.coordinated-shutdown.exit-jvm` to `off` when running in dev mode. 

Setting `exit-jvm=off` works, but then we hit the next problem.

If the proxy is running and have already discovery the user application, the restart may be so fast that the proxy don't even notice it. It won't re-discovery. 

Once we send the next request, the user application will fail with: 

```java
java.lang.IllegalStateException: Service proxy hostname and/or port are not set by proxy at discovery, too old proxy version?
        at kalix.spring.impl.WebClientProviderImpl.$anonfun$localWebClient$3(WebClientProviderHolder.scala:87)
        at scala.Option.getOrElse(Option.scala:201)
        at kalix.spring.impl.WebClientProviderImpl.<init>(WebClientProviderHolder.scala:86)
        at kalix.spring.impl.WebClientProviderHolder.<init>(WebClientProviderHolder.scala:53)
        at kalix.spring.impl.WebClientProviderHolder$.createExtension(WebClientProviderHolder.scala:48)
        at kalix.spring.impl.WebClientProviderHolder$.createExtension(WebClientProviderHolder.scala:42)
        at akka.actor.ActorSystemImpl.registerExtension(ActorSystem.scala:1165)
        at akka.actor.ExtensionId.apply(Extension.scala:78)

```

This happens because the proxy info is lost. 

Anyway, I think that all of this can be solved. We just need to explore it some more and find out the best solution. I'm leaving it on the side for now and reporting my finding here to I can reload the context in the future.

PR #1490, opened 2023-02-15
chore: load components using Main's ClassLoader

octonato:
This is in preparation for #1489.  

Not really needed for now but it is in any case a good measure to keep all project classes on the same loader. We never know which other tooling can be used by the users.

### DIFF
--- a/sdk/spring-sdk/src/main/scala/kalix/spring/impl/KalixSpringApplication.scala
+++ b/sdk/spring-sdk/src/main/scala/kalix/spring/impl/KalixSpringApplication.scala
@@ -121,8 +121,8 @@ object KalixSpringApplication {
    * Classpath scanning provider that will lookup for the original main class. Spring doesn't make the original main
    * class available in the application context, but a cglib enhanced variant.
    *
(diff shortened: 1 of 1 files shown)


franciscolopezsancho (approved):
Looking at https://github.com/lightbend/kalix-jvm-sdk/issues/1489 it makes sense to me.

franciscolopezsancho:
One question though.
> We never know which other tooling can be used by the users.
Could you elaborate?

octonato:
> One question though.
> 
> > We never know which other tooling can be used by the users.
> 
> Could you elaborate?

Well, I don't know. I never now. 😄 

My thinking is that even if we are not yet supporting Spring dev tools, a user can add it themselves. Or they can try any other tooling that does a similar trick. The trick being: sneak somewhere in the bootstrap and create different classloader for code in the project. 

This is very unlikely though. Most probably people will try Spring dev-tools and hit some other issues because we are not yet ready for that.

franciscolopezsancho:
I think I get the picture :) 👍

PR #1503, opened 2023-02-24
chore: renaming samples

efgpinto:
References https://github.com/lightbend/kalix-proxy/issues/1940


efgpinto:
I would have liked to split the samples in different folders (i.e. `samples/java-protobuf`. `samples/scala-protobuf`, `samples/java-spring`) but this was a big PR already.

aludwiko:
I like the idea of separate subfolders for samples.

efgpinto:
CI is failing because of 404 links, e.g.:
```
ERROR: 1 dead links found!
[✖] https://github.com/lightbend/kalix-jvm-sdk/tree/main/samples/java-spring-transfer-workflow → Status: 404
```

But these will be there once we merge the PR. So I would probably go with overriding the branch protections once approved rather than having to split this PR in 2.

aludwiko (approved):
Looks good, I wonder what else will break :)

efgpinto:
> Looks good, I wonder what else will break :)

Thanks. Hopefully only a couple of links in docs. 🤞  Which makes me think that those should probably link directly to the tag specific folder and not `main`.

PR #1518, opened 2023-03-06
chore: renaming java and scala modules

efgpinto:
Closes https://github.com/lightbend/kalix-proxy/issues/1938

There are 3 different commits so it's easier to review and if anyone thinks we should not rename the dirs for some reason it's easy to revert anyway.

I just realised that this change is really only a breaking change for java/protobuf sdk, not for scala/protobuf because of the plugin.


octonato (approved):
LGTM

Issue #1517, opened 2023-03-06
code first (Spring): missing event handlers validation for sealed interfaces 


aludwiko:
I've found it really frustrating that the error about the missing event handler for actions/views appears relatively late on runtime. For instance, I need to call 2/3 components to emit a new event, but I forgot about the handler, so I need to repeat the whole process all over again. For sealed interfaces, I could do the runtime check on the application start and fail much sooner.

PR #1520, opened 2023-03-07
feat: missing handlers validation for ES Entities and Subscription wh…

aludwiko:
References #1517 

Currently, tests are failing from sbt because I can't use Mixed compilation mode with Java sealed interface, Scala update is required: https://github.com/scala/scala/pull/10105 The solution works or my machine, I could even run tests with IntelliJ, so I can comment out new tests, merge, and wait for Scala release. Or I could wait with the merge for Scala release.

### DIFF
--- a/sdk/java-sdk-spring/src/main/scala/kalix/javasdk/impl/Validations.scala
+++ b/sdk/java-sdk-spring/src/main/scala/kalix/javasdk/impl/Validations.scala
@@ -189,7 +189,7 @@ object Validations {
               .map(_.getParameterTypes.last) //last because it could be a view update methods with 2 params
             val errors = eventType.getPermittedSubclasses
               .filterNot(effectMethodsInputParams.contains)
-              .map(clazz => s"Missing event handler for ${clazz.getName}")
+              .map(clazz => s"Missing event handler for '${clazz.getName}'")
             Validation(errors)
           } else {
             Valid
--- a/sdk/java-sdk-spring/src/main/scala/kalix/javasdk/impl/Validations.scala
+++ b/sdk/java-sdk-spring/src/main/scala/kalix/javasdk/impl/Validations.scala
@@ -209,7 +209,8 @@ object Validations {
           val effectOutputInputParams: Seq[Class[_]] = methods.map(_.getParameterTypes.last)
           eventType.getPermittedSubclasses
             .filterNot(effectOutputInputParams.contains)
-            .map(clazz => s"Missing ${entityClass.getSimpleName} event handler for ${clazz.getName}")
+            .map(clazz =>
+              s"Component ${component.getSimpleName} missing event handler for ${clazz.getName} from ${entityClass.getSimpleName}")
             .toList
         } else {
           List.empty
--- a/sdk/java-sdk-spring/src/main/scala/kalix/javasdk/impl/Validations.scala
+++ b/sdk/java-sdk-spring/src/main/scala/kalix/javasdk/impl/Validations.scala
@@ -215,7 +215,7 @@ object Validations {
   private def missingErrors(effectOutputInputParams: Seq[Class[_]], eventType: Class[_], component: Class[_]) = {
     eventType.getPermittedSubclasses
       .filterNot(effectOutputInputParams.contains)
-      .map(clazz => s"Component '${component.getSimpleName}' is missing event handler for '${clazz.getName}'")
+      .map(clazz => s"Component '${component.getSimpleName}' is missing an event handler for '${clazz.getName}'")
       .toList
   }
 
--- a/sdk/java-sdk-spring/src/main/scala/kalix/javasdk/impl/Validations.scala
+++ b/sdk/java-sdk-spring/src/main/scala/kalix/javasdk/impl/Validations.scala
@@ -187,10 +187,7 @@ object Validations {
             val effectMethodsInputParams: Seq[Class[_]] = methods
               .filter(updateMethodPredicate)
               .map(_.getParameterTypes.last) //last because it could be a view update methods with 2 params
-            val errors = eventType.getPermittedSubclasses
-              .filterNot(effectMethodsInputParams.contains)
-              .map(clazz => s"Missing event handler for '${clazz.getName}'")
-            Validation(errors)
+            Validation(missingErrors(effectMethodsInputParams, eventType, component))
           } else {
             Valid
           }
@@ -206,18 +203,20 @@ object Validations {
       val errors = effectOutputMethodsGrouped.flatMap { case (entityClass, methods) =>
         val eventType = getEventType(entityClass)
         if (eventType.isSealed) {
-          val effectOutputInputParams: Seq[Class[_]] = methods.map(_.getParameterTypes.last)
-          eventType.getPermittedSubclasses
-            .filterNot(effectOutputInputParams.contains)
-            .map(clazz =>
-              s"Component ${component.getSimpleName} missing event handler for ${clazz.getName} from ${entityClass.getSimpleName}")
-            .toList
+          missingErrors(methods.map(_.getParameterTypes.last), eventType, component)
         } else {
           List.empty
         }
       }
       Validation(errors.toSeq)
     }
+  }
+
+  private def missingErrors(effectOutputInputParams: Seq[Class[_]], eventType: Class[_], component: Class[_]) = {
+    eventType.getPermittedSubclasses
+      .filterNot(effectOutputInputParams.contains)
+      .map(clazz => s"Component '${component.getSimpleName}' is missing event handler for '${clazz.getName}'")
+      .toList
   }
 
   private def topicSubscriptionValidations(component: Class[_]): Validation = {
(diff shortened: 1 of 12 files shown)


sdk/java-sdk-spring/src/main/scala/kalix/javasdk/impl/Validations.scala
```diff
@@ -85,6 +95,13 @@ object Validations {
     override def failIfInvalid: Validation =
       throw InvalidComponentException(messages.mkString(", "))
 
+    override def combineWhenValid(validation: Validation): Validation = {
+      if (isValid) {
+        validation
+      } else {
+        this
+      }
+    }
```
octonato, line 104: Since `Invalid`, we don't need to check `isValid`

```suggestion
    override def combineWhenValid(validation: Validation): Validation = this
```

octonato: But I don't see why we are not using `++` instead. 

The goals of the validation framework is to accumulate errors. The combine is not accumulating, but ignoring the second validation. 

We should maybe not call it `combine` because that's exactly what we are not doing.

aludwiko: My motivation was as follows, if I join these two validations then the root cause of the second validation error is most likely the first error, thus presenting 2 errors might overwhelm the user. In this case, Subscribe annotation on the type and on the method level is the first error, but I could also add an error that there is a missing event handler. Fixing the missing event handler might not fix the problem at all, because the correct solution would be to remove the annotation. Maybe I'm overreacting and if the solution is to remove the annotation, then the missing handler might still be the case. Yeah, I can revert it.

aludwiko:
missing handlers in samples prove the value of this change :)

sdk/java-sdk-spring/src/main/scala/kalix/javasdk/impl/Validations.scala
octonato: I see. I think we should not care about this kind of situation and simply report all the errors we catch. 

Anyway, it seems you reverted it already, so 👍.

sdk/java-sdk-spring/src/test/scala/kalix/javasdk/impl/ActionDescriptorFactorySpec.scala
```diff
@@ -417,6 +420,22 @@ class ActionDescriptorFactorySpec extends AnyWordSpec with ComponentDescriptorSu
...
+      }.getMessage should include(
+        "Missing event handler for kalix.spring.testmodels.eventsourcedentity.EmployeeEvent$EmployeeEmailUpdated")
+    }
+
+    //TODO remove ignore after updating to Scala 2.13.11 (https://github.com/scala/scala/pull/10105)
+    "validates if there are missing event handlers for event sourced Entity Subscription at method level" ignore {
+      intercept[InvalidComponentException] {
+        Validations.validate(classOf[MissingHandlersWhenSubscribeToEventSourcedOnMethodLevelEntityAction]).failIfInvalid
+      }.getMessage should include(
+        "Missing EmployeeEntity event handler for kalix.spring.testmodels.eventsourcedentity.EmployeeEvent$EmployeeEmailUpdated")
```
octonato, line 436: Because we are not running the tests, we didn't catch that the test is wrong.

The message will be: 'Component EmployeeEntity missing event handler for...' instead.

But I think if should even be:

"Component 'EmployeeEntity' is missing event handler for 'Event'"

Note the single quotes and the `is` before "missing".

sdk/java-sdk-spring/src/main/scala/kalix/javasdk/impl/Validations.scala
```diff
@@ -161,6 +167,59 @@ object Validations {
...
+      eventSourcedEntitySubscription(component)
+        .map { classLevel =>
+          val eventType = getEventType(classLevel.value())
+          if (!classLevel.ignoreUnknown() && eventType.isSealed) {
+            val effectMethodsInputParams: Seq[Class[_]] = methods
+              .filter(updateMethodPredicate)
+              .map(_.getParameterTypes.last) //last because it could be a view update methods with 2 params
+            val errors = eventType.getPermittedSubclasses
+              .filterNot(effectMethodsInputParams.contains)
+              .map(clazz => s"Missing event handler for ${clazz.getName}")
```
octonato, line 192:
```suggestion
              .map(clazz => s"Missing event handler for '${clazz.getName}'")
```

Not sure if we have wrapped things in single quotes in other parts. Worth to check and align eventually.

sdk/java-sdk-spring/src/main/scala/kalix/javasdk/impl/Validations.scala
```diff
@@ -161,6 +167,59 @@ object Validations {
...
+        .groupBy(findEventSourcedEntityClass)
+
+      val errors = effectOutputMethodsGrouped.flatMap { case (entityClass, methods) =>
+        val eventType = getEventType(entityClass)
+        if (eventType.isSealed) {
+          val effectOutputInputParams: Seq[Class[_]] = methods.map(_.getParameterTypes.last)
+          eventType.getPermittedSubclasses
+            .filterNot(effectOutputInputParams.contains)
+            .map(clazz =>
+              s"Component ${component.getSimpleName} missing event handler for ${clazz.getName} from ${entityClass.getSimpleName}")
```
octonato, line 213:
```suggestion
              s"Component '${component.getSimpleName}' is missing event handler for '${clazz.getName}' from '${entityClass.getSimpleName}'")
```

octonato: Maybe the last part: "from `entityClass.getSimpleName`" can be removed?

sdk/java-sdk-spring/src/main/scala/kalix/javasdk/impl/Validations.scala
aludwiko: we have, in some proxy errors but in other cases we don't, anyway I like it with single quotes

sdk/java-sdk-spring/src/main/scala/kalix/javasdk/impl/Validations.scala
aludwiko: ok

sdk/java-sdk-spring/src/test/scala/kalix/javasdk/impl/ActionDescriptorFactorySpec.scala
aludwiko: Yeah, it sucks, but haven't figured out any workaround.

sdk/java-sdk-spring/src/test/scala/kalix/javasdk/impl/ViewDescriptorFactorySpec.scala
```diff
@@ -530,6 +531,22 @@ class ViewDescriptorFactorySpec extends AnyWordSpec with ComponentDescriptorSuit
 
       }
     }
+
+    //TODO remove ignore after updating to Scala 2.13.11 (https://github.com/scala/scala/pull/10105)
+    "validate missing handlers for method level subscription" ignore {
+      intercept[InvalidComponentException] {
+        Validations.validate(classOf[SubscribeToEventSourcedEventsWithMethodWithState]).failIfInvalid
+      }.getMessage should include(
+        "Missing EmployeeEntity event handler for kalix.spring.testmodels.eventsourcedentity.EmployeeEvent$EmployeeEmailUpdated")
```
octonato, line 540: wrong assertion here as well

sdk/java-sdk-spring/src/test/scala/kalix/javasdk/impl/ActionDescriptorFactorySpec.scala
```diff
@@ -417,6 +420,22 @@ class ActionDescriptorFactorySpec extends AnyWordSpec with ComponentDescriptorSu
       }.getMessage should include("You cannot use @Subscribe.EventSourcedEntity annotation in both methods and class.")
     }
 
+    //TODO remove ignore after updating to Scala 2.13.11 (https://github.com/scala/scala/pull/10105)
+    "validates if there are missing event handlers for event sourced Entity Subscription at type level" ignore {
+      intercept[InvalidComponentException] {
+        Validations.validate(classOf[MissingHandlersWhenSubscribeToEventSourcedEntityAction]).failIfInvalid
+      }.getMessage should include(
+        "Component `MissingHandlersWhenSubscribeToEventSourcedEntityAction` is missing event handler for `kalix.spring.testmodels.eventsourcedentity.EmployeeEvent$EmployeeEmailUpdated`")
```
octonato, line 428: Backticks are being used while the message will have single-quotes

aludwiko: fixed

sdk/java-sdk-spring/src/test/scala/kalix/javasdk/impl/ViewDescriptorFactorySpec.scala
aludwiko: fixed

octonato (approved):

PR #1521, opened 2023-03-07
fix: use correct env var syntax for docker-compose

octonato:


docs/src/modules/java-protobuf/partials/query-syntax-advanced.adoc
```diff
@@ -314,7 +314,7 @@ include::java:example$java-spring-view-store/src/it/java/store/view/TestKitConfi
 ----
 // tag::grpc[]
 
-For running a local proxy in Docker, with advanced view features enabled, use the environment variable `VIEW_FEATURES_ALL=on`.
+For running a local proxy in Docker, with advanced view features enabled, set the environment variable `VIEW_FEATURES_ALL: "on"`.
```
octonato, line 317: I'm testing if `VIEW_FEATURES_ALL: "true"` also works because that's who we document the ACL env var as well.

docs/src/modules/java-protobuf/partials/query-syntax-advanced.adoc
```diff
@@ -314,7 +314,7 @@ include::java:example$java-spring-view-store/src/it/java/store/view/TestKitConfi
 ----
 // tag::grpc[]
 
-For running a local proxy in Docker, with advanced view features enabled, use the environment variable `VIEW_FEATURES_ALL=on`.
+For running a local proxy in Docker, with advanced view features enabled, set the environment variable `VIEW_FEATURES_ALL: "true"`.
```
octonato, line 317: Should we also added it to the template, default to `false` with a message telling users to contact us about enabling it in prod? 

It will make is easier for users if they can immediately see what and where to enable it in the docker-compose file.

efgpinto: Yes, I think we should add that to the template.

efgpinto (approved):

docs/src/modules/java-protobuf/partials/query-syntax-advanced.adoc
octonato: Ok, I will merge this one and send another PR for the templates and samples. I noticed some other issues there as well.

PR #1522, opened 2023-03-07
docs: add migration guide from 1.1.x to 1.2.x

efgpinto:
References https://github.com/lightbend/kalix-proxy/issues/1936

This will be mainly for internal consumption because the major changes are for Spring SDK so I'm adding a `.md` file instead of creating a new page in docs. In any case, should be easy to convert to antora page anyway, if we prefer that.

### DIFF
--- a/docs/migration-guides.md
+++ b/docs/migration-guides.md
@@ -118,13 +118,13 @@ public class ShoppingCartEntity extends EventSourcedEntity<ShoppingCart, Shoppin
 
 Where an example of this `ShoppingCartEvent`, could be:
 ```java
-public sealed interface Events {
+public sealed interface ShoppingCartEvent {
 
   @TypeName("item-added")
-  record ItemAdded(ShoppingCart.LineItem item) implements Events { }
+  record ItemAdded(ShoppingCart.LineItem item) implements ShoppingCartEvent { }
 
   @TypeName("item-removed")
-  record ItemRemoved(String productId) implements Events { }
+  record ItemRemoved(String productId) implements ShoppingCartEvent { }
 }
 ```
 
--- a/docs/migration-guides.md
+++ b/docs/migration-guides.md
@@ -101,7 +101,7 @@ The following configurations are no longer needed and should be removed:
 
 #### EventSourcedEntity with typed events
 
-A `EventSourcedEntity<S>` is now `EventSourcedEntity<S, E>` where `E` is the top type for all the different event types such event-sourced entity will be allowed to emit. 
+An `EventSourcedEntity<S>` is now `EventSourcedEntity<S, E>` where `E` is the top type for all the different event types such event-sourced entity will be allowed to emit. Although on the protocol-first SDKs mentioned in the previous sections this is taken care of automatically, for the code-first approach you need to provide that type. 
 
 Thus, instead of :
 ```java
--- a/docs/migration-guides.md
+++ b/docs/migration-guides.md
@@ -1,0 +1,179 @@
+# Migration Guides
+
+## Migrating from 1.1.x to 1.2.x
+
+### Java SDK
+
+> **From 1.2.x onwards, the Java SDK is now called Java Protobuf SDK.**
+
+Accordingly, the package dependencies were updated accordingly:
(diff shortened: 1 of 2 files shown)


docs/migration-guides.md
```diff
@@ -0,0 +1,179 @@
...
+    </dependency>
+```
+
+That's it!
+
+### Scala SDK
+
+>**From 1.2.x onwards, the Scala SDK is now called Scala Protobuf SDK.**
+
+No changes required. The upgrade is a simple version bump here.
```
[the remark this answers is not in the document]
octonato: Correct, we also change the name of the scala artifacts, but...

This the dependency is included by the sbt plugin. So it's transparent for the user.

docs/migration-guides.md
```diff
@@ -0,0 +1,179 @@
...
+There should be now:
+```java
+public class ShoppingCartEntity extends EventSourcedEntity<ShoppingCart, ShoppingCartEvent> {
+  // command and event handlers ...
+}
+```
+
+Where an example of this `ShoppingCartEvent`, could be:
+```java
+public sealed interface Events {
```
octonato, line 121:
```suggestion
public sealed interface ShoppingCartEvent {
```

docs/migration-guides.md
```diff
@@ -0,0 +1,179 @@
...
+
+  @TypeName("item-added")
+  record ItemAdded(ShoppingCart.LineItem item) implements Events { }
+
+  @TypeName("item-removed")
+  record ItemRemoved(String productId) implements Events { }
```
octonato, line 127:
```suggestion
  record ItemAdded(ShoppingCart.LineItem item) implements ShoppingCartEvent { }

  @TypeName("item-removed")
  record ItemRemoved(String productId) implements ShoppingCartEvent { }
```

docs/migration-guides.md
```diff
@@ -0,0 +1,179 @@
...
+```
+
+Lastly, `KalixIntegrationTestKitSupport` was also moved, so you will need to replace:
+```java
+import kalix.springsdk.testkit.KalixIntegrationTestKitSupport;
+```
+With:
+```java
+import kalix.spring.testkit.KalixIntegrationTestKitSupport;
+```
```
octonato, line 179: There is one more thing to change. 

Users should delete `src/main/resources/application.properties`.

This file contains a setting to tell Spring to not start the webserver. That was needed in order to not start the spring netty.  

This now handled by the `starter` and if this old hack in place, our own server (based on akka-http). Our server is now configured as the default Spring reactive server, but the settings tell Spring to not start one if it finds one.

octonato (approved):
LGTM, with a few suggestions to add before merging.

Also, Johan's comment about the extra type param. Although we are already on the section about SpringSDK, it's good to mention it once again.

docs/migration-guides.md
efgpinto: That is mentioned in line #98 under "Spring Boot Configurations".. let me know what you think

docs/migration-guides.md
```diff
@@ -0,0 +1,179 @@
...
+
+#### Spring Boot Configurations
+
+The following configurations are no longer needed and should be removed: 
+- `spring.main.web-application-type=none` in `application.properties` (feel free to remove the file if that's the only property there)
+- in your `Main.java`, you can now remove `@Import(KalixConfiguration.class)`
+
+#### EventSourcedEntity with typed events
+
+A `EventSourcedEntity<S>` is now `EventSourcedEntity<S, E>` where `E` is the top type for all the different event types such event-sourced entity will be allowed to emit.
```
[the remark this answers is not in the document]
efgpinto: Yeah, all these are under section `### Spring SDK`.. 

Maybe something like would help?
```suggestion
An `EventSourcedEntity<S>` is now `EventSourcedEntity<S, E>` where `E` is the top type for all the different event types such event-sourced entity will be allowed to emit. Although on the protocol-first SDKs mentioned in the previous sections this is taken care of automatically, for the code-first approach you need to provide that type. 
```

docs/migration-guides.md
octonato: Oh, I missed. Thanks!

PR #1528, opened 2023-03-08
chore: adapt release draft and labeler

octonato:


.github/labeler.yml
```diff
@@ -4,25 +4,26 @@ kalix-proxy:
...
 
 
 java-sdk:
-  - sdk/spring-sdk/**/*
-  - sdk/spring-sdk-testkit/**/*
   - sdk/java-sdk-spring/**/*
   - sdk/java-sdk-spring-testkit/**/*
+  - sdk/spring-boot-starter/**/*
+  - sdk/spring-boot-starter-test/**/*
+  - sdk/spring-sdk-testkit/**/*
```
octonato, line 26: we still have a folder called `spring-sdk-testkit`, but not clear why. I will check it out, but I think more clean-up is needed here.

octonato: I can confirm that `spring-sdk-testkit` was a leftover. Removed now.

PR #1535, opened 2023-03-13
chore: template mentioning spring-boot plugin

octonato:
`mvn exec:exec` will also work, but our docs and readmes refer to the `spring-boot:run` and that's part of the spring boot plugin.


efgpinto (approved):

efgpinto:
Are there sample READMEs to update as well?

octonato:
> Are there sample READMEs to update as well?

I will check and send follow-up PR if needed.

PR #1541, opened 2023-03-17
fix: fix uberjar - use spring-boot-starter-parent

octonato:
This fix the uberjar layout. Alan hit that bug today and reported to me during a call.

It turns out that since we moved to the spring-boot-starter, the application won't start correctly when run directly from the java command, eg: `java -jar ...`. Which is how the docker image will bootstrap the application. 

The best solution I found is to rely on `spring-boot-starter-parent`. This dependency includes a few configurations to the build. Also, a shading config that builds a uberjar in the way the `starter` expects it to be.

### DIFF
--- a/maven-java/kalix-spring-boot-archetype/src/main/resources/archetype-resources/pom.xml
+++ b/maven-java/kalix-spring-boot-archetype/src/main/resources/archetype-resources/pom.xml
@@ -1,7 +1,12 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
   <modelVersion>4.0.0</modelVersion>
-
+  <parent>
+    <groupId>org.springframework.boot</groupId>
+    <artifactId>spring-boot-starter-parent</artifactId>
+    <version>3.0.4</version>
+    <relativePath/> 
+  </parent>
   <groupId>${groupId}</groupId>
   <artifactId>${artifactId}</artifactId>
   <version>${version}</version>
@@ -66,9 +71,6 @@
               <transformers>
                 <transformer implementation="org.apache.maven.plugins.shade.resource.AppendingTransformer">
                   <resource>reference.conf</resource>
-                </transformer>
-                <transformer implementation="org.apache.maven.plugins.shade.resource.ManifestResourceTransformer">
-                  <mainClass>${mainClass}</mainClass>
                 </transformer>
               </transformers>
             </configuration>
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -29,7 +29,7 @@ object Dependencies {
   val JUnitInterfaceVersion = "0.11"
   val JUnitJupiterVersion = "5.7.1"
   val SpringFrameworkVersion = "6.0.4"
-  val SpringBootVersion = "3.0.2"
+  val SpringBootVersion = "3.0.4"
 
   val CommonsIoVersion = "2.11.0"
   val MunitVersion = "0.7.29"
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -29,6 +29,7 @@ object Dependencies {
   val JUnitInterfaceVersion = "0.11"
   val JUnitJupiterVersion = "5.7.1"
   val SpringFrameworkVersion = "6.0.4"
+  // make sure to sync spring-boot-starter-parent version in samples and archetype to this version
   val SpringBootVersion = "3.0.4"
 
   val CommonsIoVersion = "2.11.0"
--- a/maven-java/kalix-spring-boot-kotlin-archetype/src/main/resources/archetype-resources/pom.xml
+++ b/maven-java/kalix-spring-boot-kotlin-archetype/src/main/resources/archetype-resources/pom.xml
@@ -2,6 +2,13 @@
 <project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
   <modelVersion>4.0.0</modelVersion>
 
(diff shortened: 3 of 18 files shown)


maven-java/kalix-spring-boot-archetype/src/main/resources/archetype-resources/pom.xml
```diff
@@ -1,7 +1,12 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
   <modelVersion>4.0.0</modelVersion>
-
+  <parent>
+    <groupId>org.springframework.boot</groupId>
+    <artifactId>spring-boot-starter-parent</artifactId>
+    <version>3.0.4</version>
```
octonato, line 7: Unfortunately, we can't reuse the version that we declare in `Dependencies.scala`. 

The `spring-boot-starter-parent` can't be a transitive deps. It needs to be declared on the pom so it includes the shading plugin config.

On the other hand, that gives users the possibility to bump the spring-boot version themselves. So, not that bad either.

maven-java/kalix-spring-boot-archetype/src/main/resources/archetype-resources/pom.xml
```diff
@@ -67,9 +72,6 @@
                 <transformer implementation="org.apache.maven.plugins.shade.resource.AppendingTransformer">
                   <resource>reference.conf</resource>
                 </transformer>
-                <transformer implementation="org.apache.maven.plugins.shade.resource.ManifestResourceTransformer">
-                  <mainClass>${mainClass}</mainClass>
-                </transformer>
```
octonato, line 72: We should let the shading config from `spring-boot-starter-parent` take over and define the manifest. We only keep the config for appending the reference.conf.

project/Dependencies.scala
```diff
@@ -29,7 +29,7 @@ object Dependencies {
   val JUnitInterfaceVersion = "0.11"
   val JUnitJupiterVersion = "5.7.1"
   val SpringFrameworkVersion = "6.0.4"
-  val SpringBootVersion = "3.0.2"
+  val SpringBootVersion = "3.0.4"
```
octonato, line 32:
```suggestion
  // make sure to sync spring-boot-starter-parent version in samples and archetype to this version
  val SpringBootVersion = "3.0.4"
```

aludwiko (commented):
sth is not right, I'm not sure if this is a part of this PR or some previous regresion, but building docker image with
```
mvn clean package docker:push -DskipTests
```
ends with an error:
```
[INFO] DOCKER> Temporary image tag skipped. Target image 'aludwiko/java-spring-transfer-workflow:latest' already has registry set or no registry is available
[ERROR] DOCKER> Unable to push 'aludwiko/java-spring-transfer-workflow:latest' : An image does not exist locally with the tag: aludwiko/java-spring-transfer-workflow  [An image does not exist locally with the tag: aludwiko/java-spring-transfer-workflow ]
```

octonato:
> Ok, spring-perf was based on some old sample, I see that with `mvn deploy` building the image works as expected.

Not sure if there is something to change here. I understand that the spring-perf readme was wrong and the readme in the jvm-sdk are correct. Do you confirm?

johanandren (approved):
LGTM (but no actual clue)

aludwiko:
Yes, I confirm that spring-perf Readme was outdated, but I'm still facing some issues with this change. I can run the app with `java -jar` but running docker image ends with:
```
Exception in thread "main" java.lang.IllegalStateException: Failed to get nested archive for entry BOOT-INF/lib/kalix-spring-boot-starter-1.2.0-M1-31-6bbba7a6-dev-SNAPSHOT.jar
	at org.springframework.boot.loader.archive.JarFileArchive.getNestedArchive(JarFileArchive.java:120)
	at org.springframework.boot.loader.archive.JarFileArchive$NestedArchiveIterator.adapt(JarFileArchive.java:274)
	at org.springframework.boot.loader.archive.JarFileArchive$NestedArchiveIterator.adapt(JarFileArchive.java:265)
	at org.springframework.boot.loader.archive.JarFileArchive$AbstractIterator.next(JarFileArchive.java:226)
	at org.springframework.boot.loader.ExecutableArchiveLauncher.createClassLoader(ExecutableArchiveLauncher.java:104)
	at org.springframework.boot.loader.Launcher.launch(Launcher.java:55)
	at org.springframework.boot.loader.JarLauncher.main(JarLauncher.java:65)
Caused by: java.io.IOException: Unable to open nested jar file 'BOOT-INF/lib/kalix-spring-boot-starter-1.2.0-M1-31-6bbba7a6-dev-SNAPSHOT.jar'
	at org.springframework.boot.loader.jar.JarFile.getNestedJarFile(JarFile.java:304)
	at org.springframework.boot.loader.jar.JarFile.getNestedJarFile(JarFile.java:290)
	at org.springframework.boot.loader.archive.JarFileArchive.getNestedArchive(JarFileArchive.java:116)
	... 6 more
Caused by: java.lang.IllegalStateException: Unable to open nested entry 'BOOT-INF/lib/kalix-spring-boot-starter-1.2.0-M1-31-6bbba7a6-dev-SNAPSHOT.jar'. It has been compressed and nested jar files must be stored without compression. Please check the mechanism used to create your executable jar file
	at org.springframework.boot.loader.jar.JarFile.createJarFileFromFileEntry(JarFile.java:330)
	at org.springframework.boot.loader.jar.JarFile.createJarFileFromEntry(JarFile.java:312)
	at org.springframework.boot.loader.jar.JarFile.getNestedJarFile(JarFile.java:301)
	... 8 more
```
Maybe it's only my local problem but it will be good to validate this. Also, I've noticed that the fat jar is pretty big now - we have unpacked jars together with packed ones in `BOOT-INF/lib` is this expected?

octonato:
> Maybe it's only my local problem but it will be good to validate this. Also, I've noticed that the fat jar is pretty big now - we have unpacked jars together with packed ones in `BOOT-INF/lib` is this expected?

I'm checking this.

octonato:
Indeed package jar is must larger. I will dig it further. 

The `BOOT-INF/lib` is created by the Spring Boot parent packaging config, but not clear why it has a mix of packaged and unpackaged jars.

PR #1542, opened 2023-03-20
feat: run proxy in testcontainer in dev-mode

octonato:
This is still in draft. I want to experiment more with a few ideas. 

I will comment on the PR to give more context.


build.sbt
```diff
@@ -26,6 +27,41 @@ def common: Seq[Setting[_]] =
     Compile / javacOptions ++= Seq("-encoding", "UTF-8"),
     Compile / scalacOptions ++= Seq("-encoding", "UTF-8", "-deprecation"))
 
+lazy val devMode = project
+  .in(file("devmode"))
```
octonato, line 31: New module to provide integration with testcontainers for dev-mode. 

Code here is a copy of KalixProxyContainer that we have for integration tests with some small differences.

build.sbt
```diff
@@ -26,6 +27,41 @@ def common: Seq[Setting[_]] =
...
+    // javadoc (I think java 9 onwards) refuses to compile javadocs if it can't compile the entire source path.
+    // but since we have java files depending on Scala files, we need to include ourselves on the classpath.
+    Compile / doc / dependencyClasspath := (Compile / fullClasspath).value,
+    Compile / doc / javacOptions ++= Seq(
+      "-Xdoclint:none",
+      "-overview",
+      ((Compile / javaSource).value / "overview.html").getAbsolutePath,
+      "-notimestamp",
+      "-doctitle",
+      "Kalix Dev Tools",
```
octonato, line 60:
```suggestion
      "Kalix Dev Mode",
```

devmode/src/main/resources/reference.conf
```diff
@@ -0,0 +1,7 @@
+kalix.dev-mode {
+    enabled = true
```
octonato, line 2: Enable by default for now, but we may consider it an opt-in in an initial phase. Users that are now used to use docker-compose will have a port conflict if we start testcontainers on their behalf.

sdk/spring-boot-starter/src/main/scala/kalix/spring/boot/KalixContainerInitializer.scala
```diff
@@ -0,0 +1,65 @@
...
+import com.typesafe.config.Config
+import com.typesafe.config.ConfigFactory
+import kalix.devmode.KalixProxyContainer
+import org.slf4j.LoggerFactory
+import org.springframework.context.ApplicationContextInitializer
+import org.springframework.context.ApplicationListener
+import org.springframework.context.ConfigurableApplicationContext
+import org.springframework.context.event.ContextClosedEvent
+
+class KalixContainerInitializer
```
octonato, line 32: For now, only integration with Spring. 

Later, we want to have a similar initializer for Java/Protobuf and Scala/Protobuf. Not yet clear to me the best way to do that though.

sdk/spring-boot-starter/src/main/scala/kalix/spring/boot/KalixContainerInitializer.scala
```diff
@@ -0,0 +1,65 @@
...
+    extends ApplicationContextInitializer[ConfigurableApplicationContext]
+    with ApplicationListener[ContextClosedEvent] {
+
+  private val logger = LoggerFactory.getLogger(this.getClass)
+
+  private val config = {
+    val mainConfig = ConfigFactory.load()
+
+    val devConfigFile = new File(System.getProperty("user.dir"), "dev-mode.conf")
```
octonato, line 41: This is a bit unorthodox. 

I'm looking for a file called `dev-mode.conf` at the root of the project, not under `src/main/resources`. This file won't be packaged, so the final jar won't have it and will just work as expected in prod. 

An alternative is to place it under `src/main/resources` and configure maven/sbt to filter it out when packaging. 

The behavior that I'm looking for is the following: 

* a user should not have to pass the file using a `-D` option.
* the application should behave the same if started with `mvn spring-boot:run` or if main function is run from inside the editor (not extra config needed).
* when packaging the application, this file should not be include and `KalixProxyContainer` should not be started.

octonato: Btw, I tested with IntelliJ and with VSCode and in both cases the editor runs the main function from the root directory and the "trick" works as expected. 

When running a package application, it also work. Meaning, dev-mode is not enabled because there is not file. However, there is one catch though. 

If you package the jar and run is with `java -jar target/my-application.jar` it will find the `dev-mode.conf` file. Instead, we need to: `(cd target && java -jar my-application.jar )`.

project/Dependencies.scala
```diff
@@ -107,6 +109,8 @@ object Dependencies {
 
   val coreSdk = deps ++= coreDeps
 
+  val devtools = deps ++= Seq(typesafeConfig, testContainers)
```
octonato, line 112: Also not ideal, we are adding testcontainers to the classpath. The prod classpath.

build.sbt
```diff
@@ -26,6 +28,41 @@ def common: Seq[Setting[_]] =
     Compile / javacOptions ++= Seq("-encoding", "UTF-8"),
     Compile / scalacOptions ++= Seq("-encoding", "UTF-8", "-deprecation"))
 
+lazy val devTools = project
+  .in(file("devtools"))
```
octonato, line 32: This is just the part to start the containers. Potentially usable from Protobuf SDKs as well.

build.sbt
```diff
@@ -236,6 +273,42 @@ lazy val springBootStarter = project
       "-noqualifier",
       "java.lang"))
 
+lazy val springBootDevToolsStarter = project
+  .in(file("sdk/spring-boot-devtools-starter"))
```
octonato, line 277: This is another started that adds some spring-boot logic to start the container when we the Spring application starts with `mvn spring-boot:run`.

I tried to shade this one and then exclude when packing, but run into too many issues.  
Shading doesn't play well with some Spring classes. In some cases, sbt-assembly cannot change the package of some spring framework classes. I could solve it after some hacks and juggling. But solution was too fragile, IMO.

Also, excluding it from packaging didn't work at all. The `spring-boot-plugin` will exclude it when running the app. Not useful for us.

devtools/src/main/resources/reference.conf
```diff
@@ -0,0 +1,35 @@
...
+    proxy-image = ""
+    proxy-image = ${?KALIX_PROXY_IMAGE}
+
+    # Allows for overriding the port the Kalix Proxy will run.
+    # Default to 9000. Useful when running more than one service in dev-mode.
+    proxy-port = 9000
+
+    # Defines a unique identification name for this service.
+    # Useful when running more than one service and testing intra-service calls.
+    service-name = ""
```
octonato, line 18: Template can fill in this with name of artifact, for example.

devtools/src/main/resources/reference.conf
```diff
@@ -0,0 +1,35 @@
...
+    # Enable advanced view features (multi-table joins).
+    view-features-all = false
+
+    # When configuring with a Kafka broker, this settings should point
+    # to a Kafka properties file, eg: /conf/kafka.properties.
+    broker-config-file = ""
+
+    # When running with a PubSub emulator, this settings must be configured
+    # to its host, eg: gcloud-pubsub-emulator.
+    pubsub-emulator-host = ""
```
octonato, line 33: Didn't really try those two. Not yet working. 

A few more things will be needed, like mounting volume to read config (Kafka) and passing pubsub config to container.

devtools/src/main/scala/kalix/devtools/impl/KalixProxyContainer.scala
```diff
@@ -0,0 +1,124 @@
...
+import org.testcontainers.containers.output.Slf4jLogConsumer
+import org.testcontainers.utility.DockerImageName
+
+object KalixProxyContainer {
+
+  object KalixProxyContainerConfig {
+    def apply(config: Config): KalixProxyContainerConfig = {
+
+      // FIXME: core Kalix should be able to read dev-mode.conf from root folder as well
+      // as such, we can kalix.user-function-port configured to something else without touching the prod config
```
octonato, line 34: Still evaluating if `ROOT/dev-model.conf` is better then `src/main/resources/dev-model.conf` + exclusion.

devtools/src/main/scala/kalix/devtools/impl/KalixProxyContainer.scala
```diff
@@ -0,0 +1,124 @@
...
+class KalixProxyContainer private (image: DockerImageName, config: KalixProxyContainerConfig)
+    extends GenericContainer[KalixProxyContainer](image) {
+
+  private val containerLogger = LoggerFactory.getLogger("kalix-proxy-server")
+  containerLogger.info("KalixProxyContainer config : {}", config)
+  withLogConsumer(new Slf4jLogConsumer(containerLogger))
+
+  val proxyPort = config.proxyPort
+  val userFunctionPort = config.userFunctionPort
+  addFixedExposedPort(proxyPort, proxyPort)
```
octonato, line 87: testcontainers folks really unrecommend fixing the port. I understand them. They are very focus on testing and in that context, fixed ports is are recipe to conflicts. 

But we need fix ports so users know there to reach the proxy, ie: localhost:9000.

Maybe, we should not use testcontainers here, but https://github.com/docker-java/docker-java directly.

devtools/src/main/scala/kalix/devtools/impl/KalixProxyContainer.scala
```diff
@@ -0,0 +1,124 @@
...
+
+  // JVM are that should be passed to the proxy container on start-up
+  withCommand("-Dconfig.resource=dev-mode.conf -Dlogback.configurationFile=logback-dev-mode.xml")
```
octonato, line 90: ah, here we can configure the pubsub if `pubsubEmulatorHost.nonEmpty`

octonato (commented):
Left a few comments to add some context. 

This is not ready, but a first review round will be nice.

sdk/spring-boot-starter/src/main/scala/kalix/spring/boot/KalixContainerInitializer.scala
franciscolopezsancho: Let me check that I understand the problem. If we let the user put the file in `src/main/resources` and then is included in the jar. Then when deployed in Kubernetes it will again create a container. Am I right? 
If so, how about this? Could we create a boolean that gets the entire `dev-mode` to false when `mvn deploy`? Meaning, it's our maven plugin that takes care of setting this creation of the container in Kubernetes off. WDYT?

octonato: For clarity, as I understand option 3, we won't though the file, but we would ensure that the container build have a -D flag added to it.

aludwiko: I would prefer 2 or 3. Let's imagine a situation in that a user will build the image manually and hack the -D flag. What are the possible consequences?

octonato: Option 3 is the one that we would use a `-D` property. So I guess you mean option 1 or 2. 

None of the solutions is waterproof. In option 2, we will need to tweak the build to exclude `dev-mode.conf` from final package. Users will be able to remove that tweak because it will be part of the pom.xml / build.sbt config. 

The consequence of packaging such a file in prod is that it will try to start a dev-mode proxy inside the user pod. That will probably fail, but the application (user function) will start anyway and the prod proxy will find it.

octonato: Now that we said that, I think Option 1 is the most safe. 

If we explicitly look-up for `dev-mode.conf` at the root of the application, then a user will have to consciously build a docker image with such a file at the root of the application. Not impossible, but quite a conscious action. 

While removing some configs from pom.xml/build.sbt can happen by accident or inadvertence.

aludwiko: My comment was more about that I'm not afraid too much of hacking a docker image (which as mentioned is possible in all 3 scenarios) I was just curious if there are any really serious consequences that I'm not aware of. 

I still prefer 2 or 3, putting config files in the root directory feels wrong.

octonato:
> Have you tried to run 2 UF processes together with direct eventing?

No, not yet and I think it might require some more changes. 
Containers should be able to 'see' each other. Maybe something we can pass in through `dev-mode.conf`, but it might be that some extra tweaks will be needed because of docker.

sdk/spring-boot-starter/src/main/scala/kalix/spring/boot/KalixContainerInitializer.scala
aludwiko: Since I don't have any merit argument against it, let's start with config in the root.

sdk/java-sdk-spring/src/main/scala/kalix/spring/boot/KalixConfiguration.scala
```diff
@@ -45,7 +46,9 @@ object KalixConfiguration {
 class KalixConfiguration(applicationContext: ApplicationContext) {
 
   @Bean
-  def config(): Config = ConfigFactory.load()
+  def config(): Config = {
+    KalixRunner.prepareConfig(ConfigFactory.load())
```
octonato, line 50: I initially wanted to run this "preparation" inside the KalixRunner, but that doesn't play well with the integration tests. 

It's ok to load the dev-mode.conf, each time. For now, a package application won't have `dev-mode.conf` and the preparation will have zero effect. The problem is when running integration tests. Then we need to disable it and the ordering of loading matters.

octonato (commented):
Moved further with this PR to allow overwrite of user function port using `dev-mode.conf`. 

As it is now, it won't interfere with integration tests, but the solution is only valid for the Java SDK as it relies on how Spring loads the class. 

Next is to find a solution that is independent of Spring and will also work for Java/Scala Protobuf SDKs.

octonato:
closing in favor of #1561

PR #1544, opened 2023-03-22
fix: remove double shading

octonato:
As reported in https://github.com/lightbend/kalix-jvm-sdk/pull/1541#issuecomment-1478157152, we were creating a hyper-fat-jar with double the size it should have. 

This PR will fix it. 

It turns out that we had a config for maven-shade-plugin coming from spring-boot-parent and our own config. Instead, we should have only one and merge them.

### DIFF
--- a/maven-java/kalix-spring-boot-archetype/src/main/resources/archetype-resources/pom.xml
+++ b/maven-java/kalix-spring-boot-archetype/src/main/resources/archetype-resources/pom.xml
@@ -42,6 +42,30 @@
       </extension>
     </extensions>
 
+    <pluginManagement>
+      <plugins>
+        <plugin>
+          <groupId>org.apache.maven.plugins</groupId>
+          <artifactId>maven-shade-plugin</artifactId>
+          <executions>
+            <execution>
+              <phase>package</phase>
+              <goals>
+                <goal>shade</goal>
+              </goals>
+              <configuration>
+                <transformers combine.children="append">
+                  <transformer implementation="org.apache.maven.plugins.shade.resource.AppendingTransformer">
+                    <resource>reference.conf</resource>
+                  </transformer>
+                </transformers>
+              </configuration>
+            </execution>
+          </executions>
+        </plugin>
+      </plugins>
+    </pluginManagement>
+
     <plugins>
       <plugin>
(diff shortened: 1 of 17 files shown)


maven-java/kalix-spring-boot-archetype/src/main/resources/archetype-resources/pom.xml
```diff
@@ -42,6 +42,30 @@
...
+          <groupId>org.apache.maven.plugins</groupId>
+          <artifactId>maven-shade-plugin</artifactId>
+          <executions>
+            <execution>
+              <phase>package</phase>
+              <goals>
+                <goal>shade</goal>
+              </goals>
+              <configuration>
+                <transformers combine.children="append">
```
octonato, line 57: This will allow the child pom to append transformers to the plugin config defined in `spring-boot-parent`.

maven-java/kalix-spring-boot-archetype/src/main/resources/archetype-resources/pom.xml
```diff
@@ -42,6 +42,30 @@
...
+                  <transformer implementation="org.apache.maven.plugins.shade.resource.AppendingTransformer">
+                    <resource>reference.conf</resource>
+                  </transformer>
+                </transformers>
+              </configuration>
+            </execution>
+          </executions>
+        </plugin>
+      </plugins>
+    </pluginManagement>
```
octonato, line 67: The other plugins can stay under `build.plugins`, but this one, need to go inside `build.pluginManagement.plugins`. Otherwise, we can't merge the configs.

maven-java/kalix-spring-boot-archetype/src/main/resources/archetype-resources/pom.xml
```diff
@@ -161,7 +164,6 @@
         <!-- configure src/it/java and src/it/resources -->
         <groupId>org.codehaus.mojo</groupId>
         <artifactId>build-helper-maven-plugin</artifactId>
-        <version>3.2.0</version>
```
octonato, line 164: Some plugins are already defined in `spring-boot-parent` and better not to override the version.

johanandren (approved):

PR #1555, opened 2023-03-30
fix(samples): s2s java example and improve docs

efgpinto:
Small fix for the s2s java sample and some docs improvement.

### DIFF
--- a/samples/java-spring-eventsourced-customer-registry/src/main/java/customer/domain/CustomerEvent.java
+++ b/samples/java-spring-eventsourced-customer-registry/src/main/java/customer/domain/CustomerEvent.java
@@ -6,13 +6,13 @@ import static customer.domain.CustomerEvent.*;
 
 public sealed interface CustomerEvent {
 
-  @TypeName("customer-created") // <1>
+  @TypeName("internal-customer-created") // <1>
   record CustomerCreated(String email, String name, Address address) implements CustomerEvent {}
 
-  @TypeName("name-changed")
+  @TypeName("internal-name-changed")
   record NameChanged(String newName) implements CustomerEvent {}
 
-  @TypeName("address-changed")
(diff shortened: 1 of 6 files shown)


octonato (approved):

samples/java-spring-eventsourced-customer-registry/src/main/java/customer/domain/CustomerEvent.java
```diff
@@ -6,13 +6,13 @@
 
 public sealed interface CustomerEvent {
 
-  @TypeName("customer-created") // <1>
+  @TypeName("internal-customer-created") // <1>
```
aludwiko, line 9: Why the prefix `internal` ?

efgpinto: There is another sealed interface used for the public api.. and they had the same names. That causes an exception as these need to be unique. So I changed the internal ones to have that prefix.

octonato: I'm curious to understand why they need to be unique. 

Are they handle in different code paths? Why the name clash?

octonato: Oh yes, that's possible. Doesn't matter which components and how to get there. We always pass by the cache. 

Which makes me think if we should not have a namespace for the cache. A cache per component. For the user, there is not reason why they could not reuse a name. And can even be confusing when seeing such an error message. Worst, that's a runtime error.

efgpinto: here it is https://github.com/lightbend/kalix-jvm-sdk/issues/1556

PR #1558, opened 2023-04-04
fix: explicit input types for workflow step definition

aludwiko:
References #1557


octonato (approved):
LGTM

The Scala variant will be cooler. 😄

Issue #1556, opened 2023-04-03
Bug: TypeInfo id clash when accessing internal cache


efgpinto:
This was noticed when running the s2s eventing sample for java ([this](https://github.com/lightbend/kalix-jvm-sdk/tree/main/samples/java-spring-eventsourced-customer-registry)).

If we use the same type identifier for multiple types, we will get some exception like:

```
14:56:53.717 ERROR k.javasdk.impl.action.ActionsImpl - Failure during handling of command customer.api.CustomerEventsService.KalixSyntheticMethodOnESCustomer
java.lang.IllegalStateException: Collision with existing existing mapping class customer.domain.CustomerEvent$CustomerCreated -> customer-created. The same type name can't be used for other class class customer.api.CustomerPublicEvent$Created
        at kalix.javasdk.impl.JsonMessageCodec.$anonfun$lookupTypeHint$3(JsonMessageCodec.scala:71)
        at java.base/java.util.concurrent.ConcurrentHashMap.compute(ConcurrentHashMap.java:1940)
        at kalix.javasdk.impl.JsonMessageCodec.lookupTypeHint(JsonMessageCodec.scala:64)
        at kalix.javasdk.impl.JsonMessageCodec.lookupTypeHint(JsonMessageCodec.scala:54)
        at kalix.javasdk.impl.JsonMessageCodec.encodeScala(JsonMessageCodec.scala:40)
        at kalix.javasdk.impl.action.ActionsImpl.effectToResponse(ActionsImpl.scala:122)
        at kalix.javasdk.impl.action.ActionsImpl.handleUnary(ActionsImpl.scala:188)
        at kalix.protocol.action.ActionsHandler$.$anonfun$partial$2(ActionsHandler.scala:118)
        at scala.concurrent.impl.Promise$Transformation.run(Promise.scala:470)
        at akka.dispatch.BatchingExecutor$AbstractBatch.processBatch(BatchingExecutor.scala:63)
        at akka.dispatch.BatchingExecutor$BlockableBatch.$anonfun$run$1(BatchingExecutor.scala:100)
        at scala.runtime.java8.JFunction0$mcV$sp.apply(JFunction0$mcV$sp.scala:18)
        at scala.concurrent.BlockContext$.withBlockContext(BlockContext.scala:94)
        at akka.dispatch.BatchingExecutor$BlockableBatch.run(BatchingExecutor.scala:100)
[...]

PR #1562, opened 2023-04-05
feat!: default to FQCN when typename is not present

efgpinto:
References https://github.com/lightbend/kalix-jvm-sdk/issues/1556

This MR:
- changes the default back to FQCN when no TypeName is provided (which also decreases chances of collision within the same service)
  - this is a breaking change for Java SDK (code-first)
- update docs to advise that TypeName should ideally be unique.

### DIFF
--- a/sdk/java-sdk-spring/src/test/scala/kalix/javasdk/impl/JsonMessageCodecSpec.scala
+++ b/sdk/java-sdk-spring/src/test/scala/kalix/javasdk/impl/JsonMessageCodecSpec.scala
@@ -73,7 +73,7 @@ class JsonMessageCodecSpec extends AnyWordSpec with Matchers {
 
     "default to FQCN for typeUrl (java)" in {
       val encoded = messageCodec.encodeJava(SimpleClass("abc", 10))
-      encoded.getTypeUrl shouldBe jsonTypeUrlWith("SimpleClass")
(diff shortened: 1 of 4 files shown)


sdk/java-sdk-spring/src/test/scala/kalix/javasdk/impl/JsonMessageCodecSpec.scala
```diff
@@ -73,7 +73,7 @@ class JsonMessageCodecSpec extends AnyWordSpec with Matchers {
 
     "default to FQCN for typeUrl (java)" in {
       val encoded = messageCodec.encodeJava(SimpleClass("abc", 10))
-      encoded.getTypeUrl shouldBe jsonTypeUrlWith("SimpleClass")
```
octonato, line 76: this test was lying 😄

octonato (approved):
LGTM

PR description is not updated though.

sdk/java-sdk-spring/src/test/scala/kalix/javasdk/impl/JsonMessageCodecSpec.scala
efgpinto: it was indeed :D

efgpinto:
> PR description is not updated though.

Updated now. Also added note in docs about the recommendation to use unique type names per Kalix Service.

octonato (approved):

PR #1564, opened 2023-04-06
docs(sample): add integration test example for s2s eventing

efgpinto:
This PR:
- adds an example for an integration test with s2s eventing for future reference
- this was mainly an exercise for me to understand what was possible at the moment and how
- Logs from the testcontainer proxy will now come formatted in plain text and not JSON

~I think it's useful to run in CI anyway because it also validates the s2s logic.~ Commented out this part because I don't want to spend much more time on this and it needs more work. Might revisit later.


octonato (approved):
LGTM

PR #1567, opened 2023-04-07
fix: retry container port until is available

octonato:
This is still not perfect, but good enough for now.

Removal of container takes too long. I think we should look for a better solution. Like tiny http proxy running on port 9000 and forwarding traffic to kalix-proxy running on random port. Then when restarting, the proxy running on 9000 will free the port immediately while testcontainers would start a new kalix-proxy on another port.


aludwiko (approved):

efgpinto (approved):
LGTM

PR #1568, opened 2023-04-07
chore: make logConfig optional

octonato:
No need to add more information to the pom.xml about logConfig. 

New projects already have a logback-dev-mode.xml file. We can default to it. Users can always use another file if they prefer.


efgpinto (approved):
LGTM

PR #1569, opened 2023-04-07
fix: pass pubsub emulator exposed port

octonato:


devtools/src/main/scala/kalix/devtools/impl/KalixProxyContainer.scala
```diff
@@ -80,10 +76,10 @@ class KalixProxyContainer private (image: DockerImageName, config: KalixProxyCon
...
 
   val pubSubContainerArg =
-    if (config.pubsubEmulatorHost.nonEmpty) {
-      withEnv("PUBSUB_EMULATOR_HOST", config.pubsubEmulatorHost)
-      Some("-Dkalix.proxy.eventing.support=google-pubsub-emulator")
-    } else None
+    config.pubsubEmulatorPort.map { port =>
+      withEnv("PUBSUB_EMULATOR_HOST", "host.testcontainers.internal")
+      "-Dkalix.proxy.eventing.support=google-pubsub-emulator"
+    }
```
octonato, line 82: When running it from kalix:run, we should do it differently. The pubsube will be running through docker-compose. And the proxy container will run inside testcontainrs. 

Therefore, we must configure it to reach pubsub on the container host + port.

maven-java/kalix-maven-plugin/src/main/scala/kalix/RunMojo.scala
```diff
@@ -89,8 +89,8 @@ class RunMojo extends AbstractMojo {
   /**
    * When running with a PubSub emulator, this settings must be configured to its host, eg: gcloud-pubsub-emulator.
    */
-  @Parameter(property = "kalix.dev-mode.pubsub-emulator-host")
-  private var pubsubEmulatorHost: String = ""
+  @Parameter(property = "kalix.dev-mode.pubsub-emulator-port")
+  private var pubsubEmulatorPort: Int = 0
```
octonato, line 93: Integer defaulting to 0, meaning no pubsub configured.

maven-java/kalix-maven-plugin/src/main/scala/kalix/RunMojo.scala
```diff
@@ -89,8 +89,8 @@ class RunMojo extends AbstractMojo {
   /**
    * When running with a PubSub emulator, this settings must be configured to its host, eg: gcloud-pubsub-emulator.
```
efgpinto, line 90:
```suggestion
   * When running with a PubSub, this setting must be configured to its port.
```

devtools/src/main/scala/kalix/devtools/impl/KalixProxyContainer.scala
```diff
@@ -95,7 +91,10 @@ class KalixProxyContainer private (image: DockerImageName, config: KalixProxyCon
     containerLogger.info("Starting Kalix Proxy Server container...")
     containerLogger.info("Using proxy image [{}]", image)
     containerLogger.info("KalixProxyContainer config : {}", config)
+
     Testcontainers.exposeHostPorts(userFunctionPort)
+    config.pubsubEmulatorPort.foreach(Testcontainers.exposeHostPorts(_))
```
efgpinto, line 96: Is this actually needed? I thought this was only needed to expose ports from the testcontainters to outside? but since the pubsub in this case is expect to run externally to the testcontainers.. is this needed? or maybe I misunderstood something.

octonato: `exposeHostPorts` is used to expose host ports to the process running inside testcontainer. 

From the perspective of the kalix proxy, both pubsub and user function are running on the host. Although pubsub runs in a container, it's accessible from the host.

I can check if that's really necessary, but this is how the KalixTestKit configures the access to the user function as well.

efgpinto: yeah, you're right. I got it wrong previously. Thanks :)

efgpinto (approved):
Apart from the tiny outdated comment mentioned earlier, this LGTM.

maven-java/kalix-maven-plugin/src/main/scala/kalix/RunMojo.scala
octonato: good catch

PR #1570, opened 2023-04-11
chore: replacing junit 4 @RunWith annotation with junit 5 @ExtendWith

aludwiko:
from spring docs:
```
If you are using JUnit 4, do not forget to also add @RunWith(SpringRunner.class) to your test, otherwise the annotations will be ignored. If you are using JUnit 5, there is no need to add the equivalent @ExtendWith(SpringExtension.class) as @SpringBootTest and the other @…​Test annotations are already annotated with it.
```


octonato:
Do we need to add `ExtendWith(SpringExtension.class)`? From the spring doc fragment you posted here I understood that it's not needed to add it because we have already `SpringBootTest`. Correct?

aludwiko:
Yes, I've noticed that after the PR creation, the fix is on the way.

octonato (approved):

PR #1573, opened 2023-04-11
fix: mount current dir as volume under /conf

octonato:
Seems to work, but needs more testing. So in draft for now. 

Btw, we must have integration tests using `mvn kalix:run`. Not yet clear how to do it though.


octonato:
> > we must have integration tests using `mvn kalix:run`.
> 
> Sorry, I think you might have explained this previously... but why is this?

We have some logic going on inside the plugin and dev-tools. And are not testing them. 

Like, if I run `mvn kalix:run` does my application really starts? 

If I run it with pubsub, is the pubsub then correctly configured? And so on...

efgpinto (approved):
LGTM

PR #1516, opened 2023-03-06
feat: sdk for workflow timeout, step timeout and recover strategy

aludwiko:
References 
 - https://github.com/lightbend/kalix-proxy/issues/1948
 - https://github.com/lightbend/kalix-proxy/issues/1950
 - https://github.com/lightbend/kalix-proxy/issues/1949
 - https://github.com/lightbend/kalix-proxy/issues/1878
 - https://github.com/lightbend/kalix-proxy/issues/1975

PR #1574, opened 2023-04-11
fix(testkit): remove modification of proxy name

efgpinto:
This change was introduced in https://github.com/lightbend/kalix-jvm-sdk/pull/1516/files#diff-1cb4454ce26068222dcb8b5e3041baf55d5b3e4c9c68349c953dd4268e464922R44

But I'm not sure what it's meant to do, feels weird. Also, it's breaking ubi tests in proxy.

We could work around it if there's a good reason for it to stay though.

### DIFF
--- a/sdk/java-sdk-protobuf-testkit/src/main/java/kalix/javasdk/testkit/KalixProxyContainer.java
+++ b/sdk/java-sdk-protobuf-testkit/src/main/java/kalix/javasdk/testkit/KalixProxyContainer.java
@@ -41,7 +41,7 @@ public class KalixProxyContainer extends GenericContainer<KalixProxyContainer> {
   static {
     String customImage = System.getenv("KALIX_TESTKIT_PROXY_IMAGE");
     if (customImage == null) {
-      DEFAULT_PROXY_IMAGE_NAME = DockerImageName.parse(BuildInfo.proxyImage()).withTag(BuildInfo.proxyVersion().replace("-SNAPSHOT", ""));
+      DEFAULT_PROXY_IMAGE_NAME = DockerImageName.parse(BuildInfo.proxyImage()).withTag(BuildInfo.proxyVersion());
     } else {
       Logger logger = LoggerFactory.getLogger(KalixProxyContainer.class);
       DEFAULT_PROXY_IMAGE_NAME = DockerImageName.parse(customImage);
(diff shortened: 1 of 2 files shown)


sdk/java-sdk-protobuf-testkit/src/main/java/kalix/javasdk/testkit/KalixProxyContainer.java
```diff
@@ -41,7 +41,7 @@ public class KalixProxyContainer extends GenericContainer<KalixProxyContainer> {
   static {
     String customImage = System.getenv("KALIX_TESTKIT_PROXY_IMAGE");
     if (customImage == null) {
-      DEFAULT_PROXY_IMAGE_NAME = DockerImageName.parse(BuildInfo.proxyImage()).withTag(BuildInfo.proxyVersion().replace("-SNAPSHOT", ""));
```
aludwiko, line 44: I changed it because when working **locally** with a proxy snapshot like: `1.1.7-M1-18-c6b96c08-SNAPSHOT` the docker image name is actually without the `-SNAPSHOT` postfix, only  `1.1.7-M1-18-c6b96c08`, so it was necessary to change evn value each time. For the ubi app test, the docker image name contains the snapshot postfix.
```
docker image tag gcr.io/kalix-public/kalix-proxy:latest gcr.io/kalix-public/kalix-proxy:${PROXY_SNAPSHOT_VERSION}
```
Since this is blocking, let's merge it, but it will be nice to somehow align the proxy docker image names, e.g. publishing locally with a snapshot postfix. I could give it a try in a separate PR.

aludwiko (approved):

sdk/java-sdk-protobuf-testkit/src/main/java/kalix/javasdk/testkit/KalixProxyContainer.java
octonato: We can overwrite the image using env var `KALIX_TESTKIT_PROXY_IMAGE`. Since it's an edge case, it might be simpler to use this workaround instead.

octonato (approved):
LGTM, but CI is failing. Probably some flakiness.

sdk/java-sdk-protobuf-testkit/src/main/java/kalix/javasdk/testkit/KalixProxyContainer.java
efgpinto: yeah I see now. I don't have a strong opinion about it to be honest.

PR #1580, opened 2023-04-17
feat: local service port mapping

octonato:


octonato:
Ready for a first review round. 

Working when using `mvn kalix:run` and Java Protobuf, but not yet when running UF in container.

sdk/java-sdk-protobuf/src/main/scala/kalix/javasdk/KalixRunner.scala
```diff
@@ -81,17 +83,40 @@ object KalixRunner {
...
+        val serviceName = entry.getKey
+        val port = entry.getValue.render()
+
+        logger.info(s"Adding dev-mode service port mapping for $serviceName:$port")
+        val mapping = ConfigFactory.parseString(s"""
+             |akka.grpc.client.$serviceName {
+             |  service-discovery {
+             |    service-name = "$serviceName"
+             |  }
+             |  host = "localhost"
```
efgpinto, line 103: Not sure if this needs to be `0.0.0.0` for cross-platform compatibility... does it? 🤔

efgpinto (commented):
looking good, left a small question

sdk/java-sdk-protobuf/src/main/scala/kalix/javasdk/KalixRunner.scala
```diff
@@ -81,17 +83,40 @@ object KalixRunner {
...
+    if (mainConfig.hasPath("kalix.dev-mode.service-port-mappings")) {
+
+      val mappings = mainConfig.getConfig("kalix.dev-mode.service-port-mappings")
+      // each kalix.dev-mode.service-port-mappings becomes a new akka.grpc.client service-discovery mapping
+      // that is then appended to the main configuration
+      mappings.entrySet().asScala.foldLeft(mainConfig) { (main, entry) =>
+        val serviceName = entry.getKey
+        val port = entry.getValue.render()
+
+        logger.info(s"Adding dev-mode service port mapping for $serviceName:$port")
```
franciscolopezsancho, line 97: Keep me honest here. Because it's `info` I understand that the message is meant for the user but on the other hand talking about `mapping`, it seems to me is meant for a developer. From the comment above I think it's creating the configuration to create a gRPC client to consume some service. 
If the message it's for the user do you think is worth giving some other info?

sdk/java-sdk-protobuf/src/main/scala/kalix/javasdk/KalixRunner.scala
octonato: Good point. I will change to `0.0.0.0`.

sdk/java-sdk-protobuf/src/main/scala/kalix/javasdk/KalixRunner.scala
octonato: Correct. It might be better to use debug. 

The goal here is indeed to print it in during developer. 

True that some one can still build a docker image that by default such a mapping and then deploy it to prod. That won't work and the user/developer will be explicitly taking a path that will cause malfunction.

Anyway, the plugin is already printing the dev-mode port mappings, so we can probably remove this logging all together instead of using `debug`.

franciscolopezsancho: I see. I understand it better now 👍 
One less line of code sounds like a plan to me.

efgpinto (approved):
LGTM

octonato:
Added one more commit in which I moved most of the logic to `coreSdk` d152ad2f634767b381f1feaa17408621dfd8083f.

I also gave up the idea of compiling coreSdk for 2.12 and 2.13. Too much trouble and impact. We will need to start to run `sbt +compile` to get a 2.12 variant to be used in tooling.

The easiest solution was to keep `coreSdk` a Java-only artefact.

build.sbt
```diff
@@ -381,6 +348,42 @@ def githubUrl(v: String): String = {
   "https://github.com/lightbend/kalix-jvm-sdk/tree/" + branch
 }
 
+lazy val devTools = project
```
octonato, line 351: no diff here, I just moved this block down in the file

octonato (commented):
I will make a follow-up PR where I will add a sample to demonstrate the usage. 

Not yet the real documentation as there are more to come, like support in Java SDK and S2S in the Proxy.

efgpinto (approved):
LGTM.

One questions though: do we plan to keep both `KalixProxyContainer`? Or can/should they be merged?

octonato:
> One questions though: do we plan to keep both KalixProxyContainer? Or can/should they be merged?

Yeah, I have been keeping it silent to see if someone would notice. 😄 

They are not quite the same, although the new one started as a copy of the Testkit. I didn't want to have something that was re-useable in both cases because that would slow me down. But I don't exclude a merge later one. We can postpone it for a while until devtools become more stable.

efgpinto:
yeah ok, it will need to be different anyway if we decide to ditch testcontainers and go to the java docker API directly on the plugin. I see.

PR #1583, opened 2023-04-20
feat: local discovery for WebClient

octonato:
Follow up of #1580 

Adds support for local discovery for the Spring WebClient


efgpinto (approved):
Nice, LGTM!

PR #1586, opened 2023-04-24
feat: eventing out integration testkit for protocol-first

efgpinto:
References https://github.com/lightbend/kalix/issues/7883

~Still in draft (API to be discussed/evolved and some other refactor/fixes) but working.~

For this PR, focusing only in eventing out for protobuf sdks:

- [x] config `grpc-backend` to boot up on testkits and use random port
- [x] eventing out testing in java/pb sdk
  - [x] sample usage for java/pb sdk
- [x] eventing out testing in scala/pb sdk 
  - [x] sample usage for scala/pb sdk


sdk/java-sdk-protobuf-testkit/src/main/scala/kalix/javasdk/testkit/impl/EventingTestKitImpl.scala
```diff
@@ -0,0 +1,179 @@
...
+  private val log = LoggerFactory.getLogger(classOf[EventingTestKit])
+  private implicit val sys = system
+  private implicit val ec = sys.dispatcher
+
+  private val eventDestinationProbe = new ConcurrentHashMap[EventDestination, TestProbe]()
+
+  override def getTopic(topic: String): Topic = {
+    val dest = EventDestination.defaultInstance.withTopic(topic)
+    eventDestinationProbe.putIfAbsent(dest, TestProbe())
+    val probe = eventDestinationProbe.get(dest)
```
aludwiko, line 129: I think that this line could be replaced with `val probe = eventDestinationProbe.putIfAbsent(dest, TestProbe())`

sdk/java-sdk-protobuf-testkit/src/main/scala/kalix/javasdk/testkit/impl/EventingTestKitImpl.scala
```diff
@@ -0,0 +1,179 @@
...
+    new TopicImpl(probe)
+  }
+
+  final class ServiceImpl extends EventingTestKitService {
+    override def emitSingle(in: EmitSingleCommand): Future[EmitSingleResult] = {
+      log.info("emitSingle request: [{}]", in)
+
+      in.destination.foreach(dest => {
+        eventDestinationProbe.putIfAbsent(dest, TestProbe())
+        eventDestinationProbe.get(dest).ref ! in
```
aludwiko, line 139: same here

sdk/java-sdk-protobuf-testkit/src/main/scala/kalix/javasdk/testkit/impl/EventingTestKitImpl.scala
```diff
@@ -0,0 +1,179 @@
...
+    val msg = probe.expectMsgType[EmitSingleCommand]
+
+    val metadata = new MetadataImpl(msg.getMessage.metadata.getOrElse(Metadata.defaultInstance).entries)
+    new TestKitMessageImpl(msg.getMessage.payload, metadata)
+  }
+
+  @ApiMayChange
+  override def expectAll(): util.Collection[TestKitMessage[ByteString]] = {
+    probe
+      .receiveWhile(3 seconds) { case m: EmitSingleCommand =>
```
aludwiko, line 166: this one will be tricky to use, how about exposing similar semantics to be clear that about the outcome?
```
override def receiveWhile(duration): util.Collection[TestKitMessage[ByteString]] = {
```

sdk/java-sdk-protobuf-testkit/src/main/scala/kalix/javasdk/testkit/impl/EventingTestKitImpl.scala
```diff
@@ -0,0 +1,179 @@
...
+        val metadata = new MetadataImpl(m.getMessage.metadata.getOrElse(Metadata.defaultInstance).entries)
+        new TestKitMessageImpl(m.getMessage.payload, metadata)
```
aludwiko, line 168: This probably could be extracted and reused in both methods.

aludwiko (commented):
minor comments, I wonder how this works on the proxy side :)

samples/java-protobuf-customer-registry-kafka-quickstart/src/it/java/customer/api/CustomerActionIntegrationTest.java
```diff
@@ -0,0 +1,96 @@
...
+
+    var outTopic = testKit.getTopic("customer_changes");
+    // wait for action to publish the change of state
+    var createdMsgOut = outTopic.expectNext();
+    var metadata = createdMsgOut.getMetadata();
+
+    assertEquals(Optional.of("customer.api.Customer"), metadata.get("ce-type"));
+    assertEquals(Optional.of("application/protobuf"), metadata.get("Content-Type"));
+    var publishedCustomer = CustomerApi.Customer.parseFrom(createdMsgOut.getPayload().toByteArray());
+    assertEquals(noAddress(toCreate), publishedCustomer);
```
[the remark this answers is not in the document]
octonato: The action is clearing the address. I think the goal was to confirm that what is published on the topic was first delivered (and transformed) by the action.

sdk/java-sdk-protobuf-testkit/src/main/java/kalix/javasdk/testkit/EventingTestKit.java
```diff
@@ -0,0 +1,45 @@
...
+
+  String getHost();
+  Integer getPort();
+
+  Topic getTopic(String topic);
+
+  interface Topic {
+    Message<ByteString> expectNext();
+
+    Collection<Message<ByteString>> expectAll();
```
octonato, line 36: It will be nice if we have the means to deserialize it automatically, something like: `Foo foo = expectNext(Foo.class)`

If we want to take that path, we might prefer to call this method: `expectNextMessage
 and `expectAllMessages` to signal that is the one returning the raw message.

octonato (commented):
Nice!

sdk/java-sdk-protobuf-testkit/src/main/scala/kalix/javasdk/testkit/impl/EventingTestKitImpl.scala
```diff
@@ -0,0 +1,179 @@
...
+    val actualPort = binding.localAddress.getPort
+    system.log.info("Eventing testkit grpc-backend started at {}:{}", host, actualPort)
+    service.port = actualPort
+    service
+  }
+
+}
+
+/**
+ * Implements the EventingTestKit protocol defined in protocols/testkit/src/main/protobuf/eventing_test_backend.proto
```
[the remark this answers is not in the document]
efgpinto: I haven't opened the PR there.. but the fact is that protocol already exists in proxy as well as the `grpc-backend`.. so the only change I'm doing there actually is to make the protocol be published to public (it is currently used only to test the proxy)

sdk/java-sdk-protobuf-testkit/src/main/scala/kalix/javasdk/testkit/impl/EventingTestKitImpl.scala
efgpinto: Yeah, at least expose the timeout for sure. I'm not quite settled on the API namings. In any case, `expectAll` is not great indeed.. Let me do another round of clean up and experimentation and see where we land..

sdk/java-sdk-protobuf-testkit/src/main/java/kalix/javasdk/testkit/KalixProxyContainer.java
```diff
@@ -35,8 +35,8 @@ public class KalixProxyContainer extends GenericContainer<KalixProxyContainer> {
   /** Default user function port (8080). */
   public static final int DEFAULT_USER_FUNCTION_PORT = 8080;
 
-  /** Default local port where the Google Pub/Sub emulator is available (8085). */
-  public static final int DEFAULT_GOOGLE_PUBSUB_PORT = 8085;
+  /** Default local port to use for eventing. */
+  public static final int DEFAULT_EVENTING_PORT = 8999;
```
efgpinto, line 39: I think defaulting to the internal grpc backend is the right thing to do.. but this change breaks backwards compatibility. Users currently relying on the defaults to use pub/sub externally will now need to either pass config or change the tests to use the new API. Thoughts?

sdk/java-sdk-protobuf-testkit/src/main/scala/kalix/javasdk/testkit/impl/EventingTestKitImpl.scala
```diff
@@ -0,0 +1,178 @@
...
+ */
+private[testkit] final class EventingTestServiceImpl(system: ActorSystem, val host: String, var port: Int)
+    extends EventingTestKit {
+
+  private val log = LoggerFactory.getLogger(classOf[EventingTestKit])
+  private implicit val sys = system
+  private implicit val ec = sys.dispatcher
+
+  private val eventDestinationProbe = new ConcurrentHashMap[EventDestination, TestProbe]()
```
efgpinto, line 121: I'm not sure using a `TestProbe` is the best for the long term... the fact that we cannot read the same message twice means users need to handle that on their own if needed.. In any case, this is external and we are free to change at any time.

efgpinto (commented):
Good a new round. I think I would probably like to refine into an acceptable state rather than adding more in this PR.

samples/java-protobuf-customer-registry-kafka-quickstart/src/main/java/customer/action/CustomerStateSubscriptionAction.java
```diff
@@ -11,24 +13,23 @@
...
+    // not populating address for public consumption
    CustomerApi.Customer customer = CustomerApi.Customer.newBuilder()
             .setCustomerId(customerState.getCustomerId())
             .setEmail(customerState.getEmail())
             .setName(customerState.getName())
-            .setAddress(address)
             .build();
 
 
+   logger.info("Publishing public customer state out: {}", customer);
```
johanandren, line 32: Indentation seems off

sdk/java-sdk-protobuf-testkit/src/main/scala/kalix/javasdk/testkit/impl/EventingTestKitImpl.scala
```diff
@@ -0,0 +1,178 @@
...
+        probe.ref ! in
+      })
+
+      if (in.destination.isEmpty) {
+        log.warn("Received a message without destination, ignoring. {}", in)
+      }
+      Future.successful(EmitSingleResult())
+    }
+
+    override def runSource(in: Source[RunSourceCommand, NotUsed]): Source[SourceElem, NotUsed] = ???
```
johanandren, line 144: FIXMEs with links to issues to support streaming in/out, or do we intend to support these right away? Probably also good with a better exception saying it is not yet supported, because it would be triggered by descriptor, right?

sdk/scala-sdk-protobuf-testkit/src/main/scala/kalix/scalasdk/testkit/EventingTestKit.scala
```diff
@@ -0,0 +1,57 @@
...
+
+  /**
+   * Waits for a default amount of time before returning all unread messages in the topic. If no message is received, a
+   * timeout exception is thrown.
+   *
+   * @return
+   *   collection of messages, each message including ByteString payload and metadata
+   */
+  def expectAll(): Seq[Message[ByteString]]
+}
```
johanandren, line 49: I think something like the typed test probe `receiveMessages(n, atMost)` is often very useful, because you know what number of events should be produced by some trigger and want to look at all of them.

Also, only providing byte string seems very low level, could we also have something where you pass the parser companion (I think there is a static parser type generated in Java as well) and get a parsed expected proto message, like the Akka `expectMessageType`

johanandren (commented):
Looks promising, added a few suggestions

sdk/scala-sdk-protobuf-testkit/src/main/scala/kalix/scalasdk/testkit/EventingTestKit.scala
johanandren: We do something like the typed parsing here: https://github.com/akka/akka-grpc/pull/1740/files#diff-559fbacacb811f269fb8a1182343f7bc82f934ad952bd0a759a65213d3a2f496R246

johanandren: I must have remembered the Java generated classes part wrong though, no interface there, only static `parseFrom(...)` methods

efgpinto: > I think something like the typed test probe `receiveMessages(n, atMost)` is often very useful, because you know what number of events should be produced by some trigger and want to look at all of them.

Yeah, I was trying to keep the API at the minimum while trying this out on samples and seeing what is useful but I agree that one is often very useful. Will add it.

> Also, only providing byte string seems very low level, could we also have something where you pass the parser companion (I think there is a static parser type generated in Java as well) and get a parsed expected proto message, like the Akka expectMessageType

Agree as well. I didn't see a way on how to do it immediately for java without code gen and assumed it was the same for scala.. but good to know it's possible in scala. thanks 👍

johanandren: Maybe a reasonable solution for Java would be a `<T> receiveMessageType(java.util.function.Function<ByteString, T>)` so that users can do something like `receiveMessageType(MyMessage::parseFrom)` with method references? (Didn't try it so maybe the overloads makes that impossible)

johanandren: No, I found it, all generated classes implement `com.google.protobuf.GeneratedMessageV3` which has a `getParserForType` 👍

octonato (commented):
some more comments here and there

sdk/java-sdk-protobuf-testkit/src/main/scala/kalix/javasdk/testkit/impl/EventingTestKitImpl.scala
efgpinto: yes, exactly. edited, tks.

samples/scala-protobuf-eventsourced-counter/src/test/scala/com/example/CounterServiceIntegrationSpec.scala
```diff
@@ -26,15 +30,37 @@ class CounterServiceIntegrationSpec extends AnyWordSpec with Matchers with Befor
...
+      val counter = client.getCurrentCounter(GetCounter(counterId)).futureValue
+      counter.value shouldBe (10 + 10 * 2)
+
+      // verify messages published to topic
+      val allMsgs = testKit.getTopic("counter-events").expectN(2)
 
-    "handle side effect that adds the initial input multiplied by two" in {
-      val counterId = "xyz"
+      val Seq(Message(payload1, md1), Message(payload2, md2)) = allMsgs
+      Increased.parseFrom(payload1.toByteArray) shouldBe Increased(10)
```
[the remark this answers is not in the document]
johanandren: That would be getting our hands the `AnySupport` and `decodeMessage` which accepts a protobuf `Any`, so no type needed for the actual decode except the string describing the payload in `Any.type_url`

sdk/java-sdk-protobuf-testkit/src/main/java/kalix/javasdk/testkit/EventingTestKit.java
```diff
@@ -63,44 +96,56 @@ interface Topic {
...
 
     /**
      * Waits for a default amount of time before returning all unread messages in the topic.
      * If no message is received, a timeout exception is thrown.
      *
-     * @return collection of messages, each message including ByteString payload and metadata
+     * @return collection of messages, each message including the deserialized payload object and metadata
      */
-    Collection<Message<ByteString>> expectN();
+    Collection<Message<Object>> expectN();
```
johanandren, line 128: Could it be `Message<?>` because we don't know what types they are? (I don't completely remember the semantics of Java wildcards), and `Message[_]` for the Scala ones

johanandren (commented):
Would be nice to also have some form of convenience on `Message` for when the type is unknown, to be able to do something like this in a test case:

```
val msgs = topic.expectN()
msgs should have size (3)
val head = msgs.head
val concrete = head.expectType[Concrete]
concrete.something should be === (4)
```

samples/scala-protobuf-eventsourced-counter/src/test/scala/com/example/CounterServiceIntegrationSpec.scala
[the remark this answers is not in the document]
efgpinto: added in https://github.com/lightbend/kalix-jvm-sdk/pull/1586/commits/16657c9337192685494b4052ba5e3bc55e35b1a5

sdk/java-sdk-protobuf-testkit/src/main/java/kalix/javasdk/testkit/EventingTestKit.java
efgpinto: fixed now in https://github.com/lightbend/kalix-jvm-sdk/pull/1586/commits/e8deab4b95871ed04fd9c1a76cdac0deca1ee3f9

sdk/java-sdk-protobuf-testkit/src/main/java/kalix/javasdk/testkit/EventingTestKit.java
efgpinto: We have typed methods now in https://github.com/lightbend/kalix-jvm-sdk/pull/1586/commits/16657c9337192685494b4052ba5e3bc55e35b1a5

sdk/scala-sdk-protobuf-testkit/src/main/scala/kalix/scalasdk/testkit/EventingTestKit.scala
efgpinto: this is added now

sdk/java-sdk-protobuf-testkit/src/main/scala/kalix/javasdk/testkit/impl/EventingTestKitImpl.scala
efgpinto: I added an exception below, this one is already changed on the other PR for eventing in.

efgpinto (commented):
I think this is now ready for final review I would say.

One outstanding question: to use case classes or not on the scala sdk `Message` (see discussion below).

samples/java-protobuf-customer-registry-kafka-quickstart/src/it/java/customer/api/CustomerActionIntegrationTest.java
```diff
@@ -0,0 +1,98 @@
...
+
+  public CustomerActionIntegrationTest() {
+    client = testKit.getGrpcClient(CustomerService.class);
+  }
+
+  @Test
+  public void createAndPublish() throws Exception {
+    var id = UUID.randomUUID().toString();
+    var toCreate = customer(id, "Johanna", "foo@example.com", "Porto", "Long Road");
+    createCustomer(toCreate);
```
aludwiko, line 48: minor thing but I would rename it a bit:
```
var customer = buildCustomer(...)
createCustomer(customer)
```

sdk/java-sdk-protobuf-testkit/src/main/java/kalix/javasdk/testkit/EventingTestKit.java
```diff
@@ -0,0 +1,168 @@
...
+  Topic getTopic(String topic);
+
+  /**
+   * Testkit utility to mock broker's topic. Useful when doing integration tests for services that do eventing (in or out) to a broker's topic.
+   */
+  @ApiMayChange
+  interface Topic {
+
+    /**
+     * Waits for predefined amount of time. If a message arrives in the meantime or
```
aludwiko, line 52: Maybe some information about the default timeout value, which could be a doc-link to the constant?

samples/java-protobuf-customer-registry-kafka-quickstart/src/it/java/customer/api/CustomerActionIntegrationTest.java
```diff
@@ -0,0 +1,98 @@
...
+
+    var outTopic = testKit.getTopic("customer_changes");
```
aludwiko, line 50: could this be a test class field?

samples/scala-protobuf-eventsourced-counter/src/test/scala/com/example/CounterServiceIntegrationSpec.scala
```diff
@@ -26,15 +27,38 @@ class CounterServiceIntegrationSpec extends AnyWordSpec with Matchers with Befor
...
+    val counterId = "xyz"
+
+    "handle side effect that adds the initial input multiplied by two and verify publishing" in {
+
+      client.increaseWithSideEffect(IncreaseValue(counterId, 10)).futureValue
+      val counter = client.getCurrentCounter(GetCounter(counterId)).futureValue
+      counter.value shouldBe (10 + 10 * 2)
+
+      // verify messages published to topic
+      val allMsgs = testKit.getTopic("counter-events").expectN(2)
```
aludwiko, line 39: similar here, could be a class field

sdk/java-sdk-protobuf-testkit/src/main/java/kalix/javasdk/testkit/EventingTestKit.java
```diff
@@ -0,0 +1,168 @@
...
+
+    /**
+     * Waits and returns the next unread message on this topic and automatically parses
+     * and casts it to the specified given type.
+     *
+     * @param instance class type to cast the received message bytes to
+     * @return a Message of type T
+     * @param <T> a given domain type
+     */
+    <T extends GeneratedMessageV3> Message<T> expectMessageType(Class<T> instance);
```
aludwiko, line 109: Naming is a bit inconsistent:
```
 - expectOne
 - expectMessageType
 ```
 my proposition:
 ```
  - expectMsg
 - expectMsgType
```
I'm not sure what to do with `expectN` because it's more about the time and not the number of msgs. @octonato suggest sth like `expectWhile` which explains better the intention. Could be renamed to `expectMsgsWhile`

sdk/java-sdk-protobuf-testkit/src/main/scala/kalix/javasdk/testkit/impl/EventingTestKitImpl.scala
```diff
@@ -0,0 +1,208 @@
...
+
+  override def expectMessageType[T <: GeneratedMessageV3](clazz: Class[T]): TestKitMessage[T] =
+    expectMessageType(clazz, DefaultTimeout)
+
+  override def expectMessageType[T <: GeneratedMessageV3](
+      clazz: Class[T],
+      timeout: time.Duration): TestKitMessage[T] = {
+    val msg = probe.expectMsgType[EmitSingleCommand]
+    val metadata = new MetadataImpl(msg.getMessage.getMetadata.entries)
+    val decodedMsg = codec.decodeMessage(ScalaPbAny(metadata.get("ce-type").orElse(""), msg.getMessage.payload))
```
aludwiko, line 156: What would happen with an empty type? Would it fail or sth?

sdk/java-sdk-protobuf-testkit/src/main/java/kalix/javasdk/testkit/EventingTestKit.java
johanandren: In Akka typed testkit we call it `receiveN` for a number and `receive` (because probe already typed) that doesn't really assert any properties of the message.

efgpinto: Glad you opened this discussion because I were to open it myself and forgot.

I think another option could be:
```
 - expectOne
 - expectOneTyped
 - expectN
 ```

> I'm not sure what to do with `expectN` because it's more about the time and not the number of msgs.

Not sure I agree here. I think it's as much about the number as about the time, depending on what kind of args you pass in. By default, the rational was to focus on the number of messages and use a big timeout.

Comment (aludwiko) [reply sdk/java-sdk-protobuf-testkit/src/main/java/kalix/javasdk/testkit/EventingTestKit.java:109]: > I'm not sure what to do with expectN because it's more about the time and not the number of msgs.

My bad, correct, it's about both.

Comment (aludwiko) [reply sdk/java-sdk-protobuf-testkit/src/main/java/kalix/javasdk/testkit/EventingTestKit.java:109]: I'm fine with any convention as long as it is consistent, my suggestion was based on untyped `akka.testkit.TestProbe`

File: sdk/java-sdk-protobuf-testkit/src/main/scala/kalix/javasdk/testkit/impl/EventingTestKitImpl.scala (hunk 20 above)
Comment (efgpinto) [reply sdk/java-sdk-protobuf-testkit/src/main/scala/kalix/javasdk/testkit/impl/EventingTestKitImpl.scala:156]: yeah, it fails.. I think I will add a TopicImplSpec to document/test such and other scenarios.

File: sdk/scala-sdk-protobuf-testkit/src/main/scala/kalix/scalasdk/testkit/EventingTestKit.scala [hunk 21]
```diff
@@ -0,0 +1,174 @@
...
+  def expectN(total: Int, timeout: FiniteDuration): Seq[Message[_]]
+}
+
+@InternalApi
+object Topic {
+  def apply(delegate: JEventingTestKit.Topic, codec: MessageCodec): Topic = TopicImpl(delegate, codec)
+}
+
+@ApiMayChange
+case class Message[P](payload: P, metadata: Metadata) {
```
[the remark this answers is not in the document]
johanandren: I think it is fine with a case class, but it should be final for good measure.

sdk/java-sdk-protobuf-testkit/src/main/java/kalix/javasdk/testkit/EventingTestKit.java
efgpinto: updated now in https://github.com/lightbend/kalix-jvm-sdk/pull/1586/commits/c053feecb06f86a4a0a9699d429da65d9dbb3a69

sdk/java-sdk-protobuf-testkit/src/main/java/kalix/javasdk/testkit/KalixTestKit.java
```diff
@@ -87,7 +93,22 @@ public static class Settings {
...
+       * running.
+       */
+      GRPC_BACKEND,
+
+      /**
+       * Used if you want to use an external Google PubSub Emulator on your tests.
+       *
+       * Note: the Google PubSub Emulator need to be started independently.
+       */
+      GOOGLE_PUBSUB_EMULATOR
```
octonato, line 111: What can I do if I'm using Kafka? 

Maybe we should use different names here. Something like: `TEST_BROKER` and `EXTERNAL_BROKER`.

Also, for the users, the fact is a GRPC_BACKEND is not relevant. That's an implementation detail on our side.

octonato: Now, we can go one step further and decide that the testkit always uses a test broker and let users use the `DockerIntegrationTest` that Andrzej added recently for tests using external brokers. 

Anyway, just an idea for later. `DockerIntegrationTest` is not yet open for general use.

octonato (commented):
Looking good. I have a few suggestions though.

sdk/java-sdk-protobuf-testkit/src/main/java/kalix/javasdk/testkit/KalixTestKit.java
efgpinto: The idea here was to allow clients that were using the googlepubsub in integration tests to still have the option to continue using it, even if now they need to pass in that config. Kafka was not supported for integration tests AFAIK.. or was it? My idea was if we wanted to support Kafka we would add a new option there and add the defaults and way to override the port by config, etc. But that would be a different issue.

Agree on `GRPC_BACKEND`, `TEST_BROKER` seems like a good one. 👍

sdk/java-sdk-protobuf/src/main/java/kalix/javasdk/Kalix.java
```diff
@@ -635,6 +636,16 @@ public KalixRunner createRunner(Config config) {
   }
 
   private AnySupport newAnySupport(Descriptors.FileDescriptor[] descriptors) {
+    allDescriptors.addAll(Arrays.asList(descriptors));
```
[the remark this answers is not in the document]
efgpinto: > It's a little annoying that we are collecting all descriptors in `allDescriptors` like that.

yeah, agree. felt like the cleanest way though. Would you do it in a different way? Adding context through comments makes sense, will update.

sdk/java-sdk-protobuf-testkit/src/main/java/kalix/javasdk/testkit/KalixTestKit.java
octonato: Correct, testing with external Kafka is not supported.
If we move further with `DockerIntegrationTest` idea, we may decide that Kafka tests should rely on docker-compose instead.

sdk/java-sdk-protobuf-testkit/src/main/java/kalix/javasdk/testkit/EventingTestKit.java
```diff
@@ -0,0 +1,176 @@
...
+
+    /**
+     * Waits and returns the next unread message on this topic and automatically parses
+     * and casts it to the specified given type.
+     *
+     * @param instance class type to cast the received message bytes to
+     * @return a Message of type T
+     * @param <T> a given domain type
+     */
+    <T extends GeneratedMessageV3> Message<T> expectOneTyped(Class<T> instance);
```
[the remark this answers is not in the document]
johanandren: For protobuf sdks that will differ from how they deal with JSON in actual code, so when we discussed this earlier we said it should reflect the manual JsonSupport decoding instead of doing magic here.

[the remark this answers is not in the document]
efgpinto: I opened this side PR https://github.com/lightbend/kalix-jvm-sdk/pull/1663 so we can discuss further and maybe have an idea of how this would work for java sdk. Please have a look 🙇

octonato (approved):
LGTM

Let's merge it as is and continue on follow-up PRs

PR #1592, opened 2023-04-25
feat: sbt runAll to start proxy and UF

octonato:


sbt-plugin/src/main/scala/kalix/sbt/KalixPlugin.scala
```diff
@@ -52,88 +59,99 @@ object KalixPlugin extends AutoPlugin {
...
-    Compile / PB.generate := (Compile / PB.generate)
-      .dependsOn(Def.task {
-        protobufDescriptorSetOut.value.getParentFile.mkdirs()
+  override def projectSettings = {
+    addCommandAlias("runAll", "runDockerCompose;run") ++
+    Seq(
+      dockerComposeFile := "docker-compose.yml",
+      logConfig := "logback-dev-mode.xml",
+      userFunctionPort := 8080,
+      runDockerCompose := dockerComposeUp(dockerComposeFile.value),
```
octonato, line 68: This is the real change. The rest is formatting.

sbt-plugin/src/main/scala/kalix/sbt/KalixPlugin.scala
```diff
@@ -52,88 +59,99 @@ object KalixPlugin extends AutoPlugin {
...
+      libraryDependencies ++= Seq(
+        "io.kalix" % "kalix-sdk-protocol" % KalixProtocolVersion % "protobuf-src",
+        "com.google.protobuf" % "protobuf-java" % "3.17.3" % "protobuf",
+        "io.kalix" %% "kalix-scala-sdk-protobuf-testkit" % KalixSdkVersion % Test),
+      run / javaOptions ++= Seq(
+        "-Dkalix.user-function-interface=0.0.0.0",
+        s"-Dlogback.configurationFile=${logConfig.value}",
+        s"-Dkalix.user-function-port=${userFunctionPort.value}"),
```
octonato, line 76: This is also a change. Previously we were passing it directly in the build.sbt file. Not needed and can we use sbt settings instead.

sbt-plugin/src/main/scala/kalix/sbt/KalixPlugin.scala
```diff
@@ -153,4 +171,20 @@ object KalixPlugin extends AutoPlugin {
...
       .toVector
   }
+
+  private def dockerComposeUp(file: String): Unit = {
+    import scala.sys.process._
+    import scala.concurrent.ExecutionContext.Implicits.global
+
+    // TODO: check if file exists
+    Future {
+      s"docker-compose -f $file up".!
```
octonato, line 181: It turns out that calling `docker-compose` directly makes it much simpler as we don't need to wrap anything with testcontainers. 

If we decide to take that directly, I will refactor the `mvn kalix:run` to do exactly the same.

aludwiko: Is the `docker-compose` available when installing docker or is this sth that must be installed separately?

efgpinto: I also have the same question as Andrzej. I _think_ you need to install the plugin but I'm not sure if that's only the case for compose V2. That's also another question, if we need to worry about [V1 vs V2](https://docs.docker.com/compose/compose-v2/) since we have been relying on V1 but until now that was clear (the user had to go through the README and explicitely do `docker-compose up` so they could easily switch to `docker compose` if that's their case.

With this, I think we might need to check for the availability of "docker-compose" or "docker compose" and warn the user when that is not found since the use of docker-compose is much more hidden.

octonato: yep, I would also go for checking what is installed and then warning. 

In any case, users need to have one or the other anyway. If they don't have and can't install it, they won't be able to run `sbt runAll` or `mvn kalix:runAll`. Actually, they won't be able to run their application locally anyway.

octonato: For the record, I removed all docker binaries from an old machine and re-installed it. `docker compose` is just included in the new installation. Also `docker-compose` or maybe it works just like with git. If there is a bin named `docker-compose` you can call it without the `-`.

octonato: some more investigations...

I have two sym-link files on my machines: `docker-compose` and `docker-compose-v1`. 

When checking the version, I get:

```shell
!❯ docker-compose-v1 --version
docker-compose version 1.29.2, build 5becea4c
```

```shell
=❯ docker-compose --version
Docker Compose version v2.12.2
```

```shell
=❯ docker compose version
Docker Compose version v2.12.2
```

I would assume that `docker-compose` is the one we should use. Doesn't matter if v1 or v2, it will just work. 


We should however, emit a warning if command fails and point user to the right place.

ennru: https://docs.docker.com/compose/

"From the end of June 2023 Compose V1 won’t be supported anymore and will be removed from all Docker Desktop versions.

Make sure you switch to [Compose V2](https://docs.docker.com/compose/compose-file/) with the docker compose CLI plugin or by activating the Use Docker Compose V2 setting in Docker Desktop. For more information, see the [Evolution of Compose](https://docs.docker.com/compose/compose-v2/)."

octonato: Sure, but the way that I understand it is that `docker-compose` works and will still work after June 2023, because `docker-compose` is an alias to `docker compose`. And that's V2. 


If we use `docker-compose` it works with V1 and with V2.

devtools/src/main/scala/kalix/devtools/impl/DockerComposeUtils.scala
```diff
@@ -0,0 +1,56 @@
...
+ */
+
+package kalix.devtools.impl
+
+import java.nio.file.Files
+import java.nio.file.Paths
+
+import scala.sys.process._
+
+object DockerComposeUtils {
```
octonato, line 24: Moved logic to this utility class and in dev-tools in order to use it from inside the maven plugin as well.

octonato (commented):
This is ready for review. 

In follow-up PR, I will add more checks to verify that `docker-compose` is on the path find a way to capture any error and warn users properly. 

But first I will change the maven plugin to use `docker-compose` as well.

efgpinto (approved):
LGTM

PR #1596, opened 2023-04-26
feat: pass service port mappings to KalixTestKit

octonato:
This is needed to make the ubiquitous shopping cart work. This app will listen to itself through s2s and therefore we should be able to configure the proxy correctly from inside the KalixTestKit.

### DIFF
--- a/sdk/java-sdk-protobuf-testkit/src/main/java/kalix/javasdk/testkit/KalixTestKit.java
+++ b/sdk/java-sdk-protobuf-testkit/src/main/java/kalix/javasdk/testkit/KalixTestKit.java
@@ -36,13 +36,12 @@ import org.slf4j.LoggerFactory;
 import java.io.IOException;
 import java.net.ServerSocket;
 import java.time.Duration;
-import java.util.HashMap;
-import java.util.Map;
-import java.util.Optional;
+import java.util.*;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CompletionStage;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeUnit;
+import java.util.stream.Collectors;
 
 /**
  * Testkit for running Kalix services locally.
@@ -77,6 +76,9 @@ public class KalixTestKit {
     /** To override workflow tick interval for integration tests */
     public final Optional<Duration> workflowTickInterval;
 
+    /** Service port mappings from serviceName to host:port */
+    public final Map<String, String> servicePortMappings;
+
     /**
      * Create new settings for KalixTestkit.
      *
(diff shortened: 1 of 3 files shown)


sdk/java-sdk-protobuf-testkit/src/main/java/kalix/javasdk/testkit/KalixTestKit.java
```diff
@@ -252,7 +274,15 @@ private void runProxy(Boolean useTestContainers, int port) {
...
       proxyContainer.addEnv("ACL_ENABLED", Boolean.toString(settings.aclEnabled));
       proxyContainer.addEnv("VIEW_FEATURES_ALL", Boolean.toString(settings.advancedViews));
-      proxyContainer.addEnv("JAVA_TOOL_OPTIONS", "-Dlogback.configurationFile=logback-dev-mode.xml");
+
+      List<String> javaOptions = new ArrayList<>();
+      javaOptions.add("-Dlogback.configurationFile=logback-dev-mode.xml");
+
+      settings.servicePortMappings.forEach((serviceName, hostPort) -> {
+        javaOptions.add("-Dkalix.dev-mode.service-port-mappings." + serviceName + "=" + hostPort);
+      });
```
octonato, line 283: @efgpinto, this will conflict with your open PR. Make sure to rebase when this one gets merged. 

As the number of `-D` is growing, better if we use a list and then join it.

franciscolopezsancho (approved):
LGTM as far as I can tell. 
I left a comment

octonato:
> I left a comment

@franciscolopezsancho, I don't see it. Maybe you forgot to push the button?

sdk/java-sdk-protobuf-testkit/src/main/java/kalix/javasdk/testkit/KalixTestKit.java
```diff
@@ -156,18 +160,33 @@ public Settings withAdvancedViews() {
...
     public Settings withWorkflowTickInterval(Duration tickInterval) {
-      return new Settings(stopTimeout, serviceName, aclEnabled, true, Optional.of(tickInterval));
+      return new Settings(stopTimeout, serviceName, aclEnabled, true, Optional.of(tickInterval), servicePortMappings);
+    }
+
+    /**
+     * Add a service port mapping from serviceName to host:port.
+     * @return The updated settings.
+     */
+    public Settings withServicePortMapping(String serviceName, String hostAndPort) {
```
franciscolopezsancho, line 170: hostAndPort looks it can leave room to error if you don't use `:` between the `host` and the `port`. I bet you considered two inputs, host and port. Why did you prefer `host` and `port` together?

octonato: Correct. I might be better indeed. 

The reason I chose `host:port` is because we pass it as such in other places as well. 

For instance, if configuring it directly in docker-compose file, 
we need to pass: `-Dkalix.dev-mode.service-port-mappings.my-service-name=host.docker.internal:port`. When passing in one `-D` is convenient.

When configuring it in the maven plugin, we can pass just the port, and then we map it automatically to `localhost:port` or we can pass `somehost:port`. 

But here it might be better indeed to split it since it's always configurable in code.

octonato: @franciscolopezsancho, done in e5200c0

PR #1597, opened 2023-04-27
feat: optional transform_updates flag

aludwiko:
SDK part of https://github.com/lightbend/kalix/issues/7956


docs/src/modules/java-protobuf/pages/views.adoc
```diff
@@ -92,7 +92,7 @@ include::example$scala-protobuf-valueentity-customer-registry/src/main/proto/cus
...
 <4> The `(kalix.method).view.update` annotation defines that this method is used for updating the View. You must define the `table` attribute for the table to be used in the query. Pick any name and use it in the query `SELECT` statement.
-<5> The second `(kalix.method).eventing.in` annotation with `handle_deletes: true` and (required) `google.protobuf.Empty` as a parameter defines that this method is used for handling Value Entity deletes. Kalix will automatically delete the View state, but you can also write your own delete handler when `transform_updates` flag is `true`. If you don't provide such method then, by default, your View state will not be deleted.
+<5> The second `(kalix.method).eventing.in` annotation with `handle_deletes: true` and (required) `google.protobuf.Empty` as a parameter defines that this method is used for handling Value Entity deletes. Kalix will automatically delete the View state, but you can also write your own delete handler when `(kalix.method).view.update.transform_updates` flag is `true`. If you don't provide such method then, by default, your View state will not be deleted.
```
octonato, line 95: Now this is looking weird. 

This is about a delete method and then we use a flag called `transform_updates` to indicate that what we want is actually to get this method generated and called whenever a delete happens. 

I know, the text was already here. It's just that now that we are re-think the semantics of this flag, it becomes obvious that the naming is not great (not to say confusing).

octonato (commented):
Looking good. 

But maybe we should consider introducing another flag to better express the intention.

docs/src/modules/java-protobuf/pages/views.adoc
aludwiko: How about introducing this flag in the 4th point:
```
<4> The `(kalix.method).view.update` annotation defines that this method is used for updating the View. You must define the `table` attribute for the table to be used in the query. Pick any name and use it in the query `SELECT` statement. When a view model is the same as the method input parameter, Kalix will optimize the processing and won't call this method. However, this behavior can be overridden with the `(kalix.method).view.update.transform_updates: true` flag. 
```

octonato: I think the most intuitive would be to have `handle_updates` and `handle_deletes` to make it clear that the method is called when an update or a delete needs to be handled. 

But we are already using `handle_deletes` to indicate that we want Kalix to delete the view. 

Maybe something like: `call_on_handle`? Not sure if make anything more obvious. 

And indeed migration is one more problem to solve. We can keep both and have a note in docs and release page.

octonato (commented):
@aludwiko, sorry for the delay. This PR completely escaped my radar.

docs/src/modules/java-protobuf/pages/views.adoc
octonato: btw, I find it indeed better to move the flag explanation to the point <4>. But it also depends if we introduce an new flag or not.

octonato: We had a chat about this and decided to now rename/introduce a new flag for now. We will go with the existing one. We just need to find the correct wording about it.

I find the last suggestion to change point <4> already an improvement. 
On point <5>, we can eventually mention that if we want the method to be called before deletion, then we must mark it with `transform_updates`. And then we say something like: 

> Equally to update methods, Kalix will optimise it and not call the method if not necessary, unless we mark it with `transform_updates` in which case we indicate to Kalix that we want the method to be called ...

docs/src/modules/java-protobuf/pages/views.adoc
```diff
@@ -125,7 +125,7 @@ When no result is found, the request fails with gRPC status code `NOT_FOUND`. A
 
 The View state corresponding to an Entity is not automatically deleted when the Entity is deleted.
 
-When the source of the `eventing.in` is a Value Entity the `handle_deletes: true` annotation of `(kalix.method).eventing.in` defines that this method is used for handling Value Entity deletes. The method must have `google.protobuf.Empty` as a parameter. Kalix will automatically delete the View state, but you can also write your own delete handler when `transform_updates` flag is `true`. If you don't provide such method then, by default, your View state will not be deleted.
```
aludwiko, line 128: The last sentence was removed, because is a duplication of the previous paragraph.

docs/src/modules/java-protobuf/pages/views.adoc
aludwiko: Ok, please check now.

docs/src/modules/java-protobuf/pages/views.adoc
```diff
@@ -91,8 +91,8 @@ include::example$scala-protobuf-valueentity-customer-registry/src/main/proto/cus
...
+<4> The `(kalix.method).view.update` annotation defines that this method is used for updating the View. You must define the `table` attribute for the table to be used in the query. Pick any name and use it in the query `SELECT` statement. When a view model is the same as the method input parameter, Kalix will optimize the processing and won't call this method. However, this behavior can be overridden with the `(kalix.method).view.update.transform_updates` flag set to `true`.
+<5> The second `(kalix.method).eventing.in` annotation with `handle_deletes: true` and (required) `google.protobuf.Empty` as a parameter defines that this method is used for handling Value Entity deletes. Similar to update methods, Kalix will optimise it and not call the method if not necessary, unless you mark it with `(kalix.method).view.update.transform_updates: true`. In this case, you indicate to Kalix that you want to provide a custom delete handler implementation.
```
octonato, line 95: Perfect!

octonato (approved):
LGTM

Let's merge it!

PR #1602, opened 2023-04-28
feat: workflow timers support

aludwiko:


sdk/java-sdk-protobuf/src/main/java/kalix/javasdk/workflowentity/WorkflowEntity.java
```diff
@@ -42,6 +45,7 @@
 
 
   private Optional<CommandContext> commandContext = Optional.empty();
+  private Optional<WorkflowEntityContext> workflowEntityContext = Optional.empty();
```
aludwiko, line 48: Not sure if an additional context is a good way, but extending the current CommandContext won't do the trick, since I also want to use timers in step actions. For the step transition timers are blocked - the context is empty.

octonato: It seems that the only reason to have `WorkflowEntityContext` is to later be able to access the `MessageCodec` and the `ActorSystem` in other to build the timer. 

So maybe instead we should just have an `Option<TimerScheduler>` that we set when handing a command and running a step transition.

sdk/java-sdk-spring/src/it/java/com/example/wiring/workflowentities/WorkflowWithTimer.java
```diff
@@ -0,0 +1,83 @@
...
+
+  @Override
+  public Workflow<FailingCounterState> definition() {
+    var counterInc =
+        step(counterStepName)
+            .asyncCall(() -> {
+              var pingWorkflow = kalixClient.put("/workflow-with-timer/" + workflowEntityContext().entityId() + "/ping",
+                  new CounterScheduledValue(12),
+                  String.class);
+              return timers().startSingleTimer("ping", Duration.ofSeconds(2), pingWorkflow).thenApply(__ -> "ok");
```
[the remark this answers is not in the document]
octonato: Agree. `Done` is part of Kalix JVM SDK API (because of timers) and therefore can receive a special treatment.

octonato (commented):
Looking good. 

Dropped a suggestion about the next context. I think we can avoid it.

sdk/java-sdk-protobuf/src/main/java/kalix/javasdk/workflowentity/WorkflowEntity.java
octonato: I see, but users should use the creation context in that case. That should be the preferred way of getting a handle on the entity id.

sdk/java-sdk-spring/src/it/java/com/example/wiring/workflowentities/WorkflowWithTimer.java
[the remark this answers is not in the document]
aludwiko: I wonder what should be the actual serialized value of a Done class, I suppose `{}` makes sense, but maybe there are better options?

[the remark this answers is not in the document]
octonato: `{}` sounds good to me. Also because can be deserialized to any other compatible type eventually.

sdk/java-sdk-protobuf/src/main/java/kalix/javasdk/workflowentity/WorkflowEntity.java
octonato: I mean `WorkflowEntityContext`. Check `KalixSpringApplication` for example. 

Users can let it be injected into constructor when they need to access the `entityId`. Since available in constructor, it will be available in `definition`.

octonato: Wow, I just realised that the `WorkflowEntityContext` that you were using is the one that we already have. I thought it was called `WorkflowEntityCreationContext`. 

Anyway, `WorkflowEntityContext` is the creation context and is expected to be injected into constructor when we need the entityId. That aligns with what we have done in other entities as well.

aludwiko: Ok, now I get it.

aludwiko: :) Actually, we've never used it in the constructor, because there is a small bug in KalixSpringApplication, fixing it right now.

sdk/java-sdk-protobuf/src/main/scala/kalix/javasdk/impl/workflowentity/WorkflowEntityImpl.scala
```diff
@@ -303,7 +305,7 @@ final class WorkflowEntityImpl(system: ActorSystem, val services: Map[String, Wo
           Future.successful(toProtoEffect(effect, command.id))
 
         case Step(executeStep) =>
-          val workflowContext = new WorkflowEntityContextImpl(entityId, system)
```
aludwiko, line 306: it was used only to get the execution context

sdk/java-sdk-protobuf/src/main/scala/kalix/javasdk/impl/workflowentity/WorkflowEntityImpl.scala
```diff
@@ -313,7 +315,8 @@ final class WorkflowEntityImpl(system: ActorSystem, val services: Map[String, Wo
                 executeStep.input,
                 executeStep.stepName,
                 service.messageCodec,
-                workflowContext)
+                timerScheduler,
+                SystemMaterializer(system).materializer.executionContext)
```
aludwiko, line 319: I have a feeling that this might be simplified.

sdk/scala-sdk-protobuf/src/test/scala/kalix/scalasdk/JsonSupportSpec.scala
```diff
@@ -38,6 +38,12 @@ class JsonSupportSpec extends AnyWordSpec with Matchers {
       any.typeUrl should ===(JavaJsonSupport.KALIX_JSON + classOf[MyJsonable].getName)
       JsonSupport.decodeJson[MyJsonable](any).field should ===("foo")
     }
+    "serialize and deserialize Akka Done class" in {
+      val done = Done
+      val any = JsonSupport.encodeJson(done)
+      any.typeUrl should ===(JavaJsonSupport.KALIX_JSON + Done.getClass.getName)
```
aludwiko, line 44: can't use `classOf[MyJsonable].getName` because for the scala object the result is different than `Done.getClass.getName` (additional `$` at the end)

sdk/java-sdk-protobuf/src/main/java/kalix/javasdk/JsonSupport.java
```diff
@@ -55,6 +66,12 @@ public final class JsonSupport {
             JsonCreator.Mode.PROPERTIES));
     objectMapper.registerModule(new com.fasterxml.jackson.datatype.jdk8.Jdk8Module());
     objectMapper.registerModule(new com.fasterxml.jackson.datatype.jsr310.JavaTimeModule());
+
+    SimpleModule module = new SimpleModule();
```
aludwiko, line 70: not sure if this is the right place for Done handling

sdk/java-sdk-protobuf/src/main/scala/kalix/javasdk/impl/workflowentity/WorkflowEntityImpl.scala
octonato, line 319: Like this, for example? 

```suggestion
                system.dispatcher
```

sdk/java-sdk-protobuf/src/main/java/kalix/javasdk/JsonSupport.java
octonato: I can't image a better place for it.

octonato (approved):
LGTM

Do you plan to add docs on this PR or on follow-up?

PR #1604, opened 2023-05-02
feat: read port configs from docker yml

octonato:


devtools/src/main/scala/kalix/devtools/impl/DockerComposeUtils.scala
```diff
@@ -19,38 +19,62 @@ package kalix.devtools.impl
...
+      .map(_.toInt)
+      .orElse(readUserFunctionPortFromFile)
+      .getOrElse(8080)
+  }
+
+  private val ExtractPort = """\D+:*(\d+)""".r
+  private def readUserFunctionPortFromFile: Option[Int] = {
+    lines
+      .find(_.trim.startsWith("USER_FUNCTION_PORT"))
+      .map(_.trim.replace("}", ""))
```
octonato, line 72:
```suggestion
      // TODO: remove this hack
      // regex doesn't match if the line contains a `}`
      // strange enough I tested it online and in other programs and it works
      // I guess this is something related to Java regex. Leaving as is for now
      .map(_.trim.replace("}", ""))
```

project/Dependencies.scala
```diff
@@ -106,7 +106,7 @@ object Dependencies {
   val coreSdk =
     deps ++= Seq("com.typesafe" % "config" % "1.4.2")
 
-  val devtools = deps ++= Seq(testContainers)
+  val devtools = deps ++= Seq(testContainers, scalaTest % Test)
```
octonato, line 109: Follow-up PR will remove `testContainers`. We need it for now because of maven plugin.

devtools/src/main/scala/kalix/devtools/impl/DockerComposeUtils.scala
```diff
@@ -19,38 +19,62 @@ package kalix.devtools.impl
...
-    execIfFileExists(file) {
-      Process(s"docker-compose -f $file up", None, "USER_FUNCTION_PORT" -> userFunctionPort.toString).run
+  // read the file once and cache the lines
+  // we will need to iterate over it more than once
+  private lazy val lines: Seq[String] = {
+    val src = Source.fromFile(file)
+    try {
+      src.getLines.toList
+    } finally {
+      src.close()
```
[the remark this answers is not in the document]
octonato: I added one more test here e8ee3bf to prove that the first port from the first proxy will be used. So eventually, we can document that in the case of multiple proxies, the first declaration is the one that should be used to define the port. 

We are also safe regarding the service-port-mappings. The effect of having many proxies is that we will collect all service-port-mappings, but that should not be a problem. Declaring more port mappings than actually needed is not a problem and the mappings should also not differ between services.

devtools/src/main/scala/kalix/devtools/impl/DockerComposeUtils.scala
```diff
@@ -35,22 +37,62 @@ object DockerComposeUtils {
...
+
+  def readUserFunctionPort: Int = {
+    envVar
+      .get("USER_FUNCTION_PORT")
+      .map(_.toInt)
+      .orElse(readUserFunctionPortFromFile)
+      .getOrElse(8080)
+  }
+
+  private val ExtractPort = """\D+:*(\d+)""".r
```
efgpinto, line 82: Is this meant to not accept something like `127.0.0.1:8899` ? I thought that would need to be accepted..
Also, what is the purpose of that `*`?

I'm wondering if something like `.+:(\d+)` does not suffice..

efgpinto (commented):
Some questions around the regex, otherwise looks good

devtools/src/main/scala/kalix/devtools/impl/DockerComposeUtils.scala
efgpinto: > Also, what is the purpose of that `*`?

Oh, I just saw now.. because its usually defined like `${USER_FUNCTION_PORT:-8081}`... I was looking at that for the normal types of input local:port... sorry. Let me see again with fresh eyes about the `}`. :)

octonato: yep, the intention is to extract the port from lines defined as:

* `USER_FUNCTION_PORT:${USER_FUNCTION_PORT:-8080}`
* `USER_FUNCTION_PORT:8080`
* `USER_FUNCTION_PORT:${MY_OWN_VARIABLE_NAME:-8080}`

octonato: Arguably, we could fix it to match `USER_FUNCTION_PORT:` at start of line.

And, also the last option is not useful because we won't be able to read `MY_OWN_VARIABLE_NAME` anyway. 

This brings us down to the following options:

* `USER_FUNCTION_PORT:${USER_FUNCTION_PORT:-8080}`
* `USER_FUNCTION_PORT:8080`

octonato: I will add an extra test case to match `USER_FUNCTION_PORT:8080`

efgpinto: ok, to remove the hack you can use `\D+:*(\d+)\}?` or `\D+:*(\d+).?`

devtools/src/main/scala/kalix/devtools/impl/DockerComposeUtils.scala
```diff
@@ -35,22 +37,62 @@ object DockerComposeUtils {
...
+      // strange enough I tested it online and in other programs and it works
+      // I guess this is something related to Java regex. Leaving as is for now
+      .map(_.trim.replace("}", ""))
+      .collect { case ExtractPort(port) => port.toInt }
+  }
+
+  def readServicePortMappings: Seq[String] =
+    lines
+      .map(_.trim)
+      .filter(_.startsWith("-Dkalix.dev-mode.service-port-mappings"))
```
octonato, line 97: This is not sufficient. 
One can configure their docker with:

`-Dsomething=bla -Dkalix.dev-mode.service-port-mappings.foo=9001` 

And then we won't find the service-port-mappings because line is not starting with it.

Need to make it more robust.

octonato:
@efgpinto, I pushed two new commits. One to improve the extraction of UF port and one for the extraction of port mappings.

efgpinto (approved):
LGTM

PR #1606, opened 2023-05-03
feat: run docker-compose from kalix:runAll

octonato:
This PR aligns the mvn plugin with the sbt. 

We have now `mvn kalix:run` and `mvn kalix:runAll`. The former runs only the user function and user need to start the proxy by hand. The former runs both.

Built on top of #1604, but ready for review.


aludwiko (approved):
looks good, but maybe someone with more sbt experience should have a look :)

octonato:
> looks good, but maybe someone with more sbt experience should have a look :)

I merged without asking for a sbt expert since the PR is about maven. 😆

PR #1582, opened 2023-04-20
chore: add multi-service testing example

octonato:
This is will be the basis of the full docs on how to run multiple services. I will come back to this when we have the Proxy ready.

PR #1608, opened 2023-05-04
chore: move sys props to JAVA_TOOL_OPTIONS section

octonato:
This is in preparation for the changes to document `sbt runAll`/`mvn kalix:runAll`.

As the number of properties might grow, it's inconvenient to pass it as a single line. 

When we break it on different lines, we can't have uncommented lines. Typical for the google-pubsub-emulator. So, instead, we instruct how to add the PubSub extra property.

Last but not least, we should avoid using `command:` and prefer `JAVA_TOOL_OPTIONS`. The reason is that `command` works because of the way `sbt-native-pacakger` builds the docker image, but it's not necessarily a standard. If we want to pass `-D` properties to other JVM containers (for example one running the user function), then we need to use `JAVA_TOOL_OPTIONS`. 

As we document how to build a large docker-compose containing multiple proxies and user functions, we don't want to have two ways of passing `-D` properties. Therefore, we should standardize on `JAVA_TOOL_OPTIONS`

### DIFF
--- a/docs/src/modules/java/pages/access-control.adoc
+++ b/docs/src/modules/java/pages/access-control.adoc
@@ -124,7 +124,6 @@ When running a service during local development, it may be useful to enable ACL 
 ```yaml
 kalix-proxy:
   image: gcr.io/kalix-public/kalix-proxy:latest
-  command: -Dconfig.resource=dev-mode.conf -Dlogback.configurationFile=logback-dev-mode.xml -Dkalix.proxy.eventing.support=google-pubsub-emulator
   ports:
     - "9000:9000"
   environment:
@@ -144,7 +143,6 @@ If running multiple services in local development, you may want to run with ACLs
 ```yaml
 kalix-proxy:
   image: gcr.io/kalix-public/kalix-proxy:latest
-  command: -Dconfig.resource=dev-mode.conf -Dlogback.configurationFile=logback-dev-mode.xml -Dkalix.proxy.eventing.support=google-pubsub-emulator
   ports:
     - "9000:9000"
   environment:
--- a/maven-java/kalix-maven-archetype-event-sourced-entity/src/main/resources/archetype-resources/docker-compose.yml
+++ b/maven-java/kalix-maven-archetype-event-sourced-entity/src/main/resources/archetype-resources/docker-compose.yml
@@ -1,17 +1,19 @@
 #[[
 # If you're looking to use eventing with Google PubSub, to get an emulator running:
-# - uncomment property "-Dkalix.proxy.eventing.support=google-pubsub-emulator"
+# - add property "-Dkalix.proxy.eventing.support=google-pubsub-emulator" to the JAVA_TOOL_OPTIONS environment map under the kalix-proxy service
 # - uncomment the env var PUBSUB_EMULATOR_HOST and the section below for gcloud-pubsub-emulator service
 version: "3"
 services:
   kalix-proxy:
     image: gcr.io/kalix-public/kalix-proxy:1.1.8
-    command: -Dconfig.resource=dev-mode.conf -Dlogback.configurationFile=logback-dev-mode.xml #-Dkalix.proxy.eventing.support=google-pubsub-emulator
     ports:
       - "9000:9000"
     extra_hosts:
       - "host.docker.internal:host-gateway"
     environment:
+      JAVA_TOOL_OPTIONS: >
+        -Dconfig.resource=dev-mode.conf
+        -Dlogback.configurationFile=logback-dev-mode.xml
       USER_FUNCTION_HOST: ${USER_FUNCTION_HOST:-host.docker.internal}
       USER_FUNCTION_PORT: ${USER_FUNCTION_PORT:-8080}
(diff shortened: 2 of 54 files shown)


docs/src/modules/java/pages/access-control.adoc
```diff
@@ -144,7 +143,6 @@ If running multiple services in local development, you may want to run with ACLs
 ```yaml
 kalix-proxy:
   image: gcr.io/kalix-public/kalix-proxy:latest
-  command: -Dconfig.resource=dev-mode.conf -Dlogback.configurationFile=logback-dev-mode.xml -Dkalix.proxy.eventing.support=google-pubsub-emulator
```
octonato, line 147: Here and above, the fragment on how to pass `-D` is irrelevant. The goal of this two documentation section is to document how to enable the ACL and how to name the service. 

Removing just to now bloat the docs with unnecessary info. 

Of course we need some context to users can recognize that we are point to the kalix-proxy service configuration in docker-compose.

maven-java/kalix-maven-archetype-event-sourced-entity/src/main/resources/archetype-resources/docker-compose.yml
```diff
@@ -1,17 +1,19 @@
 #[[
 # If you're looking to use eventing with Google PubSub, to get an emulator running:
-# - uncomment property "-Dkalix.proxy.eventing.support=google-pubsub-emulator"
+# - add property "-Dkalix.proxy.eventing.support=google-pubsub-emulator" to the JAVA_TOOL_OPTIONS environment map under the kalix-proxy service
```
octonato, line 3: Instruct how to add a property, instead of uncommenting it.

samples/java-protobuf-eventsourced-customer-registry-subscriber/docker-compose.yml
```diff
@@ -1,15 +1,16 @@
 version: "3"
 services:
   # note the ports being different from other sample docker-compose files to allow this service to run
-  # on the same local machine as the scala-customer-registry-
+  # on the same local machine as the java-protobuf-eventsourced-customer-registry
   kalix-proxy:
     image: gcr.io/kalix-public/kalix-proxy:1.1.8
-    command: -Dconfig.resource=dev-mode.conf -Dlogback.configurationFile=logback-dev-mode.xml -Dkalix.proxy.http-port=9001 -Dkalix.proxy.eventing.support=google-pubsub-emulator
```
octonato, line 7: removed `kalix.proxy.eventing.support=google-pubsub-emulator`, not needed in this service

samples/java-spring-eventsourced-customer-registry-subscriber/docker-compose.yml
```diff
@@ -4,12 +4,13 @@ services:
   # on the same local machine as the java-spring-customer-registry
   kalix-proxy:
     image: gcr.io/kalix-public/kalix-proxy:1.1.8
-    command: -Dconfig.resource=dev-mode.conf -Dlogback.configurationFile=logback-dev-mode.xml #-Dkalix.proxy.eventing.support=google-pubsub-emulator
```
octonato, line 7: removed `kalix.proxy.eventing.support=google-pubsub-emulator`, not needed in this service

samples/scala-protobuf-eventsourced-customer-registry-subscriber/docker-compose.yml
```diff
@@ -4,12 +4,13 @@ services:
   # on the same local machine as the scala-customer-registry-
   kalix-proxy:
     image: gcr.io/kalix-public/kalix-proxy:1.1.8
-    command: -Dconfig.resource=dev-mode.conf -Dlogback.configurationFile=logback-dev-mode.xml -Dkalix.proxy.http-port=9001 -Dkalix.proxy.eventing.support=google-pubsub-emulator
```
octonato, line 7: removed `kalix.proxy.eventing.support=google-pubsub-emulator`, not needed in this service

samples/scala-protobuf-eventsourced-customer-registry-subscriber/docker-compose.yml
```diff
@@ -4,12 +4,13 @@ services:
...
     ports:
-      - "9001:9001"
+      - "9001:9000"
```
octonato, line 8: We don't need to change the port inside the container, only the exposed port.

On property less to configure.

samples/java-protobuf-eventsourced-customer-registry-subscriber/docker-compose.yml
```diff
@@ -1,15 +1,16 @@
...
     ports:
-      - "9001:9001"
+      - "9001:9000"
```
octonato, line 8: We don't need to change the port inside the container, only the exposed port.

On property less to configure.

octonato:
For the records, this is a spin-off from #1582 

I'm breaking it up to make it easier to manage and have some parts already out.

aludwiko (approved):

efgpinto (approved):
LGTM

PR #1612, opened 2023-05-04
ci: run tests in GH workflow

ennru:
Run all steps for PR validation in GH after the publish local step.

I have some ideas to improve the caching as we currently store too much: "Approaching total cache storage limit (26.18 GB of 10 GB Used)"


johanandren (approved):

ennru:
Latest run of these workflows

image

PR #1616, opened 2023-05-05
ci: reuse a single sbt cache

ennru:
`coursier/cache-action` uses the job name as key for the cache and it can't be overridden, AFAICS.

This uses `actions/cache` directly so that many jobs rely on the same cache.

Also, moving the caching before the `coursier/setup-action` will cache the JVM installation, IIUC.


ennru:
The second run, with the existing cache, looked like this
image

johanandren (approved):

PR #1619, opened 2023-05-09
fix: handling path variables and query parameters with proto default values

aludwiko:
References https://github.com/lightbend/kalix-proxy/issues/2003

### DIFF
--- a/sdk/java-sdk-protobuf/src/main/scala/kalix/javasdk/impl/EntityExceptions.scala
+++ b/sdk/java-sdk-protobuf/src/main/scala/kalix/javasdk/impl/EntityExceptions.scala
@@ -89,17 +89,10 @@ object EntityExceptions {
   }
 
   def failureMessageForLog(cause: Throwable): String = cause match {
-    case EntityException(_, _, _, _, Some(BadRequestException(msg))) => msg
     case EntityException(entityId, commandId, commandName, _, _) =>
       val commandDescription = if (commandId != 0) s" for command [$commandName]" else ""
       val entityDescription = if (entityId.nonEmpty) s" [$entityId]" else ""
       s"Terminating entity$entityDescription due to unexpected failure$commandDescription"
     case _ => "Terminating entity due to unexpected failure"
-  }
-
-  def failureResponse(correlationId: String, cause: Throwable): Failure = cause match {
-    case EntityException(_, _, _, _, Some(BadRequestException(_))) =>
-      Failure(0, s"Bad request [$correlationId]", Status.Code.INVALID_ARGUMENT.value())
-    case _ => Failure(description = s"Unexpected failure [$correlationId]")
   }
 }
--- a/sdk/java-sdk-protobuf/src/main/scala/kalix/javasdk/impl/EntityExceptions.scala
+++ b/sdk/java-sdk-protobuf/src/main/scala/kalix/javasdk/impl/EntityExceptions.scala
@@ -16,7 +16,10 @@
 
 package kalix.javasdk.impl
 
+import io.grpc.Status
+import kalix.javasdk.impl.ErrorHandling.BadRequestException
 import kalix.javasdk.{ eventsourcedentity, valueentity }
+import kalix.protocol.component.Failure
 import kalix.protocol.entity.Command
 import kalix.protocol.event_sourced_entity.EventSourcedInit
 import kalix.protocol.replicated_entity.ReplicatedEntityInit
@@ -86,10 +89,17 @@ object EntityExceptions {
   }
 
   def failureMessageForLog(cause: Throwable): String = cause match {
+    case EntityException(_, _, _, _, Some(BadRequestException(msg))) => msg
     case EntityException(entityId, commandId, commandName, _, _) =>
       val commandDescription = if (commandId != 0) s" for command [$commandName]" else ""
       val entityDescription = if (entityId.nonEmpty) s" [$entityId]" else ""
       s"Terminating entity$entityDescription due to unexpected failure$commandDescription"
     case _ => "Terminating entity due to unexpected failure"
+  }
+
+  def failureResponse(correlationId: String, cause: Throwable): Failure = cause match {
+    case EntityException(_, _, _, _, Some(BadRequestException(_))) =>
+      Failure(0, s"Bad request [$correlationId]", Status.Code.INVALID_ARGUMENT.value())
+    case _ => Failure(description = s"Unexpected failure [$correlationId]")
   }
 }
--- a/sdk/java-sdk-protobuf/src/main/scala/kalix/javasdk/impl/action/ActionsImpl.scala
+++ b/sdk/java-sdk-protobuf/src/main/scala/kalix/javasdk/impl/action/ActionsImpl.scala
@@ -90,7 +90,7 @@ private[javasdk] object ActionsImpl {
       command: ActionCommand,
       ex: Throwable): ActionResponse = {
     ex match {
-      case badReqEx: BadRequestException => handleBadRequest(service, command, badReqEx)
+      case badReqEx: BadRequestException => handleBadRequest(badReqEx.getMessage)
       case _ =>
         ErrorHandling.withCorrelationId { correlationId =>
           service.log.error(s"Failure during handling of command ${command.serviceName}.${command.name}", ex)
@@ -99,13 +99,8 @@ private[javasdk] object ActionsImpl {
     }
   }
 
-  private def handleBadRequest(service: ActionService, command: ActionCommand, ex: Throwable): ActionResponse =
-    ErrorHandling.withCorrelationId { correlationId =>
-      service.log.error(s"Bad request when handling of command ${command.serviceName}.${command.name}", ex)
-      ActionResponse(
-        ActionResponse.Response.Failure(
-          Failure(0, s"Bad request [$correlationId]", Status.Code.INVALID_ARGUMENT.value())))
-    }
+  private def handleBadRequest(description: String): ActionResponse =
+    ActionResponse(ActionResponse.Response.Failure(Failure(0, description, Status.Code.INVALID_ARGUMENT.value())))
 
   private def protocolFailure(correlationId: String): ActionResponse = {
     ActionResponse(ActionResponse.Response.Failure(Failure(0, s"Unexpected error [$correlationId]")))
--- a/sdk/java-sdk-protobuf/src/main/scala/kalix/javasdk/impl/action/ActionsImpl.scala
+++ b/sdk/java-sdk-protobuf/src/main/scala/kalix/javasdk/impl/action/ActionsImpl.scala
@@ -26,9 +26,11 @@ import akka.stream.scaladsl.Sink
 import akka.stream.scaladsl.Source
 import com.google.protobuf.Descriptors
 import com.google.protobuf.any.Any
+import io.grpc.Status
 import kalix.javasdk._
 import kalix.javasdk.action._
 import kalix.javasdk.impl.ActionFactory
+import kalix.javasdk.impl.ErrorHandling.BadRequestException
 import kalix.javasdk.impl._
 import kalix.javasdk.impl.effect.EffectSupport.asProtocol
 import kalix.protocol.action.ActionCommand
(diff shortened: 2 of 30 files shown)


sdk/java-sdk-protobuf/src/main/scala/kalix/javasdk/impl/EntityExceptions.scala
```diff
@@ -86,10 +89,17 @@ object EntityExceptions {
...
     case EntityException(entityId, commandId, commandName, _, _) =>
       val commandDescription = if (commandId != 0) s" for command [$commandName]" else ""
       val entityDescription = if (entityId.nonEmpty) s" [$entityId]" else ""
       s"Terminating entity$entityDescription due to unexpected failure$commandDescription"
     case _ => "Terminating entity due to unexpected failure"
   }
+
+  def failureResponse(correlationId: String, cause: Throwable): Failure = cause match {
+    case EntityException(_, _, _, _, Some(BadRequestException(_))) =>
+      Failure(0, s"Bad request [$correlationId]", Status.Code.INVALID_ARGUMENT.value())
```
aludwiko, line 102: This is not working correctly on the proxy side, because we are ignoring the status code for failures at this level, so currently it will be 500.

aludwiko (commented):
Ok, ready for review, not sure if we want to map errors about missing params to 400 http status code both for action and entities. I suppose we should.

aludwiko (commented):
Ok, ready for review, not sure if we want to map errors about missing params to 400 http status code both for action and entities. I suppose we should.

sdk/java-sdk-spring/src/it/java/com/example/wiring/EventSourcedEntityIntegrationTest.java
```diff
@@ -102,6 +107,59 @@ public void verifyCounterEventSourcedAfterRestartFromSnapshot() {
...
+      .onErrorResume(WebClientResponseException.class, error -> {
+        if (error.getStatusCode().is4xxClientError()) {
+          return Mono.just(ResponseEntity.status(error.getStatusCode()).body(null));
+        } else {
+          return Mono.error(error);
+        }
+      })
+      .block(timeout);
+
+    assertThat(result.getStatusCode()).isEqualTo(HttpStatus.BAD_REQUEST);
```
aludwiko, line 159: currently, this is failing because of missing handling on the proxy side

sdk/java-sdk-protobuf/src/main/scala/kalix/javasdk/impl/action/ActionsImpl.scala
```diff
@@ -83,10 +85,26 @@ final class ActionService(
...
+          protocolFailure(correlationId)
+        }
+    }
+  }
+
+  private def handleBadRequest(service: ActionService, command: ActionCommand, ex: Throwable): ActionResponse =
     ErrorHandling.withCorrelationId { correlationId =>
-      service.log.error(s"Failure during handling of command ${command.serviceName}.${command.name}", ex)
-      protocolFailure(correlationId)
+      service.log.error(s"Bad request when handling of command ${command.serviceName}.${command.name}", ex)
```
johanandren, line 104: Should be an error log, it's a malformed request from the user and should say what is wrong in the response to the user instead.

sdk/java-sdk-protobuf/src/main/scala/kalix/javasdk/impl/action/ActionsImpl.scala
```diff
@@ -83,10 +85,26 @@ final class ActionService(
 private[javasdk] object ActionsImpl {
   private[action] val log = LoggerFactory.getLogger(classOf[ActionsImpl])
 
-  private def handleUnexpectedException(service: ActionService, command: ActionCommand, ex: Throwable): ActionResponse =
+  private def handleUnexpectedException(
+      service: ActionService,
+      command: ActionCommand,
+      ex: Throwable): ActionResponse = {
+    ex match {
+      case badReqEx: BadRequestException => handleBadRequest(service, command, badReqEx)
```
johanandren, line 93: We'd need the same kind of handling for all components, because all components can have path/query param annotations, and need to turn this exception into a failure response with INVALID_ARGUMENT.

That's probably why you say the error code does not reach the user, the test is for an entity, not because the proxy is missing logic.

aludwiko: It is already handled for all entities, but the proxy ignores the status code for Failures different than the ones from the effect builder.

sdk/java-sdk-protobuf/src/main/scala/kalix/javasdk/impl/action/ActionsImpl.scala
johanandren: Yes, validation errors in HTTP (and gRPC) APIs should generally not lead to any logging on the server side, the error is at the client side so needs to go in response so the client can fix their request.

aludwiko: Ok, so it `Shouldn't` :)

sdk/java-sdk-protobuf/src/main/scala/kalix/javasdk/impl/action/ActionsImpl.scala
aludwiko: for instance here: https://github.com/lightbend/kalix-jvm-sdk/pull/1619/files#diff-9a0d8d88f0ecd39710e5c4242e4115bb9300efeed6112df3949db5859805d1aeR116

johanandren: I don't get it, we have test coverage here https://github.com/lightbend/kalix-proxy/blob/main/proxy/core/src/test/scala/kalix/proxy/action/ActionServiceSpec.scala#L140

For the exception we send a failure response with a custom code and that should reach the client. Or are you saying that doesn't work?

sdk/java-sdk-protobuf/src/main/scala/kalix/javasdk/impl/action/ActionsImpl.scala
johanandren: Ah, runaway `n`t`, sorry didn't notice.

sdk/java-sdk-protobuf/src/main/scala/kalix/javasdk/impl/action/ActionsImpl.scala
aludwiko: It works for actions, but not for entities: https://github.com/lightbend/kalix-proxy/blob/main/proxy/core/src/main/scala/kalix/proxy/eventsourcedentity/EventSourcedEntity.scala#L381

johanandren: Aha, so what is missing is error code propagation for entities, but it's at least the same protocol so it's just about picking the `grpc_status_code` up from the `Failure` message. 

Does that mean we don't have the custom error codes effect API in the entities either (probably, if we had it we would have noticed it didn't work)

aludwiko: ok, so we have two possible sources of failures:
unexpected failures, on the very top level of the protocol:
```
message EventSourcedStreamOut {
  oneof message {
    EventSourcedReply reply = 1;
    Failure failure = 2;
    EventSourcedSnapshotReply snapshot_reply = 3;
  }
}
```
user-defined failures:
```
message EventSourcedReply {
  // The id of the command being replied to. Must match the input command.
  int64 command_id = 1;
  // The action to take
  ClientAction client_action = 2;
  ...
  }

message ClientAction {
  oneof action {
    // Send a reply
    Reply reply = 1;
    // Forward to another component
    Forward forward = 2;
    // Send a failure to the client
    Failure failure = 3;
  }
}
```

for the second one, you can propagate the error code as expected. The question is which kind of a failure I should use for request validation errors?

johanandren: I think the client action level of failure is fine here, a bad request doesn't break state or affect subsequent requests to the same entity, it's just something we want to communicate to the client, just like an error effect from the user.

aludwiko: Ok, VE, ES, WE, Action done. I've put the catch for entities in a different place than for the action, feels more suitable to check it there, but I am open to suggestions.

sdk/java-sdk-spring/src/main/scala/kalix/javasdk/impl/reflection/ParameterExtractor.scala
```diff
@@ -78,7 +78,7 @@ object ParameterExtractors {
     override def extract(context: DynamicMessageContext): T = {
       (required, field.isRepeated || context.message.hasField(field)) match {
         case (_, true)      => deserialize(context.message.getField(field))
-        case (true, false)  => throw BadRequestException(s"Message missing required field: ${field.getName}")
+        case (true, false)  => throw BadRequestException(s"Message is missing required field: ${field.getName}")
```
johanandren, line 81: Do we know if it is a path or query parameter here and can we improve the error message to describe that, or some step further out, "message is missing required field" are a bit misleading since the user does not know anything about us encoding the path hand query params as message fields internally.

aludwiko: True, I don't like it, it was a copy-paste from some other place. And yes, we know that currently, this applies only to request parameters. Changed.

sdk/java-sdk-protobuf/src/main/scala/kalix/javasdk/impl/action/ActionsImpl.scala
johanandren: Looks good to me

sdk/java-sdk-spring/src/main/scala/kalix/javasdk/impl/reflection/KalixMethod.scala
```diff
@@ -181,6 +180,13 @@ case class SyntheticRequestServiceMethod(
       "Missing request method mapping. Kalix Java SDK methods must have a request method defined.")
   }
 
+  if (javaMethod.getParameterAnnotations.toSet.flatten
+      .collect { case (p: PathVariable) => p }
+      .exists(_.required() == false)) {
+
+    throw ServiceIntrospectionException(javaMethod, "Currently all @PathVariables must be defined as required.")
+  }
```
octonato, line 188: annoying that we need `collect` and `exists`. 

Maybe something like: 

```scala
javaMethod.getParameterAnnotations.toSet.flatten.exists {
  case p: PathVariable => !p.required
  case _ => false
}
```

Not sure if make it more readable though. 
Maybe if we move it out of the `if`?

octonato (approved):
LGTM

with some minor comments and suggestions

Regarding the imports, since we are not being consistent anyway, feel free to ignore it. 
If should better have a scalafmt rule for it instead.

octonato (approved):
LGTM

with some minor comments and suggestions

Regarding the imports, since we are not being consistent anyway, feel free to ignore it. 
If should better have a scalafmt rule for it instead.

sdk/java-sdk-spring/src/main/scala/kalix/javasdk/impl/reflection/KalixMethod.scala
aludwiko: Personally, I think that collect and exists is more readable, so I will leave it as it is.

octonato:
> @octonato I have this problem, that from time to time I'm using "optimize import" from IntelliJ, which doesn't respect scalafmt settings: https://youtrack.jetbrains.com/issue/SCL-20791 so it looks like the only workaround is to configure it correctly. Let me check what I can do...

I can export my intellij settings and send to you

Issue #1626, opened 2023-05-16
supporting custom JVM argument in kalix-maven-plugin


aludwiko:
Extending the kalix-maven-plugin with additional configuration properties like:
```
<configuration>
   <jvmArgs>
      <arg>--enable-preview<arg>
      <arg>-Dspring.profiles.active=all<arg>
   </jvmArgs>
<configuration>
```

PR #1627, opened 2023-05-16
fix: supporting custom JVM args in kalix-maven-plugin

aludwiko:
References https://github.com/lightbend/kalix-jvm-sdk/issues/1626


octonato (approved):
LGTM

In the end, it doesn't look as complicated as I thought.

PR #1629, opened 2023-05-17
fix: enable todos in dev

octonato:
Partially reverting #1625. 

These settings are used when building the docs locally in dev mode. Something else is enabling this in prod.


efgpinto (approved):

octonato:
For the record, nothing was showing in the prod docs. We were looking at the dev docs. 🤦

Issue #1631, opened 2023-05-17
Running docker-compose from inside KalixRunner


octonato:
We have recently extended the Kalix plugin (both sbt and mvn) to start docker-compose and therefore remove the need to start it on a separate terminal. 

However, this solution only works if users run the project using the plugin. Very often a user will simply start the application from the IDE and run the Main function directly. Also, when debugging an application, users will probably do it from the IDE and the Main function. 

When started in that way, the user function will run in 'standalone' and users will need to be aware that they need to run `docker-compose up` themselves. And to make it worst, we don't have the means to detect that the Main function was executed by hand and therefore we can't even print a warning or instruction to the users. 

One possible solution to that problem is to integrate the `DockerComposeUtils` inside the `KalixRunner`. 

The plan would be:

1. add a dev-mode config property: `kalix.dev-mode.docker-compose-file=docker-compose.yml`. Defaulting to `docker-compose.yml`, but users can change it if they want.
2. when the user function runs, we read that property and look for the file at the root of the process. If there is such a file, we know that we are in dev mode and we can start `DockerComposeUtils`
3. when the user function runs in prod, the file won't be present at the root (we only package `src/main/{resources,java,scala}`) and therefore we won't start `DockerComposeUtils`.
4. When running `mvn kalix:run` or `sbt run`, we set `kalix.dev-mode.docker-compose-file=none` as the user is explicitly asking to not start docker.
5. When running `mvn kalix:runAll` or `sbt runAll`, we leave it as is and we let the `KalixRunner` bootstrap the UF and the docker.
6. For integration tests using the Teskit, we need to set `kalix.dev-mode.docker-compose-file=none` as the proxy is managed by Testcontainers and we don't want to run docker-compose in such a case.

aludwiko:
> 4. When running mvn kalix:run or sbt run, we set kalix.dev-mode.docker-compose-file=none as the user is explicitly asking to not start docker.

Isn't kalix:run equal to running Main from the IDE?

octonato:
Yes, `kalix:run` is now the equivalent of running the Main on the IDE. And that's exactly the problem. We document `kalix:runAll` and we expect users to use it. As soon as they start the Main from the IDE, they won't immediately see that something is missing. 

Users will always have the option to run docker-compose by hand and then the Main. 
Btw, we kind of support the mix already. If you start docker-compose twice on the same dir, the second time won't try to start it but will attach to the running containers and start to print the logs.

[...]

PR #1637, opened 2023-05-22
feat: integrate DockerComposeUtils into KalixRunner

octonato:
References https://github.com/lightbend/kalix-jvm-sdk/issues/1631

Still draft, but almost there.

- [x] `docker-compose stop` in CoordinatedShutdown
- [x] sbt run should not start docker, runAll should


octonato:
There are lots of clean-up to be done. Previously we were accepting mapping from `-D` properties or from docker-compose. 

Now, we generate a new config if docker-compose.yml is available. There is still some code supporting both cases that's now obsolete.

Also, the integration tests from the subscriber samples can be revisited now.

maven-java/kalix-maven-plugin/src/main/scala/kalix/RunAllMojo.scala
```diff
@@ -11,19 +10,13 @@ import org.apache.maven.plugins.annotations.ResolutionScope
  */
 @Mojo(name = "runAll", defaultPhase = LifecyclePhase.VALIDATE, requiresDependencyResolution = ResolutionScope.RUNTIME)
 @Execute(phase = LifecyclePhase.COMPILE)
-class RunAllMojo extends RunParameters with DockerParameters {
+class RunAllMojo extends RunParameters {
 
   override def execute(): Unit = {
-
-    val dockerComposeUtils = DockerComposeUtils(dockerComposeFile, sys.env)
-    dockerComposeUtils.start()
```
octonato, line 19: When running Kalix using the plugin, we might want to start the docker-compose from inside the plugin though. So we might want to revert this change later.

The rationale is that if we want to have some hot-reload feature (like Spring DevTools), we don't want to have the docker-compose process attached to the process running `Main`. 

If we have them separate, the hot-reload will mean only reload the user function and proxy will be kept in the background. 

If we run docker-compose from behind the `Main`, then it will be shutdown and restart on each reload. We don't need to decide it now. We don't have hot-reload anyway, but something to keep in mind.

sdk/java-sdk-protobuf/src/main/scala/kalix/javasdk/KalixRunner.scala
```diff
@@ -233,6 +244,10 @@ final class KalixRunner private[javasdk] (
 
     logJvmInfo()
 
+    // start containers if application (only possible when running locally)
+    // TODO: stop it using CoordinateShutdown
+    dockerComposeUtils.foreach(_.start())
```
octonato, line 249: With this change, we will start the proxy container (and whatever other containers the user configures) when we run the `Main` function. That covers the starting it up from the IDE and from the plugin. 

This code won't be executed in a few situations though:

* In prod, because we don't have a docker-compose file at the root.
* If users explicitly set `kalix.dev-mode.docker-compose-file=none`.
* If users delete their `docker-compose.yml` file.
* When running integration tests using KalixTestkit, in which case the proxy is run with Testcontainers.
* When running `sbt run` or `mvn kalix:run` instead of the `runAll` variation. 

And here is the point where I think we can let it sink a little bit. 

Do we want to have `sbt run` and `mvn kalix:run` behave differently than running the `Main` function directly? 

The other alternative is to say: `sbt run` and `mvn kalix:run` behaves the same as running `Main`. Meaning, it also runs docker-compose. 
If we believe this is a good approach, then we can simply remove the `runAll`. 

And, in case someone wants to run docker-compose manually, then they will have to set `kalix.dev-mode.docker-compose-file=none`.

There is a catch though. `mvn kalix:run` is not yet released so it's easier for us to add `docker-compose` to it, but `sbt run` is just how sbt run any project. Existing users are already used to run `sbt run` + `docker-compose up`. 
This is not such a big deal though. Running `docker-compose` twice won't make it fail. The second run will simply attach to the existing container instances and start to print the logs.

Anyway, something we can let sink for a while.

devtools/src/main/scala/kalix/devtools/impl/DockerComposeUtils.scala
```diff
@@ -61,6 +73,8 @@ case class DockerComposeUtils(file: String, envVar: Map[String, String]) {
       Seq.empty
     }
 
+  // FIXME: process output is being printed in sbt console as error
+  // note to self: this is seems to be similar to sbt native packager printing errors when build docker images
```
octonato, line 77: I still need to understand why this happens.

sbt-plugin/src/main/scala/kalix/sbt/KalixPlugin.scala
```diff
@@ -46,44 +48,108 @@ object KalixPlugin extends AutoPlugin {
...
+
+      // workaround for https://github.com/sbt/sbt/issues/4038
+      Compile / runAll / forkOptions := {
+        val forkOpts = (Compile / run / forkOptions).value
+        forkOpts.withRunJVMOptions((Compile / runAll / javaOptions).value.toVector)
+      },
+      // redefine the runner for runAll
+      // this is needed in order to have the forkOptions applied
+      // this is a simplified clone of Defaults.runnerInit
+      Compile / runAll / runner := {
```
octonato, line 119: This was tricky. We want different jvm args depending on the tasks that we call. 

I got strange errors with classpath not containing some akka classes. I also tried, with help from Dale, to redefine the runner in a more elegant way. But it didn't work as expected and I decide to go for a hardcoded redefinition.

sbt-plugin/src/main/scala/kalix/sbt/KalixPlugin.scala
```diff
@@ -46,44 +48,108 @@ object KalixPlugin extends AutoPlugin {
...
+      // -------------------------------------------------------------------------------------------
+      // runAll task
+      Compile / runAll / fork := true,
+      // pass Jvm Args to forked process with default docker-compose settings
+      Compile / runAll / javaOptions ++= defaultJvmArgs ++ passedKalixArgs,
+
+      // workaround for https://github.com/sbt/sbt/issues/4038
+      Compile / runAll / forkOptions := {
+        val forkOpts = (Compile / run / forkOptions).value
+        forkOpts.withRunJVMOptions((Compile / runAll / javaOptions).value.toVector)
```
octonato, line 114: We need to manually redefine `forkOptions` as explained in https://github.com/sbt/sbt/issues/4038

octonato (commented):
Ready for final review. 

Not happy with end result in sbt plugin. I wish I could make it more elegant and idiomatic.

samples/java-spring-eventsourced-customer-registry-subscriber/src/it/java/customer/api/CustomerIntegrationTest.java
```diff
@@ -61,8 +61,8 @@ public CustomerIntegrationTest(ApplicationContext applicationContext) {
     confMap.put("kalix.user-function-port", dockerComposeUtils.userFunctionPort());
     // don't kill the test JVM when terminating the KalixRunner
     confMap.put("kalix.system.akka.coordinated-shutdown.exit-jvm", "off");
-    // dev-mode should be false when running integration tests
-    confMap.put("kalix.dev-mode.enabled", false);
+    // FIXME: should point to docker-compose-integration.yml instead
+    conf.put("kalix.dev-mode.docker-compose-file", "none");
```
efgpinto, line 65: Not sure I get this FIXME.. why should it point to "docker-compose-integration.yml" and what is preventing it?

efgpinto (approved):
AFAICS (which is not much in this case) LGTM. With a small question.

samples/java-spring-eventsourced-customer-registry-subscriber/src/it/java/customer/api/CustomerIntegrationTest.java
octonato: I forgot about this FIXME. 

I set it explicitly to `none` because I didn't want the KalixRunner to read the default. Which is, `docker-compose.yml`. 

Instead, we should configure it to use `docker-compose-integration.yml` and all the DockerComposeUtils that we are explicitly calling in this test should be removed. 

This test should start Klaix using `docker-compose-integration.yml` in dev-mode, run the assertions and let the containers be shutdown during coordinated shutdown.

devtools/src/main/scala/kalix/devtools/impl/DockerComposeUtils.scala
octonato: I did some more tests on this. Here is what I observed. 

When running `docker-compose up` directly, we don't have any error output. At least not on my shell. 

```shell
=❯ docker-compose up
[+] Running 1/0
 ✔ Container scala-protobuf-fibonacci-action-kalix-proxy-1  Created                                                                                                    0.0s
Attaching to scala-protobuf-fibonacci-action-kalix-proxy-1
scala-protobuf-fibonacci-action-kalix-proxy-1  | Picked up JAVA_TOOL_OPTIONS: -Dconfig.resource=dev-mode.conf -Dlogback.configurationFile=logback-dev-mode.xml
...
^CGracefully stopping... (press Ctrl+C again to force)
Aborting on container exit...
[+] Running 1/1
 ✔ Container scala-protobuf-fibonacci-action-kalix-proxy-1  Stopped                                                                                                    0.6s
canceled
```

When running with the current version (v1.2.1), we also don't get any error logger. 
In the current version, `DockerComposeUtils` runs in the same process as the plugin.

```shell
=❯ sbt runAll
...
 Container scala-protobuf-fibonacci-action-kalix-proxy-1  Created
Attaching to scala-protobuf-fibonacci-action-kalix-proxy-1
scala-protobuf-fibonacci-action-kalix-proxy-1  | Picked up JAVA_TOOL_OPTIONS: -Dconfig.resource=dev-mode.conf -Dlogback.configurationFile=logback-dev-mode.xml
...
^CGracefully stopping... (press Ctrl+C again to force)
Aborting on container exit...
 Container scala-protobuf-fibonacci-action-kalix-proxy-1  Stopping
```

But, when running with the code in this PR, we get:

```shell
[error]  Container scala-protobuf-fibonacci-action-kalix-proxy-1  Created
[info] Attaching to scala-protobuf-fibonacci-action-kalix-proxy-1
[error] scala-protobuf-fibonacci-action-kalix-proxy-1  | Picked up JAVA_TOOL_OPTIONS: -Dconfig.resource=dev-mode.conf -Dlogback.configurationFile=logback-dev-mode.xml
[error] scala-protobuf-fibonacci-action-kalix-proxy-1  |
[info] scala-protobuf-fibonacci-action-kalix-proxy-1  | 15:56:38.892 INFO  kalix.proxy.KalixProxyMain - Starting Kalix Proxy version [1.1.13] Dev Mode
...
^C[error] Gracefully stopping... (press Ctrl+C again to force)
[error] Aborting on container exit...
[info] shutting down sbt server
[error]  Container scala-protobuf-fibonacci-action-kalix-proxy-1  Stopping
```

Here the behaviour is quite different. First, the lines get a log-level tag (`[info]` and `[error]`). And some of the lines are printed as `error` others as `info`. 

In this run, `DockerComposeUtils` runs from inside the `KalixRunner` and is in a forked process. 

The process we use to spawn docker-compose from inside the JVM seems to be emitting some of the logs to stderr. Then sbt runs a forked process, it captures everything that is emitted to stdout and stderr and log it with the log level as above.

I did a quick test with this:

```scala
object Hello extends  App {
  System.out.println("this goes to stdout")
  System.err.println("this goes to stderr")
}
```

When run with `run/fork := true`, we get:

```scala
[error] this goes to stderr
[info] this goes to stdout
```

When run with `run/fork := false`, we get:

```scala
this goes to stderr
this goes to stdout
```

So, first question is, why docker-compose is sending this output to stderr? 
We could try to forward this output to stdout, but then in case of a real error, sbt will print it as `info`.

octonato: Sorry for the wall of text. Now I realise how large it became. 
Anyway, it's useful to me to organise my thoughts and register somewhere.

aludwiko (approved):
LGTM, but I'm not qualified to review the sbt part :)

devtools/src/main/scala/kalix/devtools/impl/DevModeSettings.scala
```diff
@@ -44,6 +46,60 @@ object DevModeSettings {
...
+            mainConfig.getConfig(portMappingsKeyPrefix).entrySet().asScala
+          portMappings.foldLeft(mainConfig) { case (main, entry) =>
+            val (host, port) = HostAndPort.extract(entry.getValue.unwrapped().toString)
+            main.withFallback(grpcClientConfig(entry.getKey, host, port))
+          }
+
+        } else
+          mainConfig
+      }
+  }
```
octonato, line 101: This is the missing part that was making the build to fail. 

It goes as such:

If running locally (exec Main) or through plugin, we read the docker-compose.yml file to learn where the other services are running. That means that we are running in a non-dockerized environment and we need to reach whatever other services' proxy through localhost (0.0.0.0).

So, first we read the file and build the mappings. Then we need to create gRPC client configs and web client configs for each of them. But in both cases, we need to ignore the host declared in the file and replace it with 0.0.0.0. 

On the other hand, if we run the user function from inside a container, we need to have a full host:port setup. In such a case, the port mapping will need to be something like: 
`kalix.dev-mode.service-port-mappings.foo=host.docker.internal:9001`. Where `foo` is the name of the service an `host.docker.internal` the address to access (intra container calls).

Note that the settings is initially target to the proxy. Between two proxy, we still need to use `host.docker.internal`.

sdk/java-sdk-spring/src/it/java/com/example/wiring/pubsub/DockerIntegrationTest.java
```diff
@@ -49,25 +49,15 @@ public abstract class DockerIntegrationTest {
...
 
   private KalixSpringApplication kalixSpringApplication;
 
   public DockerIntegrationTest(ApplicationContext applicationContext) {
     Map<String, Object> confMap = new HashMap<>();
-    confMap.put("kalix.user-function-port", dockerComposeUtils.userFunctionPort());
     // don't kill the test JVM when terminating the KalixRunner
     confMap.put("kalix.system.akka.coordinated-shutdown.exit-jvm", "off");
-    // dev-mode should be false when running integration tests
-    confMap.put("kalix.dev-mode.enabled", false);
```
octonato, line 62: we have been copying this over and over, but there is no such a thing as `dev-mode.enabled`. We probably had it in the past, but it's gone.

devtools/docker-compose.yml
```diff
@@ -0,0 +1,3 @@
+# this is a dummy yml file we need in other to test that DockerComposeUtils loads a docker-compose.yml
```
efgpinto, line 1:
```suggestion
# this is a dummy yml file we need in order to test that DockerComposeUtils loads a docker-compose.yml
```

devtools/src/main/scala/kalix/devtools/impl/DevModeSettings.scala
```diff
@@ -0,0 +1,109 @@
...
+          ConfigFactory
+            .parseString(s"kalix.user-function-port = ${dcu.userFunctionPort}")
+            .withFallback(mainConfig)
+
+        dcu.servicesHostAndPortMap.foldLeft(adaptedConfig) { case (main, (serviceName, hostAndPort)) =>
+          // DockerComposeUtils (dcu) is only available when running locally,
+          // therefore should always use localhost (0.0.0.0)
+          val host = "0.0.0.0"
+          val port = HostAndPort.extractPort(hostAndPort)
+          // when configuring through DockerComposeUtils, we need to parts.
```
efgpinto, line 80:
```suggestion
          // when configuring through DockerComposeUtils, we need two parts:
```

devtools/src/test/scala/kalix/devtools/impl/DevModeSettingsSpec.scala
```diff
@@ -0,0 +1,115 @@
...
+      val config = ConfigFactory.parseString(s"""
+           |kalix.dev-mode.docker-compose-file = none
+           |kalix.dev-mode.service-port-mappings.foo = "foo.docker.internal:9001"
+           |kalix.dev-mode.service-port-mappings.bar = "barhost:9002"
+           |kalix.dev-mode.service-port-mappings.baz = "baz-host:9003"
+           |""".stripMargin)
+
+      val enrichedConfig = DevModeSettings.addDevModeConfig(config)
+
+      // when relying on docker-compose file, hosts are replaced by 0.0.0.0
```
efgpinto, line 105:
```suggestion
      // when not relying on docker-compose file, hosts are kept as as originally
```

efgpinto (approved):
LGTM, left suggestions for typos in comments.

Issue #1646, opened 2023-05-29
support @Value annotation in Spring SDK


aludwiko:
We need to provide a way to configure Kalix components. In Spring, usually, this is done via `@Value` annotation.
```
public class EchoAction extends Action {

  public EchoAction(@Value("${test.config.int}") int value) {
    System.out.println(value);
  }
  ```

this doesn't work because we have a custom `DefaultListableBeanFactory` without additional configuration.

aludwiko:
I was able to fix that: https://github.com/lightbend/kalix-jvm-sdk/pull/1647/files#diff-145fb34add220ff4456929bdec52eb12dd0b294dad2c6d6fcde36ff8674c8e87R262 but I'm not comfortable with that change. It feels hacky and maybe there is a more elegant solution.

PR #1647, opened 2023-05-29
experimenting with @Value annotation

aludwiko:
References #1646


franciscolopezsancho:
I read a bit about this. Would you consider having a class that can be accessed from any component? I just had a look at @EnableConfigurationProperties? Like [this](https://www.baeldung.com/spring-enable-config-properties).  
Although now that I mention I remember we actively avoid the usage of Spring boot annotation outside of our control. Specifically @Autowired.

aludwiko:
I'm open to all options. For sure we need to provide a way to configure the application in some (Spring-like) way for the code first SDK.

Issue #1644, opened 2023-05-26
Improve docs on how to subscribe to a Topic from a View


octonato:
I think we should document this somewhere. Subscribing to a topic from a view from docs looks like a super straightforward task: https://docs.kalix.io/java/views.html#topic-view, but it requires manual metadata management. There is sth about ce-subject here: https://docs.kalix.io/projects/message-brokers.html#_delivery_characteristics but it's not clear that this is required for views.

_Originally posted by @aludwiko in https://github.com/lightbend/kalix-jvm-sdk/pull/1642#discussion_r1206957207_

efgpinto:
I think this is done now.

PR #1651, opened 2023-05-30
doc: view from topic documentation

aludwiko:
References #1644 

After the initial approach, I changed my mind and decided to put additional view-from-topic example in 
 - *eventsourced-customer-registry
 - *eventsourced-customer-registry-subscriber 
 samples. It feels better to have it there. Currently, only java-protobuf sample is updated, once we agree that it is ok, I will do similar changes to scala-protobuf and java-spring.

### DIFF
--- a/docs/src/modules/java-protobuf/pages/views.adoc
+++ b/docs/src/modules/java-protobuf/pages/views.adoc
@@ -321,7 +321,34 @@ include::example$scala-protobuf-eventsourced-customer-registry/src/main/scala/cu
 [#topic-view]
 == Creating a View from a topic
 
-The source of a View can be an eventing topic. You define it in the same way as shown in <<event-sourced-entity>> or <<value-entity>>, but leave out the `eventing.in` annotation in the Protobuf file.
+The source of a View can be a broker topic.
+You define it in the same way as shown in <<event-sourced-entity>> or <<value-entity>>.
+Note that, on your producer side you must manually pass the `ce-subject` metadata, required by the View component.
+
+[.tabset]
+Java::
++
+[source,proto,indent=0]
+.src/main/java/customer/api/CustomerEventsToTopicServiceAction.java
+----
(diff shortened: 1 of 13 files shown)


docs/src/modules/java-protobuf/pages/views.adoc
```diff
@@ -321,25 +321,54 @@ include::example$scala-protobuf-eventsourced-customer-registry/src/main/scala/cu
...
+The View definition must set `eventing.in` to a given topic name.
 
 [.tabset]
 Java::
 +
 [source,proto,indent=0]
 .src/main/proto/customer/view/customer_view.proto
 ----
-include::example$java-protobuf-eventsourced-customer-registry/src/main/proto/customer/view/customer_view.proto[tags=declarations;service-topic]
+include::example$java-protobuf-eventsourced-customer-registry-subscriber/src/main/proto/customer/view/customer_view.proto[tags=view-from-topic]
```
aludwiko, line 359: I'm not sure if we even need this code. The most important part is the producer, so the things related to the view can be summarized in one sentence: "The View definition must set `eventing.in` to a given topic name."

docs/src/modules/java-protobuf/pages/views.adoc
```diff
@@ -321,25 +321,54 @@ include::example$scala-protobuf-eventsourced-customer-registry/src/main/scala/cu
 [#topic-view]
 == Creating a View from a topic
 
-The source of a View can be an eventing topic. You define it in the same way as shown in <<event-sourced-entity>> or <<value-entity>>, but leave out the `eventing.in` annotation in the Protobuf file.
+The source of a View can be a broker topic.
+You define it in the same way as shown in <<event-sourced-entity>> or <<value-entity>>.
+Note that, on your producer side you must manually pass the `ce-subject` metadata, required by the View component.
```
aludwiko, line 326: That's frustrating, intellij editor is messing the the new lines since the latest update :/

PR #1658, opened 2023-05-31
docs: make workflow documentation publicly available

aludwiko:


docs/src/modules/java/nav.adoc
```diff
@@ -5,6 +5,7 @@
 *** xref:java:spring-boot-integration.adoc[Spring Boot Integration]
 *** xref:java:value-entity.adoc[Implementing Value Entities]
 *** xref:java:event-sourced-entities.adoc[Implementing Event Sourced Entities]
+*** xref:java:workflow-entities.adoc[Implementing Workflow Entities]
```
aludwiko, line 8: Do I need to do anything else?

efgpinto: yes, this does not matter much actually (only used for building docs locally).. the actual change needs to be done on `kalix-docs` repo

efgpinto (approved):

docs/src/modules/java/nav.adoc
aludwiko: like this: https://github.com/lightbend/kalix-docs/pull/1787/files ?

PR #1659, opened 2023-05-31
chore: update proxy version to 1.1.12

octonato:
New docker-compose-integration.yml were left out from bot PR


efgpinto (approved):
We might need to update the bot script to include this file. Will open a issue for that.

octonato:
> We might need to update the bot script to include this file. Will open a issue for that.

Correct. Can try to match any file like `docker-compose*.yml` and then we will kind of future-proof.

PR #1663, opened 2023-06-01
feat: eventing teskit for java sdk

efgpinto:
Opened this draft PR on the side to track the discussion on what would mean to have no upper type on the `expect` methods.

This would mean that both users from protocol-first or code-first would use the same API. I think this could work but opening for discussion.

~This is draft only, so might be missing some edge cases.~

### DIFF
--- a/samples/java-spring-eventsourced-counter/src/it/java/com/example/CounterIntegrationTest.java
+++ b/samples/java-spring-eventsourced-counter/src/it/java/com/example/CounterIntegrationTest.java
@@ -74,7 +74,7 @@ public class CounterIntegrationTest extends KalixIntegrationTestKitSupport {
     }
 
     @Test
-    public void verifyCounterEventSourcedPublishToTopic() throws JsonProcessingException {
+    public void verifyCounterEventSourcedPublishToTopic() {
 
         var counterId = "pubsub-test";
         var increaseCmd1 = new CounterCommandFromTopicAction.IncreaseCounter(counterId, 3);
--- a/samples/java-spring-eventsourced-counter/src/it/java/com/example/CounterIntegrationTest.java
+++ b/samples/java-spring-eventsourced-counter/src/it/java/com/example/CounterIntegrationTest.java
@@ -1,17 +1,24 @@
 package com.example;
 
+import com.fasterxml.jackson.core.JsonProcessingException;
+import kalix.javasdk.testkit.EventingTestKit;
+import kalix.javasdk.testkit.KalixTestKit;
 import kalix.spring.testkit.KalixIntegrationTestKitSupport;
 
 import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.BeforeAll;
 import org.junit.jupiter.api.Test;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.context.annotation.Import;
 import org.springframework.boot.test.context.SpringBootTest;
+import org.springframework.test.annotation.DirtiesContext;
 import org.springframework.web.reactive.function.client.WebClient;
 
 import java.time.Duration;
 
 import static java.time.temporal.ChronoUnit.SECONDS;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
 // tag::class[]
 @SpringBootTest(classes = Main.class)
 @Import(TestkitConfig.class)  // <1>
@@ -22,7 +29,18 @@ public class CounterIntegrationTest extends KalixIntegrationTestKitSupport {
     @Autowired
     private WebClient webClient;
 
+    @Autowired
+    private KalixTestKit kalixTestKit;
+
     private Duration timeout = Duration.of(10, SECONDS);
+
+    private EventingTestKit.Topic outTopic;
+
+    @BeforeAll
+    public void beforeAll() {
+        outTopic = kalixTestKit.getTopic("counter-events");
+    }
+
 
     @Test
     public void verifyCounterEventSourcedWiring() {
@@ -51,6 +69,34 @@ public class CounterIntegrationTest extends KalixIntegrationTestKitSupport {
                 webClient.get().uri("/counter/hello").retrieve().bodyToMono(String.class).block(timeout);
 
         Assertions.assertEquals("\"200\"", counterGet);
+    }
+
+    @Test
+    public void verifyCounterEventSourcedPublishToTopic() throws JsonProcessingException {
+
+        String counterIncrease =
+            webClient
+                .post()
+                .uri("/counter/to-topic/increase/10")
+                .retrieve()
+                .bodyToMono(String.class)
+                .block(timeout);
+        assertEquals("\"10\"", counterIncrease);
+
+        String counterMultiply =
+            webClient
+                .post()
+                .uri("/counter/to-topic/multiply/20")
+                .retrieve()
+                .bodyToMono(String.class)
+                .block(timeout);
+        assertEquals("\"200\"", counterMultiply);
+
+        var eventIncreased = outTopic.expectOneTyped(CounterEvent.ValueIncreased.class);
+        assertEquals(new CounterEvent.ValueIncreased(10), eventIncreased.getPayload());
+
+        var eventMultiply = outTopic.expectOneTyped(CounterEvent.ValueMultiplied.class);
+        assertEquals(new CounterEvent.ValueMultiplied(20), eventMultiply.getPayload());
     }
 // tag::class[]
 }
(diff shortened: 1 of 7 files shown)


samples/java-spring-eventsourced-counter/src/it/java/com/example/CounterIntegrationTest.java
```diff
@@ -52,6 +70,34 @@ public void verifyCounterEventSourcedWiring() {
...
+                .retrieve()
+                .bodyToMono(String.class)
+                .block(timeout);
+        assertEquals("\"200\"", counterMultiply);
+
+        var eventIncreased = outTopic.expectOneTyped(CounterEvent.ValueIncreased.class);
+        assertEquals(new CounterEvent.ValueIncreased(10), eventIncreased.getPayload());
+
+        var eventMultiply = outTopic.expectOneTyped(CounterEvent.ValueMultiplied.class);
+        assertEquals(new CounterEvent.ValueMultiplied(20), eventMultiply.getPayload());
```
efgpinto, line 99: note this is a java sdk and we are able to use the same exact API

sdk/java-sdk-protobuf-testkit/src/main/scala/kalix/javasdk/testkit/impl/EventingTestKitImpl.scala
```diff
@@ -150,37 +150,48 @@ class TopicImpl(private val probe: TestProbe, codec: MessageCodec) extends Topic
...
+    val contentType = metadata.get("Content-Type").asScala
+
+    (ceType, contentType) match {
+      case (_, Some("text/plain; charset=utf-8")) => "type.kalix.io/string"
+      case (_, Some("application/octet-stream"))  => "type.kalix.io/bytes"
+      case (Some(t), Some("application/json"))    => s"json.kalix.io/$t"
+      case (Some(t), _)                           => s"type.googleapis.com/$t"
+      case (t, ct) =>
+        log.warn(s"Could not extract typeUrl from ce-type=$t content-type=$ct")
+        ""
```
efgpinto, line 193: we are probably still missing some edge cases here I think.. I will need to double check and cover with some tests.

sdk/java-sdk-protobuf-testkit/src/main/scala/kalix/javasdk/testkit/impl/EventingTestKitImpl.scala
```diff
@@ -150,37 +150,48 @@ class TopicImpl(private val probe: TestProbe, codec: MessageCodec) extends Topic
...
-  override def expectOneTyped[T <: GeneratedMessageV3](clazz: Class[T], timeout: time.Duration): TestKitMessage[T] = {
+  override def expectOneTyped[T](clazz: Class[T], timeout: time.Duration): TestKitMessage[T] = {
     val msg = probe.expectMsgType[EmitSingleCommand]
     val metadata = new MetadataImpl(msg.getMessage.getMetadata.entries)
-    val decodedMsg = codec.decodeMessage(ScalaPbAny(typeUrlFor(metadata), msg.getMessage.payload))
+    val scalaPb = ScalaPbAny(typeUrlFor(metadata), msg.getMessage.payload)
+
+    val decodedMsg = if (typeUrlFor(metadata).startsWith(JsonSupport.KALIX_JSON)) {
+      val objectMapper = JsonSupport.getObjectMapper
+      objectMapper.readerFor(clazz).readValue(msg.getMessage.payload.toByteArray)
```
octonato, line 163: We can use `JsonSupport.decodeJson` directly and it has already some extra logic to deal with primitive bytes

efgpinto: Yeah, was the first approach actually but somehow it fails when deserializing the message.. as if the format was not correct. Haven't found the reason why. Will give it another try.

octonato: Have you tried it again? 
Was that related to the extra characters in front of the payload?

sdk/java-sdk-protobuf-testkit/src/main/scala/kalix/javasdk/testkit/impl/EventingTestKitImpl.scala
```diff
@@ -322,15 +335,23 @@ private[testkit] class TopicImpl(
   override def publish(message: ByteString, metadata: SdkMetadata): Unit =
     brokerProbe.emit(message, metadata)
 
-  override def publish[T <: GeneratedMessageV3](message: TestKitMessage[T]): Unit =
-    publish(message.getPayload.toByteString, message.getMetadata)
+  override def publish(message: TestKitMessage[?]): Unit = message.getPayload match {
```
octonato, line 338: I don't remember to have ever seen `[?]` usage in Scala. Is that the same as `[_]`?

sdk/java-sdk-protobuf-testkit/src/main/scala/kalix/javasdk/testkit/impl/EventingTestKitImpl.scala
```diff
@@ -322,15 +335,23 @@ private[testkit] class TopicImpl(
...
+    case javaPb: GeneratedMessageV3 => publish(javaPb.toByteString, message.getMetadata)
+    case scalaPb: GeneratedMessage  => publish(scalaPb.toByteString, message.getMetadata)
+    case str: String                => publish(ByteString.copyFromUtf8(str), message.getMetadata)
+    case _ =>
+      val encodedMsg = JsonSupport.getObjectMapper
+        .writerFor(message.getPayload.getClass)
+        .writeValueAsBytes(message.getPayload)
+      publish(ByteString.copyFrom(encodedMsg), message.getMetadata)
```
octonato, line 346: Same comment as above, we should use `JsonSupport.encodeJson` instead. We must understand why this is not working.

octonato (commented):
Looking good. 

Good to not depend on `GeneratedMessageV3`

sdk/java-sdk-protobuf-testkit/src/main/scala/kalix/javasdk/testkit/impl/EventingTestKitImpl.scala
efgpinto: Me neither, it was a typo but now I learnt that's what we'll actually need to use in scala 3... see [this](https://docs.scala-lang.org/scala3/reference/changed-features/wildcards.html#)

sdk/java-sdk-protobuf-testkit/src/main/scala/kalix/javasdk/testkit/impl/EventingTestKitImpl.scala
efgpinto: So, after some experiments, I think it makes sense to not use `JsonSupport.decodeJson` here as that expects the payload to be encoded with the special tag. However, when we publish the messages out to a broker we don't include such tag, and that's why using the `decodeJson` would break here. To make it work, we would need to manually wrap the payload with `AnySupport.encodingBytesPrimitive` in line 271 and then we could use the `decodeJson` but I don't see a point in doing that.

sdk/java-sdk-protobuf-testkit/src/main/scala/kalix/javasdk/testkit/impl/EventingTestKitImpl.scala
efgpinto: Similarly to the other occurrence but reversed, here we are offering to serialize the object for the user and in a real scenario where the user is writing to a topic, the payload will not be preceded by the special proto tag that "encodeJson" includes, so I don't think we should use it here at all. Also, note that the user is not forced to use our serialization, it can use `publish(ByteString, Metadata)` and push whatever bytes they want directly.

johanandren: This seams reasonable, given that `publish(Array[Byte])` puts exactly those bytes in the pretend topic, and presents those bytes to eventing in the proxy, just like if they came from a message broker. The proxy will then TLA-encode the bytes for passing around and possibly hand those bytes to an action for example.

Same should go in the other direction, the proxy should decode `Any` with our custom type_url into non-prefixed bytes when publishing to a topic.

There might be something slightly related to attaching metadata with content type `application/json` when auto-transforming like this and just publishing the raw bytes like if to a broker topic.

johanandren: I see that metadata magic is in the TestkitMessageImpl

octonato (approved):
LGTM

PR #1664, opened 2023-06-01
chore: add env var to debug tests

efgpinto:
This is very useful when debugging failures in integration tests and having a env var makes it easier to turn on/off. I guess we can use this ENV var to turn on other things as well in future, thus why I made it about debugging and not about forwarding logs specifically.

### DIFF
--- a/sdk/java-sdk-protobuf-testkit/src/main/java/kalix/javasdk/testkit/KalixProxyContainer.java
+++ b/sdk/java-sdk-protobuf-testkit/src/main/java/kalix/javasdk/testkit/KalixProxyContainer.java
@@ -20,6 +20,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.testcontainers.Testcontainers;
 import org.testcontainers.containers.GenericContainer;
+import org.testcontainers.containers.output.Slf4jLogConsumer;
 import org.testcontainers.containers.wait.strategy.Wait;
 import org.testcontainers.utility.DockerImageName;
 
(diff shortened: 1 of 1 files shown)


efgpinto:
We could probably document these small tricks that will help debugging locally, maybe in the upcoming section of the docs for tests? refs https://github.com/lightbend/kalix/issues/8598

sdk/java-sdk-protobuf-testkit/src/main/java/kalix/javasdk/testkit/KalixProxyContainer.java
```diff
@@ -92,8 +93,10 @@ public void start() {
     Testcontainers.exposeHostPorts(googlePubSubPort);
     super.start();
     // Debug tooling: pass the Proxy logs into the client SLF4J
-    // Slf4jLogConsumer logConsumer = new Slf4jLogConsumer(LoggerFactory.getLogger("proxy-logs"));
-    // followOutput(logConsumer);
+    if ("true".equals(System.getenv("KALIX_TESTKIT_DEBUG"))) {
+      Slf4jLogConsumer logConsumer = new Slf4jLogConsumer(LoggerFactory.getLogger("proxy-logs"));
+      followOutput(logConsumer);
```
octonato, line 98: Looks good, but I remember having done something similar in the past and struggling to get the output in a convenient format. 

If I'm not mistaken, I had to alter the docker to include another logback.xml with debug enabled. We also want the logs to not use json. The proxy has its own logback-dev-mode.xml, but we need to configure it explictly.

octonato: Well, maybe changing the debug level for the proxy doesn't make much sense for the users as they won't be able to do much, only infer what might be wrong on their own application.

octonato (approved):
LGTM

but might be needed to configure the proxy to use logback-dev-mode.xml

efgpinto:
Let me double check, I could have been using a ENV VAR with a modified proxy version and thats why I was assuming this was the only thing needed.

efgpinto:
AFAIK, it seems to work fine without any other change (see 2nd line below).

```shell
2023-06-01 15:05:38,539 INFO  kalix.javasdk.impl.DiscoveryImpl - Info reported from Kalix system:  Service started and reachable at 0.0.0.0:9000
2023-06-01 15:05:38,580 INFO  proxy-logs - STDOUT: 14:05:38.431 INFO  kalix.proxy.eventing.EventingManager - Eventing Manager StartConsumer for component [com.example.action.DoubleCounterAction], event source options: [EventSource(,ValueEntity(counter),false,None,None,UnknownFieldSet(Map()))], topicSupport: [Some(kalix.proxy.eventing.GooglePubSubEventingSupport@1f0cf458)], eventSourcedEntitySupport: [Some(kalix.proxy.eventing.EventSourcedEntityEventing@536680ce)], valueEntitySupport: [Some(kalix.proxy.eventing.ValueEntityEventing@5631c7e8)]

```

PR #1665, opened 2023-06-01
chore: fix double counter action in sample

efgpinto:
This action was not working (found an error on the logs when running the IT test) because it didn't match the Entity state nor the endpoint it exposes.


octonato (approved):
LGTM

Issue #1329, opened 2022-11-03
feat: Allow to consume raw bytes or text on Subscription in the Spring SDK


franciscolopezsancho:
The proxy redirects the raw bytes to a gRPC method in a component if it has input type `type.googleapis.com/com.google.protobuf.BytesValue` or in the case of `text` to a method that has input type`"type.googleapis.com/com.google.protobuf.StringValue`.  Currently, the Spring SDK doesn't generate Descriptors with these input types. 

To make this work, when a user annotates a method with a Subscription that has input type `Array[Byte]` or `String` the Spring SDK needs to generate an RPC method that has  "com.google.protobuf.BytesValue" or "com.google.protobuf.BytesValue".

Issue #1024, opened 2022-06-15
CloudEvent API doesn't make sense


jroper:
We always add CloudEvent information to outgoing events, including generating an id, setting the type, and setting the source. A user should be able to rely on that being set by us, and should be able to set additional cloud event metadata, without having to set that themselves. The problem is that right now, in order for a user to set CloudEvent metadata on an outgoing event using our API, they need to set the id, type and source.

There are three different methods we provide that allow them to get an instance of CloudEvent.

The first is `Metadata.asCloudEvent()`. This method will throw an exception if the underlying metadata doesn't contain a CloudEvent id, type or source. Since when publishing, you will typically start with `Metadata.EMPTY`, this is going to be the case.

The second is `Metadata.asCloudEvent(String, String, String)`. This requires explicitly passing an id, type or source.

The third is `CloudEvent.of(String, String, String)`, which again requires explicitly passing an id, type or source.

So, if all I want to do is set the CloudEvent subject on an outgoing message, and let Kalix set the remaining attributes - I can't.

johanandren:
So, we deprecate, and add a `Metadata#withSubjectId(id)` or something like that?

patriknw:
This is what I had to do:
```
    var metadata = actionContext().metadata().set("ce-subject", request.getTransactionId());

    return effects()
        .reply(topicTransaction, metadata);
```

jroper:
But what about other CloudEvent metadata? There's all sorts of CloudEvent data users might want to set.

I'm not sure though if we want to pass all metadata along as is - technically I think these are different events and should have different cloud event IDs, plus the incoming source references the event source entity type, not something we probably want to expose publicly.

Also, the fact that we encode the CloudEvent subject as `ce-subject`, while consistent with most CloudEvent header based encodings, is supposed to be more of an implementation detail.

johanandren:
Very problematic to pass along incoming cloud metadata to the response as it includes `ce-type` for the incoming message type which when returned in metadata overrides the type of the response message and leads to a downstream cloud event where metadata does not match the payload.

jroper:
Yes, so this is why I think we should make it easy for users to just set a subset of CloudEvent metadata that they want to send, and allow other CloudEvent metadata, such as `ce-type`, be set by the proxy. The current API doesn't allow that, it forces the user to set a type, id and source.

aludwiko:
[...]

PR #1668, opened 2023-06-02
feat: code first sdk raw bytes pub/sub

aludwiko:
References #1329

### DIFF
--- a/docs/src/modules/java/pages/serialization.adoc
+++ b/docs/src/modules/java/pages/serialization.adoc
@@ -29,7 +29,7 @@ As we mentioned it works out of the box.
 
 == Bytes
 
-If the content type is `application/octet-stream`, no content type is present, or the type is unknown to Kalix the message is treated as a binary message. The topic subscriber method must accept the `byte[]` message.
+If the content type is `application/octet-stream`, no content type is present, or the type is unknown to Kalix, the message is treated as a binary message. The topic subscriber method must accept the `byte[]` message.
 
 [source,java]
 .src/main/java/com/example/actions/SubscribeRawBytesAction.java
--- a/docs/src/modules/java/pages/serialization.adoc
+++ b/docs/src/modules/java/pages/serialization.adoc
@@ -27,6 +27,19 @@ include::example$java-spring-eventsourced-counter/src/main/java/com/example/acti
 
 As we mentioned it works out of the box.
 
+== Bytes
+
+If the content type is `application/octet-stream`, no content type is present, or the type is unknown to Kalix the message is treated as a binary message. The topic subscriber method must accept the `byte[]` message.
+
+[source,java]
+.src/main/java/com/example/actions/SubscribeRawBytesAction.java
+----
+include::example$java-spring-eventsourced-counter/src/main/java/com/example/actions/SubscribeToRawBytesAction.java[tag=class]
+----
+<1> When consuming raw bytes messages from a topic the input type must be `byte[]`.
+
+If an action has a return type of `byte[]` and publishes to a topic, the events published to the topic will have content-type `application/octet-stream`.
+
 == Receiving CloudEvents
 Kalix uses the https://cloudevents.io/[CloudEvents] standard when receiving from and publishing to topics. The CloudEvents specification standardizes message metadata so that systems can integrate more easily.
 
--- a/sdk/java-sdk-spring/src/it/java/com/example/wiring/pubsub/PubSubIntegrationTest.java
+++ b/sdk/java-sdk-spring/src/it/java/com/example/wiring/pubsub/PubSubIntegrationTest.java
@@ -29,6 +29,9 @@ import java.time.Instant;
 import java.util.List;
 import java.util.concurrent.TimeUnit;
 
+import static com.example.wiring.pubsub.PublishBytesToTopic.CUSTOMERS_BYTES_TOPIC;
+import static com.example.wiring.pubsub.PublishTopicToTopic.CUSTOMERS_2_TOPIC;
+import static com.example.wiring.pubsub.PublishVEToTopic.CUSTOMERS_TOPIC;
 import static java.time.temporal.ChronoUnit.SECONDS;
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.awaitility.Awaitility.await;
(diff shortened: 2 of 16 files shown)


sdk/java-sdk-spring/src/it/java/com/example/wiring/pubsub/PubSubIntegrationTest.java
```diff
@@ -129,6 +129,30 @@ public void shouldVerifyActionSubscribingToCustomersTopic() {
       });
   }
 
+  @Test
+  public void shouldVerifyActionSubscribingAndPublishingRawBaytes() {
```
octonato, line 133:
```suggestion
  public void shouldVerifyActionSubscribingAndPublishingRawBytes() {
```

sdk/java-sdk-spring/src/it/java/com/example/wiring/pubsub/SubscribeToBytesFromTopic.java
```diff
@@ -0,0 +1,62 @@
...
+@Subscribe.Topic(CUSTOMERS_BYTES_TOPIC)
+public class SubscribeToBytesFromTopic extends Action {
+
+  private Logger logger = LoggerFactory.getLogger(getClass());
+
+  public Effect<Done> handleChange(byte[] payload) {
+    logger.info("Consuming raw bytes: " + new String(payload));
+    try {
+      Object o = JsonSupport.getObjectMapper().readerFor(CustomerEntity.Customer.class).readValue(payload);
+      System.out.println(o);
```
octonato, line 46: leftover?

octonato: btw, don't you need to add the customer in the DummyCustomerStore?

That was the goal of this test, no?

sdk/java-sdk-spring/src/it/java/com/example/wiring/pubsub/SubscribeToBytesFromTopic.java
```diff
@@ -0,0 +1,62 @@
...
+@Profile("docker-it-test")
+@Subscribe.Topic(CUSTOMERS_BYTES_TOPIC)
+public class SubscribeToBytesFromTopic extends Action {
+
+  private Logger logger = LoggerFactory.getLogger(getClass());
+
+  public Effect<Done> handleChange(byte[] payload) {
+    logger.info("Consuming raw bytes: " + new String(payload));
+    try {
+      Object o = JsonSupport.getObjectMapper().readerFor(CustomerEntity.Customer.class).readValue(payload);
```
octonato, line 45:
```suggestion
      Customer customer = JsonSupport.getObjectMapper().readerFor(CustomerEntity.Customer.class).readValue(payload);
```
and we should use JsonSupport.decode method

octonato (commented):
looking good

sdk/java-sdk-spring/src/it/java/com/example/wiring/pubsub/PublishBytesToTopic.java
```diff
@@ -0,0 +1,47 @@
...
+@Subscribe.ValueEntity(CustomerEntity.class)
+public class PublishBytesToTopic extends Action {
+
+  public static final String CUSTOMERS_BYTES_TOPIC = "customers_bytes";
+  private Logger logger = LoggerFactory.getLogger(getClass());
+
+  @Publish.Topic(CUSTOMERS_BYTES_TOPIC)
+  public Action.Effect<byte[]> handleChange(CustomerEntity.Customer customer) {
+    try {
+      var payload = JsonSupport.getObjectMapper().writerFor(CustomerEntity.Customer.class).writeValueAsBytes(customer);
```
aludwiko, line 40: @octonato you suggested using `JsonSupport.decode` but then I would need to wrap bytes with Any, which I think makes no sense for this one.

octonato: I see, but my comment was also misplaced. It is not `decode`, but `encode`. But doesn't matter. I see your point. How the user will take a payload and put it in bytes, it's their choice. Here we are just testing it.

octonato: one more thought...

What happens if I subscriber to a topic that is emitting Protobuf (messages `Foo` and `Bar`)? 

I have the proto descriptor. I generate the message in my project and I create an Action with two methods receiving `Foo` and `Bar`. 

The Java SDK converts it to one method receiving `Any`, but it will expect it to be serialized to Json. The SDK will take the incoming Any payload and try to use the types `Foo.class` and `Bar.class` to deserialize it using the `JsonMessageCodec`.

We look already inside the Any to see if it is a json, but it's not, it fails. We need to be able to fallback to proto in that case.

aludwiko: Just to sure that I understand your use case. One Kalix service produces proto messages to a topic. The second one, which is code-first, wants to consume from that topic. Am I correct? Looks like a separate issue to solve.

octonato: Yes, more or less that. Don't even need to be another Kalix service. It suffices to have a Topic with proto and the consumer has the spec to generate the messages to consume. 

A Kalix user might expect it to work out of the box by adding methods receiving the proto messages directly.

aludwiko: Now I'm confused. If the producer is not a Kalix service then we can't force to publish additional metadata like `ce-type` or `ce-datacontenttype` required for routing to a specific handler.

octonato: Ok, so maybe I'm confusing something there. I thought that the proxy would be able to dispatch the messages based on the type, but that's won't work out to the box and the expected behavior in such a case is to receive the raw bytes and the user have to deal with it themselves.

aludwiko: Yes, the proxy is able to do this based on additional metadata. Moreover, currently, we are passing only the metadata from the CloudEvent spec. For raw consumption, I would expect to get all the metadata from the topic. https://github.com/lightbend/kalix-jvm-sdk/issues/1024#issuecomment-1576327397

docs/src/modules/java/pages/serialization.adoc
```diff
@@ -27,6 +27,19 @@ include::example$java-spring-eventsourced-counter/src/main/java/com/example/acti
 
 As we mentioned it works out of the box.
 
+== Bytes
```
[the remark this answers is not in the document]
octonato: I think it's a sensible one.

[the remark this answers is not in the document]
octonato: btw, can be added in follow-up PR

octonato (approved):
LGTM

octonato (approved):
LGTM

PR #1671, opened 2023-06-03
Document use of logging.config JVM option.

max8github:
Option `-Dlogging.config=src/main/resources/logback-dev-mode.xml` was undocumented.


docs/src/modules/java/pages/running-locally.adoc
```diff
@@ -197,3 +197,19 @@ This time, `kalix:runAll` will start `docker-compose-integration.yml` instead. T
 Note that the configuration is exactly the same except that now we are running one single `mvn kalix:runAll` command and the `docker-compose` file we are using contains all the dependencies required by *customer-registry-subscriber*.
 
 This approach can be extended to any service you might want to integrate with. It can be any other Kalix service that you plan to deploy to the same Kalix project or even external services. The only requirement is to have a Docker image for it.
+
+== Logging Settings when Running Locally
```
octonato, line 201: It's good to document it, but it is given the impression that you always need it. This is only needed when running `Main.class` directly from an IDE. 

When running `mvn kalix:run` or `mvn kalix:runAll` (available since v1.2.1), you don't need to pass anything. It will configure the service to use `logback-dev-mode.xml` automatically. 

That said, I think this part of the docs must be more explicitly about when running from the IDE.

Issue #1667, opened 2023-06-02
error when having two Action components with the same name in two different packages


aludwiko:
action1:
```
package com.example.wallet.application;
@Subscribe.EventSourcedEntity(value = WalletEntity.class, ignoreUnknown = true)
public class NotifyAboutLowBalance extends Action {
```
action2
```
package com.example.wallet.ve;
@Subscribe.ValueEntity(WalletValueEntity.class)
public class NotifyAboutLowBalance extends Action {
```
exception:
```
2023-06-02 09:05:37,283 ERROR kalix.javasdk.impl.DiscoveryImpl - Error reported from Kalix system: KLX-00004 Unable to find service com.example.wallet.application.NotifyAboutLowBalance in the passed in descriptors.

The services found in the descriptor at discovery were:
  * com.example.wallet.api.WalletController
  * com.example.wallet.application.TransferFunds
  * com.example.wallet.application.WalletByBalance
  * com.example.wallet.application.WalletByOwner
  * com.example.wallet.application.WalletEntity
  * com.example.wallet.ve.NotifyAboutLowBalance
  * com.example.wallet.ve.WalletValueEntity
```

octonato:
Can it be that we have a map like `Map[SimpleClassName, Descriptor]` in the code-first SDK? 

I vaguely remember we needed to keep descriptors on a Map to find them later.

aludwiko:
It's more complex. On the sdk side, we have this change, from package + name to just a name:
https://github.com/lightbend/kalix-jvm-sdk/commit/5c9b2caa#diff-902d2d3964621479fb41d622d3fb3fcd8d07c15003d20862cc730bfe355669d1L138

Reverting SDK is not enough, the proxy must be adjusted as well: https://github.com/lightbend/kalix-proxy/blob/main/proxy/core/src/main/scala/kalix/proxy/FileDescriptorBuilder.scala#L21

Yeah, before I change anything I want to make sure that this is not a trap or sth.

PR #1675, opened 2023-06-06
fix: file description name with package name to avoid collisions

aludwiko:
References #1667

### DIFF
--- a/sdk/java-sdk-spring/src/it/java/com/example/wiring/actions/headers/EchoAction.java
+++ b/sdk/java-sdk-spring/src/it/java/com/example/wiring/actions/headers/EchoAction.java
@@ -22,7 +22,7 @@ import org.springframework.web.bind.annotation.GetMapping;
 import org.springframework.web.bind.annotation.PathVariable;
 
 /**
- * Action with the same in a different package.
+ * Action with the same name in a different package.
  */
 public class EchoAction extends Action {
 
--- a/sdk/java-sdk-spring/src/it/java/com/example/wiring/actions/headers/EchoAction.java
+++ b/sdk/java-sdk-spring/src/it/java/com/example/wiring/actions/headers/EchoAction.java
@@ -1,0 +1,33 @@
+/*
+ * Copyright 2021 Lightbend Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.example.wiring.actions.headers;
+
+import com.example.wiring.actions.echo.Message;
+import kalix.javasdk.action.Action;
+import org.springframework.web.bind.annotation.GetMapping;
+import org.springframework.web.bind.annotation.PathVariable;
+
+/**
+ * Action with the same in a different package.
+ */
+public class EchoAction extends Action {
+
+  @GetMapping("/echo2/message/{msg}")
+  public Effect<Message> stringMessage(@PathVariable String msg) {
+    return effects().reply(new Message(msg));
+  }
+}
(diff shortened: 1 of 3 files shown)


sdk/java-sdk-spring/src/main/scala/kalix/javasdk/impl/ProtoDescriptorGenerator.scala
```diff
@@ -48,7 +48,7 @@ private[impl] object ProtoDescriptorGenerator {
 
     val protoBuilder = DescriptorProtos.FileDescriptorProto.newBuilder
     protoBuilder
-      .setName(name + ".proto") // FIXME: snake_case this ?!
+      .setName(packageName.replace(".", "/") + "/" + name + ".proto") // FIXME: snake_case this ?!
```
aludwiko, line 51: I could do sth with the snake_case fixme, but I'm not if it's worth it.

aludwiko: Maybe some context would be useful to review it. During the discovery process, we are deduplicating file descriptors in `kalix.javasdk.impl.AnySupport#flattenDescriptors`. Without the package prefix, the first component with a given name wins.

sdk/java-sdk-spring/src/it/java/com/example/wiring/actions/headers/EchoAction.java
```diff
@@ -0,0 +1,33 @@
...
+
+package com.example.wiring.actions.headers;
+
+import com.example.wiring.actions.echo.Message;
+import kalix.javasdk.action.Action;
+import org.springframework.web.bind.annotation.GetMapping;
+import org.springframework.web.bind.annotation.PathVariable;
+
+/**
+ * Action with the same in a different package.
```
franciscolopezsancho, line 25:
```suggestion
 * Action with the same name in a different package.
```

Or maybe something similar to `Action with the same name as com.example.wiring.actions.echo.ActionEcho` shouldn't raise `kalix.proxy.UserFunctionErrors$.unableToFindServiceInDescriptor`. I'm saying this to keep some context around for the future reader.

sdk/java-sdk-spring/src/main/scala/kalix/javasdk/impl/ProtoDescriptorGenerator.scala
franciscolopezsancho: If we don't tackle [the snake case](https://protobuf.dev/programming-guides/style/#file-structure) I wouldn't even do the `replace`.

franciscolopezsancho (approved):
LGTM

sdk/java-sdk-spring/src/main/scala/kalix/javasdk/impl/ProtoDescriptorGenerator.scala
octonato: I think I added this FIXME. Not sure what I was thinking then. I was probably trying to be overly precise (and failing) on the generation and had the intention to make it look like a real proto file. 

Do we need to append `.proto`? If not, I would simply use the FQCN without bothering with replacing the dots. This is for internal usage, so we don't need to comply to any convention.

If we need the extension, then using `/` is preferable as it should look like a file location.

sdk/java-sdk-spring/src/it/java/com/example/wiring/actions/headers/EchoAction.java
```diff
@@ -0,0 +1,33 @@
...
+
+package com.example.wiring.actions.headers;
+
+import com.example.wiring.actions.echo.Message;
+import kalix.javasdk.action.Action;
+import org.springframework.web.bind.annotation.GetMapping;
+import org.springframework.web.bind.annotation.PathVariable;
+
+/**
+ * Action with the same name in a different package.
```
octonato, line 25: Could we have a test that verifies that the service has two Actions sharing the same name? I don't know how hard it will be to verify it, but it will be nice if we make it waterproof.

aludwiko: This class in some sense testing this already. We have the same class in the `com.example.wiring.actions.echo` package so without the fix, all integrations tests will fail, because of proxy exception. Is that enough?

octonato: If you remove this class tests will still pass. 

What I mean by waterproofing is to have a check that indeed the final descriptor do have two services with the same name.

But that might be hard to achieve.

aludwiko: ok, I get it, let me check what I can do

aludwiko: After a few attempts - I give up, maybe I will figure this out later. I only added one assertion for the filedescriptor name.

PR #1677, opened 2023-06-07
chore: logs in integration tests be plain text by default

### LINKED ISSUE #1676: Java sdk archetype not defaulting to plain text logs for Integration tests
Since we generate and example for an IntegrationTest we should also include the logback-test.xml file that makes the logs be in JSON format by default.

efgpinto:
Close #1676

### DIFF
--- a/maven-java/kalix-spring-boot-archetype/src/main/resources/archetype-resources/src/it/resources/logback-test.xml
+++ b/maven-java/kalix-spring-boot-archetype/src/main/resources/archetype-resources/src/it/resources/logback-test.xml
@@ -1,0 +1,18 @@
+<configuration>
+    <appender name="STDOUT" class="ch.qos.logback.core.ConsoleAppender">
+        <encoder>
+            <pattern>%date{ISO8601} %-5level %logger - %msg%n</pattern>
+        </encoder>
+    </appender>
+
+    <logger name="akka" level="INFO"/>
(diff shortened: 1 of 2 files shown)


aludwiko (approved):

octonato (approved):
LGTM

Can you also add this file to the Kotlin template?

efgpinto:
> Can you also add this file to the Kotlin template?

Added for kotlin template now.

> I just saw that we have it in src/test/resources in the protobuf templates, but we also need it in src/it/resources.

We just need it in one place I think. For code-first I put in the IT folder because we generate a sample test there by default (but could move it to `test`folder if you think it's better). For protocol-first, the one we have under `test` is also picked up when running integration tests. Thus, I'm not sure we necessarily need to have them in both places, do we?

octonato:
> Thus, I'm not sure we necessarily need to have them in both places, do we?

It depends if the IntegrationTest includes the Test resources. From what you describe, it seems it does.

efgpinto:
> > Thus, I'm not sure we necessarily need to have them in both places, do we?
> 
> It depends if the IntegrationTest includes the Test resources. From what you describe, it seems it does.

It does with this PR for code-first sdk. For protobuf, we have it under `test`. There is no specific folder for integration tests.

efgpinto:
I will merge this now and follow up if needed.

PR #1682, opened 2023-06-15
Docs

max8github:
Documentation corrections.

Notes:
- I tried to guess and put together two sentences that were not holding syntactically: you need to check if they do convey what they originally intended to convey.
- there was a naming conflict, which gets resolved by changing either the code or the doc: I chose to change the name in the code example


efgpinto (approved):
LGTM, thanks Max.

PR #1687, opened 2023-06-16
chore: renaming Workflows

octonato:
Draft because we still need to decide on the final name for the new annotations. 

`@EntityType` => `@Type` or `@TypeName`
`@EntityKey` => `@Key` or `@Id`

Also missing is the deprecation of `EntityType` and `EntityKey`. We might decide to introduce `WorkflowType` and `WorkflowKey` instead.


octonato:
My latest commit unintentionally mentioned users `@Key` and `@Id`. Sorry for that.

octonato:
Ok, we settled for TypeId and Id. 

I'm marking this PR as ready for review now. I believe that what we have here is good enough for the renaming part.  

I tried to keep it in three commits to make it easier to review. 
* Renaming workflows(0d54625ae0cfe3a4c02dc52e5d9ba63186f034b4)
* use TypeId and Id in the code base (084c9e2f4097831ca38f29e3797ec1ec7ce48ec3)
* adapt the samples to use new annotations (788d1eace15ff0917b92e96f8c2793904a1c5c37)

Here is a plan for follow-up PR and actions:

- [ ] Proxy must support two new gRPC annotations: `id` and `type_id`
- [ ] Update SDK with new proxy protocol and let code-first emit new annotations
- [ ] codegen should support new annotations and emit a warning if `entity_key` and `entity_type` is in use. 
- [ ] `ViewId` can eventually be deprecated in favour of `TypeId`

All of the above can be done later. We don't need to block a release. 

The deprecation of `ViewId` may be optional. We can give us some more time to think about it. I reason that `TypeId` + `Id` is the unique identifier of a stateful component. Views are a little different. There is no instance. We can let it sink for a while.

docs/src/modules/java/pages/event-sourced-entities.adoc
```diff
@@ -61,12 +61,12 @@ Let's have a look at what our shopping cart entity will look like for the first
 include::example$java-spring-eventsourced-shopping-cart/src/main/java/com/example/shoppingcart/ShoppingCartEntity.java[tag=class]
 ----
 <1> Create a class that extends `EventSourcedEntity<S, E>`, where `S` is the state type this entity will store (i.e. `ShoppingCart`) and `E` is the top type for the events it emits (i.e. `ShoppingCartEvent`).
-<2> Annotate such class with `@EntityKey`  and pass the name of the key that will be used as the entity unique identifier.
-<3> Make sure to annotate such class with `@EntityType` and pass a unique name for this entity type.
+<2> Annotate such class with `@Id`  and pass the name of the id that will be used as the entity instance unique identifier.
```
aludwiko, line 64:
```suggestion
<2> Annotate such class with `@Id` and pass the name of the id that will be used as the entity instance unique identifier.
```

docs/src/modules/java/pages/value-entity.adoc
```diff
@@ -41,8 +41,8 @@ Let us start by showing how to create the Value Entity:
 ----
 include::example$java-spring-valueentity-counter/src/main/java/com/example/CounterEntity.java[tags=declarations]
 ----
-<1> Every Entity must be annotated with `@EntityType` with a stable name. This name should be unique among the different existing entities within a Kalix application.
-<2> The `@EntityKey` value should be unique per entity and map to some field being received on the route path, in this example it's the `counter_id`.
+<1> Every Entity must be annotated with `@TypeId` with a stable identifier. This identifier should be unique among the different existing entities within a Kalix application.
```
aludwiko, line 44: I would change the order

aludwiko: But that's a lot of work. It's just a visual aspect.

aludwiko (approved):
LGTM, minor suggestions

docs/src/modules/java/pages/value-entity.adoc
octonato: Yeah, I agree, but will leave it as is. Not work to go change the callouts.

PR #1697, opened 2023-06-23
docs: use entity id instead of key

octonato:
I spotted a few more usages of 'key'.

We still mention it in the protobuf docs and glossary, we will be able to change it once in have the new annotations in the proxy protocol.

### DIFF
--- a/docs/src/modules/java/partials/entity-ids.adoc
+++ b/docs/src/modules/java/partials/entity-ids.adoc
@@ -15,7 +15,7 @@ For instance, `@Id("id")` will instruct Kalix to look up a matching path variabl
 
 === Composite identifiers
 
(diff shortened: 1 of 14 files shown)


efgpinto (approved):
LGTM

docs/src/modules/java/partials/entity-ids.adoc
```diff
@@ -8,16 +8,16 @@ In order to interact with an Entity in Kalix, we need to assign an *type id* and
...
 The entity id can be defined in different ways, as detailed below.
 
-=== Single keys
+=== Single identifier
 
 The most common use is to annotate the class with `@Id` and assign one path variable name to it.
 For instance, `@Id("id")` will instruct Kalix to look up a matching path variable. For an endpoint defined with `@RequestMapping("/users/\{id}")`, Kalix will extract whatever path segment is used to replace `\{id}` and treat it as the Entity unique identifier.
 
-=== Composite keys
+=== Composite identifiers
```
octonato, line 16:
```suggestion
=== Composite identifier
```

PR #1699, opened 2023-06-23
docs: add section on how to identify Workflows

octonato:


docs/src/modules/java/pages/workflows.adoc
```diff
@@ -6,6 +6,33 @@ include::ROOT:partial$include.adoc[]
...
 
+
+== Identifying the Workflow
+
+In order to interact with a Workflow in Kalix, we need to assign an *type id* and one or more instance *ids*:
+
+* *type id* is a unique identifier to the workflow. To define the workflow type id, the workflow class must be annotated with `@TypeId` and have a unique and stable identifier assigned.
```
octonato, line 14: @aludwiko, here is the point I realised that workflows are sharing the same space as ES entities in the Proxy. 

It will be better if we prefix the persistence id for workflows internally (in the proxy) so we are sure that they never conflict (until a user prefixes their own entity type id themselves with the same prefix 🤦 ).

aludwiko: As discussed, a prefix and a quick deployment are possible, because I'm the only one using this entity on prod (confirmed based on metrics). On the other hand, maybe we should not prefix it and create a separate table for events from workflows. I think it's possible, but I'm not sure about the overall ops cost of such a change.

octonato: Another table would be nicer but increases of ops cost on our side with no real benefit for users. Not only we will need to manage another table, but there are also other things to put in place, like tracking usage, billing, etc.

efgpinto (approved):
LGTM

PR #1718, opened 2023-06-29
fix!: update enum for gpubsub and add docs

### LINKED ISSUE #1694: Document how to run integration tests against a real broker
The default broker used for integration tests has changed to the internal test broker but we added an option so each tests can still be configured to continue using google pub sub emulator as previously.

We need to document it though.

efgpinto:
Close #1694 

While adding docs on how to configure an integration test to run against a real pubsub instance, I noticed that the enum value we provide on the settings is wrong (we've called it `GOOGLE_PUBSUB_EMULATOR` instead of `GOOGLE_PUBSUB`).. Using "emulator" suffix does not make much sense because in theory you could also run against a real gpubsub instance.

So along with docs, I'm changing the enum as well.

This is a breaking change but.. this option/enum was something recently introduced in 1.3.0, was not documented properly and is only for testkits, so I believe it's not a big deal to change it. But let me know your thoughts.

### DIFF
--- a/docs/src/modules/java-protobuf/pages/actions-publishing-subscribing.adoc
+++ b/docs/src/modules/java-protobuf/pages/actions-publishing-subscribing.adoc
@@ -636,3 +636,23 @@ include::example$scala-protobuf-eventsourced-counter/src/test/scala/com/example/
 <2> Clear the topic ignoring any unread messages.
 
 NOTE: Despite the example, you are neither forced to clear all topics nor to do it before each test. You can do it selectively, or you might not even need it depending on your tests and the flows they test.
+
+=== External Broker
+
+To run an integration test against a real instance of Google PubSub (or its Emulator), use the TestKit settings to override the default eventing support, as shown below:
+
+[.tabset]
+Java::
++
+[source,java]
+private static final KalixTestKitResource testKit = new KalixTestKitResource(
+        Main.createKalix(),
+        Settings.DEFAULT.withEventingSupport(EventingSupport.GOOGLE_PUBSUB)
+    );
+Scala::
++
+[source,scala,indent=0]
+private val testKit = KalixTestKit(
+    Main.createKalix(),
+    DefaultSettings.withEventingSupport(GooglePubSub)
+  ).start()
--- a/samples/java-spring-eventsourced-counter/src/it/java/com/example/CounterIntegrationWithRealPubSubTest.java
+++ b/samples/java-spring-eventsourced-counter/src/it/java/com/example/CounterIntegrationWithRealPubSubTest.java
@@ -72,18 +72,21 @@ public class CounterIntegrationWithRealPubSubTest extends KalixIntegrationTestKi
     // builds a message in PubSub format, ready to be injected
     private String buildMessageBody(String jsonMsg, String ceType) {
         var data = Base64.getEncoder().encodeToString(jsonMsg.getBytes());
-        return "{\n" +
-            "    \"messages\": [\n" +
-            "        {\n" +
-            "            \"data\": \"" + data + "\",\n" +
-            "            \"attributes\": {\n" +
-            "                \"Content-Type\": \"application/json\",\n" +
-            "                \"ce-specversion\": \"1.0\",\n" +
-            "                \"ce-type\": \"" + ceType + "\"\n" +
-            "            }\n" +
-            "        }\n" +
-            "    ]\n" +
-            "}";
+
+        return """
+            {
+                "messages": [
+                    {
+                        "data": "%s",
+                        "attributes": {
+                            "Content-Type": "application/json",
+                            "ce-specversion": "1.0",
+                            "ce-type": "%s"
+                        }
+                    }
+                ]
+            }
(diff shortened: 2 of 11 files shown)


samples/java-spring-eventsourced-counter/src/it/java/com/example/CounterIntegrationWithRealPubSubTest.java
```diff
@@ -0,0 +1,91 @@
...
+            "                \"ce-type\": \"" + ceType + "\"\n" +
+            "            }\n" +
+            "        }\n" +
+            "    ]\n" +
+            "}";
+    }
+
+// tag::class[]
+}
+// end::class[]
```
efgpinto, line 91: This file is not used in documentation, I just figured it could be good to have a sample were we actually mix both tests using the test broker and a real instance. And since for this sample we actually already run the gpubsub on the pipeline, I thought it was good to add it here.

samples/java-spring-eventsourced-counter/src/it/java/com/example/CounterIntegrationWithRealPubSubTest.java
```diff
@@ -0,0 +1,91 @@
...
+            "        {\n" +
+            "            \"data\": \"" + data + "\",\n" +
+            "            \"attributes\": {\n" +
+            "                \"Content-Type\": \"application/json\",\n" +
+            "                \"ce-specversion\": \"1.0\",\n" +
+            "                \"ce-type\": \"" + ceType + "\"\n" +
+            "            }\n" +
+            "        }\n" +
+            "    ]\n" +
+            "}";
```
efgpinto, line 86: Would this be better in a resources file? I opted like this because it's easier to read the test but it's not pretty.

samples/java-spring-eventsourced-counter/src/it/java/com/example/TestkitConfig.java
```diff
@@ -1,14 +1,33 @@
...
+
+    @Profile("with-acls")
+    // tag::acls[]
     @Bean
     public KalixTestKit.Settings settings() {
         return KalixTestKit.Settings.DEFAULT.withAclEnabled(); // <1>
     }
+    // end::acls[]
+
+    @Profile("with-pubsub")
```
efgpinto, line 22: note the use of profiles so we can using different settings for different tests

samples/java-spring-eventsourced-counter/src/it/java/com/example/CounterIntegrationWithRealPubSubTest.java
```diff
@@ -0,0 +1,91 @@
...
+
+import static java.time.temporal.ChronoUnit.SECONDS;
+import static org.awaitility.Awaitility.await;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+// tag::class[]
+@SpringBootTest(classes = Main.class)
+@Import(TestkitConfig.class)
+@ActiveProfiles("with-pubsub")
```
efgpinto, line 28: configuring the proper settings with pubsub eventing support

efgpinto (commented):
I'm not sure we need to do an immediate release with this. I think it's fine if this waits for 1.3.1 to go out TBH.

docs/src/modules/java-protobuf/pages/actions-publishing-subscribing.adoc
```diff
@@ -636,3 +636,23 @@ include::example$scala-protobuf-eventsourced-counter/src/test/scala/com/example/
...
+        Main.createKalix(),
+        Settings.DEFAULT.withEventingSupport(EventingSupport.GOOGLE_PUBSUB)
+    );
+Scala::
++
+[source,scala,indent=0]
+private val testKit = KalixTestKit(
+    Main.createKalix(),
+    DefaultSettings.withEventingSupport(GooglePubSub)
+  ).start()
```
efgpinto, line 658: Turns out that for protocol-first sdks the config is easier so I don't think we need to actually have samples for this. The snippets seem enough to me.

samples/java-spring-eventsourced-counter/src/it/java/com/example/CounterIntegrationWithRealPubSubTest.java
octonato: You can use triple quotes here. Available since Java 15.
https://www.baeldung.com/java-multiline-string#text-blocks

octonato (approved):
LGTM, but I would cut a 1.3.1 soonish. The sooner the better.

samples/java-spring-eventsourced-counter/src/it/java/com/example/CounterIntegrationWithRealPubSubTest.java
efgpinto: of course.. thanks, somehow I thought this was not available in this version.

octonato (approved):
LGTM

Issue #1710, opened 2023-06-27
Revisit how we are manage logback for dev and prod


octonato:
We currently ship a logback.xml file that is tailored for prod and an alternative one (logback-dev-mode.xml) tailored for local development. 

Our plugins (mvn and sbt) take care of it for the users. When starting the service on your local machine, we configured it to use `logback-dev-mode.xml`. 

Unfortunately, this file won't be used when running the Main function directly from inside the IDE. 

So, instead, we should have logback.xml configured for dev-mode and instead pass logback-prod.xml as an argument when we build the container. 

Related to https://github.com/lightbend/kalix-jvm-sdk/pull/1671

johanandren:
There is an SPI for config in logback, so it could perhaps be possible to use some Kalix JVM SDK impl of that, to do clever detection when starting: https://logback.qos.ch/manual/configuration.html#auto_configuration

PR #1719, opened 2023-06-29
fix: load logback-dev-mode.xml using Configurator

octonato:
Fixes https://github.com/lightbend/kalix-jvm-sdk/issues/1710

Also relates to https://github.com/lightbend/kalix-jvm-sdk/pull/1671
We don't need to document if we can load the right file in dev-mode.


maven-java/kalix-maven-archetype-event-sourced-entity/pom.xml
```diff
@@ -2,12 +2,12 @@
     <modelVersion>4.0.0</modelVersion>
 
     <artifactId>kalix-maven-archetype-event-sourced-entity</artifactId>
-    <version>1.3.0</version>
+    <version>1.3.0-4-deee4bb0-SNAPSHOT</version>
```
octonato, line 5: hmm, wrong commit

maven-java/kalix-maven-plugin/src/main/scala/kalix/RunMojo.scala
```diff
@@ -49,18 +48,7 @@ object RunMojo {
...
-    val loggingArgs: Seq[Element] = {
-      val logConfigFile = new File(logConfig.trim)
-      if (logConfigFile.exists) {
-        log.info(s"Using logging configuration file: '$logConfigFile' ")
-        // when using SpringBoot, logback config is passed using logging.config
-        element(name("argument"), "-Dlogging.config=" + logConfig) ::
-        element(name("argument"), "-Dlogback.configurationFile=" + logConfig) :: Nil
-      } else {
-        log.warn(s"Dev mode logging configuration file '$logConfig' not found")
-        List.empty
```
octonato, line 61: we were previously doing this kind of hack in the plugin. We don't need it anymore as the JoranConfigurator will do it as soon as LoggerFactory is used. And this is valid for both java and scala.

sdk/java-sdk-protobuf/src/main/scala/kalix/javasdk/impl/logging/KalixLogbackConfigurator.scala
```diff
@@ -0,0 +1,45 @@
...
+  System.setProperty("spring.main.banner-mode", "off")
+
+  if (devModeLogback.exists) {
+    // only set it if not already set by user
+    if (System.getProperty("logging.config") == null)
+      System.setProperty("logging.config", devModeLogbackFile)
+
+    // only set it if not already set by user
+    if (System.getProperty("logback.configurationFile") == null)
+      System.setProperty("logback.configurationFile", devModeLogbackFile)
```
octonato, line 41: When running with spring-boot, the fact that we `logback.configurationFile` cause a warning in the logs. We should find a proper way to identify that we are running with SpringBoot and skip this.

sdk/java-sdk-protobuf/src/main/scala/kalix/javasdk/impl/logging/KalixLogbackConfigurator.scala
```diff
@@ -0,0 +1,45 @@
...
+
+package kalix.javasdk.impl.logging
+
+import java.io.File
+
+import ch.qos.logback.classic.LoggerContext
+import ch.qos.logback.classic.spi.Configurator
+import ch.qos.logback.core.spi.ContextAwareBase
+
+class KalixLogbackConfigurator extends ContextAwareBase with Configurator {
```
patriknw, line 25: now I understand, I thought we set the System.property from `main`, but I guess this is guaranteed to be loaded first before logback is initialized

sdk/java-sdk-protobuf/src/main/scala/kalix/javasdk/impl/logging/KalixLogbackConfigurator.scala
```diff
@@ -0,0 +1,45 @@
...
+
+  private val devModeLogbackFile = "src/main/resources/logback-dev-mode.xml"
+  // this file path only exist in dev-mode. Once packaged we don't have src/main/resources anymore
+  private val devModeLogback = new File(sys.props("user.dir"), devModeLogbackFile)
+
+  // not really related to logging, but we can disable it here as well
+  System.setProperty("spring.main.banner-mode", "off")
+
+  if (devModeLogback.exists) {
```
patriknw, line 34: what happens when running tests? not using logback-test.xml any more, or are these system properties set by the build tool when running tests?

patriknw (commented):
looking good

sdk/java-sdk-protobuf/src/main/scala/kalix/javasdk/impl/logging/KalixLogbackConfigurator.scala
octonato: Hmm, I think that might be an issue. 

If I recall correctly the ordering is: 

* check system property
* if absent, `logback-test.xml`
* if absent use `logback.xml`

I will verify this.

octonato: yep, that's the case. If `-Dlogback.configurationFile` is set, it gets precedence.

And we have been doing so for a while. 

Before, we were passing `-Dlogback.configurationFile` in the pom.xml/build.sbt. Now we are passing it from inside the maven/sbt plugins. In this PR, we move it to the logback configurator. 

So, `src/main/resources/logback-dev-mode.xml` is always getting precedence over `src/test/resources/logback-test.xml`.

We can fix it by using `src/test/resources/logback-test.xml`, if present, in dev-mode, and falling back to `src/main/resources/logback-dev-mode.xml` if not present. 

Next, we can update our templates to not include `logback-dev-mode.xml` anymore, but always include a `logback-test.xml`. Arguably, dev-mode is also a kind of test run.

octonato: an alternative is to only configure `logback-dev-mode.xml` if `logback-test.xml` is not in the classpath. If `logback-test.xml` is in the classpath, it means that we are running the tests and therefore we should favour `logback-test.xml`.

sdk/java-sdk-protobuf/src/main/resources/META-INF/services/ch.qos.logback.classic.spi.Configurator
```diff
@@ -1 +1 @@
-kalix.javasdk.impl.logging.KalixLogbackConfigurator
\ No newline at end of file
+kalix.javasdk.impl.logging.KalixJoranConfigurator
```
octonato, line 1: Made it a JoranConfigurator and extend the DefaultJoranConfigurator

sdk/java-sdk-protobuf/src/main/scala/kalix/javasdk/impl/logging/KalixJoranConfigurator.scala
```diff
@@ -0,0 +1,79 @@
...
+      val myClassLoader = Loader.getClassLoaderOfObject(this)
+      val url = Loader.getResource(DefaultJoranConfigurator.TEST_AUTOCONFIG_FILE, myClassLoader)
+      url == null
+    }
+
+    // this file path only exist when running in dev-mode.
+    // once packaged we don't have src/main/resources anymore
+    def devModeLogbackExists = new File(sys.props("user.dir"), devModeLogbackFile).exists()
+
+    noLogbackTest && devModeLogbackExists
```
octonato, line 47: giving precedence to `logback-test.xml` if on the classpath.

sdk/java-sdk-protobuf/src/main/scala/kalix/javasdk/impl/logging/KalixJoranConfigurator.scala
```diff
@@ -0,0 +1,79 @@
...
+        logbackConfigurationFile = true
+      }
+
+      // let the DefaultJoranConfigurator do its thing
+      val status = super.configure(loggerContext)
+
+      // we can only log after we run 'configure'
+      addInfo("Kalix application running in dev-mode");
+      if (loggingConfigSet) addInfo("Setting logging.config to " + devModeLogbackFile);
+      if (logbackConfigurationFile) addInfo("Setting logback.configurationFile to " + devModeLogbackFile);
```
octonato, line 71: I thought we would get some logging here, but apparently, nothing is being printed. 

What is interesting is that when using `DefaultJoranConfigurator` instead of our extension, we do get some logs.

sdk/java-sdk-protobuf/src/main/scala/kalix/javasdk/impl/logging/KalixLogbackConfigurator.scala
patriknw: yeah, could be confusing with many different files and unclear which is used when.

note that logback-test.xml is probably not in the classpath when running the app because it is typically placed in `src/test/resources`. At least it is not included when running from a packaged image. Could be included if running from the IDE (if we support that).

> always include a logback-test.xml. Arguably, dev-mode is also a kind of test run

still somewhat strange to place that in main/resources

logback-test.xml is probably well know and somewhat strange if we wouldn't support it, but maybe too much to include it in our templates. Then we could prefer logback-test in dev mode (if they have added it themselves) otherwise we use logback-dev-mode.xml

octonato: Actually, our templates include a `src/test/resources/logback-test.xml` and `src/main/resources/logback-dev-mode.xml`. 

Current PR is now looking for `logback-test.xml` in classpath, therefore only present when running the tests. 

Otherwise, it will pick `src/main/resources/logback-dev-mode.xml`, but not from the classpath, but directly fro the dir. Therefore only present if running on local machine and not when running a packaged jar. 

The current code is somewhat future-proof in the sense that we can later decide to only provide `logback-test.xml` and drop `logback-dev-mode.xml`. Then, when running application locally, we would point `logback.configurationFile` to `src/test/resources/logback-test.xml`.

sdk/java-sdk-protobuf/src/main/scala/kalix/javasdk/impl/logging/KalixJoranConfigurator.scala
```diff
@@ -0,0 +1,79 @@
...
+import ch.qos.logback.classic.spi.Configurator.ExecutionStatus
+import ch.qos.logback.classic.util.DefaultJoranConfigurator
+import ch.qos.logback.core.util.Loader
+
+class KalixJoranConfigurator extends DefaultJoranConfigurator {
+
+  private val devModeLogbackFile = "src/main/resources/logback-dev-mode.xml"
+
+  // not really related to logging, but we can disable it here as well
+  System.setProperty("spring.main.banner-mode", "off")
```
aludwiko, line 31: looks strange here

sdk/java-sdk-protobuf/src/main/scala/kalix/javasdk/impl/logging/KalixJoranConfigurator.scala
```diff
@@ -0,0 +1,79 @@
...
+    // this is testing if logback-test.xml is on the classpath, which is the case when running tests
+    val noLogbackTest = {
+      // using as much as possible Logback classes to find the resource in the classpath
+      val myClassLoader = Loader.getClassLoaderOfObject(this)
+      val url = Loader.getResource(DefaultJoranConfigurator.TEST_AUTOCONFIG_FILE, myClassLoader)
+      url == null
+    }
+
+    // this file path only exist when running in dev-mode.
+    // once packaged we don't have src/main/resources anymore
```
aludwiko, line 44: `don't have src/main/resources` do you mean the whole folder? what about .properties files?

aludwiko (approved):
Looks good, some minor comments. Maybe it's worth documenting the `logging.config` option anyway.

sdk/java-sdk-protobuf/src/main/scala/kalix/javasdk/impl/logging/KalixJoranConfigurator.scala
octonato: Yes, the whole folder is not available in a package app. The content is packaged into meta-inf. 

When the jar is executed, there are no `src/main/resources`, therefore the file is not found. And that's intentional.

Issue #1702, opened 2023-06-26
support metadata in deferred calls execution


aludwiko:
At this point, executing a deferred call with `.execute()` method will ignore metadata.

PR #1724, opened 2023-06-30
fix: support metadata in deferred calls execution

aludwiko:
References #1702

### DIFF
--- a/codegen/java-gen/src/main/scala/kalix/codegen/java/ComponentsSourceGenerator.scala
+++ b/codegen/java-gen/src/main/scala/kalix/codegen/java/ComponentsSourceGenerator.scala
@@ -144,7 +144,15 @@ object ComponentsSourceGenerator {
              |    MetadataImpl.Empty(),
              |    "${component.service.messageType.fullyQualifiedProtoName}",
              |    "${command.name}",
-             |    (Metadata metadata) -> addHeaders(((${messageType.fullyQualifiedGrpcServiceInterfaceName}Client) getGrpcClient(${messageType.fullyQualifiedGrpcServiceInterfaceName}.class)).$commandMethod(), metadata).invoke($paramName)
+             |    (Metadata metadata) -> {
+             |      ${messageType.fullyQualifiedGrpcServiceInterfaceName} client = getGrpcClient(${messageType.fullyQualifiedGrpcServiceInterfaceName}.class);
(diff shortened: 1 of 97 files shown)


codegen/java-gen/src/main/scala/kalix/codegen/java/ComponentsSourceGenerator.scala
```diff
@@ -173,6 +176,16 @@ object ComponentsSourceGenerator {
        |    return context.getComponentGrpcClient(serviceClass);
        |  }
        |
+       |  private <Req, Res> SingleResponseRequestBuilder<Req, Res> addHeaders(SingleResponseRequestBuilder<Req, Res> requestBuilder, Metadata metadata){
+       |    var updatedBuilder = requestBuilder;
+       |    for (Metadata.MetadataEntry entry: metadata){
+       |      if (entry.isText()) {
```
aludwiko, line 182: Actually, maybe for grpc based sdk, I should support binary headers as well?

samples/java-protobuf-valueentity-shopping-cart/src/test/java/com/example/shoppingcart/ShoppingCartActionImplTest.java
```diff
@@ -47,7 +48,7 @@ public void tearDown() throws Exception {
   // tag::initialize[]
   @Test
   public void initializeCartTest() throws ExecutionException, InterruptedException, TimeoutException {
-    when(shoppingCartService.create(notNull()))
+    when(shoppingCartService.create().invoke(notNull()))
```
aludwiko, line 51: Under the hood we have:
```
(Metadata metadata) -> addHeaders(((com.example.shoppingcart.ShoppingCartServiceClient) getGrpcClient(com.example.shoppingcart.ShoppingCartService.class)).create(), metadata).invoke(createCart)
```
thanks to `RETURNS_DEEP_STUBS` I was able to fix it in the java version, but with mockito scala I'm not sure if this is an option. 

Maybe we shouldn't encourage users to write unit tests for side effects. Personally I would go with IT tests for such cases.

octonato (approved):
LGTM

codegen/java-gen/src/main/scala/kalix/codegen/java/ComponentsSourceGenerator.scala
johanandren: I think we can skip those for now.

octonato (approved):

Issue #1734, opened 2023-07-19
finished workflow should not update the state on the sdk side


aludwiko:
When a workflow is finished users can still run any request against it. This request can return an effect with state update and transition. Both will be ignored on the proxy side but on the sdk side the state is updated in the `effectMessage` method before returning the response to the proxy. The next request could be a get request that will return state that was never persisted.

PR #1735, opened 2023-07-19
fix: skipping state update for finished workflow

aludwiko:
References #1734

### DIFF
--- a/sdk/java-sdk-protobuf/src/main/scala/kalix/javasdk/impl/workflow/WorkflowImpl.scala
+++ b/sdk/java-sdk-protobuf/src/main/scala/kalix/javasdk/impl/workflow/WorkflowImpl.scala
@@ -196,7 +196,7 @@ final class WorkflowImpl(system: ActorSystem, val services: Map[String, Workflow
     init.userState match {
       case Some(state) =>
         val decoded = service.messageCodec.decodeMessage(state)
-        router._internalSetInitState(decoded)
+        router._internalSetInitState(decoded, finished = false) //TODO get this from the init message
       case None => // no initial state
     }
 
@@ -207,7 +207,7 @@ final class WorkflowImpl(system: ActorSystem, val services: Map[String, Workflow
         val protoEffect =
           persistence match {
             case UpdateState(newState) =>
-              router._internalSetInitState(newState)
+              router._internalSetInitState(newState, transition.isInstanceOf[End.type])
               WorkflowEffect.defaultInstance.withUserState(service.messageCodec.encodeScala(newState))
             // TODO: persistence should be optional, but we must ensure that we don't save it back to null
             // and preferably we should not even send it over the wire.
@@ -307,7 +307,7 @@ final class WorkflowImpl(system: ActorSystem, val services: Map[String, Workflow
           val stepResponse =
             try {
               val decoded = service.messageCodec.decodeMessage(executeStep.userState.get)
-              router._internalSetInitState(decoded)
+              router._internalSetInitState(decoded, false) // here we know that workflow is still running
               router._internalHandleStep(
                 executeStep.commandId,
                 executeStep.input,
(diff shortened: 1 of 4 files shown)


sdk/java-sdk-protobuf/src/main/scala/kalix/javasdk/impl/workflow/WorkflowImpl.scala
```diff
@@ -196,7 +196,7 @@ final class WorkflowImpl(system: ActorSystem, val services: Map[String, Workflow
     init.userState match {
       case Some(state) =>
         val decoded = service.messageCodec.decodeMessage(state)
-        router._internalSetInitState(decoded)
+        router._internalSetInitState(decoded, finished = false) //TODO get this from the init message
```
aludwiko, line 199: If we are ok, with the solution I will update the proxy and then use the proper value it in this or some next PR.

sdk/java-sdk-protobuf/src/main/scala/kalix/javasdk/impl/workflow/WorkflowImpl.scala
```diff
@@ -207,7 +207,7 @@ final class WorkflowImpl(system: ActorSystem, val services: Map[String, Workflow
         val protoEffect =
           persistence match {
             case UpdateState(newState) =>
-              router._internalSetInitState(newState)
+              router._internalSetInitState(newState, transition.isInstanceOf[End.type])
```
octonato, line 210: Maybe it will be better to fail if the user returns an Effect with persistence set to `UpdateState`. 

I would say, if finished, only `NoPersistence` and `DeleteState` (not yet supported), should be allowed.

octonato: The same applies to the Transition. Once finished, all Effects should have a `NoTransition` and fail otherwise. 

As such we close the doors and only allow methods that are effectively read-only.

aludwiko: The thing is that if I throw an exception here then the proxy will just return a generic "unexpected error" to the user. We can log a warning here (or info), but I would create a separate issue to focus on the response on the proxy side.

octonato: Can't we introduce a specific failure for this so that the proxy is then able to convert to something more sensible?

aludwiko: That would require an additional protocol change. I think that the proxy has already all the information to construct a detailed response.

aludwiko: And the logic to validate the effect must be present in the proxy anyway, since in theory, anyone can create an SDK.

sdk/java-sdk-protobuf/src/main/scala/kalix/javasdk/impl/workflow/WorkflowRouter.scala
```diff
@@ -81,8 +82,11 @@ abstract class WorkflowRouter[S, W <: Workflow[S]](protected val workflow: W) {
 
   /** INTERNAL API */
   // "public" api against the impl/testkit
-  def _internalSetInitState(s: Any): Unit = {
-    state = Some(s.asInstanceOf[S])
+  def _internalSetInitState(s: Any, finished: Boolean): Unit = {
```
aludwiko, line 85: ok, this will work also for recovering a finished workflow, since then the flag is false and then true

octonato:
@aludwiko, I leave it for you to merge at your own discretion. As discussed today, we can wait for the proxy or not. Your call. 😄

Issue #1332, opened 2022-11-17
Java code-first SDK: support for Akka's Jackson migration


octonato:

efgpinto:
I was doing some experiments mainly for my own understanding of what was possible already or not, so here are some notes here:
- add/removing fields from state types seems mostly fine: the fields will be `null` when non-existing and logic to handle that needs to live on the UF.
- class renaming is fine as long as structure matches new type (or type info is used for ES events)

Akka Jackson migration will be needed to:
- support renaming fields
- support transforming fields and changing type (works for some primitive type conversions but fails for composite ones as expected)

octonato:
Json behaves similar to proto with respect to unknown fields. The main issue is renaming indeed. 

Renaming is no issue for gRPC, only ordering matters. Changing type is a major problem for gRPC as well. And I don't think we should even try to solve this kind of problems. 

Good that Akka Jackson Migration have already some support for it, even if limited that's already something. We are fine as long as we document the limitations. 

Speaking about documenting migration, I don't think we have anything about protobuf and we simply expect that users will know how to evolve their protobuf. That's probably too optimistic. 

I hear often that schema evolution is not an issue for protobuf. That's true only if you know what you are doing. The reality, is that very often people change their format without much care.

johanandren:
Side note: Re the protobuf schema evolution, we don't really link to it from our docs (we should), but there is information in the protobuf documentation: https://protobuf.dev/programming-guides/proto3/#updating

aludwiko:
Ok, I'm experimenting with a similar approach to Akka Jackson Migration. The biggest challenge is how to save the version number. My idea is to use the `typeUrl` for that, e.g. `json.kalix.io/com.example.SimpleClass#1` after hash I will add the version number. So far it looks promising, but if you see any blockers, please let me know:)

aludwiko:
One interesting thing about Views + Migration is that we can now support the schema evolution of VE state model. For instance, we could have:
```
public class CustomerEntity extends ValueEntity<Customer> 
```
after some time this could be changed to:
```
@Migration(CustomerUpdatedMigration.class)
public record CustomerUpdated()
...
public class CustomerEntity extends ValueEntity<CustomerUpdated> 
```

A view can be created with a dummy transformation, that just returns the deserialized state, and the `CustomerUpdatedMigration` will do the job on the fly.

```
public class CustomerByEmailView extends View<CustomerUpdated> {
[...]

PR #1760, opened 2023-08-08
feat: JSON schema evolution support

aludwiko:
reference #1332


build.sbt
```diff
@@ -126,7 +126,7 @@ lazy val javaSdkProtobufTestKit = project
 
 lazy val javaSdkSpring = project
   .in(file("sdk/java-sdk-spring"))
-  .dependsOn(javaSdkProtobuf)
+  .dependsOn(javaSdkProtobuf % "compile->compile;test->test")
```
aludwiko, line 129: I'm reusing some test classes.

sdk/java-sdk-protobuf/src/main/java/kalix/javasdk/JacksonMigration.java
```diff
@@ -0,0 +1,38 @@
...
+ * limitations under the License.
+ */
+
+package kalix.javasdk;
+
+import com.fasterxml.jackson.databind.JsonNode;
+
+import java.util.List;
+
+public abstract class JacksonMigration {
```
aludwiko, line 23: I'm unsure where to put this class and `Migration` annotation. It should be close to `JsonSupport`, but maybe there are better options.

sdk/java-sdk-spring/src/main/scala/kalix/javasdk/impl/JsonMessageCodec.scala
```diff
@@ -53,20 +57,34 @@ private[kalix] class JsonMessageCodec extends MessageCodec {
...
       .collect { case ann if ann.value().trim.nonEmpty => ann.value() }
       .getOrElse(clz.getName)
-    cache.computeIfAbsent(clz, _ => typeName)
+
+    val (version, supportedClassNames) = getVersionAndSupportedClassNames(clz)
+    val typeNameWithVersion = typeName + (if (version == 0) "" else "#" + version)
+
     //TODO verify if this could be replaced by sth smarter/safer
-    reversedCache.compute(
+    addToReversedCache(clz, typeName)
```
aludwiko, line 80: While updating `typeHints` map, I'm also updating another map `reversedTypeHints`, this feels wrong, but I haven't figured out a better option for that.

sdk/java-sdk-protobuf/src/main/java/kalix/javasdk/JsonSupport.java
```diff
@@ -130,11 +132,27 @@ public static <T> Any encodeJson(T value, String jsonType) {
...
+   * the JSON string as bytes as value and a type URL starting with "json.kalix.io/".
+   *
+   * @param valueClass       The type of class to deserialize the object to, the class must have the
+   *                         proper Jackson annotations for deserialization.
+   * @param any              The protobuf Any object to deserialize.
+   * @param jacksonMigration The optional @{@link JacksonMigration} implementation used for deserialization.
+   * @return The decoded object
+   * @throws IllegalArgumentException if the given value cannot be decoded to a T
+   */
+  public static <T> T decodeJson(Class<T> valueClass, Any any, Optional<? extends JacksonMigration> jacksonMigration) {
```
aludwiko, line 155: A `jacksonMigration` additional param to pass the migration from the upper level, otherwise I would need to move the @Migration annotation here.

sdk/java-sdk-protobuf/src/main/java/kalix/javasdk/JsonSupport.java
```diff
@@ -158,14 +193,29 @@ public static <T> T decodeJson(Class<T> valueClass, Any any) {
...
   }
 
+  private static <T> T migrate(Class<T> valueClass, ByteString decodedBytes, int fromVersion, JacksonMigration jacksonMigration) throws IOException {
+    JsonNode jsonNode = objectMapper.readTree(decodedBytes.toByteArray());
+    JsonNode newJsonNode = jacksonMigration.transform(fromVersion, jsonNode);
+    return objectMapper.treeToValue(newJsonNode, valueClass);
+  }
+
+  private static int parseVersion(String typeUrl) {
+    if (typeUrl.contains("#")) { //TODO can we assume that there will be ony one "#" ??
```
aludwiko, line 203: I suppose we can, right?

aludwiko:
PR is ready for the first review round.

sdk/java-sdk-protobuf/src/main/java/kalix/javasdk/JacksonMigration.java
```diff
@@ -0,0 +1,68 @@
...
+
+/**
+ * Allows to specify dedicated strategy for JSON schema evolution.
+ * <p>
+ * It is used when deserializing data of older version than the
+ * {@link JacksonMigration#currentVersion}. You implement the transformation of the
+ * JSON structure in the {@link JacksonMigration#transform} method. If you have changed the
+ * class name you should add it to {@link JacksonMigration#supportedClassNames}.
+ */
+public abstract class JacksonMigration {
```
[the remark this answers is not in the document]
johanandren: Sounds better to me 👍

sdk/java-sdk-protobuf/src/main/java/kalix/javasdk/JsonSupport.java
johanandren: Make the signature with Optional param internal and just have one overload with migration and one without, for Java?

sdk/java-sdk-protobuf/src/main/java/kalix/javasdk/JsonSupport.java
johanandren: Safer to do a regex match for something like `.*#(\d+)$`?

johanandren (commented):
Did an initial skim-review, looking good

sdk/java-sdk-protobuf/src/main/java/kalix/javasdk/JacksonMigration.java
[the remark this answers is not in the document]
aludwiko: Renamed

sdk/java-sdk-protobuf/src/main/java/kalix/javasdk/JsonSupport.java
aludwiko: I thought about this, but then all the invocation would look like:
```
        extractMigration(typeClass)
          .map(JsonSupport.decodeJson(typeClass, any, _))
          .getOrElse(JsonSupport.decodeJson(typeClass, any))
``` which is also smelly.

File: sdk/java-sdk-protobuf/src/main/java/kalix/javasdk/JsonSupport.java (hunk 5 above)
Comment (aludwiko) [reply sdk/java-sdk-protobuf/src/main/java/kalix/javasdk/JsonSupport.java:203]: Yeah, regexp is an option, but I'm always worried that it might be more costly. Maybe I shouldn't care too much in this case...

Comment (patriknw) [reply sdk/java-sdk-protobuf/src/main/java/kalix/javasdk/JsonSupport.java:203]: `lastIndexOf('#')` should be efficient

File: sdk/java-sdk-protobuf/src/test/java/kalix/javasdk/DummyClassRenamed.java [hunk 7]
```diff
@@ -0,0 +1,49 @@
...
+import java.util.Objects;
+import java.util.Optional;
+
+public class DummyClassRenamed {
+  public String stringValue;
+  public int intValue;
+  public Optional<String> optionalStringValue;
+
+  @JsonCreator
+  public DummyClassRenamed(@JsonProperty("stringValue") String stringValue, @JsonProperty("intValue") int intValue, @JsonProperty("optionalStringValue") Optional<String> optionalStringValue) {
```
patriknw, line 31: are these JsonProperty needed if you have `javacOptions += "-parameters"`?

patriknw (commented):
looking good, recognising much from Akka (which is a good thing)

sdk/java-sdk-protobuf/src/main/java/kalix/javasdk/JsonSupport.java
aludwiko: updated

sdk/java-sdk-protobuf/src/test/java/kalix/javasdk/DummyClassRenamed.java
aludwiko: nope, removed

sdk/java-sdk-protobuf/src/main/java/kalix/javasdk/JsonSupport.java
johanandren: I thought user API for proto would skip the reflection-migration-lookup and be that the user supplies migrations manually, for that it just seems messy that they'd need to deal with optional, but for the reflective may-have-migration use I understand why it needs to be optional.

johanandren: Another thought: if we support annotation-detection of migration also for proto, can we just hide it inside the JsonSupport instead and the user doesn't have to pass anything?

aludwiko: "I thought user API for proto would skip the reflection-migration-lookup" - in some sense, it will, because the user will just use the `public static <T> T decodeJson(Class<T> valueClass, Any any) {` method. Although nothing will prevent him to use the signature with migration.

"can we just hide it inside the" - exactly that's my point one method without any extra params. The "only" drawback is that some code-first annotations will be from the `java-sdk-spring` and some (`@Migration`, `@TypeName` ?) from the `java-sdk-protobuf`.

aludwiko:
I will create a separate PR with the documentation.

johanandren (approved):
LGTM, I think the protobuf side usage can be revisited in the future and is good enough for now.

sdk/java-sdk-spring/src/main/scala/kalix/javasdk/impl/JsonMessageCodec.scala
octonato: If I recall correctly, we register the types when we register the components. This is supposed to be done by a single thread running over all components and finding out the types. In that sense, it's safe to update the main map and then the reversed one. 

For good measure, we could synchronize this method to make sure we are safe. Just in case something changes one day.

sdk/java-sdk-spring/src/test/scala/kalix/javasdk/impl/JsonMessageCodecSpec.scala
```diff
@@ -76,6 +108,19 @@ class JsonMessageCodecSpec extends AnyWordSpec with Matchers {
...
+    "add version number to typeUrl" in {
+      //new codec to avoid collision with SimpleClass
+      val encoded = new JsonMessageCodec().encodeJava(SimpleClassUpdated("abc", 10, 123))
+      encoded.getTypeUrl shouldBe jsonTypeUrlWith("kalix.javasdk.impl.JsonMessageCodecSpec$SimpleClassUpdated#1")
+    }
+
+    "decode with new schema version" in {
+      val encoded = messageCodec.encodeJava(SimpleClass("abc", 10))
+      val decoded =
+        JsonSupport.decodeJson(classOf[SimpleClassUpdated], encoded, Optional.of(new SimpleClassUpdatedMigration))
```
octonato, line 120: I guess we are adding this extra method receiving `Optional<JsonMigration>` because we are using classes that belong to the Protobuf package and we don't have access to `MigrationExtractor.extractMigration` nor the `Migration` annotation. 

IMO, this is another reason to remove this method. We should not increase the public API just because we want to reuse some classes.

octonato (commented):
Late for the party, but I think there are a few things that we should change before releasing it. See my comments.

octonato:
One more thing...

I see comments about Protobuf migration. If we plan to do it, we should build another abstraction. 

The class name is JsonMigration and it depends on Jackson classes (JsonNode). Not something that we want or can reuse for Protobuf payloads.

sdk/java-sdk-protobuf/src/main/java/kalix/javasdk/JsonSupport.java
```diff
@@ -130,11 +132,27 @@ public static <T> Any encodeJson(T value, String jsonType) {
...
+   * the JSON string as bytes as value and a type URL starting with "json.kalix.io/".
+   *
+   * @param valueClass       The type of class to deserialize the object to, the class must have the
+   *                         proper Jackson annotations for deserialization.
+   * @param any              The protobuf Any object to deserialize.
+   * @param jacksonMigration The optional @{@link JsonMigration} implementation used for deserialization.
+   * @return The decoded object
+   * @throws IllegalArgumentException if the given value cannot be decoded to a T
+   */
+  public static <T> T decodeJson(Class<T> valueClass, Any any, Optional<? extends JsonMigration> jacksonMigration) {
```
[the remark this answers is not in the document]
aludwiko: It is a problematic situation. Initially, this method was private, but to make it private I would need to move the `@Migration` to java-sdk-protobuf.

sdk/java-sdk-spring/src/it/java/com/example/wiring/SpringSdkIntegrationTest.java
```diff
@@ -242,25 +242,37 @@ public void verifyEchoActionRequestParamWithTypedForward() {
...
 
+  @Test
+  public void verifyCounterEventSourceSubscription2() throws InterruptedException {
+
+    String entityId = "hello1";
+    execute(componentClient.forEventSourcedEntity(entityId)
+        .call(CounterEntity::increase)
+        .params(777));
+
+    Thread.sleep(60000);
```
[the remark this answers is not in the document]
aludwiko: Yeah, just ignore it, I removed it in some other PRs.

sdk/java-sdk-spring/src/it/java/com/example/wiring/SpringSdkIntegrationTest.java
```diff
@@ -242,25 +242,37 @@ public void verifyEchoActionRequestParamWithTypedForward() {
...
   @Test
   public void verifyCounterEventSourceSubscription() {
     // GIVEN IncreaseAction is subscribed to CounterEntity events
     // WHEN the CounterEntity is requested to increase 42\
     String entityId = "hello1";
     execute(componentClient.forEventSourcedEntity(entityId)
-      .call(CounterEntity::increase)
-      .params(42));
+        .call(CounterEntity::increase)
+        .params(42));
```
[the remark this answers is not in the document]
aludwiko: probably the order of tests is different than the order in the class

sdk/java-sdk-protobuf/src/main/java/kalix/javasdk/JsonSupport.java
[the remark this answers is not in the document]
octonato: I think it's better to have the Migration annotation in the java-sdk-protobuf package, no? 

Or do we plan to use this method in other contexts? 

The `decodeJson` method is also used when consuming json from a topic, but in such case, I doubt that it will be useful to pass a migration class as the type URL won't have the version number attached to it.

sdk/java-sdk-spring/src/it/java/com/example/wiring/SpringSdkIntegrationTest.java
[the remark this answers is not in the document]
octonato: But that has the potential to introduce flakiness. Better to use a different entityId.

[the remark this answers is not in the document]
octonato: Also, I don't see how that could even work. If the ordering is different, it should also fail. One of the tests must fail, not clear why both are passing. 🤷

PR #1763, opened 2023-08-10
feat: adding traces to UF components

franciscolopezsancho:
References https://github.com/lightbend/kalix/issues/9380


sdk/java-sdk-protobuf/src/main/scala/kalix/javasdk/impl/eventsourcedentity/EventSourcedEntityRouter.scala
```diff
@@ -88,6 +93,8 @@ abstract class EventSourcedEntityRouter[S, E, ES <: EventSourcedEntity[S, E]](pr
       context: CommandContext,
       snapshotEvery: Int,
       eventContextFactory: Long => EventContext): CommandResult = {
+    //do I have the metadata here?
```
octonato, line 96: It seems you have it, no? You accessed it on the command context a few lines later.

codegen/java-gen/src/main/scala/kalix/codegen/java/EventSourcedEntitySourceGenerator.scala
```diff
@@ -88,7 +88,7 @@ object EventSourcedEntitySourceGenerator {
        |  }
        |
        |  @Override
-       |  public $stateType handleEvent($stateType state, Object event) {
+       |  public $stateType handleEvent($stateType state, Object event, Metadata metadata) {
```
octonato, line 91: Do we need to add it to this method signature? 

Users can access it from the EventContext and the addition here is not even being used. The `eventCases` are not using this information.

octonato (commented):
I did a first pass. Mainly on the codegen part. Left some comments.

codegen/java-gen/src/main/scala/kalix/codegen/java/EventSourcedEntitySourceGenerator.scala
johanandren: I agree, don't change this signature, it would break every existing service built with the protocol first JVM SDKs

sdk/java-sdk-spring/src/main/scala/kalix/javasdk/impl/Telemetry.scala
```diff
@@ -0,0 +1,147 @@
...
+        .setSpanKind(SpanKind.SERVER)
+        .startSpan()
+      Some(
+        span
+          .setAttribute(SemanticAttributes.HTTP_METHOD, getHttpMethod(method))
+          .setAttribute("server.address", "localhost") // TODO use Kubernetes .svc name?
+          .setAttribute("server.port", 8080) // TODO read the port from properties
+          .setAttribute("server.scheme", "http")
+          .setAttribute("service.name", method.getName)
+          .setAttribute("kalix.inputType", inputTypeUrl)
```
johanandren, line 140: Note that this will probably be our internal synthetic protobuf message type and doesn't make sense for the user at all.

sdk/java-sdk-spring/src/main/scala/kalix/javasdk/impl/Telemetry.scala
```diff
@@ -0,0 +1,147 @@
...
+        .spanBuilder(s"""${component.getClass.getName}.${method.getName}""")
+        .setParent(context)
+        .setSpanKind(SpanKind.SERVER)
+        .startSpan()
+      Some(
+        span
+          .setAttribute(SemanticAttributes.HTTP_METHOD, getHttpMethod(method))
+          .setAttribute("server.address", "localhost") // TODO use Kubernetes .svc name?
+          .setAttribute("server.port", 8080) // TODO read the port from properties
+          .setAttribute("server.scheme", "http")
```
johanandren, line 138: More likely HTTPS, are these expected to be the public endpoint that the request came in through? Note that the request is not only from the public internet but could also be from another service internally to Kalix. 

Since it will always be from the proxy actually accepting that request, that's where you could possibly know these (but not entirely we know for sure there either, because service mesh in front)

sdk/java-sdk-spring/src/main/scala/kalix/javasdk/impl/Telemetry.scala
```diff
@@ -0,0 +1,147 @@
...
+          .setAttribute("service.name", method.getName)
```
johanandren, line 139: Is this the Java class name for the user class handling the request or is it the synthetic protobuf service method name (and is that really the "service.name")?

sdk/java-sdk-spring/src/main/scala/kalix/javasdk/impl/eventsourcedentity/ReflectiveEventSourcedEntityRouter.scala
```diff
@@ -84,11 +91,18 @@ class ReflectiveEventSourcedEntityRouter[S, E, ES <: EventSourcedEntity[S, E]](
...
-      .invoke(entity, invocationContext)
-      .asInstanceOf[EventSourcedEntity.Effect[_]]
+
+    val span = telemetry.buildSpan(entity, inputTypeUrl, Some(methodInvoker), commandContext.metadata)
+    try {
+      methodInvoker
+        .invoke(entity, invocationContext)
+        .asInstanceOf[EventSourcedEntity.Effect[_]]
+    } finally {
+      span.map(s => s.end())
```
johanandren, line 103: Is that really the span end, this is as soon as the user method returns, wouldn't it be after handling the returned effect, for example after persist and event has been applied? The effect could also be async and not have completed when the methor returns.

codegen/java-gen/src/main/scala/kalix/codegen/java/EventSourcedEntitySourceGenerator.scala
johanandren: Ah, sorry, I thought it was the abstract event sourced entity that the user extends but it is the generated route-to-the-right-command/event-handler method.

franciscolopezsancho: I'll take it from the context.

franciscolopezsancho: Actually I will remove the metadata all together

franciscolopezsancho:
The new integration test won't pass until https://github.com/lightbend/kalix-proxy/pull/2096 is merged

sdk/java-sdk-protobuf/src/main/resources/reference.conf
```diff
@@ -93,4 +93,11 @@ kalix {
       }
     }
   }
+
+  telemetry {
+     tracing {
+        collector-endpoint = "http://localhost:4317"
+        collector-endpoint =  ${?COLLECTOR_ENDPOINT}
+     }
+  }
```
johanandren, line 102: So, this is always enabled against a local endpoint when user is running tests, integration tests, locally etc? Shouldn't it rather be disabled by default and then opt in? Will the user be running something on port 4317?

sdk/java-sdk-protobuf/src/main/scala/kalix/javasdk/impl/action/ActionsImpl.scala
```diff
@@ -117,6 +119,10 @@ private[javasdk] final class ActionsImpl(
   import ActionsImpl._
   import _system.dispatcher
   implicit val system: ActorSystem = _system
+  //TODO
```
johanandren, line 122: What is there to do?

sdk/java-sdk-protobuf/src/main/scala/kalix/javasdk/impl/action/ActionsImpl.scala
```diff
@@ -203,6 +210,8 @@ private[javasdk] final class ActionsImpl(
           case NonFatal(ex) =>
             // command handler threw an "unexpected" error
             Future.successful(handleUnexpectedException(service, in, ex))
+        } finally {
+          span.map(_.end())
```
johanandren, line 214: This is not where the call ends - `Future[ActionResponse]`, so should be once that future completes.

Also, note that now we are back into the question about passing trace context through thread local for non async and how to do it for async calls to propagate context to things the user call (other Kalix services for example).

sdk/java-sdk-protobuf/src/main/scala/kalix/javasdk/impl/telemetry/Telemetry.scala
```diff
@@ -0,0 +1,153 @@
...
+      .builder()
+      .setTracerProvider(sdkTracerProvider)
+      .setPropagators(ContextPropagators.create(W3CTraceContextPropagator.getInstance()))
+      .build()
+    //    Runtime.getRuntime().addShutdownHook(new Thread(sdkTracerProvider.close)) TODO add shutdown
+    sdk
+  }
+
+  /**
+   * TODO
```
johanandren, line 85: What is left to do?

sdk/java-sdk-protobuf/src/main/scala/kalix/javasdk/impl/telemetry/Telemetry.scala
```diff
@@ -0,0 +1,153 @@
...
+        .addSpanProcessor(
+          SimpleSpanProcessor.create(OtlpGrpcSpanExporter.builder().setEndpoint("http://jaeger:4317").build()))
+        .setResource(resource)
+        .build()
+    val sdk = OpenTelemetrySdk
+      .builder()
+      .setTracerProvider(sdkTracerProvider)
+      .setPropagators(ContextPropagators.create(W3CTraceContextPropagator.getInstance()))
+      .build()
+    //    Runtime.getRuntime().addShutdownHook(new Thread(sdkTracerProvider.close)) TODO add shutdown
```
johanandren, line 80: Is the shutdown really needed?

sdk/java-sdk-protobuf/src/main/scala/kalix/javasdk/impl/telemetry/Telemetry.scala
```diff
@@ -0,0 +1,153 @@
...
+        .setParent(context)
+        .setSpanKind(SpanKind.SERVER)
+        .startSpan()
+      Some(
+        span
+          //          .setAttribute(SemanticAttributes.HTTP_METHOD, httpMethod.name)
+          .setAttribute("server.address", "localhost")
+          // TODO use Kubernetes .svc name?
+          .setAttribute("server.port", 8080) // TODO read the port from properties
+          .setAttribute("server.scheme", "http")
```
johanandren, line 108: Still not likely to be http, also, again what is the port, the public one that client used to hit the proxy or the internal one for the SDK?

If we don't have anything more correct/better here, maybe skip the attributes completely instead? The trace is anyway identified by the service name?

sdk/java-sdk-spring/docker-compose-integration.yml
```diff
@@ -28,4 +31,18 @@ services:
...
+    volumes:
+      - ./tempo.yaml:/etc/tempo.yaml
+      - ./tempo-data:/tmp/tempo
+    ports:
+      - "14268:14268"  # jaeger ingest
+      - "3200:3200"   # tempo
+      - "9095:9095" # tempo grpc
+      - "4317:4317"  # otlp grpc
+      - "4318:4318"  # otlp http
+      - "9411:9411"   # zipkin
```
johanandren, line 48: Ok, so the plan is to always run a server. Does that work? What about integration tests/test containers?

It also means the docker compose with both will be needed for all samples (and the g8 template).

sdk/java-sdk-spring/src/it/java/com/example/wiring/tracing/TracingIntegratonTest.java
```diff
@@ -0,0 +1,88 @@
...
+        );
+
+    }
+
+    private Integer callTCounter(String counterId, Integer increase) {
+        return webClient.post().uri("/tcounter/" + counterId + "/increase/" + increase).retrieve().bodyToMono(Integer.class).block(timeout);
+    }
+ // TODO investigate this path   String response = WebClient.create("http://0.0.0.0:3200").get().uri(
+ //                uriBuilder -> uriBuilder.path("/api/search")
+ //                        .queryParam("q","{ }").build()).retrieve().bodyToMono(String.class).block(timeout)
```
johanandren, line 74: Do it or issue

sdk/java-sdk-spring/src/main/scala/kalix/javasdk/impl/eventsourcedentity/ReflectiveEventSourcedEntityRouter.scala
```diff
@@ -36,6 +36,10 @@ class ReflectiveEventSourcedEntityRouter[S, E, ES <: EventSourcedEntity[S, E]](
     messageCodec: JsonMessageCodec)
     extends EventSourcedEntityRouter[S, E, ES](entity) {
 
+  private val logger = LoggerFactory.getLogger(this.getClass)
+
+//  private val telemetry = new Telemetry(entity.getClass.getName)
```
johanandren, line 41: ?

sdk/java-sdk-spring/src/main/scala/kalix/javasdk/impl/action/ReflectiveActionRouter.scala
```diff
@@ -33,6 +35,7 @@ class ReflectiveActionRouter[A <: Action](
     ignoreUnknown: Boolean)
     extends ActionRouter[A](action) {
 
+  private val telemetry = new Telemetry(action.getClass.getName)
```
johanandren, line 38: Is this used anywhere?

sdk/java-sdk-protobuf/src/main/resources/reference.conf
franciscolopezsancho: Right. Did opt-in in the proxy but forgot to do it here.
The UF will be pointing to an OpenTelemetry container sidecar

franciscolopezsancho: Actually, I remember now that I thought that one shouldn't enable or disable the tracing in the SDK. It will propagate tracing if it receives traces from the proxy otherwise won't do nothing.

franciscolopezsancho: see [here](https://github.com/lightbend/kalix-jvm-sdk/pull/1763/files/a59c5f1c108c310537b5817d4f9669387f716cf5#diff-0eafcaa0a75ac0670d8a260a9d73a07dd2e7aba36d82e183e6b67762a360a7e4R89)  please

sdk/java-sdk-protobuf/src/main/scala/kalix/javasdk/impl/action/ActionsImpl.scala
franciscolopezsancho: cleaned

sdk/java-sdk-protobuf/src/main/scala/kalix/javasdk/impl/telemetry/Telemetry.scala
franciscolopezsancho: I think this will flush the traces in case of a shutdown. I left it there as a reminder.

sdk/java-sdk-protobuf/src/main/scala/kalix/javasdk/impl/telemetry/Telemetry.scala
franciscolopezsancho: adding the api docs. Done

sdk/java-sdk-spring/src/it/java/com/example/wiring/tracing/TracingIntegratonTest.java
franciscolopezsancho: issued

sdk/java-sdk-spring/src/main/scala/kalix/javasdk/impl/action/ReflectiveActionRouter.scala
franciscolopezsancho: deleted.

sdk/java-sdk-protobuf/src/main/scala/kalix/javasdk/impl/telemetry/Telemetry.scala
franciscolopezsancho: I was going to checkout this attributes today, but got busy with the operator. Yeah, let's ditch these attributes by now and we can add them later. Once the spike gets some maturity.

sdk/java-sdk-spring/docker-compose-integration.yml
franciscolopezsancho: This change is made exclusively for the integration tests/testContainer. See [this](https://github.com/lightbend/kalix-jvm-sdk/pull/1763/files#diff-7c37fa3ed1fba565ba77f927e8f7f8d82e92bc274d54130ef8d2ed4a72ade74f), please.

Not sure that I follow that "the plan is to always run a server". If the proxy telemetry.tracing is disabled (which is by default) no traces should be created nor pushed.

sdk/java-sdk-spring/src/main/scala/kalix/javasdk/impl/Telemetry.scala
franciscolopezsancho: supersesed [below](https://github.com/lightbend/kalix-jvm-sdk/pull/1763#discussion_r1328378700)

sdk/java-sdk-spring/src/main/scala/kalix/javasdk/impl/Telemetry.scala
franciscolopezsancho: changed now

sdk/java-sdk-spring/src/main/scala/kalix/javasdk/impl/eventsourcedentity/ReflectiveEventSourcedEntityRouter.scala
franciscolopezsancho: went one level deeper, to the ActionImpl but still not enough. [Follow](https://github.com/lightbend/kalix-jvm-sdk/pull/1763#discussion_r1328376206).

sdk/java-sdk-protobuf/src/main/resources/reference.conf
johanandren: That means it should lazily create the Telemetry instances the first time it sees that they are needed rather than up front like it is currently done.

sdk/java-sdk-protobuf/src/main/scala/kalix/javasdk/impl/action/ActionsImpl.scala
franciscolopezsancho: changed now

sdk/java-sdk-protobuf/src/main/scala/kalix/javasdk/impl/eventsourcedentity/EventSourcedEntityRouter.scala
franciscolopezsancho: deleted

sdk/java-sdk-protobuf/src/main/resources/reference.conf
franciscolopezsancho: I see. For that I would need a mutable Map (sort of a cache) because we don't know if tracing is coming from the proxy until we process the first command. We find one way or another inside the metadata. Hence I need to delay the construction until then. The problem I see is that such a cache  won't be thread safe. WDYT?

I looked at http://eed3si9n.com/learning-scalaz/Memo.html but I'm not sure we want this either. 

That said I think this optimization may not be necessary. Even if we have tenths of services I don't think it will be a to create that amount of Telemetry objects.

franciscolopezsancho: I can use a [TrieMap](https://www.scala-lang.org/api/current/scala/collection/concurrent/TrieMap.html) in case you still think is worth to optimize this creation.

franciscolopezsancho: Maybe better a `java.util.concurrent.ConcurrentHashMap`

sdk/java-sdk-spring/src/main/scala/kalix/javasdk/impl/Telemetry.scala
franciscolopezsancho: We'll register the service.name in the proxy and this span will register the entities type along the entity's id.

sdk/java-sdk-protobuf/src/main/scala/kalix/javasdk/impl/telemetry/Telemetry.scala
franciscolopezsancho: added the concurrent set in a Runtime global shutdown. I'll add it to an Akka coordinated shutdown if @octonato confirms we restart the SDK in the same JVM.

franciscolopezsancho: done

sdk/java-sdk-protobuf/src/main/resources/reference.conf
franciscolopezsancho: I'm implementing it and I think it will perform worse when checking every request if the Telemetry has been created on not in a `ConcurrentHashMap`. Not only it need to do the check first (in the best case scenario, i.e. the Telemetry instance exists already), it will sync (block) with any other thread that is accessing that `ConcurrentHashMap`. 

Does that makes sense?

johanandren (commented):
Do you expect to try tackle propagating the context to whatever the UF does separately, and this only adds spans for the SDK side of the components? Are we sure we are interested in the SDK spans or is the propagation possibly the valuable thing?

I think I mentioned this in the other PR but separate SDK spans for calls we already have spans for in the proxy is probably more interesting to us than it is to the users.

Compare the value for a user of:
```
    [-------- proxy --------]
      [------- sdk --------]
```

Vs
```
   [-----   service/component A -----------------------------------------]
          [ ---- service/component B]   [----- service/component C –––] 
```

Where they can see what component talked to what other component and which of them took time etc.

sdk/java-sdk-protobuf/src/main/scala/kalix/javasdk/impl/telemetry/Telemetry.scala
```diff
@@ -58,17 +58,24 @@ object Telemetry {
   }
 }
 
-class Telemetry(serviceName: String, system: ActorSystem) {
+class Telemetry[T](serviceName: String, system: ActorSystem, commandType: T) {
```
johanandren, line 61: Maybe make it an internal ADT/enum for the different types of components instead of an arbitrary typed param? 

Also, it's not really the type of the command, it's rather something like "component category" (I guess "component type" is already used for the user categorising their concrete types of components).

sdk/java-sdk-protobuf/src/main/scala/kalix/javasdk/impl/telemetry/Telemetry.scala
```diff
@@ -107,14 +114,14 @@ class Telemetry(serviceName: String, system: ActorSystem) {
...
-        .spanBuilder(s"""${service.serviceName}.${command.entityId}""")
+        .spanBuilder(s"""${command.entityId}""")
         .setParent(context)
         .setSpanKind(SpanKind.SERVER)
         .startSpan()
       Some(
         span
           .setAttribute("service.name", s"""${service.serviceName}.${command.entityId}""")
-          .setAttribute(s"kalix.${service.componentType}", command.entityId))
+          .setAttribute(s"${service.componentType}", command.entityId))
```
johanandren, line 124: Isn't it strange to have an attribute key defined by the user/unknown up front?

Vs something like:
```
.setAttribute("component.type", service.componentType)
.setAttribute("entity.id", command.entityId)
```

If we don't know the attribute name up front, how can it be used anywhere?

sdk/java-sdk-protobuf/src/main/scala/kalix/javasdk/impl/telemetry/Telemetry.scala
franciscolopezsancho: Of course. I didn't like it but didn't came up with anything better.

sdk/java-sdk-protobuf/src/main/scala/kalix/javasdk/impl/telemetry/Telemetry.scala
franciscolopezsancho: Good point

sdk/java-sdk-protobuf/src/main/scala/kalix/javasdk/impl/telemetry/Telemetry.scala
franciscolopezsancho: done

sdk/java-sdk-protobuf/src/main/scala/kalix/javasdk/impl/telemetry/Telemetry.scala
franciscolopezsancho: done

sdk/java-sdk-protobuf/src/main/scala/kalix/javasdk/impl/eventsourcedentity/EventSourcedEntityRouter.scala
```diff
@@ -88,13 +92,15 @@ abstract class EventSourcedEntityRouter[S, E, ES <: EventSourcedEntity[S, E]](pr
...
       eventContextFactory: Long => EventContext): CommandResult = {
+
     val commandEffect =
       try {
         entity._internalSetCommandContext(Optional.of(context))
         entity._internalSetCurrentState(_stateOrEmpty())
         handleCommand(commandName, _stateOrEmpty(), command, context).asInstanceOf[EventSourcedEntityEffectImpl[Any, E]]
       } catch {
         case CommandHandlerNotFound(name) =>
+          //TODO add status failed
```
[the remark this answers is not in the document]
franciscolopezsancho: deleted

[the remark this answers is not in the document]
franciscolopezsancho: removed

sdk/java-sdk-protobuf/src/main/scala/kalix/javasdk/impl/eventsourcedentity/EventSourcedEntitiesImpl.scala
```diff
@@ -195,13 +200,15 @@ final class EventSourcedEntitiesImpl(
...
             } catch {
+              //TODO close span in case of failure
               case BadRequestException(msg) =>
                 val errorReply = ErrorReplyImpl(msg, Some(Status.Code.INVALID_ARGUMENT), Vector.empty)
                 CommandResult(Vector.empty, errorReply, None, context.sequenceNumber, false)
               case e: EntityException => throw e
               case NonFatal(error) =>
                 throw EntityException(command, s"Unexpected failure: $error", Some(error))
             } finally {
+              span.map(_.end())
```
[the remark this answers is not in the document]
franciscolopezsancho: done

sdk/java-sdk-protobuf/src/main/scala/kalix/javasdk/impl/action/ActionsImpl.scala
```diff
@@ -191,18 +196,23 @@ private[javasdk] final class ActionsImpl(
...
-            in.payload.getOrElse(throw new IllegalArgumentException("No command payload")))
-          val effect = service.factory
-            .create(creationContext)
-            .handleUnary(in.name, MessageEnvelope.of(decodedPayload, context.metadata()), context)
-          effectToResponse(service, in, effect, service.messageCodec)
-        } catch {
-          case NonFatal(ex) =>
-            // command handler threw an "unexpected" error
-            Future.successful(handleUnexpectedException(service, in, ex))
+        val context = createContext(in, service.messageCodec)
```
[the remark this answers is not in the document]
franciscolopezsancho: It was. I thought the construction `createContext` of the object `ActionContextImpl` couldn't raise an error. But now that you mention better safe than sorry.

Reset.

sdk/java-sdk-protobuf/src/main/resources/reference.conf
johanandren: Just to sum up what we discussed latest, it seems we want one OpenTelemetry sdk instance per component, but we can probably eager create those optionally in `...Impl` constructor so that no concurrency is needed, and only if tracing is enabled/needed for a given SDK process.

sdk/java-sdk-protobuf/src/main/scala/kalix/javasdk/impl/telemetry/Telemetry.scala
```diff
@@ -58,14 +58,18 @@ object Telemetry {
...
   import Telemetry._
 
-  val tracePrefix = commandType match {
-    case _: Command.type       => "Event Sourced Entity"
-    case _: ActionCommand.type => "Action"
-    case other                 => logger.warn("Command type not implemented [{}].", other.getClass)
+  val tracePrefix = componentCategory match {
+    case _: EventSourcedEntityCategory.type => "Event Sourced Entity"
+    case _: ActionCategory.type             => "Action"
+    case other                              => logger.warn("Command type not implemented [{}].", other.getClass)
```
johanandren, line 72: Can't be any other now since sealed. Also you could put the prefix in a method on each and not have to switch at all.

franciscolopezsancho: done

sdk/java-sdk-protobuf/src/main/scala/kalix/javasdk/impl/DiscoveryImpl.scala
```diff
@@ -76,7 +76,7 @@ class DiscoveryImpl(
    */
   override def discover(in: ProxyInfo): scala.concurrent.Future[Spec] = {
     log.info(
-      "Received discovery call from [{} {}] at [{}]:[{}] supporting Kalix protocol {}.{}",
+      "ServiceInfo [{} {}] at [{}]:[{}] supporting Kalix protocol {}.{}",
```
aludwiko, line 79: The previous version was more informative to me.

sdk/java-sdk-spring/src/it/java/com/example/wiring/eventsourcedentities/tracingcounter/TCounterEvent.java
```diff
@@ -0,0 +1,28 @@
...
+
+import kalix.javasdk.annotations.TypeName;
+
+public interface TCounterEvent {
+
+
+    @TypeName("tincreased")
+    record ValueIncreased(Integer value) implements TCounterEvent{}
+
+    ;
```
aludwiko, line 27: formatting

aludwiko (commented):
minor comments

sdk/java-sdk-spring/src/it/java/com/example/wiring/eventsourcedentities/tracingcounter/TCounterEvent.java
franciscolopezsancho: done

sdk/java-sdk-protobuf/src/main/scala/kalix/javasdk/impl/DiscoveryImpl.scala
franciscolopezsancho: It is also now for me. Changed back.

sdk/java-sdk-spring/tempo-query.yaml
```diff
@@ -0,0 +1 @@
+backend: "tempo:3200"
```
johanandren, line 1: Are these only for the spring sdk integration tests, if not, should they be in more places?

sdk/java-sdk-spring/docker-compose-integration.yml
```diff
@@ -1,7 +1,7 @@
 version: "3"
 services:
   kalix-proxy:
-    image: gcr.io/kalix-public/kalix-proxy:1.1.19
+    image: gcr.io/kalix-workbench/kalix-proxy:1.1.20
```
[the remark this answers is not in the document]
franciscolopezsancho: darn, I thought I changed that back. 

Sure. Looking for your feedback, guys.

sdk/java-sdk-spring/tempo-query.yaml
franciscolopezsancho: So far these are only for the Spring SDK. At the moment, that's the only one I have integration tests for.

johanandren: Ok 👍

sdk/java-sdk-spring/docker-compose-integration.yml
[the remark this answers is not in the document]
franciscolopezsancho: ... I did change it. But didn't `push`. Not very useful 'it works in my machine'

sdk/java-sdk-spring/src/it/java/com/example/wiring/pubsub/DockerIntegrationTest.java
```diff
@@ -52,18 +50,27 @@ public abstract class DockerIntegrationTest {
 
   private KalixSpringApplication kalixSpringApplication;
 
-  public DockerIntegrationTest(ApplicationContext applicationContext) {
+
```
franciscolopezsancho, line 53: This changes are related to the `TracingIntegrationTest`. I needed to pass some extra configuration.

sdk/java-sdk-protobuf/src/main/scala/kalix/javasdk/impl/eventsourcedentity/EventSourcedEntitiesImpl.scala
```diff
@@ -181,63 +189,65 @@ final class EventSourcedEntitiesImpl(
...
-            } finally {
-              context.deactivate() // Very important!
-            }
-
-          val serializedSecondaryEffect = secondaryEffect match {
-            case MessageReplyImpl(message, metadata, sideEffects) =>
-              MessageReplyImpl(service.messageCodec.encodeJava(message), metadata, sideEffects)
-            case other => other
-          }
+          val span = instrumentations(service.serviceName).buildSpan(service, command)
```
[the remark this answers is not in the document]
franciscolopezsancho: Good point. done

johanandren (approved):

PR #1765, opened 2023-08-17
fix: removing kalix dependency from Spring grpc client

aludwiko:
in the documentation, we inform users to remove all Kalix annotations and imports from proto files, so I don't see any point to keep the dependency to the sdk


efgpinto (approved):
Makes sense to me!

PR #1775, opened 2023-08-25
fix: catch npe when trying to load file that does not exist

sebastian-alfers:
Before this change,  running `mvn verify` in this sample failed with an npe.

### DIFF
--- a/samples/java-protobuf-web-resources/src/main/java/com/example/FileServiceAction.java
+++ b/samples/java-protobuf-web-resources/src/main/java/com/example/FileServiceAction.java
@@ -9,6 +9,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.io.IOException;
+import java.io.InputStream;
 import java.nio.file.Files;
 import java.nio.file.NoSuchFileException;
 import java.nio.file.Paths;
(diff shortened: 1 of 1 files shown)


samples/java-protobuf-web-resources/src/main/java/com/example/FileServiceAction.java
```diff
@@ -47,7 +48,11 @@ private  Effect<HttpBody> loadFile(String dir, String file) {
     String fullPath = baseDir+dir+"/"+file;
     try {
       // tag::200-ok[]
-      byte[] byteArray = getClass().getResourceAsStream(fullPath).readAllBytes();
+      InputStream inputStream = getClass().getResourceAsStream(fullPath);
+      if(null == inputStream) {
+        throw new NoSuchFileException("File " + fullPath + " not found");
+      }
```
johanandren, line 54: Should really be a return with a not-found grpc status code (which should lead to a 404 response from the proxy)  rather than an exception leading to 500 internal.

sebastian-alfers: Maybe a `throw` should not be used here, but right now this one gets caught below and turned into a `effects().ignore();` which gets a 404 in the browser, which I was hoping to express that the requested file does not exist.

johanandren: Ah, I missed the end of the method with the specific catch.

johanandren (approved):

Issue #1774, opened 2023-08-24
EventingTestkit should respect TypeName annotation


aludwiko:
When publishing the type is always `class.getName`, which is fine for proto-sdk. For the code first sdk we should use the logic from JsonMessageCodec.

PR #1777, opened 2023-08-25
fix: respecting TypeName annotations when using eventing testkit

aludwiko:
References #1774

### DIFF
--- a/sdk/java-sdk-protobuf-testkit/src/main/java/kalix/javasdk/testkit/EventingTestKit.java
+++ b/sdk/java-sdk-protobuf-testkit/src/main/java/kalix/javasdk/testkit/EventingTestKit.java
@@ -16,19 +16,14 @@
 
 package kalix.javasdk.testkit;
 
-import akka.NotUsed;
 import akka.actor.ActorSystem;
 import akka.annotation.ApiMayChange;
 import akka.annotation.InternalApi;
-import akka.stream.javadsl.Source;
-import akka.testkit.TestProbe;
 import com.google.protobuf.ByteString;
-import com.google.protobuf.GeneratedMessageV3;
 import kalix.javasdk.Metadata;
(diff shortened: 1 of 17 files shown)


sdk/java-sdk-protobuf-testkit/src/main/java/kalix/javasdk/testkit/EventingTestKit.java
```diff
@@ -208,44 +203,53 @@ interface Topic {
...
      *
-     * @param payload the message payload
+     * @param payload  the message payload
      * @param metadata the metadata associated with the message
-     * @return a Message object to be used in the context of the Testkit
      * @param <T>
+     * @return a Message object to be used in the context of the Testkit
      */
-    static <T> Message<T> of(T payload, Metadata metadata) {
+    public <T> Message<T> of(T payload, Metadata metadata) {
```
aludwiko, line 234: Moving these methods here is not backward compatible, not sure how relevant that is.

efgpinto: I think it's probably fine since the feature is very recent.

efgpinto (approved):
LGTM

sdk/java-sdk-protobuf-testkit/src/main/java/kalix/javasdk/testkit/EventingTestKit.java
aludwiko: Ok, just don't forget to update the webinar (if necessary) :)

efgpinto: luckily I did not use the publisher side :D

Issue #1689, opened 2023-06-20
Support running integration tests against real Kafka instance


efgpinto:
AFAIR, it is indeed impossible to run integration tests against Kafka at the moment becuase the proxy container run from the integration tests supports either the mock broker or the google pub sub (before the eventing testkit it would only support google pub sub). This is somehting we can (and should) easily change... it's just a matter of allowing that proxy to be configured for kafka as well (meaning we set a differnet eventing support method and a different port by default). But at the moment, I don't think we have configs for that.

_Originally posted by @efgpinto in https://github.com/lightbend/kalix-jvm-sdk/pull/1688#discussion_r1234148797_

johanandren:
With the changes for Kafka auth, it should at least be easier to set Kafka up, new config key/or env var to set up what server to connect to (`kalix.proxy.eventing.kafka.bootstrap-servers`/`BROKER_SERVERS`) instead of the old full kafka client props file.

PR #1779, opened 2023-08-28
feat: support running integration tests against real kafka

efgpinto:
Close https://github.com/lightbend/kalix-jvm-sdk/issues/1689


samples/java-protobuf-customer-registry-kafka-quickstart/docker-compose.yml
```diff
@@ -13,31 +13,38 @@ services:
...
+      # configuring kafka broker used for eventing
+      BROKER_SERVERS: kafka:29092
+      # Alternatively:
+      # - for easy local/dev mode - plaintext, for non dev-mode TLS, use:
+      #BROKER_AUTH_MECHANISM: "PLAIN"
+      #BROKER_USERNAME: [your-username]
+      #BROKER_PASS: [your-pass]
+      # - for SCRAM-SHA-256 or SCRAM-SHA-512, use:
+      #BROKER_AUTH_MECHANISM: "SCRAM-SHA-256" #OR SCRAM-SHA-512
+      #BROKER_CA_CERT: [your-cert-location]
```
johanandren, line 25: Note that for scram they would actually have to script adding users to zookeeper interacting with the zookeeper node and a whole mess, so maybe skip mentioning that

samples/java-protobuf-customer-registry-kafka-quickstart/docker-compose.yml
```diff
@@ -13,31 +13,38 @@ services:
...
-      BROKER_CONFIG_FILE: /conf/my-local.kafka.properties
-    volumes:
-      - .:/conf
+      # configuring kafka broker used for eventing
+      BROKER_SERVERS: kafka:29092
+      # Alternatively:
+      # - for easy local/dev mode - plaintext, for non dev-mode TLS, use:
+      #BROKER_AUTH_MECHANISM: "PLAIN"
+      #BROKER_USERNAME: [your-username]
+      #BROKER_PASS: [your-pass]
```
johanandren, line 22: Note that for TLS they would both have to map those certs into Kafka and configure it to use them and also have to mount the CA cert into the proxy and configure it to use that in the truststore for the Kafka client, not entirely sure we should mention it. Or if we do we should document all of that.

samples/java-protobuf-customer-registry-kafka-quickstart/src/it/java/customer/api/CustomerActionWithKafkaIntegrationTest.java
```diff
@@ -0,0 +1,118 @@
...
+        .get(5, SECONDS);
+
+
+    // Example Kafka consumer configuration
+    Properties props = new Properties();
+    props.put(ConsumerConfig.BOOTSTRAP_SERVERS_CONFIG, "0.0.0.0:9093");
+    props.put(ConsumerConfig.GROUP_ID_CONFIG, "it-test-" + System.currentTimeMillis()); // using new consumer group every run
+    props.put(ConsumerConfig.KEY_DESERIALIZER_CLASS_CONFIG, StringDeserializer.class.getName());
+    props.put(ConsumerConfig.VALUE_DESERIALIZER_CLASS_CONFIG, ByteArrayDeserializer.class.getName());
+    props.put(ConsumerConfig.AUTO_OFFSET_RESET_CONFIG, "earliest");
```
johanandren, line 77: Messy that it has to be repeated here, but maybe enough. Probably not _that_ common that users will think they need Kafka integration tests?

sdk/java-sdk-protobuf-testkit/src/main/java/kalix/javasdk/testkit/KalixTestKit.java
```diff
@@ -297,6 +305,17 @@ public KalixTestKit start(final Config config) {
     return this;
   }
 
+  private int getEventingBackendPort(Settings.EventingSupport eventingSupport) {
+    switch(eventingSupport) {
+      case KAFKA:
+        return DEFAULT_KAFKA_PORT;
```
johanandren, line 311: So we actually only support it on the default port?

samples/java-protobuf-customer-registry-kafka-quickstart/docker-compose.yml
efgpinto: Yeah, I had the same questions, if we needed to actually have these different options here. I ended up adding them because we previously also had this `my-dev.kafka.properties` using user/pass... so I thought we better keep a similar example with the new approach.. But I'm not really sure. You've been working on this integration, I think I will just follow your whatever you feel is best.. should we just remove and keep this simple? Leave the details for docs?

samples/java-protobuf-customer-registry-kafka-quickstart/src/it/java/customer/api/CustomerActionWithKafkaIntegrationTest.java
efgpinto: Yeah, my rational is that users will only need one of this when they are interested in testing something specific related with Kafka and its clients. e.g. make sure that certain messages go to certain partitions. In these cases, I think it's reasonable to expose users to these details.

For general business related tests, testkit mocked topic is the way to go.

sdk/java-sdk-protobuf-testkit/src/main/java/kalix/javasdk/testkit/KalixTestKit.java
efgpinto: I do think being able to provide a different port would be good. But we don't support for pubsub either, so I thought about opening a follow-up issue and allow a different port for both.

samples/java-protobuf-customer-registry-kafka-quickstart/docker-compose.yml
johanandren: Let's drop and don't mention and follow up the day there is a user/customer who wants to run integration tests with credentials/TLS.

sdk/java-sdk-protobuf-testkit/src/main/java/kalix/javasdk/testkit/KalixTestKit.java
johanandren: Let's leave it as is. I was thinking about the scenario where a user wants to connect to a cloud Kafka, where they often use some arbitrary port, but we can add support the day a user actually wants to do that.

efgpinto:
moving to draft while I debug the issues with CI

efgpinto:
working now in CI. Ready for final review @johanandren

johanandren (approved):

PR #1780, opened 2023-08-29
fix: moving @Migration to java-sdk-protobuf

aludwiko:
reverting the code to the initial idea


sdk/java-sdk-protobuf/src/main/java/kalix/javasdk/JsonSupport.java
```diff
@@ -163,9 +151,12 @@ public static <T> T decodeJson(Class<T> valueClass, Any any, Optional<? extends
     } else {
       try {
         ByteString decodedBytes = ByteStringEncoding.decodePrimitiveBytes(any.getValue());
-        if (jacksonMigration.isPresent()) {
+        if (valueClass.getAnnotation(Migration.class) != null) {
+          JsonMigration migration = valueClass.getAnnotation(Migration.class)
```
aludwiko, line 155: Not sure how costly this is, but we can add a cache in the future.

octonato (approved):
LGTM

Agree that it's also not great to move the annotation, but I think it's better than augment the public API.

PR #1787, opened 2023-09-05
docs: inject jwt token in integration tests

efgpinto:
Closes https://github.com/lightbend/kalix-proxy/issues/884

Adds documentation with snippets on how to inject the jwt token in integrations tests for:
- [x] java sdk
- [x] java pb sdk
- [x] scala pb sdk


samples/java-protobuf-doc-snippets/src/it/java/com/example/JwtIntegrationTest.java
```diff
@@ -0,0 +1,68 @@
...
+  }
+
+  private String bearerTokenWith(Map<String, String> claims) throws JsonProcessingException {
+    // setting algorithm to none
+    String alg = "ewogICJhbGciOiAibm9uZSIKfQo"; // <4>
+    byte[] jsonClaims = new ObjectMapper().writeValueAsBytes(claims);
+
+    // no validation is done for integration tests, thus no valid signature required
+    return alg + "." +  Base64.getEncoder().encodeToString(jsonClaims); // <5>
+  }
```
efgpinto, line 66: we could eventually build some small test utility to provide this part out of the box.. but I'm not so convinced it's worth the increase in tesktit API surface. I think this example in docs might be enough. LKM if disagree.

samples/java-protobuf-doc-snippets/src/it/java/com/example/JwtIntegrationTest.java
```diff
@@ -0,0 +1,68 @@
...
+      new KalixTestKitResource(Main.createKalix());
+
+  /**
+   * Use the generated gRPC client to call the service through the Kalix proxy.
+   */
+  // tag::bearer-token-claims-test[]
+  private final JwtServiceActionClient client;
+
+  public JwtIntegrationTest() {
+    client = (JwtServiceActionClient) testKit.getGrpcClient(CounterService.class); // <1>
```
efgpinto, line 40: This bit feels a bit weird.. This is due to the fact that we need to be able to inject headers and that is available otherwise. Is this worth a different alternative?

efgpinto (commented):
I think for java sdk, the docs + sample work fine as the `ComponentClient` already provides ability to inject headers. For PB sdks, it's not so natural due to a forced cast to be able to inject headers.

All in all, I think I would probably go with docs + snippets but see below and let me know your thoughts.

samples/java-protobuf-doc-snippets/src/it/java/com/example/JwtIntegrationTest.java
johanandren: I think that is good enough

samples/java-protobuf-doc-snippets/src/it/java/com/example/JwtIntegrationTest.java
```diff
@@ -0,0 +1,68 @@
...
+        .invoke(req)
+        .toCompletableFuture()
+        .get();
+
+    assertTrue(response.getMsg().contains(msg));
+  }
+
+  private String bearerTokenWith(Map<String, String> claims) throws JsonProcessingException {
+    // setting algorithm to none
+    String alg = "ewogICJhbGciOiAibm9uZSIKfQo"; // <4>
```
octonato, line 61: as discussed on Slack, better to be more explicit that this is the Base64 encoding of `{ "alg": "none" }`

octonato (approved):
LGTM with one minor change request

samples/java-protobuf-doc-snippets/src/it/java/com/example/JwtIntegrationTest.java
```diff
@@ -0,0 +1,68 @@
...
+        .invoke(req)
+        .toCompletableFuture()
+        .get();
+
+    assertTrue(response.getMsg().contains(msg));
+  }
+
+  private String bearerTokenWith(Map<String, String> claims) throws JsonProcessingException {
+    // setting algorithm to none
+    String alg = Base64.getEncoder().encodeToString("{\"alg\":\"none\"}".getBytes()); // <4>
```
octonato, line 61: we can use triple quotes here, eg: `"""{ "alg": "none" }"""`.

PR #1799, opened 2023-10-10
docs: document service level jwt

octonato:
References https://github.com/lightbend/kalix-proxy/issues/1594

Also, document multi-issuer configuration as seen here https://github.com/lightbend/kalix-docs/pull/1906

### DIFF
--- a/docs/src/modules/java/pages/using-jwts.adoc
+++ b/docs/src/modules/java/pages/using-jwts.adoc
@@ -58,13 +58,13 @@ include::example$java-spring-doc-snippets/src/main/java/com/example/jwt/JWTActio
 
 The above examples show how to configure a JWT token on a method.
 
-JWT configurations can also be shared between all methods on a component by specifying them on the class, at type level. The annotation is then applied on all public methods in the current component. Method level annotations have precedence over type level annotations.
+JWT configurations can also be shared between all methods of a component by specifying them at type level. The annotation will then be applied to all methods in that component. Method-level annotations take precedence over type-level annotations.
 
 [source, java, indent=0]
 ----
 include::example$java-spring-doc-snippets/src/main/java/com/example/jwt/HelloJwtAction.java[tag=bearer-token]
 ----
-<1> A type level JWT annotation is added to the class requiring a bearer token issuer.
+<1> A type level JWT annotation is added to the class requiring a bearer token issuer named `my-issuer`.
 <2> The method `message` inherits the JWT configuration from the type level annotation and therefore will require a token issued by `my-issuer`.
 <3> The method `messageWithIssuer` overwrites the type level annotation and requires a different token issuer, i.e.: `my-other-issuer`.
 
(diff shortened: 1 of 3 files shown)


docs/src/modules/java/pages/using-jwts.adoc
```diff
@@ -46,6 +54,20 @@ Kalix will place the claims from the validated token into the request metadata,
 include::example$java-spring-doc-snippets/src/main/java/com/example/jwt/JWTAction.java[tag=bearer-token-claims]
 ----
 
+=== Configuring JWT at component level
+
+The above examples show how to configure a JWT token on a method.
+
+JWT configurations can also be shared between all methods of a component by specifying them at type level. The annotation will then be applied to all methods in that component. Method-level annotations take precedence over type-level annotations.
```
ennru, line 61:
```suggestion
JWT configurations can also be shared between all methods of a component by specifying them on the class. The annotation will then be applied to all methods in that component. Method-level annotations take precedence over class-level annotations.
```

docs/src/modules/java/pages/using-jwts.adoc
```diff
@@ -46,6 +54,20 @@ Kalix will place the claims from the validated token into the request metadata,
...
+
+[source, java, indent=0]
+----
+include::example$java-spring-doc-snippets/src/main/java/com/example/jwt/HelloJwtAction.java[tag=bearer-token]
+----
+<1> A type level JWT annotation is added to the class requiring a bearer token issuer named `my-issuer`.
+<2> The method `message` inherits the JWT configuration from the type level annotation and therefore will require a token issued by `my-issuer`.
+<3> The method `messageWithIssuer` overwrites the type level annotation and requires a different token issuer, i.e.: `my-other-issuer`.
```
ennru, line 69:
```suggestion
<2> The method `message` inherits the JWT configuration from the class-level annotation and therefore will require a token issued by `my-issuer`.
<3> The method `messageWithIssuer` overwrites the class-level annotation and requires a different token issuer, i.e.: `my-other-issuer`.
```

docs/src/modules/java/pages/using-jwts.adoc
```diff
@@ -5,7 +5,7 @@ include::ROOT:partial$include.adoc[]
 
 This section describes the practical aspects of using JSON Web Tokens (JWTs) with the Java SDK, if you are not sure what JWTs are, how they work or how to generate them, see https://docs.kalix.io/security/jwts.html[JSON Web Tokens] first.
 
-Kalix’s JWT support is configured by placing annotations on methods in your endpoints.
+Kalix's JWT support is configured by placing annotations on methods in your endpoints or at the component type level.
```
ennru, line 8:
```suggestion
Kalix's JWT support is configured by placing annotations on methods in your endpoints or at the component class level.
```

octonato: I have switched back and forth a few times. When defining an annotation, we need to say that it can be added to a type as oppose to a method or field. Hence the idea of referring to it as type-level annotations. 

But let's go with class. It's probably better.

PR #1813, opened 2023-10-19
test: update some test timing for projections

pvlugter:
Updating the timing for some tests that failed in #1810, with the switch to H2 and some longer delays for projections. There may be others that need adjusting. Separate from the bump PR, but will also test there.


octonato (approved):
LGTM

PR #1816, opened 2023-10-20
bug: disabling integration tracing test after pull/#1810

franciscolopezsancho:
References https://github.com/lightbend/kalix-jvm-sdk/pull/1810

### DIFF
--- a/sdk/java-sdk-spring/src/it/java/com/example/wiring/tracing/TracingIntegratonTest.java
+++ b/sdk/java-sdk-spring/src/it/java/com/example/wiring/tracing/TracingIntegratonTest.java
@@ -20,6 +20,7 @@ import com.example.Main;
 import com.example.wiring.pubsub.DockerIntegrationTest;
 import com.typesafe.config.Config;
 import com.typesafe.config.ConfigFactory;
+import org.junit.Ignore;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.TestInstance;
 import org.slf4j.Logger;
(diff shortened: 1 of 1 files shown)


sdk/java-sdk-spring/src/it/java/com/example/wiring/tracing/TracingIntegratonTest.java
```diff
@@ -50,7 +51,7 @@ public TracingIntegratonTest(ApplicationContext applicationContext) {
         super(applicationContext, config);
     }
 
-    @Test
+    @Ignore //disabled ATM for https://github.com/lightbend/kalix-jvm-sdk/pull/1810
```
octonato, line 54: I'm confused now. 
In #1810, we added the Ignore annotation and that's what we have now in main. 

I think the diff should not even show here.

franciscolopezsancho: Right. I mess the order I guess. I would have seen this if I would have made a rebase from main. Never mind.

octonato: But that's also surprising. Ok that you needed to rebase, but GitHub compares your branch to `main` since you were planning to merge it into `main` and therefore it should not show any diff. 

Possibly, the PR was made before #1810 got merged. Still, I thought that GH would refresh the diff anyway, but I might be wrong. 🤷

PR #1821, opened 2023-10-25
fix: workflow input decoding

octonato:
Fixes https://github.com/lightbend/kalix-proxy/issues/2135

### DIFF
--- a/sdk/java-sdk-protobuf/src/main/scala/kalix/javasdk/impl/workflow/WorkflowRouter.scala
+++ b/sdk/java-sdk-protobuf/src/main/scala/kalix/javasdk/impl/workflow/WorkflowRouter.scala
@@ -41,6 +41,7 @@ import kalix.javasdk.workflow.Workflow.Effect
 import Workflow.AsyncCallStep
 import Workflow.CallStep
 import Workflow.WorkflowDef
+import kalix.javasdk.JsonSupport
 import kalix.javasdk.impl.WorkflowExceptions.WorkflowException
 import kalix.protocol.workflow_entity.StepDeferredCall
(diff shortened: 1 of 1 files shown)


ennru (approved):
LGTM.

sdk/java-sdk-protobuf/src/main/scala/kalix/javasdk/impl/workflow/WorkflowRouter.scala
```diff
@@ -121,6 +122,16 @@ abstract class WorkflowRouter[S, W <: Workflow[S]](protected val workflow: W) {
 
   protected def handleCommand(commandName: String, state: S, command: Any, context: CommandContext): Workflow.Effect[_]
 
+  // in same cases, the Proxy may send a message with typeUrl set to object.
+  // if that's the case, we need to patch the message using the typeUrl from the expected input class
+  private def decodeInput(messageCodec: MessageCodec, result: ScalaPbAny, expectedInputClass: Class[_]) =
+    if (result.typeUrl == JsonSupport.KALIX_JSON + "object") {
+      val typeUrl = messageCodec.typeUrlFor(expectedInputClass)
```
aludwiko, line 129: Would that work when the workflow queries the view, and the view itself was not used yet, so the messaceCodec might not have the `expectedInputClass` entry?

octonato: The input classes are registered in the messageCodec when we register the component. And that's valid for all components. So, the answer is yes. 

The problem here is that we were always relying on the incoming `typeUrl` which is simply not reliable enough in this situation.

Issue #1844, opened 2023-11-06
Set a more user friendly name for the kalix-runtime in docker-compose files


aludwiko:
Autogenerated names from the docker-compose, for the CLI `kalix sandbox service list` are too verbose. Would be nice to fix all the samples and archetypes.

PR #1847, opened 2023-11-07
fix: user friendly name of the kalix proxy docker container

aludwiko:
References #1844


efgpinto:
Shouldn't we append a `-runtime` or something? otherwise, when we package the userfunction to run alongside which name would it have?

octonato:
@efgpinto, we need to experiment with this. But I think it's ok to have it as such. The image name of the service won't be the same as the name chosen here. It will be similar though. 

The image will be like: `user-selected-docker-register/artifactId` while the kalix-runtime when running locally will be `artifactId`. 

What we want is to have a name that emulates the name of the service in prod. At least to start with. User's could eventually change that name and not change it in all places. But that's not in our control anymore.

octonato (approved):
LGTM

Issue #1849, opened 2023-11-10
component client - calling a View with more than one parametr fails


aludwiko:
Exception:
```
java.lang.IllegalArgumentException: Can only resolve method references from serializable SAMs, class was: class java.lang.reflect.Method

	at kalix.javasdk.impl.client.MethodRefResolver$.resolveMethodRef(MethodRefResolver.scala:32)
	at kalix.javasdk.client.ComponentCall$.invoke(ComponentCall.scala:73)
	at kalix.javasdk.client.ComponentCall2.params(ComponentCall.scala:206)
	at be.....myTest.java:86)
```

PR #1850, opened 2023-11-10
fix: calling a view with more than one parameter

aludwiko:
References #1849


johanandren (approved):
LGTM, 21 arguments is all one could ever expect anyone to use! :D 

Brings back good old Scala case class memories.

PR #1853, opened 2023-11-13
fix: Discover and health check with service incarnation

patriknw:


project/Dependencies.scala
```diff
@@ -6,7 +6,7 @@ object Dependencies {
   object Kalix {
     val ProtocolVersionMajor = 1
     val ProtocolVersionMinor = 1
-    val ProxyVersion = System.getProperty("kalix-proxy.version", "1.1.22")
+    val ProxyVersion = System.getProperty("kalix-proxy.version", "1.1.22-8-7d112bc6-dev-SNAPSHOT")
```
patriknw, line 9: need a release first

johanandren (approved):

octonato (approved):

patriknw:
As far as I can see this change is even compatible to use with an old runtime version. The new field will be ignored (included in unknown fields) and can be parsed to Empty.

PR #1861, opened 2023-11-16
fix: ensure KalixClient is configured before used

octonato:


octonato:
> good catch

Yeah, but I which that the integration tests would have catched it first. 
That mean that none of our tests are using the ComponentClient and that's quite surprising, no?

PR #1864, opened 2023-11-17
Correct GetCustomer to return non-null Address.

max8github:
If you follow the quickstart tutorial and [run it](https://docs.kalix.io/java-protobuf/quickstart/cr-value-entity-scala.html#_invoke_your_service), by issuing `grpcurl` commands, `Address` will always come back as `null`, even when setting it, which may be perceived as a bug by the user experimenting with this.

References
- [docs](https://docs.kalix.io/java-protobuf/quickstart/cr-value-entity-scala.html)


octonato (approved):
Thanks @max8github!

PR #1865, opened 2023-11-17
bump in sample: current Maven plugins

ennru:
Bump to current versions of Maven plugins, some with significant jumps.

Other samples and the archetype need bumping.


octonato:
any reason why you only changed it in one single sample?

ennru:
Superseded by 
- #1866

PR #1867, opened 2023-11-17
Reminder to announce in the public Slack channel

ennru:


efgpinto (approved):

octonato:
What if we had this automated with a Slack / GitHub integration?

octonato:
And if we want to have a little bit of control about it we could build a Kalix app that does the integration.

efgpinto:
> And if we want to have a little bit of control about it we could build a Kalix app that does the integration.

I'm all in for any and every Kalix app. Dogfooding FTW.

PR #1869, opened 2023-11-17
docs: improved component's javadoc and scaladoc

octonato:


docs/src/modules/java-protobuf/partials/actions.adoc
```diff
@@ -4,7 +4,9 @@ Actions are stateless functions that can be used to implement different uses cas
 
 - a pure function
 - request conversion - you can use Actions to convert incoming data into a different format before forwarding a call to a different component.
-- publish and subscribe to events
```
franciscolopezsancho, line 7: As a facade or controller to fan out multiple calls to different components?

docs/src/modules/java-protobuf/partials/actions.adoc
```diff
@@ -4,7 +4,9 @@ Actions are stateless functions that can be used to implement different uses cas
 
 - a pure function
 - request conversion - you can use Actions to convert incoming data into a different format before forwarding a call to a different component.
-- publish and subscribe to events 
+- publish messages to a topic
+- subscribe to events from an event-sourced entity
```
franciscolopezsancho, line 8:
```suggestion
- subscribe to events from an Event Sourced Entity.
```
I see this "Event Sourced Entity" is the standard way of writing it
Also I can see most the bullet points in other pages ending with a period.

docs/src/modules/java-protobuf/partials/actions.adoc
```diff
@@ -13,4 +15,6 @@ Actions can be triggered in multiple ways. For example, by:
 - an HTTP service call
 - a forwarded call from another component
 - a scheduled call from a timer
+- an incoming message from a topic
 - an incoming event from within the same service or a from different service
```
franciscolopezsancho, line 19:
```suggestion
- an incoming event from an Event Sourced Entity, from within the same service or from a different service. 
```
Making explicit ESE and adding "a" before "different service".

docs/src/modules/java/partials/actions.adoc
franciscolopezsancho, line 1: Could be share this file between, `java-protobuf` and `java`? instead of duplicating it?

sdk/java-sdk-protobuf/src/main/java/kalix/javasdk/action/Action.java
```diff
@@ -30,6 +30,34 @@
 import java.util.Optional;
 import java.util.concurrent.CompletionStage;
 
+/**
```
franciscolopezsancho, line 33: I really like this comment making explicit the intent of the class. I think we could have many more and not only in the user facing APIs but the internal classes. To understand an implementation, specially if the class is complex, is necessary to understand its intention. Otherwise deviation of such intention is most likely not to be noticed.

sdk/java-sdk-protobuf/src/main/java/kalix/javasdk/action/Action.java
```diff
@@ -65,15 +95,34 @@ public final Effect.Builder effects() {
...
+   * operations that match the capabilities of that component.
+   * <p>
+   * An Action Effect can either:
+   * <p>
+   * <ul>
+   *   <li>reply with a message to the caller
+   *   <li>reply with a message to be published to a topic (in case the method is a publisher)
+   *   <li>forward the message to another component
+   *   <li>return an error
+   *   <li>ignore the call
```
franciscolopezsancho, line 123: I think this doesn't align with "what Kalix needs to do after the command is handled". If the command has been handled how can it be ignored after the fact?
How about this? Ignore further processing of the message, drop it, and process next element in the topic or return NotFound for a Request/Reponse. 

I would add that and effect is the last thing an action can do and must do for when processing a message.

sdk/java-sdk-protobuf/src/main/java/kalix/javasdk/eventsourcedentity/EventSourcedEntity.java
```diff
@@ -31,6 +31,25 @@
 import java.util.function.Function;
 
 /**
+ * The Event Sourced state model captures changes to data by storing events in a journal.
```
franciscolopezsancho, line 34: again lovely jubbly.

sdk/java-sdk-protobuf/src/main/java/kalix/javasdk/eventsourcedentity/EventSourcedEntity.java
```diff
@@ -116,6 +135,23 @@ protected final Effect.Builder<S, E> effects() {
...
+   * behalf and ensure that any data that needs to be persisted will be persisted.
+   * <p>
+   * Each Kalix component defines its own effects, which are a set of predefined
+   * operations that match the capabilities of that component.
+   * <p>
+   * An EventSourcedEntity Effect can either:
+   * <p>
+   * <ul>
+   *   <li>emit events and send a reply to the caller
+   *   <li>ignore the command by simply returning to the caller without emitting any events
```
franciscolopezsancho, line 149: There is not `ignore` in this class.

sdk/java-sdk-protobuf/src/main/java/kalix/javasdk/valueentity/ValueEntity.java
```diff
@@ -22,13 +22,34 @@
...
+import kalix.javasdk.eventsourcedentity.EventSourcedEntity;
 import kalix.javasdk.impl.valueentity.ValueEntityEffectImpl;
 import io.grpc.Status;
 
 import java.util.Collection;
 import java.util.Optional;
 
-/** @param <S> The type of the state for this entity. */
+/**
+ * Value Entities persist their state on every change. You can think of them as a Key-Value entity where
```
franciscolopezsancho, line 33: I would love if we call them "KeyValueEntity" instead of "ValueEntity".  I think is more expressive.

sdk/java-sdk-protobuf/src/main/java/kalix/javasdk/valueentity/ValueEntity.java
```diff
@@ -22,13 +22,34 @@
...
+ * the key is the entity id and the value is the state of the entity.
+ * <p>
+ * Kalix Value Entities have nothing in common with the domain-driven design concept of Value Objects.
+ * The Value in the name refers to the direct modification of the entity's state.
+ *
+ * When implementing a Value Entity, you first define what will be its internal state (your domain model),
+ * and the commands it will handle (mutation requests).
```
franciscolopezsancho, line 40: "mutation requests" doesn't go well with the below `directly reply to the caller if the command is not requesting any state change`. Not always the request are mutation ones.

sdk/java-sdk-protobuf/src/main/java/kalix/javasdk/eventsourcedentity/EventSourcedEntity.java
```diff
@@ -31,6 +31,25 @@
 import java.util.function.Function;
 
 /**
+ * The Event Sourced state model captures changes to data by storing events in a journal.
+ * The current entity state is derived from the emitted events.
+ * <p>
+ * When implementing an Event Sourced Entity, you first define what will be its internal state (your domain model),
+ * the commands it will handle (mutation requests) and the events it will emit (state changes).
```
franciscolopezsancho, line 38: "acceptable requests"? instead of "mutation requests"

sdk/java-sdk-protobuf/src/main/java/kalix/javasdk/valueentity/ValueEntity.java
franciscolopezsancho: Review to be continued.

sdk/java-sdk-protobuf/src/main/java/kalix/javasdk/view/View.java
```diff
@@ -17,11 +17,26 @@
...
 
 import akka.annotation.ApiMayChange;
+import kalix.javasdk.eventsourcedentity.EventSourcedEntity;
 import kalix.javasdk.impl.view.ViewUpdateEffectImpl;
 
 import java.util.Optional;
 
-/** @param <S> The type of the state for this view. */
+/**
+ * Views allows you to build new representations for the state of your entities that are indexed by fields different from the entity id.
```
aludwiko, line 26: I would mention the CQRS pattern.

docs/src/modules/java/partials/actions.adoc
octonato: Probably. I will look at it. Thanks for the suggestion.

sdk/java-sdk-protobuf/src/main/java/kalix/javasdk/eventsourcedentity/EventSourcedEntity.java
octonato: Agree, we can also have read-only commands. 

Not sure about "acceptable requests" though. I will try to improve it.

sdk/java-sdk-protobuf/src/main/java/kalix/javasdk/eventsourcedentity/EventSourcedEntity.java
octonato: No, maybe my sentence is not clear enough. I will remove the 'ignore' to avoid confusion. I mean to say, you can handle a command but decide to not mutate the state and simply return an acknowledgement.

sdk/java-sdk-protobuf/src/main/java/kalix/javasdk/valueentity/ValueEntity.java
octonato: We have been discussing renaming it. But for the moment, we will start by better documenting it.

docs/src/modules/java/partials/actions.adoc
```diff
@@ -4,7 +4,9 @@ Actions are stateless functions that can be used to implement different uses cas
 
 - a pure function
 - request conversion - you can use Actions to convert incoming data into a different format before forwarding a call to a different component.
-- publish and subscribe to events 
+- publish messages to a topic
+- subscribe to events from an event-sourced entity
```
octonato, line 8:
```suggestion
- subscribe to events from an Event Sourced Entity
```

docs/src/modules/java-protobuf/partials/actions.adoc
```diff
@@ -4,7 +4,9 @@ Actions are stateless functions that can be used to implement different uses cas
 
 - a pure function
 - request conversion - you can use Actions to convert incoming data into a different format before forwarding a call to a different component.
-- publish and subscribe to events 
+- publish messages to a topic
```
octonato, line 7:
```suggestion
- publish messages to a topic.
```

docs/src/modules/java-protobuf/partials/actions.adoc
```diff
@@ -4,7 +4,9 @@ Actions are stateless functions that can be used to implement different uses cas
...
+- subscribe to events from an event-sourced entity
+- subscribe to state changes from a value entity
```
octonato, line 9:
```suggestion
- subscribe to state changes from a value entity.
```

docs/src/modules/java/partials/actions.adoc
```diff
@@ -4,7 +4,9 @@ Actions are stateless functions that can be used to implement different uses cas
...
+- subscribe to state changes from a value entity
 - schedule and cancel timers
```
octonato, line 10:
```suggestion
- publish messages to a Topic
- subscribe to events from an Event Sourced Entity.
- subscribe to state changes from a Value Entity.
- schedule and cancel Timers.
```

sdk/java-sdk-protobuf/src/main/java/kalix/javasdk/action/Action.java
octonato: I prefer to not be too verbose on this description because we have more information down on the method itself. 

For example, the `ignore` method is explaining in detail what are the semantics.

sdk/java-sdk-protobuf/src/main/java/kalix/javasdk/eventsourcedentity/EventSourcedEntity.java
octonato: I'm replacing it by: 

> the commands it will handle and the events it will emit to modify its state

So not being explicit about what a command handler is. It can be a read-only or a command that emits events.

ennru (commented):
Great improvements.

sdk/java-sdk-protobuf/src/main/java/kalix/javasdk/eventsourcedentity/EventSourcedEntity.java
octonato: I removed this sentence. Confusing and the use case that I was trying to explain is something that the user can easily learn by themselves.

docs/src/modules/java-protobuf/partials/actions.adoc
octonato: added

sdk/java-sdk-protobuf/src/main/java/kalix/javasdk/valueentity/ValueEntity.java
franciscolopezsancho: No more comments. The rest is a repetition of the previous or I have nothing to add.

franciscolopezsancho (approved):
LGTM

sdk/java-sdk-protobuf/src/main/java/kalix/javasdk/eventsourcedentity/EventSourcedEntity.java
franciscolopezsancho: yeah, simple and expressive.

PR #1876, opened 2023-11-21
Octonato/move to junit5

octonato:
Experiment to support JUnit5


samples/java-protobuf-customer-registry-kafka-quickstart/pom.xml
```diff
@@ -284,9 +284,21 @@
...
+    </dependency>
+    <dependency>
+      <groupId>org.junit.jupiter</groupId>
+      <artifactId>junit-jupiter</artifactId>
+      <version>5.10.1 </version>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.junit.vintage</groupId>
+      <artifactId>junit-vintage-engine</artifactId>
```
octonato, line 300: Using the vintage-engine, a can continue to use the junit4. However, the reality is that to move to junit5, a user needs to update their pom.xml anyway. So, it's a conscious decision. 

A user can eventually update the SDK version and stay on JUnit 4. If later they add a new component, the generate tests for the new component will have imports for JUnit5 and the project will fail to compile. 

Alternatively, we can add junit4, junit5 and junit-vintage-engine as transitive dependency of `kalix-java-sdk-protobuf-testkit` as such the users will always have the right dependencies on their classpath.

octonato: btw, junit4 and junit5 are already include in the testkit, but marked as provided.

sdk/java-sdk-protobuf-testkit/src/main/java/kalix/javasdk/testkit/junit/KalixTestKitResource.java
```diff
@@ -60,7 +63,7 @@
  * }
  * </pre>
  */
-public final class KalixTestKitResource extends ExternalResource {
+public final class KalixTestKitResource extends ExternalResource implements BeforeAllCallback, AfterAllCallback {
```
octonato, line 66: When implementing this two interfaces (`BeforeAllCallback` and `AfterAllCallback`), we can use this class as a Registered Extension. Junit will take care of the lifecycle for us. This is the equivalent of `ClassRule` in junit4.


`KalixTestKitResource` has already a hard dependency on junit4. In the sbt project, junit4 is marked as provided and then we add it to the pom explicitly. 

With this change, the `KalixTestKitResource` will also depend on junit5 and we will need both in the pom. That's problematic for existing projects. If a user update the SDK version, but doesn't touch the pom, they will get an error.

Therefore, as mentioned above, we can solve it by added junit libraries as transitive dependencies.

samples/java-protobuf-customer-registry-kafka-quickstart/src/it/java/customer/api/CustomerActionWithKafkaIntegrationTest.java
```diff
@@ -38,7 +38,7 @@ public class CustomerActionWithKafkaIntegrationTest {
   /**
    * The test kit starts both the service container and the Kalix proxy.
    */
-  @ClassRule
+  @RegisterExtension
```
octonato, line 41: This will behave as the `ClassRule` annotation if the class (ie `KalixTestKitResource`) implements `BeforeAllCallback` and `AfterAllCallback`.

octonato (commented):
@ennru, here is the results of first experiments. 

I modified only one sample. The integration tests are using junit5, but the unit tests are still on junit4. It works with both if we add the vintage engine.

We can add support for junit5 and still support junit4 at the same time. 

Newly generated test classes will be on junit5 though. Users really willing to stick with junit4 will have to change the generated tests.

ennru:
Possibly interesting changes in Testcontainers v1.18.0:
- https://github.com/testcontainers/testcontainers-java/pull/5985

project/Dependencies.scala
```diff
@@ -125,8 +126,9 @@ object Dependencies {
 
   val javaSdkTestKit = deps ++= Seq(
     testContainers,
-    junit4 % Provided,
-    junit5 % Provided,
```
octonato, line 129: This is the most impactful change. 

So far, our testkit was depending directly on a class from junit4. It has junit5 as dependency, but it was not used. 

Users were required to have junit4 in their pom because it was `Provided`. 

It's annoying now to keep them as `Provided` because the testkit now requires both libraries. That means that users will need to add both to their pom.xml.

By making them transitive through the testkit, we ensure that users will have both and can pick the one they prefer. 

On our side, new projects (created from mvn archetype) and new generated code will use junit5.

sdk/java-sdk-spring/src/it/java/com/example/wiring/tracing/TracingIntegratonTest.java
```diff
@@ -20,7 +20,7 @@
 import com.example.wiring.pubsub.DockerIntegrationTest;
 import com.typesafe.config.Config;
 import com.typesafe.config.ConfigFactory;
-import org.junit.Ignore;
```
franciscolopezsancho, line 23: still using Ignore in L54

project/Dependencies.scala
aludwiko: I don't see any point in supporting both junit versions. JUnit 5 was released in 2017. We could support both for a short transition time, but we should be very clear that migration is indispensable.

aludwiko (approved):
LGTM

samples/java-protobuf-doc-snippets/src/main/java/com/example/Main.java
```diff
@@ -0,0 +1,33 @@
+package com.example;
```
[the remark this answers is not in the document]
octonato: No, that was a mistake. And I'm surprised that they are not checked in. But I get why. We only build it to verify correctness and then the generated classed are each time recreated.

At some point, when chasing for compilation errors, I wrote a script to compile all projects under samples. That's how those got generated on my machine.

I will revert it.

sdk/java-sdk-protobuf-testkit/src/main/java/kalix/javasdk/testkit/junit/KalixTestKitResource.java
octonato: Hmm, that could be an option. 

That will make it easier for us to deprecate junit4 support.

I will PR this.

samples/java-protobuf-valueentity-counter/src/test/java/com/example/actions/ExternalCounterActionTest.java
```diff
@@ -32,12 +33,12 @@ public class ExternalCounterActionTest {
 
   private AutoCloseable closeable;
 
-  @Before
+  @BeforeEach
   public void setUp() {
     closeable = MockitoAnnotations.openMocks(this);
   }
 
-  @Before
```
octonato, line 40: For the record, that was my WTF moment of the day. 

I don't know how this was working before, but the `tearDown` method should be annotated with `After`, not `Before`. 

What I'm guessing is that Junit4 will only run the first method it finds annotated with `Before`. And the `tearDown` where never called.

PR #1883, opened 2023-11-23
chore: remove obsolete jvm properties

octonato:


ennru (approved):
LGTM.

maven-java/kalix-maven-archetype-event-sourced-entity/src/main/resources/archetype-resources/docker-compose.yml
```diff
@@ -13,8 +13,7 @@ services:
       - "host.docker.internal:host-gateway"
     environment:
       JAVA_TOOL_OPTIONS: >
-        -Dconfig.resource=dev-mode.conf
-        -Dlogback.configurationFile=logback-dev-mode.xml
+      # jvm -D properties can be under this environment map (note: remove this comment when adding properties)
```
aludwiko, line 16: It's too late, but it would be nice to inform what env variable name should be used for that.

octonato: what you mean? they are many one can use?

aludwiko: ahh, nevermind, for a second, I thought that you also removed this line `JAVA_TOOL_OPTIONS: >`

octonato: I was planning to remove it, but then I saw that we have a comment at the top of the file mentioning this env var. If would be confusing if we didn't have a placeholder. Also confusing without any comment. Hence my comment just under it.

PR #1886, opened 2023-11-27
feat: allow Config injection

octonato:


maven-java/kalix-maven-plugin/src/main/scala/kalix/RunMojo.scala
```diff
@@ -29,29 +29,30 @@ object RunMojo {
...
-    }
+    def collectSysProperties: Seq[String] =
+      sys.env
+        .get("MAVEN_CMD_LINE_ARGS")
+        .map { args =>
+          args.split("\\s+").toSeq.collect {
+            case arg if arg.startsWith("-D") => arg
+          }
+        }
+        .getOrElse(Seq.empty)
```
octonato, line 47: We were previously picking all properties starting with `kalix` and making it a `-D` property to pass to the underlying process. 

We cannot take all sys properties and pass them down as `-D` args. Instead, we need to read the env var MAVEN_CMD_LINE_ARGS. This contains all the args passed to mvn. We can then take everything that has an explicit `-D` in front of it and consider as an argument to the use application.

sdk/java-sdk-spring/src/main/scala/kalix/spring/impl/KalixSpringApplication.scala
```diff
@@ -378,6 +378,8 @@ case class KalixSpringApplication(applicationContext: ApplicationContext, config
           case p if p == classOf[KalixClient]           => kalixClient(context)
           case p if p == classOf[ComponentClient]       => componentClient(context)
           case p if p == classOf[WebClientProvider]     => webClientProvider(context)
+          // we wire the Akka Config instance as it's the final one (see KalixRunner.prepareConfig)
+          case p if p == classOf[Config] => context.materializer().system.settings.config
```
octonato, line 382: Allowing users to inject Typesafe config in their components.

sdk/java-sdk-spring/src/main/scala/kalix/spring/impl/KalixSpringApplication.scala
```diff
@@ -424,26 +428,45 @@ case class KalixSpringApplication(applicationContext: ApplicationContext, config
...
   private def valueEntityProvider[S, VE <: ValueEntity[S]](clz: Class[VE]): ValueEntityProvider[S, VE] =
     ReflectiveValueEntityProvider.of(
       clz,
       messageCodec,
-      context => wiredInstance(clz) { case p if p == classOf[ValueEntityContext] => context })
+      context =>
+        wiredInstance(clz) {
+          case p if p == classOf[ValueEntityContext] => context
+          // we wire the Akka Config instance as it's the final one (see KalixRunner.prepareConfig)
+          case p if p == classOf[Config] => context.materializer().system.settings.config
```
aludwiko, line 446: I think this line can be moved to `wiredInstance`. This way, we will avoid code duplication.

aludwiko (approved):
Approved, with a small improvement suggestion.

sdk/java-sdk-spring/src/main/scala/kalix/spring/impl/KalixSpringApplication.scala
octonato: Of course! 

I will clean this up.

sdk/java-sdk-spring/src/test/java/kalix/spring/badwiring/workflow/Main.java
```diff
@@ -31,6 +31,6 @@ public class Main {
 
   public static void main(String[] args) {
     logger.info("Starting Kalix Application");
-    SpringApplication.run(kalix.spring.badwiring.view.Main.class, args);
+    SpringApplication.run(Main.class, args);
```
octonato, line 34: This was a mistake that I just noticed and fixing now. 

However, I'm quite confused by this. I'm sure we had test that would effectively run this classes to verify if our wiring rules are correct. But I can't remember how we did that and where is this code.

octonato: Oh, I found it. Nevermind. 
It's in KalixSpringApplicationSpec

But it's not covering Workflows. I will fix it and add some more tests to prove that we can wire config + context for all components and component clients for Actions and Workflows.

octonato: I didn't add extra tests here. It's easier to test if something can't be wired because Spring fails, but if it can be wired, Spring will just start as normal and doing that inside a unit test is producing all sort of other errors. 

We should certainly build something to assert it, but it will require some more investigation.

octonato: created an issue for this: https://github.com/lightbend/kalix-jvm-sdk/issues/1888

Issue #1898, opened 2023-12-07
use more generic typeUrl for Java primitives for code first sdk


aludwiko:
The idea is to change for instance, `json.kalix.io/java.lang.String` to `json.kalix.io/string`.

PR #1899, opened 2023-12-07
fix: generic typeUrls for Java primitives

aludwiko:
References #1898

### DIFF
--- a/sdk/java-sdk-spring/src/main/scala/kalix/javasdk/impl/JsonMessageCodec.scala
+++ b/sdk/java-sdk-spring/src/main/scala/kalix/javasdk/impl/JsonMessageCodec.scala
@@ -16,6 +16,7 @@
 
 package kalix.javasdk.impl
 
+import java.lang
 import java.util.concurrent.ConcurrentHashMap
 import java.util.concurrent.ConcurrentMap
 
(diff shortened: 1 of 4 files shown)


sdk/java-sdk-spring/src/main/scala/kalix/javasdk/impl/JsonMessageCodec.scala
```diff
@@ -82,18 +81,43 @@ private[kalix] class JsonMessageCodec extends MessageCodec {
...
+      } else if (clz.isAssignableFrom(classOf[Character])) {
+        TypeHint("char", List("char", "java.lang.Character"))
+      } else if (clz.isAssignableFrom(classOf[lang.Byte])) {
+        TypeHint("byte", List("byte", "java.lang.Byte"))
+      } else if (clz.isAssignableFrom(classOf[Short])) {
+        TypeHint("short", List("short", "java.lang.Short"))
+      } else {
+        TypeHint(clz.getName, List(clz.getName))
+      }
+      typeHint.allTypeHints.foreach(className => addToReversedCache(clz, className))
```
octonato, line 106: Whoever designed this codec must be a genius. `allTypeHints` was already to cover this kind of situation! 😄

sdk/java-sdk-spring/src/main/scala/kalix/javasdk/impl/JsonMessageCodec.scala
```diff
@@ -82,18 +81,43 @@ private[kalix] class JsonMessageCodec extends MessageCodec {
...
+      } else if (clz.isAssignableFrom(classOf[Float])) {
+        TypeHint("float", List("float", "java.lang.Float"))
+      } else if (clz.isAssignableFrom(classOf[Character])) {
+        TypeHint("char", List("char", "java.lang.Character"))
+      } else if (clz.isAssignableFrom(classOf[lang.Byte])) {
+        TypeHint("byte", List("byte", "java.lang.Byte"))
+      } else if (clz.isAssignableFrom(classOf[Short])) {
+        TypeHint("short", List("short", "java.lang.Short"))
+      } else {
+        TypeHint(clz.getName, List(clz.getName))
```
octonato, line 104: We are missing a json array. I never tried it, but would it be possible to have an entity with the state set to `List<Foo>`? That will (should) be converted to a json array of `Foo` objects.

octonato: Ok, let's leave it. As long as we don't support it as entity state, we don't need to have support for this here.

PR #1901, opened 2023-12-11
chore: some more proxy > runtime renaming

octonato:


sdk/java-sdk-protobuf-testkit/src/main/java/kalix/javasdk/testkit/KalixProxyContainer.java
octonato, line 1: This class references 'proxy' a few times and it also have some constants mentioning proxy and user_function. 

Not change it right now, because this class is public. Although it should not be public.
I would move this class to `.testkit.impl`. No reason to have it as public.

sdk/java-sdk-protobuf/src/main/scala/kalix/javasdk/impl/ProxyInfoHolder.scala
```diff
@@ -63,10 +63,10 @@ class ProxyInfoHolder(system: ExtendedActorSystem) extends Extension {
     _identificationInfo = proxyInfo.identificationInfo
     _proxyTracingCollectorEndpoint = Some(proxyInfo.tracingCollectorEndpoint)
 
-    log.debug("Proxy hostname: [{}]", chosenProxyName)
-    log.debug("Proxy port to: [{}]", proxyInfo.proxyPort)
```
octonato, line 67: The protobuf type is still called `ProxyInfo`, but I think we can safely rename it in the runtime to `RuntimeInfo` and it will work anyway, no?

@johanandren could you confirm this? 
If I have two messages: `Foo` and `Bar` and they have the same fields in the same order, doesn't matter the name of the messages in the descriptor, we will always be able to deserialize a payload from `Foo` into `Bar`, correct?

Issue #1858, opened 2023-11-15
Run something at startup after discovery


patriknw:
For example creating a timer at startup would not work until the runtime has made the discovery call and initialized the service.

johanandren:
Important aspect: should it be run on every node startup, or once per deploy of a given version of the user service?

johanandren:
For the record, on every node startup is relatively easy and can be done in SDK only, while once per service version deploy is something that needs support in protocol and runtime.

efgpinto:
I was thinking on doing this based on the 1st heartbeat received. The sdk would trigger the "onStartup" when receiving it. Seems like the easiest way we have currently to know if discovery succeded.

Parsing logs coming from the runtime would be another option but I think it's more error prone.
This is if we want to avoid having changes on the Runtime side of course.

johanandren:
That's not 100% safe because the runtime hits the service healthcheck in parallel with sharding getting ready, but maybe it is still good enough vs the work to add a new discovery endpoint method to signal ready (or parse logs fed from the runtime)

efgpinto:
Some notes from live SUP brainstorming:
- should the "onStartup" be defined right on `Main` or rather as a specific `Action`?
  - within `Main` there is no access to component calls, timers, etc. Having this modelled as an Action seems a good fit.
  - Action / Action methods would be annotated so Runtime knows when to call a specific one. For now, only a type "onStartup" but could have others in future for similar "onRollingUpdate", "beforeReady", etc. 
- support similar execution guarantees than timers: action method is called by runtime when time comes and can be retried for a specific amount of time  (depending on config) until it suceeds. We should probably allow for fire-and-forget as well.

octonato:
One thing that we didn't explicit mention. Such an action method will need to receive proto `Empty` or be a parameterless method in code-first SDK. Because the runtime can't build an specific message to it.

ceecer1:
Westpac Architect's email description:

> Here is a simple example of the requirement to cache config/reference data:
>  
> Our entities support internal versioning. We need to be able to configure which versions are active so we can do a form of blue/green deployment.
>  
> This config data rarely changes. It applies equally to every instance of the same entity.
>  
> The only current way to provide the entity with such config data is by including it in the relevant commands. In this case it would need to be added to every single command! If that is not done correctly for each and every command then the consequences could be catastrophic.
>  
> Multiply this problem by the dozens of config/ref data in our solution and the issue becomes even worse.
>  
[...]

PR #1909, opened 2023-12-12
docs: actions as service life-cycle hooks

efgpinto:
Refs https://github.com/lightbend/kalix-jvm-sdk/issues/1858

### DIFF
--- a/docs/src/modules/java-protobuf/pages/actions.adoc
+++ b/docs/src/modules/java-protobuf/pages/actions.adoc
@@ -362,7 +362,9 @@ Such validation that depends on state can only be done safely while handling the
 
 == Actions as Life-cycle Hooks
 
-An Action's gRPC method can be triggered automatically when some predefined service life-cycle event happens (currently, only on startup is available), serving as a custom hook. For such use, the method needs be annotated with a specific flag as shown below and its input type must be `google.protobuf.Empty`.
+An Action's gRPC method can be triggered automatically when some predefined service life-cycle event happens (currently, only on startup is available), serving as a custom hook. For such use, the method needs be annotated with a specific flag and its input type must be `google.protobuf.Empty` as shown below.
+
+IMPORTANT: The on startup hook is called every time a service instance boots up. This can happen for very different reasons: restarting / redeploying the service, scaling up to more instances or even without any user-driven action (e.g. Kalix Runtime versions being rolled out, infrastruture-related incidents, etc.). Therefore, you should carefully consider how you use this hook and its implementation.
 
 [source,proto]
 ----
@@ -374,6 +376,7 @@ include::example$java-protobuf-doc-snippets/src/main/proto/com/example/on_startu
 <4> optionally, set the max amount of retries to 3.
 
 NOTE: If the call to the hook returns a failure and the `max_retries` is set to a value greater than the default value (`0`), a number of retry calls will be executed with a fixed delay up until the configure amount is reached.
+
 
 == Running Side Effects
 
--- a/docs/src/modules/java-protobuf/pages/actions.adoc
+++ b/docs/src/modules/java-protobuf/pages/actions.adoc
@@ -360,6 +360,20 @@ action returning and the `addItem` call from the action would lead to more items
 
 Such validation that depends on state can only be done safely while handling the command inside the entity.
 
+== Actions as Life-cycle Hooks
+
+An Action's gRPC method can be triggered automatically when some predefined service life-cycle event happens, serving as a custom hook. For such use, the method needs be annotated with a specific flag as shown below and its must be of type `google.protobuf.Empty`.
+
+[source,proto]
+----
+include::example$java-protobuf-doc-snippets/src/main/proto/com/example/on_startup.proto[tag=hook]
+----
+<1> only methods belonging to an Action can be configured as a hook;
+<2> method must receive `google.protobuf.Empty` as input type;
+<3> select which event triggers this hook (in this case, this is once instance startup is concluded);
+<4> optionally, set the max amount of retries to 3.
+
+NOTE: If the call to the hook returns a failure and the `max_retries` is set to a value greater than the default value (`0`), a number of retry calls will be executed with a fixed delay up until the configure amount is reached.
 
 == Running Side Effects
 
--- a/docs/src/modules/java-protobuf/pages/actions.adoc
+++ b/docs/src/modules/java-protobuf/pages/actions.adoc
@@ -362,14 +362,14 @@ Such validation that depends on state can only be done safely while handling the
 
 == Actions as Life-cycle Hooks
 
-An Action's gRPC method can be triggered automatically when some predefined service life-cycle event happens, serving as a custom hook. For such use, the method needs be annotated with a specific flag as shown below and its must be of type `google.protobuf.Empty`.
+An Action's gRPC method can be triggered automatically when some predefined service life-cycle event happens, serving as a custom hook. For such use, the method needs be annotated with a specific flag as shown below and it must be of type `google.protobuf.Empty`.
 
 [source,proto]
 ----
 include::example$java-protobuf-doc-snippets/src/main/proto/com/example/on_startup.proto[tag=hook]
 ----
 <1> only methods belonging to an Action can be configured as a hook;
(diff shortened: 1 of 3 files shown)


samples/java-protobuf-doc-snippets/src/main/proto/com/example/on_startup.proto
```diff
@@ -0,0 +1,34 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// tag::counter-topic[]
+syntax = "proto3";
+package com.example.actions;
+
+import "kalix/annotations.proto";
+import "google/protobuf/empty.proto";
```
efgpinto, line 20:
```suggestion
import "google/protobuf/empty.proto";

```

docs/src/modules/java-protobuf/pages/actions.adoc
```diff
@@ -360,6 +360,20 @@ action returning and the `addItem` call from the action would lead to more items
 
 Such validation that depends on state can only be done safely while handling the command inside the entity.
 
+== Actions as Life-cycle Hooks
+
+An Action's gRPC method can be triggered automatically when some predefined service life-cycle event happens, serving as a custom hook. For such use, the method needs be annotated with a specific flag as shown below and its must be of type `google.protobuf.Empty`.
```
octonato, line 365:
```suggestion
An Action's gRPC method can be triggered automatically when some predefined service life-cycle event happens, serving as a custom hook. For such use, the method needs be annotated with a specific flag as shown below and it must be of type `google.protobuf.Empty`.
```

docs/src/modules/java-protobuf/pages/actions.adoc
```diff
@@ -360,6 +360,20 @@ action returning and the `addItem` call from the action would lead to more items
...
+
+[source,proto]
+----
+include::example$java-protobuf-doc-snippets/src/main/proto/com/example/on_startup.proto[tag=hook]
+----
+<1> only methods belonging to an Action can be configured as a hook;
+<2> method must receive `google.protobuf.Empty` as input type;
```
octonato, line 372:
```suggestion
<2> the method must receive `google.protobuf.Empty` as input type;
```

docs/src/modules/java-protobuf/pages/actions.adoc
```diff
@@ -360,6 +360,20 @@ action returning and the `addItem` call from the action would lead to more items
...
+<3> select which event triggers this hook (in this case, this is once instance startup is concluded);
```
octonato, line 373: We only have STARTUP, so it can be misleading to say that we can select the event.

docs/src/modules/java-protobuf/pages/actions.adoc
octonato, line 365: > when some predefined service life-cycle event happens

This gives the impression that we can chose from a list of life-cycle events, but we can only do it on start-up. Better to be explicit about it.

docs/src/modules/java-protobuf/pages/actions.adoc
```diff
@@ -360,6 +360,20 @@ action returning and the `addItem` call from the action would lead to more items
...
+<4> optionally, set the max amount of retries to 3.
+
+NOTE: If the call to the hook returns a failure and the `max_retries` is set to a value greater than the default value (`0`), a number of retry calls will be executed with a fixed delay up until the configure amount is reached.
```
octonato, line 376: Just confirming.
We don't support infinite retries, correct? 
User need to use max int if they want to try many many times, but not infinite times. 

It's actually a good strategy. Something failing forever is clearly broken. Makes no sense to have infinite retries.

efgpinto: thats correct. 👍

docs/src/modules/java-protobuf/pages/actions.adoc
efgpinto: yeah, I was preparing for other events being added.. but maybe better to be explicit for now. I will update.

docs/src/modules/java-protobuf/pages/actions.adoc
```diff
@@ -360,6 +360,20 @@ action returning and the `addItem` call from the action would lead to more items
 
 Such validation that depends on state can only be done safely while handling the command inside the entity.
 
+== Actions as Life-cycle Hooks
+
+An Action's gRPC method can be triggered automatically when some predefined service life-cycle event happens (currently, only on startup is available), serving as a custom hook. For such use, the method needs be annotated with a specific flag as shown below and it must be of type `google.protobuf.Empty`.
```
aludwiko, line 365: how about "and the input type must be `...`"

docs/src/modules/java-protobuf/pages/actions.adoc
```diff
@@ -360,6 +360,20 @@ action returning and the `addItem` call from the action would lead to more items
...
+
+[source,proto]
+----
+include::example$java-protobuf-doc-snippets/src/main/proto/com/example/on_startup.proto[tag=hook]
+----
+<1> only methods belonging to an Action can be configured as a hook;
```
aludwiko, line 371: Do we use `;` in other parts of the documentation?

docs/src/modules/java-protobuf/pages/actions.adoc
```diff
@@ -360,6 +360,20 @@ action returning and the `addItem` call from the action would lead to more items
...
+<2> the method must receive `google.protobuf.Empty` as input type;
+<3> this hook will be triggered once instance startup is concluded;
```
aludwiko, line 373: I'm sorry, I haven't tracked the actual implementation, is it once per each instance, or per kalix service (singleton)?

docs/src/modules/java-protobuf/pages/actions.adoc
efgpinto, line 365:
```suggestion
An Action's gRPC method can be triggered automatically when some predefined service life-cycle event happens (currently, only on startup is available), serving as a custom hook. For such use, the method needs be annotated with a specific flag as shown below and its input type must be `google.protobuf.Empty`.
```

docs/src/modules/java-protobuf/pages/actions.adoc
efgpinto, line 373:
```suggestion
<1> Only methods belonging to an Action can be configured as a hook.
<2> The method must receive `google.protobuf.Empty` as input type.
<3> This hook will be triggered once the instance startup is concluded (i.e. will be called 3 times if 3 instances are running).
```

efgpinto: It is indeed once per instance. I noticed the phrase is a bit ambigious, see the new version below.

docs/src/modules/java-protobuf/pages/actions.adoc
efgpinto: fixed now, tks.

docs/src/modules/java-protobuf/pages/actions.adoc
```diff
@@ -360,6 +360,20 @@ action returning and the `addItem` call from the action would lead to more items
...
+
+An Action's gRPC method can be triggered automatically when some predefined service life-cycle event happens (currently, only on startup is available), serving as a custom hook. For such use, the method needs be annotated with a specific flag as shown below and its input type must be `google.protobuf.Empty`.
+
+[source,proto]
+----
+include::example$java-protobuf-doc-snippets/src/main/proto/com/example/on_startup.proto[tag=hook]
+----
+<1> Only methods belonging to an Action can be configured as a hook.
+<2> The method must receive `google.protobuf.Empty` as input type.
+<3> This hook will be triggered once the instance startup is concluded (i.e. will be called 3 times if 3 instances are running).
```
aludwiko, line 373: I think this deserves even more explanation it's not only about the startup but also restarts (sometimes not triggered by the user) and scaling up.

efgpinto: yeah, I think you're right, let me give it another try. :)

aludwiko (approved):
LGTM

PR #1918, opened 2023-12-14
docs: documenting persistent dev mode

aludwiko:


samples/java-protobuf-customer-registry-kafka-quickstart/docker-compose.yml
```diff
@@ -3,11 +3,15 @@ services:
   kalix-runtime:
```
aludwiko, line 3: Once we agree that the docker-compose change is ok, I will change all compose files in samples.

efgpinto: unless in a specific sample we demonstrate something that needs persistence, I would not change all the sample compose files to include this. I think maybe in the archetype we can put this (we have other flags there as well) but on the samples I don't see much benefit.

aludwiko: I'm a bit biased because I was using this a lot the last few days, but for me, the benefit is that the config is there, I don't have to look for it in the docs and copy-paste some lines. It's just a matter of uncommenting and I'm ready to go.

docs/src/modules/java-protobuf/pages/running-locally.adoc
```diff
@@ -123,7 +123,20 @@ grpcurl ^
 
 == Shut down the service
```
ennru, line 124:
```suggestion
== Shutting down the service
```

docs/src/modules/java-protobuf/pages/running-locally.adoc
```diff
@@ -123,7 +123,20 @@ grpcurl ^
...
 
-Use `Ctrl+C` to shut down the service. When stopping your service, it will also shutdown the Kalix Runtime container and any other container you have defined in `docker-compose.yml`. Unless you have chosen to start `docker-compose` manually in a separate terminal, in which case you will need to stop it manually as well. 
+Use `Ctrl+C` to shut down the service. When stopping your service, it will also shutdown the Kalix Runtime container and any other container you have defined in `docker-compose.yml`. Unless you have chosen to start `docker-compose` manually in a separate terminal, in which case you will need to stop it manually as well.
+
+== Running service with persistence enabled
+
+By default, Kalix services are configured to run with disabled persistence. This means that the Kalix Runtime will use an in-memory database to store the state of your service. This is useful for local development since it allows you to quickly start and stop your service without having to worry about cleaning up the database.
```
ennru, line 130:
```suggestion
By default, Kalix services are configured to run locally with persistence disabled. This means the Kalix Runtime will use an in-memory datastore for the state of your services. This is useful for local development since it allows you to quickly start and stop your service without having to worry about cleaning the database.
```

docs/src/modules/java-protobuf/pages/running-locally.adoc
```diff
@@ -123,7 +123,20 @@ grpcurl ^
...
+
+However, if you want to run your service with persistence enabled, you can do so by changing the `docker-compose` file as below.
```
ennru, line 132:
```suggestion
However, if you want to run your service with persistence enabled to keep the data when restarting, you may update the `docker-compose` file as below.
```

docs/src/modules/java/pages/running-locally.adoc
```diff
@@ -87,6 +87,19 @@ curl -XGET localhost:9000/customer/one
 
 Use `Ctrl+C` to shut down the service. When stopping your service, it will also shutdown the Kalix Runtime container and any other container you have defined in `docker-compose.yml`. Unless you have chosen to start `docker-compose` manually in a separate terminal, in which case you will need to stop it manually as well.
 
+== Running service with persistence enabled
+
+By default, Kalix services are configured to run with disabled persistence. This means that the Kalix Runtime will use an in-memory database to store the state of your service. This is useful for local development since it allows you to quickly start and stop your service without having to worry about cleaning up the database.
```
ennru, line 92:
```suggestion
By default, Kalix services are configured to run locally with persistence disabled. This means the Kalix Runtime will use an in-memory datastore for the state of your services. This is useful for local development since it allows you to quickly start and stop your service without having to worry about cleaning the database.
```

docs/src/modules/java/pages/running-locally.adoc
```diff
@@ -87,6 +87,19 @@ curl -XGET localhost:9000/customer/one
...
+
+However, if you want to run your service with persistence enabled, you can do so by changing the `docker-compose` file as below.
```
ennru, line 94:
```suggestion
However, if you want to run your service with persistence enabled to keep the data when restarting, you may update the `docker-compose` file as below.
```

ennru (approved):
LGTM.

PR #1920, opened 2023-12-15
fix: Hello world action exposed to the internet in code first archetypes

aludwiko:


maven-java/kalix-spring-boot-kotlin-archetype/src/main/resources/archetype-resources/pom.xml
```diff
@@ -24,7 +24,7 @@
     <dockerImage>${D}{kalixContainerRegistry}/${D}{kalixOrganization}/${D}{project.artifactId}</dockerImage>
     <dockerTag>${project.version}-${build.timestamp}</dockerTag>
     <maven.build.timestamp.format>yyyyMMddHHmmss</maven.build.timestamp.format>
-    <mainClass>${package}.Main</mainClass>
+    <mainClass>${package}.MainKt</mainClass>
```
aludwiko, line 27: I'm not a Kotling expert, but without this, the `mvn kalix:runAll` command doesn't work.

maven-java/kalix-spring-boot-archetype/src/main/resources/archetype-resources/src/main/java/__packageInPathFormat__/hello/HelloWorld.java
```diff
@@ -0,0 +1,20 @@
...
+import kalix.javasdk.action.Action;
+import kalix.javasdk.annotations.Acl;
+import org.springframework.web.bind.annotation.GetMapping;
+import org.springframework.web.bind.annotation.RequestMapping;
+
+@RequestMapping
+/**
+ * This is a simple Action that returns "Hello World!" and it's exposed to the Internet.
+ * Locally, you can access it by running `curl http://localhost:9000/hello`.
+ */
```
ennru, line 12: Is it intentional to have the annotation before the class comment?
Look non-standard to me.

maven-java/kalix-spring-boot-kotlin-archetype/src/main/resources/archetype-resources/pom.xml
ennru, line 27: Is this correct? AFAICS, it is `Main` in `Main.kt`.

maven-java/kalix-spring-boot-archetype/src/main/resources/archetype-resources/src/main/java/__packageInPathFormat__/hello/HelloWorld.java
aludwiko: yeah, I saw the same in the main class:
```
@SpringBootApplication
// Allow all other Kalix services deployed in the same project to access the components of this
// Kalix service, but disallow access from the internet. This can be overridden explicitly
// per component or method using annotations.
// Documentation at https://docs.kalix.io/services/using-acls.html
@Acl(allow = @Acl.Matcher(service = "*"))
public class Main {

  private static final Logger logger = LoggerFactory.getLogger(Main.class);

  public static void main(String[] args) {
    logger.info("Starting Kalix Application");
    SpringApplication.run(Main.class, args);
  }
}
```

ennru (approved):
LGTM.

maven-java/kalix-spring-boot-kotlin-archetype/src/main/resources/archetype-resources/pom.xml
aludwiko: Yeah, Kotlin magic, I've tested this locally.

maven-java/kalix-spring-boot-archetype/src/main/resources/archetype-resources/src/main/java/__packageInPathFormat__/hello/HelloWorld.java
octonato: This is different. 

The following is a javadoc.
```
/**
 *
 */
```

While the comment in the Main.class is a comment about the `Acl` annotation that is just under it.

PR #1924, opened 2023-12-15
feat: support on startup hook for java sdk

efgpinto:
Close https://github.com/lightbend/kalix-jvm-sdk/issues/1858

Depends on https://github.com/lightbend/kalix-runtime/pull/2253

### DIFF
--- a/sdk/java-sdk-spring/src/main/java/kalix/javasdk/annotations/Trigger.java
+++ b/sdk/java-sdk-spring/src/main/java/kalix/javasdk/annotations/Trigger.java
@@ -1,0 +1,54 @@
+/*
+ * Copyright 2021 Lightbend Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package kalix.javasdk.annotations;
+
+import java.lang.annotation.*;
+
+/**
+ * This annotation is used to mark a method as a hook that will be triggered automatically uppon certain events.
+ * The method must be public and have no parameters.
+ * If the call fails, it will be retried up to the number of times specified by the maxRetries parameter.
+ */
+@Target(ElementType.METHOD)
+@Retention(RetentionPolicy.RUNTIME)
+@Documented
+public @interface Trigger {
+  enum TriggerEvent {
+    UNSPECIFIED,
+
+    /**
+     * The on startup hook is called every time a service instance boots up.
+     * This can happen for very different reasons: restarting / redeploying the service,
+     * scaling up to more instances or even without any user-driven action
+     * (e.g. Kalix Runtime versions being rolled out, infrastructure-related incidents, etc.).
+     * Therefore, one should carefully consider how to use this hook and its implementation.
+     */
+    STARTUP,
+  }
+
+
+  /**
+   * The service life-cycle event for which this hook will be triggered.
+   */
+  TriggerEvent on() default TriggerEvent.UNSPECIFIED;
+
+  /**
+   * The maximum number of retries we will do upon failure of the method hook calls.
+   * The default value 0 means no retries are done.
+   */
+  int maxRetries() default 0;
+}
--- a/sdk/java-sdk-spring/src/main/java/kalix/javasdk/annotations/Trigger.java
+++ b/sdk/java-sdk-spring/src/main/java/kalix/javasdk/annotations/Trigger.java
@@ -28,8 +28,6 @@ import java.lang.annotation.*;
 @Documented
 public @interface Trigger {
   enum TriggerEvent {
-    UNSPECIFIED,
-
     /**
      * The on startup hook is called every time a service instance boots up.
      * This can happen for very different reasons: restarting / redeploying the service,
@@ -44,7 +42,7 @@ public @interface Trigger {
   /**
    * The service life-cycle event for which this hook will be triggered.
    */
-  TriggerEvent on() default TriggerEvent.UNSPECIFIED;
+  TriggerEvent on();
 
   /**
    * The maximum number of retries we will do upon failure of the method hook calls.
--- a/sdk/java-sdk-spring/src/main/java/kalix/javasdk/annotations/Trigger.java
+++ b/sdk/java-sdk-spring/src/main/java/kalix/javasdk/annotations/Trigger.java
@@ -23,30 +23,24 @@ import java.lang.annotation.*;
  * The method must be public and have no parameters.
  * If the call fails, it will be retried up to the number of times specified by the maxRetries parameter.
  */
-@Target(ElementType.METHOD)
-@Retention(RetentionPolicy.RUNTIME)
-@Documented
 public @interface Trigger {
-  enum TriggerEvent {
+
+  /**
+   * The on startup hook is called every time a service instance boots up.
+   * This can happen for very different reasons: restarting / redeploying the service,
+   * scaling up to more instances or even without any user-driven action
+   * (e.g. Kalix Runtime versions being rolled out, infrastructure-related incidents, etc.).
+   * Therefore, one should carefully consider how to use this hook and its implementation.
+   */
+  @Target(ElementType.METHOD)
+  @Retention(RetentionPolicy.RUNTIME)
+  @Documented
+  @interface OnStartup {
     /**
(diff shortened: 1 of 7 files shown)


sdk/java-sdk-spring/src/main/java/kalix/javasdk/annotations/Trigger.java
```diff
@@ -0,0 +1,54 @@
...
+     * Therefore, one should carefully consider how to use this hook and its implementation.
+     */
+    STARTUP,
+  }
+
+
+  /**
+   * The service life-cycle event for which this hook will be triggered.
+   */
+  TriggerEvent on() default TriggerEvent.UNSPECIFIED;
```
efgpinto, line 47: I have also tried with this being `value()` and using `@TriggerOn`, so it would be `@TriggerOn(STARTUP)` instead of `@Trigger(on = STARTUP)` which is the current form.

I ended up with the latest but let me know your thoughts.

efgpinto:
This will require a new Runtime version. Depends on https://github.com/lightbend/kalix-runtime/pull/2253

sdk/java-sdk-spring/src/main/java/kalix/javasdk/annotations/Trigger.java
```diff
@@ -0,0 +1,54 @@
...
+ * This annotation is used to mark a method as a hook that will be triggered automatically uppon certain events.
+ * The method must be public and have no parameters.
+ * If the call fails, it will be retried up to the number of times specified by the maxRetries parameter.
+ */
+@Target(ElementType.METHOD)
+@Retention(RetentionPolicy.RUNTIME)
+@Documented
+public @interface Trigger {
+  enum TriggerEvent {
+    UNSPECIFIED,
```
aludwiko, line 31: why do we need unspecified here?

efgpinto: good point, I was just folliwing the proto specification without thinking too much.. removed now, thanks

sdk/java-sdk-spring/src/test/java/kalix/spring/testmodels/action/ActionsTestModels.java
```diff
@@ -218,4 +221,12 @@ public Flux<Effect<Message>> message(@RequestBody Flux<Message> messages) {
       return messages.map(msg -> effects().reply(msg));
     }
   }
+
+  public static class OnStartupHookAction extends Action {
+    @PostMapping("/message")
+    @Trigger(on = STARTUP, maxRetries = 2)
```
octonato, line 227: I find it better to use `on` instead of `value` as it reads better. 

On the other hand, I feel that we are making a future-proof API without having a clear plan. At the moment, we don't have any reason to think that there will be another TriggerEvent type.

But the decision we are taking is forcing the user to choose it from a list of one item. 

From the point of view of the user's API, we force them to write `\@Trigger`, then type `on = `, then look up for the enum to pass. And that will be as such for the foreseeable future. While an annotation as `\@TriggerOnStartUp(maxReqrites = 2)` would be easier for the user to write (no enum to look after) and not give the impression there is an option for event trigger. 

Unless we have a very clear use case for other TriggerEvents.

efgpinto: Although we don't have immediate plans for that, we already have some customer feedback wanting other events:
> This was previously reported and the discussion yielded the idea to support full lifecycle event triggering so not just service startup, shutdown, node failure, cluster member failure, shard rebalancing, or whatever makes sense from a Kalix point of view. In our view, modern cloud native applications need to be aware of the state of the resources they are using. It doesn’t need to be super-fine-grained however.

While I don't think all of those are reasonable, I do think there is a high probability of needing to expand this for more events.. some of the more obvious for me are "before readiness" and "on shutdown".

For that reason, I think it is better to have the user spending the extra few seconds but allow for future evolution rather than having them/us go through deprecation cycles. Or would you consider then having a different annotation for each event?

octonato: Having specialized annotations can be an option as well, but I was not thinking about that at all, because I have the feel that we will only have `OnStartUp` anyway. 

But fine. I just wanted to raise the question. Actually, I should have started this debate when we were working on the protocol, but it only occurred to me now.

efgpinto: What if it was `Trigger.OnStartup`? Would it make a difference in your view?

I think we are still on time to discuss it btw.. I mean, this is for code-first, we are not limited by what we did at the protocol-first one anyway.

octonato: Hmm, I like this because it make it quite discoverable if we add other trigger types. It suffices to type `Trigger.` and get a list of all possible event triggers. And then we can add specific params on each of them. 

> I think we are still on time to discuss it btw.. I mean, this is for code-first, we are not limited by what we did at the protocol-first one anyway.

I see your point and I've done similar in the past. For code-first we should not try to reproduce the same as Protobuf, because Java annotations can be more expressive. So we can make it more convenient. 

But on the other hand, if we believe it's better to have specialized annotations, why shouldn't we have for the Protobuf? In that particular case, there is not restriction.

octonato: Let me clarify the above with an example.

Protobuf has `eventing.in` and that receives the type name of the entity. And that's a String. 

In Java code-first, we could force some typing. That's why we have: `@Subscribe.EventSourcedEntity(MyEntity.class)`. So we could make something more much expressive and not stringly-typed. 

While here, we could have specialized annotations on both sides.

I'm fine if we don't do it in Protobuf and have `Trigger.OnStartup` for java.

efgpinto: > But on the other hand, if we believe it's better to have specialized annotations, why shouldn't we have for the Protobuf? In that particular case, there is no restriction.

yeah, sure. I agree, if it makes sense for code-first it would probably make sense for protocol-first as well.
I just think it's one of those things that the user will do once every 6 months and never touch again...

I will update this to `Trigger.OnStartup` and then we can decide what to do with protocol first.

efgpinto: Btw, the advantage of using specialized interfaces is that fact that we can define different properties depending on the event, the disadvantage being that we will probably repeat some of the fields (i.e. maxRetries). In any case, I think it's a good tradeoff.

efgpinto: Refactored in https://github.com/lightbend/kalix-jvm-sdk/pull/1924/commits/2b666cb623d09efc3694e4ca977e0c1969d21988

sdk/java-sdk-spring/src/main/scala/kalix/javasdk/impl/ActionDescriptorFactory.scala
```diff
@@ -47,12 +47,34 @@ import kalix.javasdk.impl.reflection.NameGenerator
...
+import scala.PartialFunction.condOpt
 
 private[impl] object ActionDescriptorFactory extends ComponentDescriptorFactory {
 
+  private def hasTriggerMethodOptions(javaMethod: Method): Boolean = {
+    javaMethod.isPublic && javaMethod.hasAnnotation[Trigger.OnStartup] // this is the only event available at the moment
+  }
+
+  private def triggerOptions(javaMethod: Method): Option[TriggerOptions] = {
+    condOpt(hasTriggerMethodOptions(javaMethod)) { case true =>
```
octonato, line 67: Here we can use Option.when instead.

```suggestion
    Option.when(hasTriggerMethodOptions(javaMethod)) {
```

sdk/java-sdk-spring/src/main/scala/kalix/javasdk/impl/ActionDescriptorFactory.scala
```diff
@@ -47,12 +47,34 @@ import kalix.javasdk.impl.reflection.NameGenerator
...
-import java.lang.reflect.Method
 
+import java.lang.reflect.Method
 import kalix.javasdk.impl.ComponentDescriptorFactory.mergeServiceOptions
+import kalix.javasdk.impl.Reflect.Syntax.AnnotatedElementOps
+import kalix.javasdk.impl.Reflect.Syntax.MethodOps
+import kalix.TriggerOptions
+import kalix.javasdk.annotations.Trigger
+
+import scala.PartialFunction.condOpt
```
octonato, line 58:
```suggestion
```
see below

octonato (approved):
LGTM, with a minor suggestion regarding usages of `condOpt` vs. `Option.when`.

I find it much better with Trigger.OnStartup. Easy to find and the door is open for adding other types.

PR #1926, opened 2023-12-19
feat(java-sdk): support for pattern in jwt static claim

efgpinto:
Refs https://github.com/lightbend/kalix-runtime/issues/2192

This is open against the previous PT for multiple values since it changes the same files.. will rebase once former is merged.

This also depends on Runtime release but ready for review otherwise.


octonato (approved):
LGTM

PR #1927, opened 2023-12-19
feat: pulling runtime latest to ensure SDK compatibility

franciscolopezsancho:
https://github.com/lightbend/kalix-runtime/issues/2239

Looking for some feedback.

### DIFF
--- a/devtools/src/main/scala/kalix/devtools/impl/DockerComposeUtils.scala
+++ b/devtools/src/main/scala/kalix/devtools/impl/DockerComposeUtils.scala
@@ -19,12 +19,17 @@ package kalix.devtools.impl
 import java.io.File
 import java.nio.file.Files
 import java.nio.file.Paths
-
 import scala.io.Source
 import scala.jdk.CollectionConverters._
 import scala.sys.process._
-
 import com.typesafe.config.Config
+
+import java.net.HttpURLConnection
+import java.net.URL
+import java.time.LocalTime
+import java.time.format.DateTimeFormatter
+import scala.util.Success
+import scala.util.Try
 
 object DockerComposeUtils {
 
@@ -82,7 +87,15 @@ case class DockerComposeUtils(file: String, envVar: Map[String, String]) {
   // note to self: this is seems to be similar to sbt native packager printing errors when build docker images
   def start(): Unit =
     execIfFileExists {
-      val proc = Process(s"docker-compose -f $file up", None).run(processLogger)
+      val pullFlag = dockerImagesAvailable match {
+        case Success(true) => "--pull=always"
+        case _ =>
+          val fmt = DateTimeFormatter.ofPattern("HH:mm:ss:SSS") //In line with logback-dev-mode.xml
+          println(
+            s"${LocalTime.now().format(fmt)} WARN Kalix plugin can NOT `docker pull` the `latest` kalix-runtime. Using your local version. The plugin is trying to download the latest `kalix-runtime` to ensure that you can use the latest `kalix-sdk`. Check your internet connection.")
+          ""
+      }
+      val proc = Process(s"docker-compose -f $file up $pullFlag", None).run(processLogger)
       started = proc.isAlive()
       // shutdown hook to down containers when jvm exits
       sys.addShutdownHook {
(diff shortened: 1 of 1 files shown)


devtools/src/main/scala/kalix/devtools/impl/DockerComposeUtils.scala
```diff
@@ -82,7 +87,15 @@ case class DockerComposeUtils(file: String, envVar: Map[String, String]) {
   // note to self: this is seems to be similar to sbt native packager printing errors when build docker images
   def start(): Unit =
     execIfFileExists {
-      val proc = Process(s"docker-compose -f $file up", None).run(processLogger)
+      val pullFlag = dockerImagesAvailable match {
+        case Success(true) => "--pull=always"
+        case _ =>
+          val fmt = DateTimeFormatter.ofPattern("HH:mm:ss:SSS") //In line with logback-dev-mode.xml
+          println(
+            s"${LocalTime.now().format(fmt)} WARN Kalix plugin can NOT `docker pull` the `latest` kalix-runtime. Using your local version. The plugin is trying to download the latest `kalix-runtime` to ensure that you can use the latest `kalix-sdk`. Check your internet connection.")
```
franciscolopezsancho, line 95: Happy to change this text.

octonato: I find it good to have the `--pull=always`, but I think it will only make sense if the docker-compose is pointing to a image tagged with `latest`. 

If it's pointing to a fix version image, `--pull=always` won't make a difference. 

Also, the `dockerImagesAvailable` check is unnecessary. The image must be there, if not it means that the user is using a wrong number.

octonato: Ok, I see now your comment in the ticket. Your goal is to check if the user has internet connection.

octonato: Maybe an alternative approach could be to do a `docker-compose pull`. That forces the pulling of  new `latest`. 

We can wrap it on a Future and if it fails, we log a message to the user and we recover the future. 

The future is then flatMap with the one that run `docker-compose up`.

octonato: On the other hand, I think we should first change the runtime build to not automatically tag the image with latest.

Then, we need to change the CI pipeline to tag it with `latest` once we deploy it into Prod. 

Only when all this is ready, we can change the plugin.

franciscolopezsancho: How about we tag latest as the one  we have in production in the plugin itself and we don't touch the CI? In pro of separation of concerns we could deal with all the requirements of this issue in the plugin. And leave the CI unaware of this requirements.

franciscolopezsancho: I'll sketch what I mean.

octonato: > How about we tag latest as the one we have in production in the plugin itself and we don't touch the CI? In pro of separation of concerns we could deal with all the requirements of this issue in the plugin. And leave the CI unaware of this requirements.

We can't do this because there is often a very long time between the moment the image is published and the moment it's deployed in prod. During this time, we don't want users to use the image as they might use features that are not yet in prod. If they deploy their application in while Prod is still on the previous image, they might break their application.

octonato: Sorry, I think I misunderstood you. But we can't tag in the plugin. Tagging happens in the container registry. Anyway, let's talk later. We are probably talking about two different things.

franciscolopezsancho: After talking to Renato we decided to add another tag `latest-prod` that will be created after latest is used in prod. We'll pull force that image.

One issue we should take into account is if that pull might erase the H2 database and lose all the state in Kalix. Ma

efgpinto:
This can be closed now, right?

octonato:
Yes!

octonato:
Oh, I thought it was the original issue. It was not my intention to sound enthusiastic about closing a PR without merging it. Sorry @franciscolopezsancho! 

My "Yes!" was more, "Yes, we have a solution in place now".

PR #1928, opened 2023-12-20
infra: GH workflow to update kalix-io repo

ennru:
I'm quite sure this will not work out this simple -- this user can't open a PR on the other organisation and repo.

### DIFF
--- a/.github/workflows/sample-repo-sync.yml
+++ b/.github/workflows/sample-repo-sync.yml
@@ -24,8 +24,8 @@ jobs:
         uses: actions/checkout@93ea575cb5d8a053eaa0ac8fa3b40d7e05a33cc8
         with:
           # https://github.com/kalix-io/user-registry-orchestration-saga-workflow
-          repo: kalix-io/user-registry-orchestration-saga-workflow
-          path: user-registry-orchestration-saga-workflow/
+          repo: kalix-io/user-registry-choreography-saga
+          path: user-registry-choreography-saga/
 
       - name: Copy example
         run: |-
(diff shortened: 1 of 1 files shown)


octonato:
Won't it fork and send a PR by default? By the way, which user will do it? A CI user?

octonato:
It seems that we can configure the user that will do it and then we can use the kalixbot one.

.github/workflows/sample-repo-sync.yml
```diff
@@ -0,0 +1,45 @@
...
+          sparse-checkout: |
+            samples/java-spring-choreography-saga-quickstart
+
+      - name: Checkout sample in kalix.io
+        # v3.1.0
+        uses: actions/checkout@93ea575cb5d8a053eaa0ac8fa3b40d7e05a33cc8
+        with:
+          # https://github.com/kalix-io/user-registry-orchestration-saga-workflow
+          repo: kalix-io/user-registry-orchestration-saga-workflow
+          path: user-registry-orchestration-saga-workflow/
```
octonato, line 28: this is wrong, we should target `kalix-io/user-registry-choreography-saga` instead

octonato: But I have a second thought about this approach. 

If we use such an action, we will need lots of CI configs to cover all the samples that we want to sync. 

Instead, we should have a matrix from source to target repos and a shell script that receiving the pairs and doing all the rest. Then, we only need to add/remove lines from the matrix.

ennru: Before I matrix, I go flat.

efgpinto:
This is looking good. I'm looking into having the required secret generated and added in this repo and then I'll try to run the workflow and see how it goes.

efgpinto (approved):

PR #1933, opened 2023-12-21
chore: use USER_SERVICE instead

octonato:
References https://github.com/lightbend/kalix-runtime/issues/2139

### DIFF
--- a/devtools/src/main/scala/kalix/devtools/impl/DockerComposeUtils.scala
+++ b/devtools/src/main/scala/kalix/devtools/impl/DockerComposeUtils.scala
@@ -107,7 +107,8 @@ case class DockerComposeUtils(file: String, envVar: Map[String, String]) {
 
   def userFunctionPort: Int =
     envVar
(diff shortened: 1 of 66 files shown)


devtools/src/main/scala/kalix/devtools/impl/DockerComposeUtils.scala
```diff
@@ -107,13 +107,14 @@ case class DockerComposeUtils(file: String, envVar: Map[String, String]) {
 
   def userFunctionPort: Int =
     envVar
-      .get("USER_FUNCTION_PORT")
+      .get("USER_SERVICE_PORT")
+      .orElse(envVar.get("USER_FUNCTION_PORT")) // legacy name
```
octonato, line 111: If USER_SERVICE_PORT is not an env var, we try USER_FUNCTION_PORT

Otherwise we read from file.

efgpinto (approved):

PR #1934, opened 2023-12-21
chore: move container class to impl package

octonato:
Build on top of https://github.com/lightbend/kalix-jvm-sdk/pull/1933

This is a little more complicated. We have a KalixProxyContainer class that is public, but should have been in impl package. 

It is instantiated inside the testkit and kept private. So, from that perspective it is effectively private.

Leaving it as draft until we merge #1933 and decide what to do with this class. From my part, we can simply move it to the impl package and consider it private.


efgpinto (approved):
Moving `kalixProxyContainer` to `.impl` seems ok, unlikely someone decided to use it directly I think.

LGTM

Issue #1851, opened 2023-11-13
High level API for setting _kalix-http-code


jroper:
We should expose a method on `Metadata` to set the HTTP response code when using HTTP transcoding, without having to rely on the internal `_kalix-http-code` name. ie:

```java
/**
 * Set the HTTP response code for the response.
 *
 * This will only take effect when HTTP transcoding is in use. It will be ignored for gRPC requests.
 */
public Metadata withHttpResponseCode(int code) {
  if (code < 200 || code > 599) {
    throw new RuntimeException("Invalid HTTP status code: " + code);
  }
  return set("_kalix-http-code", code.toString());
}
```

efgpinto:
We should probably move forward with this.. We have had multiple users making use of our internal header which is not ideal. Another [example](https://kalixworld.slack.com/archives/C05CBK0VCKH/p1702323429975409).

octonato:
Somewhat related to https://github.com/lightbend/kalix-jvm-sdk/issues/1639

efgpinto:
I think the strategy might be slightly different between code-first and protocol-first sdks.

For protocol-first a simple method on the Metadata as shown above could be enough, since users might not even be using HTTP, they don't need to be exposed to it by default (which would happen if we allowed to pass the status code directly in a new `reply()´ method overload or in specific methods `replyNoContent()`).

For code-first I think there will be a need for a different API, one that maybe not only allows you to define the HTTP code but maybe some associated data (e.g. when setting 301 as return code I should probably set a value for a "Location" header as well). Should this be done leveraging the existing Spring specific classes (e.g. ResponseEntity, ResponseStatus) or using internal ones that could be reused for scala code-first in the future? This is a similar discussion to what we are doing [here](https://github.com/lightbend/kalix-jvm-sdk/issues/1639).

I'm tempted to implement the simpler helper method which can be used in both protocol-first and code-first SDK as a quicker way to move users off using our internal header directly. As for the long term solution, I think we will still need to discuss further.

efgpinto:
Also discovered today that no only Spring users use the `ResponseEntity` to set the status code, but they can also shortcut it with `ResponseCode` at method, class and exception level (see [here](https://www.baeldung.com/spring-response-status)). Seemed like an annotation we could potentially support as well to please people coming from Spring background.

octonato:
[...]

Issue #1639, opened 2023-05-23
No content responses


jroper:
It's common to return no content in REST APIs. For example, in response to a delete request, often the response will be 204 NoContent. The Spring SDK should support this idiomatically. Not sure of the best way to support this, one way could be for us to provide a `NoContent` class, eg:

```java
return effects().emit(...).thenReply(state -> NoContent.getInstance());
```

or even:

```java
return effects().emit(...).thenReplyNoContent();
```

The proxy already allows setting metadata to set a custom HTTP status code, and if that status code doesn't allow content, will not send any content in the HTTP body. This is just about creating an idiomatic way to do that in the Spring SDK.

jroper:
Example of how this can be achieved today:

```java
    @DeleteMapping()
    public Effect<Asset> disposeOfAsset() {
        if (currentState().status() != Asset.AssetStatus.EXISTING) {
            return effects().error("Asset not found", StatusCode.ErrorCode.NOT_FOUND);
        }
        return effects()
                .emitEvent(new Asset.Event.AssetDisposed())
                .thenReply(Function.identity(), Metadata.EMPTY.add("_kalix-http-code", "204"));
    }
```

What I would rather is something to the effect of:

```java
    @DeleteMapping()
    public Effect<NoContent> disposeOfAsset() {
        if (currentState().status() != Asset.AssetStatus.EXISTING) {
            return effects().error("Asset not found", StatusCode.ErrorCode.NOT_FOUND);
        }
        return effects()
                .emitEvent(new Asset.Event.AssetDisposed())
                .thenReplyNoContent();
    }
```

octonato:
I like the idea. 
Would it make sense to open it up and offer something more generic? 

A call to a REST endpoint may return many things. If we introduce `NoContent`, why don't we introduce `Created` and `Accepted`? And should they share a common trait like `Response`? 

Users can still return `Foo` if they want, but they could also type it to `Effect<Response>` and return `effects().theReply(Ok(foo))`.

We can have `Response` impl covering some of the 20x status code range and keep the errors in `effects().error(...)`. As such, it won't be possible to emit events and return an error, for example.

jroper:
All good questions/suggestions.

One thing to note, we already support returning a `google.api.HttpBody` protobuf message in the proxy, which allows returning arbitrary HTTP bodies. I don't know if that's at all relevant here, if there's anything we want to use with that, but just something to be aware of in case we want it.

[...]

PR #1936, opened 2023-12-21
feat: provide http status code API

efgpinto:
Refs https://github.com/lightbend/kalix-jvm-sdk/issues/1851

This is still very early draft version. I still have some questions to revisit:
- Should this go directly into Metadata or into something like ResponseCode (similar to CloudEvents and JwtClaims)?
- It could also go directly on the `reply` method (similar to the `error` that receives an error code), but maybe this way is better
- scala API

### DIFF
--- a/sdk/java-sdk-protobuf/src/main/java/kalix/javasdk/Metadata.java
+++ b/sdk/java-sdk-protobuf/src/main/java/kalix/javasdk/Metadata.java
@@ -239,6 +239,15 @@ public interface Metadata extends Iterable<Metadata.MetadataEntry> {
    */
   Principals principals();
 
+  /**
+   * Add an HTTP response code to this metadata.
+   * This will only take effect when HTTP transcoding is in use. It will be ignored for gRPC requests.
+   *
+   * @param code The status code to add.
+   * @return a copy of this metadata with the HTTP response code set.
+   */
+  Metadata withHttpResponseCode(StatusCode code);
+
   /** A metadata entry. */
   interface MetadataEntry {
     /**
--- a/sdk/java-sdk-protobuf/src/main/java/kalix/javasdk/Metadata.java
+++ b/sdk/java-sdk-protobuf/src/main/java/kalix/javasdk/Metadata.java
@@ -243,19 +243,19 @@ public interface Metadata extends Iterable<Metadata.MetadataEntry> {
    * Add an HTTP response code to this metadata.
    * This will only take effect when HTTP transcoding is in use. It will be ignored for gRPC requests.
    *
-   * @param code The success status code to add.
+   * @param httpStatusCode The success status code to add.
    * @return a copy of this metadata with the HTTP response code set.
    */
-  Metadata withHttpResponseCode(StatusCode.Success code);
+  Metadata withStatusCode(StatusCode.Success httpStatusCode);
 
   /**
    * Add an HTTP response code to this metadata.
    * This will only take effect when HTTP transcoding is in use. It will be ignored for gRPC requests.
    *
-   * @param code The redirect status code to add.
+   * @param httpStatusCode The redirect status code to add.
    * @return a copy of this metadata with the HTTP response code set.
    */
-  Metadata withHttpResponseCode(StatusCode.Redirect code);
+  Metadata withStatusCode(StatusCode.Redirect httpStatusCode);
 
   /** A metadata entry. */
   interface MetadataEntry {
--- a/sdk/java-sdk-protobuf/src/main/java/kalix/javasdk/Metadata.java
+++ b/sdk/java-sdk-protobuf/src/main/java/kalix/javasdk/Metadata.java
@@ -243,10 +243,19 @@ public interface Metadata extends Iterable<Metadata.MetadataEntry> {
    * Add an HTTP response code to this metadata.
    * This will only take effect when HTTP transcoding is in use. It will be ignored for gRPC requests.
    *
-   * @param code The status code to add.
+   * @param code The success status code to add.
    * @return a copy of this metadata with the HTTP response code set.
    */
-  Metadata withHttpResponseCode(StatusCode code);
+  Metadata withHttpResponseCode(StatusCode.Success code);
+
+  /**
+   * Add an HTTP response code to this metadata.
+   * This will only take effect when HTTP transcoding is in use. It will be ignored for gRPC requests.
+   *
+   * @param code The redirect status code to add.
+   * @return a copy of this metadata with the HTTP response code set.
+   */
+  Metadata withHttpResponseCode(StatusCode.Redirect code);
 
   /** A metadata entry. */
   interface MetadataEntry {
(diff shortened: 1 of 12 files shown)


sdk/java-sdk-protobuf/src/main/scala/kalix/javasdk/impl/MetadataImpl.scala
```diff
@@ -194,6 +196,13 @@ private[kalix] class MetadataImpl(val entries: Seq[MetadataEntry]) extends Metad
 
   override def clearTime(): MetadataImpl = remove(MetadataImpl.CeTime)
 
+  override def withHttpResponseCode(code: StatusCode): MetadataImpl = {
+    if (code.isInstanceOf[ErrorCode])
+      throw new RuntimeException("Invalid HTTP status code: " + code + ". Use .error method for replying with errors.")
+
+    set("_kalix-http-code", code.value.toString)
+  }
```
johanandren, line 204: One thing to consider, if given direct official access to this: redirects for example requires a `Location` header where the redirect points to, but now a user could say `301 Moved permanently` without such a header resulting in an invalid HTTP response. There are also a bunch of codes that are more protocol level and should never be emitted by a user endpoint (100, 103 for example).

That might be fine (with a "if it hurts, don't do it"-mentality) but could also mean we should rather have higher level APIs when we make this more officially supported.

efgpinto: Yep, definitely good points.

I was doing some experiments today but I'm not so confident we know exactly where we want to go, mainly for code-first sdk (depending on spring classes vs having our own annotations). Wrote some more [here](https://github.com/lightbend/kalix-jvm-sdk/issues/1851#issuecomment-1875576570).

efgpinto:
After some more thought, I think this simpler approach could be the right move at the moment. Allows users to stop using the internal header, works for all sdks and does not add any external dependency (e.g. from Spring).

We can continue to discuss the higher level API in the related ticket. In the future, we can deprecate this method or just live with both options. 

@octonato let me also know your thoughts here also.

sdk/java-sdk-protobuf/src/main/scala/kalix/javasdk/impl/MetadataImpl.scala
```diff
@@ -194,6 +196,13 @@ private[kalix] class MetadataImpl(val entries: Seq[MetadataEntry]) extends Metad
 
   override def clearTime(): MetadataImpl = remove(MetadataImpl.CeTime)
 
+  override def withHttpResponseCode(code: StatusCode): MetadataImpl = {
```
octonato, line 199: We can overload this method. One receiving `Success` and one receiving `Redirect`. As such, we don't need to add a runtime exception. 

Alternatively, we can have it as `withSuccessStatusCode` and `withRedirectStatusCode`, but I'm not convinced myself with the names. 🤷 

As a side comment, why are we planning to support redirects? Feels like unusual for a service that is mainly about APIs and less about http resources, but I might be missing some context.

octonato (commented):
@efgpinto, looking good. 

We do need a low-level API for it and the Metadata is the right place to add it (for now). 

We can later introduce types likes `Response` (as discussed in https://github.com/lightbend/kalix-jvm-sdk/issues/1639) and enrich the Metadata behind the scenes.

sdk/java-sdk-protobuf/src/main/scala/kalix/javasdk/impl/MetadataImpl.scala
efgpinto: > As a side comment, why are we planning to support redirects? Feels like unusual for a service that is mainly about APIs and less about http resources, but I might be missing some context.

How else can I support my super fun KLX URL shortener? Just kidding.

In any case, since we are providing a tool for (also) building HTTP APIs, I think we need to give people that flexibility. It will be used sparsely but it has applications I think. Even for things like API deprecation cycles, it can be useful to redirect clients to newer versions.

efgpinto:
This is ready for review.

sdk/java-sdk-protobuf/src/main/java/kalix/javasdk/Metadata.java
```diff
@@ -239,6 +239,24 @@ public interface Metadata extends Iterable<Metadata.MetadataEntry> {
...
   Principals principals();
 
+  /**
+   * Add an HTTP response code to this metadata.
+   * This will only take effect when HTTP transcoding is in use. It will be ignored for gRPC requests.
+   *
+   * @param code The success status code to add.
+   * @return a copy of this metadata with the HTTP response code set.
+   */
+  Metadata withHttpResponseCode(StatusCode.Success code);
```
aludwiko, line 249: How about adding some integration tests?

efgpinto: of course, will update. tks

sdk/java-sdk-spring/src/it/java/com/example/wiring/valueentities/user/UserEntity.java
```diff
@@ -49,7 +50,8 @@ public Effect<User> getUser() {
 
   @PostMapping("/{email}/{name}")
   public Effect<String> createOrUpdateUser(@PathVariable String email, @PathVariable String name) {
-    return effects().updateState(new User(email, name)).thenReply("Ok");
+    return effects().updateState(new User(email, name)).thenReply("Ok",
+        Metadata.EMPTY.withHttpResponseCode(StatusCode.Success.CREATED));
```
aludwiko, line 54: pretty long name:) maybe `.withResponseCode` and the fact that it's HTTP can be derived from the parameter, also, similar like here:
```
      <T> Effect<T> error(String description, Status.Code grpcErrorCode);
      <T> Effect<T> error(String description, StatusCode.ErrorCode httpErrorCode);
```

efgpinto: ok, yeah.. now I'm in doubt between `.withStatusCode` or `.withResponseCode`.. wikipedia calls them [HTTP status codes](https://en.wikipedia.org/wiki/List_of_HTTP_status_codes). so, maybe `.withStatusCode`?

aludwiko: yeah `withStatusCode`

aludwiko (approved):
go with the merge after renaming methods

sdk/java-sdk-protobuf/src/main/java/kalix/javasdk/StatusCode.java
```diff
@@ -17,20 +17,81 @@
 package kalix.javasdk;
 
 /**
- * Interface used to represent a status code, typically used when replying with an error. **NOT**
- * for user extension.
+ * Interface used to represent HTTP status code. **NOT** for user extension.
```
octonato, line 20: can we sealed it to avoid user extension?

efgpinto: tried it.. but I think we are still in java 11 here, right?

octonato (approved):
LGTM

PR #1937, opened 2023-12-21
feat: allow metadata in asyncReply

### LINKED ISSUE #1935: Allow passing Metadata for asyncReply
We currently don't support passing metadata when using `asyncReply`. I don't see a reason for not to allow that.

It's as useful as in the context of a normal `reply`.

efgpinto:
Close #1935 

I'm a bit surprised we haven't heard complains about this before.. Am I missing something?

### DIFF
--- a/sdk/java-sdk-protobuf/src/main/java/kalix/javasdk/action/Action.java
+++ b/sdk/java-sdk-protobuf/src/main/java/kalix/javasdk/action/Action.java
@@ -145,7 +145,7 @@ public abstract class Action {
       <S> Effect<S> reply(S message);
 
       /**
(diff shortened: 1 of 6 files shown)


johanandren (approved):
LGTM (if we don't want test coverage for good measure)

I guess you could always use asyncEffect with a reply with metadata before to work around it.

efgpinto:
> LGTM (if we don't want test coverage for good measure)

Added a test in the java spring IT suite... felt a bit too high-level but since I could't actually find any test there about headers on the response, I think it's fine.

johanandren (approved):

Issue #1477, opened 2023-02-13
Single handler for events in code-first sdk


efgpinto:
Nice! Is single handler for events planned in some incoming changes?

_Originally posted by @aludwiko in https://github.com/lightbend/kalix-jvm-sdk/pull/1468#pullrequestreview-1291429049_

efgpinto:
@aludwiko @octonato opened this side issue for discussion. I don't have a very good view on what this would entail so feel free to expand on your thoughts.

aludwiko:
Ok, my 2 cents. Currently, we have sth like this:
```
public class ShowEntity extends EventSourcedEntity<Show> {

  @EventHandler
  public Show onEvent(ShowCreated showCreated) {
    return Show.create(showCreated);
  }

  @EventHandler
  public Show onEvent(SeatReserved seatReserved) {
    return currentState().apply(seatReserved);
  }

  @EventHandler
  public Show onEvent(SeatReservationCancelled seatReservationCancelled) {
    return currentState().apply(seatReservationCancelled);
  }
}
```
After the change I would expect sth like:
```
public class ShowEntity extends EventSourcedEntity<Show, ShowEvent> {

  @EventHandler
  public Show onEvent(ShowEvent showEvent) {
    if (showEvent instanceof ShowCreated showCreated){
      return currentState().apply(showCreated);
    }else if (showEvent instanceof SeatReserved seatReserved){
      return currentState().apply(seatReserved);
    }else if (showEvent instanceof SeatReservationCancelled seatReservationCancelled){
      return currentState().apply(seatReservationCancelled);
    }else {
      throw new IllegalStateException("missing handler");
    }
  }
}
```
Still not great because of the runtime exception, but from my perspective, it's more concise. If we embrace pattern matching in java:
```
public class ShowEntity extends EventSourcedEntity<Show, ShowEvent> {

    @EventHandler
  public Show onEvent(ShowEvent showEvent) {
    return switch (showEvent) {
      case ShowCreated showCreated -> currentState().apply(showCreated);
      case SeatReserved seatReserved -> currentState().apply(seatReserved);
      case SeatReservationCancelled seatReservationCancelled -> currentState().apply(seatReservationCancelled);
    };
  }
}
```
Now the compiler will check if we didn't forget anything, and personally, I would move this matching to the domain aggregate itself (why I think it's better is a separate discussion - more than happy to discuss this as well):
```
[...]

PR #1938, opened 2023-12-23
feat: single event handler for sealed interface

aludwiko:
My main motivation is that we want to migrate java sdk sample to Java 21, where we can fully embrace a sealed interface pattern-matching. Although is only about ES entity event handlers, we should promote this approach not only for ES but also for Views or any subscription that consumes all events because then the compiler will check if all cases are handled.

### DIFF
--- a/sdk/java-sdk-spring/src/it/java/com/example/wiring/eventsourcedentities/counter/Counter.java
+++ b/sdk/java-sdk-spring/src/it/java/com/example/wiring/eventsourcedentities/counter/Counter.java
@@ -29,4 +29,16 @@ public record Counter(Integer value) {
   public Counter onValueMultiplied(CounterEvent.ValueMultiplied evt) {
     return new Counter(this.value * evt.value());
   }
+
+  public Counter apply(CounterEvent counterEvent) {
+    if (counterEvent instanceof CounterEvent.ValueIncreased increased) {
+      return onValueIncreased(increased);
+    } else if (counterEvent instanceof CounterEvent.ValueSet set) {
+      return onValueSet(set);
+    } else if (counterEvent instanceof CounterEvent.ValueMultiplied multiplied) {
+      return onValueMultiplied(multiplied);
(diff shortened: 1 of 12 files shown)


sdk/java-sdk-spring/src/it/java/com/example/wiring/eventsourcedentities/counter/Counter.java
```diff
@@ -29,4 +29,16 @@ public Counter onValueSet(CounterEvent.ValueSet evt) {
   public Counter onValueMultiplied(CounterEvent.ValueMultiplied evt) {
     return new Counter(this.value * evt.value());
   }
+
+  public Counter apply(CounterEvent counterEvent) {
+    if (counterEvent instanceof CounterEvent.ValueIncreased increased) {
```
aludwiko, line 34: this will be replaced with pattern matching after switching to JDK 21

efgpinto (approved):
This is about supporting a single event handler but also moves the validations for ES entities to the existing Validations framework, right? Haven't touched this code in a while thus my question.. 

In any case, LGTM.

aludwiko:
Yes, I'm trying to move the validation on every possible occasion.

johanandren:
One pretty cool thing with JDK sealed is that you can access a set of all subtypes through reflection `Class#getPermittedSubclasses`, so it would be possible to let users do method-per-event and still have a check that all event types are handled (runtime though, but we are already in things-may-break-at-runtime with spring annotation style driven things) .

aludwiko:
> so it would be possible to let users do method-per-event and still have a check that all event types are handled 

It's done already :)

johanandren:
Ah, nice, I think I mis-skimmed the Counter and thought it moved event handling into a single method to get that as a compile time guarantee.

efgpinto:
Also, refs https://github.com/lightbend/kalix-jvm-sdk/issues/1477

PR #1939, opened 2024-01-02
feat: support HttpBody as a response type (code first sdk)

aludwiko:


sdk/java-sdk-spring/src/it/java/com/example/wiring/actions/echo/ActionWithPrimitives.java
```diff
@@ -35,6 +37,16 @@ public Effect<Message> stringMessageWithOptionalParams(@RequestParam Long longVa
     return effects().reply(new Message(response));
   }
 
+  @GetMapping("/text-body")
+  public Effect<HttpBody> textBody() {
+    return effects().reply(HttpBody.newBuilder().setContentType("text/plain").setData(ByteString.copyFromUtf8("test")).build());
+  }
```
johanandren, line 43: Is there no more spring-native return type/or annotation to do the same? Something like: `@GetMapping("/text-body", produces = MediaType.TEXT_PLAIN_VALUE)`

aludwiko: Yes, but I do think we are aiming to support them in 100%, that could open the Pandora box. Currently, we ignore everything except the path.

johanandren: I think supporting just a subset (maybe just text/plain and application/json for now) and failing fast on other would be nicer than pulling in protoubuf message types/APIs.

aludwiko: But then, what should be the Effect type? Always Effect<String>? How to support an empty body for application/json?

johanandren: What is an empty body that is valid in JSON? Is it an object with no fields `{}`, or an empty list `[]`, or is it a completely empty response (I don't _think_ that is valid JSON).

In Java land I'd probably expect just `effects().reply(null)` without any extra ceremony would lead to empty page, and then perhaps a user _really_ wanting empty object could create a class with no fields and rely on Jackson turning that into `{}`?

Checking how it is done in spring and doing the same probably makes most sense for JVM code first?

aludwiko: Ok, as discussed during the daily. An empty body is only one part of the story. The main driver behind HttpBody was to return any payload, for instance, a raw JSON as String. In vanilla Spring, this could be done in several ways:
```
  @GetMapping(value = "/test", produces = "application/json")
  public ResponseEntity<String> test(){
    return ResponseEntity.ok("{\"response\": \"123\"}");
  }

  @GetMapping(value = "/test2", produces = "application/json")
  @ResponseBody
  public String test2(){
    return "{\"response\": \"123\"}";
  }
```
In Kalix we could (probably) go with sth similar to the first one:
```
  @GetMapping(value = "/test", produces = "application/json")
  public Effect<String> test() {
    return effects().reply("{\"response\": \"123\"}";);
  }
```
and use `HttpBody` under the hood. 

But there is also an idea to allow any proto type in the code first SDK (cc @octonato ), which will open a door for proto based serialization of events/states.

efgpinto: > In Kalix we could (probably) go with sth similar to the first one:
> 
> ```
>   @GetMapping(value = "/test", produces = "application/json")
>   public Effect<String> test() {
>     return effects().reply("{\"response\": \"123\"}";);
>   }
> ```

This seems like the most convenient way indeed. It will also automatically solve the issue when you reply with string `"ok"` and receive `""ok""` on the other side. Since `"ok"` is valid JSON, that would be exactly the output.

octonato (commented):
Looking good to me. 

As discussed on call today, I think we should give more time to these ideas and explore the options in https://github.com/lightbend/kalix-jvm-sdk/pull/1936 and discussed in https://github.com/lightbend/kalix-jvm-sdk/issues/1639.

I feel we are moving into the good direction but we need to have all puzzle pieces on the table first and decide for a strategy.

sdk/java-sdk-spring/src/it/java/com/example/wiring/actions/echo/ActionWithPrimitives.java
octonato: If I'm not mistaking, there are annotations for `@Produces` and `@Consumes`, but we block them as we want to force it to be json anyway. We might need to revisit it.

efgpinto:
This can be closed now right?

aludwiko:
Closing, done in: https://github.com/lightbend/kalix-jvm-sdk/pull/1965

PR #1954, opened 2024-01-09
docs: describe what an effect is

octonato:
Add docs about what an Effect is.

Related to https://github.com/lightbend/kalix-docs/pull/1995

### DIFF
--- a/docs/src/modules/developing/pages/development-process-proto.adoc
+++ b/docs/src/modules/developing/pages/development-process-proto.adoc
@@ -4,6 +4,7 @@ include::ROOT:partial$include.adoc[]
 The main steps in developing a service include:
 
 . <<_specify_service_interface_and_domain_model>>
+. <<_effects_api>>
 . <<_implement_components>>
 . <<_create_unit_tests>>
 . <<_package_service>>
@@ -26,11 +27,13 @@ The gRPC definitions in `.proto` files describe the external interface, messages
 
 TIP: Kickstart a project using the xref:java-protobuf:quickstart-template.adoc[project template].
 
+include::java-protobuf:partial$effects-api.adoc[]
 
 [#_implement_components]
 == Implement components
 
 Stateful services can implement https://docs.kalix.io/reference/glossary.html#value_entity[Value Entity] or https://docs.kalix.io/reference/glossary.html#event_sourced_entity[Event Sourced Entity] and https://docs.kalix.io/reference/glossary.html#view[View] components. Stateless services implement https://docs.kalix.io/reference/glossary.html#action[Actions]. Typically, a stateful service should only have one Entity type, but can also include Actions and Views.
+
 
 === Actions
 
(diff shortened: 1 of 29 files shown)


docs/src/modules/developing/pages/development-process-proto.adoc
```diff
@@ -48,6 +51,10 @@ TIP: For more information see xref:java-protobuf:actions-as-controller.adoc[Acti
 
 Services can interact asynchronously with other services and with external systems. Event Sourced Entities emit events to a journal, to which other services can subscribe. By configuring your own publish/subscribe (pub/sub) mechanism, any service can publish their own events and subscribe to events published by other services or external systems.
 
+=== Views
+
+A View provides a way to retrieve state from multiple Entities based on a query. You can query non-key data items. You can create views from Value Entity state, Event Sourced Entity events, and by subscribing to topics. For more information about writing views see xref:java:views.adoc[Implementing Views].
```
aludwiko, line 56: `You can query non-key data items.` - I would remove this sentence.

docs/src/modules/java-protobuf/partials/effects-api.adoc
```diff
@@ -0,0 +1,11 @@
+
+[#_effects_api]
+== Understanding what an Effect is
+
+Each component defines a set of operations through its application programming interface (API). These operations are specific to the semantics of each component. For the JVM SDKs, these APIs take the shape of an Effect.
+
+An Effect is a description of what Kalix needs to do after an incoming command is handled. You can think of it as a set of instructions you are passing to Kalix. Kalix will process the instructions on your behalf.
```
aludwiko, line 7:
```suggestion
An Effect is a description of what Kalix needs to do to handle an incoming command. You can think of it as a set of instructions you are passing to Kalix. Kalix will process the instructions on your behalf.
```

aludwiko: after command is handler, sounds strange, the effect is the instruction how the command should be handled

aludwiko (approved):
LGTM, with some comments

docs/src/modules/java-protobuf/partials/effects-api.adoc
octonato: I see your point, but it might be confusing. 

The command handler is the user defined method that can do many things, validation, call other components in case of Action, etc. It does handle the command and run business logic. 

The result of this handling is passed to Kalix for further processing. In a sense, the handling of a command is a shared responsibility where part of it is defined and executed by the user's code and part of it is handled by Kalix.

The choice of using the phrasing "after an incoming command is handled" is intentional because it emphasizes that after the user code is executed, something else will be executed on their behalf.

docs/src/modules/developing/pages/development-process-proto.adoc
octonato: and there was a mistake on this text fragment, the link was pointing to the java page for Implementing Views, not the protobuf one. Fixed now.

PR #1963, opened 2024-01-10
chore: improve logging about outdate runtime image

octonato:
References https://github.com/lightbend/kalix-runtime/issues/2239

The strategy here is to disable the warning we get from the runtime and check the current versions in the docker-compose file against the version used to build the SDK.


devtools/src/test/scala/kalix/devtools/impl/RuntimeVersionCheckerSpec.scala
```diff
@@ -0,0 +1,179 @@
...
+          |
+          |  kalix-runtime:
+          |    image: gcr.io/kalix-public/kalix-runtime:1.1.27
+          |    ports:
+          |      - "9000:9000"
+          |
+          |  kalix-runtime:
+          |    image: gcr.io/kalix-public/kalix-runtime:1.1.26
+          |    ports:
+          |      - "9000:9000"
```
octonato, line 153: Here I'm taking a shortcut. 

When running multi-services tests, a user may have more than one image in the docker-compose file. We can't know which image will be serving the current application (running locally) and which images will be serving the other services (running in containers). 

So the shortcut is to collect all old versions and recommend to be updated them to the version used to build the current SDK. 

If can be that a service running in a container built with 1.1.28, gets a runtime updated to 1.1.29. The user can choose to ignore the warning or only update it the image for the local run. 

Anyway, this is such a border case that I don't think it's worth to invest more time to make it 100% perfect. Also because the multi-service testing will probably be superseded soonish by `kalix local run`.

sdk/java-sdk-protobuf/src/main/scala/kalix/javasdk/KalixRunner.scala
```diff
@@ -225,6 +223,16 @@ final class KalixRunner private[javasdk] (
...
     // start containers if application (only possible when running locally)
     dockerComposeUtils.foreach { dcu =>
+      // check out runtime container version before starting it
+      val outdatedVersions = dcu.collectOutdatedVersions(BuildInfo.runtimeVersion)
+      if (outdatedVersions.nonEmpty) {
+        log.warn(
+          "The docker-compose file '{}' contains outdated runtime image(s): ({}). It's recommended to update your image(s) to '{}'",
+          dcu.file,
+          outdatedVersions.mkString(","),
+          s"${BuildInfo.runtimeImage}:${BuildInfo.runtimeVersion}")
```
octonato, line 233: Here we could check and recommend the version running in prod, but for the moment I'm limiting it to the version used to build the SDK.

efgpinto:
I actually thought we could do the check against the information coming from the runtime discovery call but maybe checking against the docker compose works too.

octonato:
> I actually thought we could do the check against the information coming from the runtime discovery call but maybe checking against the docker compose works too.

Hmm, I will check it. I think I'm adding too much code. If we have this info already digested we should use it. And we will probably solve the issue related with multi-service tests. Each running service checks against its associated container.

Less code and more robust solution.

octonato:
@efgpinto, I simplified this PR (a lot) by following your suggestion. ProxyInfo contains the version of the running runtime and that's all we need. Thanks!

sdk/java-sdk-protobuf/src/main/scala/kalix/javasdk/impl/DiscoveryImpl.scala
```diff
@@ -87,6 +87,13 @@ class DiscoveryImpl(
       in.protocolMajorVersion,
       in.protocolMinorVersion)
 
+    if (in.devMode && BuildInfo.runtimeVersion.compareTo(in.proxyVersion) > 0) {
+      log.warn(
+        "Your service is using an outdated runtime image (version: {}). It's recommended to update your image to '{}' in your docker-compose.yml",
+        in.proxyVersion,
+        s"${BuildInfo.runtimeImage}:${BuildInfo.runtimeVersion}")
+    }
```
octonato, line 95: This is the main change, the other changes in this file are more about renaming 'proxy' to 'runtime'.

sdk/java-sdk-protobuf/src/main/scala/kalix/javasdk/impl/DiscoveryImpl.scala
```diff
@@ -87,6 +87,13 @@ class DiscoveryImpl(
       in.protocolMajorVersion,
       in.protocolMinorVersion)
 
+    if (in.devMode && BuildInfo.runtimeVersion.compareTo(in.proxyVersion) > 0) {
```
octonato, line 90: Strictly speaking the version will only be behind in dev-mode and when running using the docker-compose file. But checking the `in.devMode` for extra protection.

efgpinto (approved):
Nice, LGTM!

PR #1965, opened 2024-01-11
feat: http response as endpoint return type

aludwiko:
References
-

### DIFF
--- a/sdk/java-sdk-protobuf/src/main/java/kalix/javasdk/HttpResponse.java
+++ b/sdk/java-sdk-protobuf/src/main/java/kalix/javasdk/HttpResponse.java
@@ -1,0 +1,139 @@
+/*
+ * Copyright 2024 Lightbend Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package kalix.javasdk;
+
+
+import com.fasterxml.jackson.core.JsonProcessingException;
+
+import java.io.IOException;
+
+import static java.nio.charset.StandardCharsets.UTF_8;
+
+/**
+ * Represents an HTTP response with more level control over the status code, content type and body.
+ */
+public class HttpResponse {
+
+  public static final String STATUS_CODE_EXTENSION_TYPE_URL = "Status-Code";
+
+  private final StatusCode.Success statusCode;
+  private final String contentType;
+  private final byte[] body;
+
+  private HttpResponse(StatusCode.Success statusCode, String contentType, byte[] body) {
+    if (statusCode == null) throw new IllegalArgumentException("statusCode must not be null");
+    if (contentType == null) throw new IllegalArgumentException("contentType must not be null");
+    if (body == null) throw new IllegalArgumentException("body must not be null");
+    this.statusCode = statusCode;
+    this.contentType = contentType;
+    this.body = body;
+  }
+
+  private HttpResponse(StatusCode.Success statusCode) {
+    this(statusCode, "application/octet-stream", new byte[0]);
+  }
+
+  public StatusCode.Success getStatusCode() {
+    return statusCode;
+  }
+
+  public String getContentType() {
+    return contentType;
+  }
+
+  public byte[] getBody() {
+    return body;
+  }
+
+  /**
+   * Creates a 200 OK response.
+   */
+  public static HttpResponse ok() {
+    return new HttpResponse(StatusCode.Success.OK);
+  }
+
+  /**
+   * Creates a 201 CREATED response.
+   */
+  public static HttpResponse created() {
+    return new HttpResponse(StatusCode.Success.CREATED);
+  }
+
+  /**
+   * Creates a 202 ACCEPTED response.
+   */
+  public static HttpResponse accepted() {
+    return new HttpResponse(StatusCode.Success.ACCEPTED);
+  }
+
+  /**
+   * Creates a 204 NO CONTENT response.
+   */
+  public static HttpResponse noContent() {
+    return new HttpResponse(StatusCode.Success.NO_CONTENT);
+  }
+
+  /**
+   * Creates a 200 OK response with a text/plain body.
+   */
+  public static HttpResponse ok(String text) {
+    if (text == null) throw new IllegalArgumentException("text must not be null");
+    return new HttpResponse(StatusCode.Success.OK, "text/plain", text.getBytes(UTF_8));
+  }
+
+  /**
+   * Creates a 200 OK response with a application/json body. Object is encoded using Jackson serializer.
+   */
+  public static HttpResponse ok(Object object) {
+    if (object == null) throw new IllegalArgumentException("object must not be null");
+    try {
+      byte[] body = JsonSupport.encodeToBytes(object).toByteArray();
+      return new HttpResponse(StatusCode.Success.OK, "application/json", body);
+    } catch (JsonProcessingException e) {
+      throw new RuntimeException(e);
+    }
+  }
+
+  /**
+   * Creates a 200 OK response with a application/octet-stream body.
+   */
+  public static HttpResponse ok(byte[] body) {
+    return new HttpResponse(StatusCode.Success.OK, "application/octet-stream", body);
+  }
+
+  /**
+   * Creates an HTTP response with specified status code, content type and body.
+   *
+   * @param statusCode  HTTP status code
+   * @param contentType HTTP content type
+   * @param body        HTTP body
+   */
+  public static HttpResponse of(StatusCode.Success statusCode, String contentType, byte[] body) {
+    return new HttpResponse(statusCode, contentType, body);
+  }
+
+  /**
+   * Parses an HTTP body to a specified JSON type using Jackson deserializer.
+   */
+  public <T> T bodyAsJson(Class<T> clazz) {
+    try {
+      return JsonSupport.parseBytes(body, clazz);
+    } catch (IOException e) {
+      throw new RuntimeException(e);
+    }
+  }
+}
--- a/sdk/java-sdk-protobuf/src/main/java/kalix/javasdk/JsonSupport.java
+++ b/sdk/java-sdk-protobuf/src/main/java/kalix/javasdk/JsonSupport.java
@@ -119,15 +119,18 @@ public final class JsonSupport {
    */
   public static <T> Any encodeJson(T value, String jsonType) {
     try {
-      ByteString bytes =
-          UnsafeByteOperations.unsafeWrap(
-              objectMapper.writerFor(value.getClass()).writeValueAsBytes(value));
+      ByteString bytes = encodeToBytes(value);
       ByteString encodedBytes = ByteStringEncoding.encodePrimitiveBytes(bytes);
       return Any.newBuilder().setTypeUrl(KALIX_JSON + jsonType).setValue(encodedBytes).build();
     } catch (JsonProcessingException ex) {
       throw new IllegalArgumentException(
           "Could not encode [" + value.getClass().getName() + "] as JSON", ex);
     }
+  }
+
+  public static <T> ByteString encodeToBytes(T value) throws JsonProcessingException {
+    return UnsafeByteOperations.unsafeWrap(
+        objectMapper.writerFor(value.getClass()).writeValueAsBytes(value));
   }
 
   /**
@@ -162,7 +165,7 @@ public final class JsonSupport {
           if (fromVersion < currentVersion) {
             return migrate(valueClass, decodedBytes, fromVersion, migration);
           } else if (fromVersion == currentVersion) {
-            return objectMapper.readValue(decodedBytes.toByteArray(), valueClass);
+            return parseBytes(decodedBytes.toByteArray(), valueClass);
           } else if (fromVersion <= supportedForwardVersion) {
             return migrate(valueClass, decodedBytes, fromVersion, migration);
           } else {
@@ -170,7 +173,7 @@ public final class JsonSupport {
                 "behind version " + fromVersion + " of deserialized type [" + valueClass.getName() + "]");
           }
         } else {
-          return objectMapper.readValue(decodedBytes.toByteArray(), valueClass);
+          return parseBytes(decodedBytes.toByteArray(), valueClass);
         }
       } catch (JsonProcessingException e) {
         throw jsonProcessingException(valueClass, any, e);
@@ -179,6 +182,10 @@ public final class JsonSupport {
         throw genericDecodeException(valueClass, any, e);
       }
     }
+  }
+
+  public static <T> T parseBytes(byte[] bytes, Class<T> valueClass) throws IOException {
+    return objectMapper.readValue(bytes, valueClass);
   }
 
   private static <T> IllegalArgumentException jsonProcessingException(Class<T> valueClass, Any any, JsonProcessingException e) {
--- a/sdk/java-sdk-protobuf/src/main/java/kalix/javasdk/StatusCode.java
+++ b/sdk/java-sdk-protobuf/src/main/java/kalix/javasdk/StatusCode.java
@@ -46,6 +46,15 @@ public interface StatusCode {
     public int value() {
       return value;
     }
+
+    public static Success from(int statusCode) {
+      for (Success s : Success.values()) {
+        if (s.value() == statusCode) {
+          return s;
+        }
+      }
+      throw new IllegalArgumentException("No matching status for [" + statusCode + "]");
+    }
   }
 
   enum Redirect implements StatusCode {
--- a/sdk/java-sdk-protobuf/src/main/scala/kalix/javasdk/impl/workflow/WorkflowRouter.scala
+++ b/sdk/java-sdk-protobuf/src/main/scala/kalix/javasdk/impl/workflow/WorkflowRouter.scala
@@ -16,12 +16,14 @@
 
 package kalix.javasdk.impl.workflow
 
+import java.nio.ByteBuffer
 import java.util.Optional
 import java.util.concurrent.CompletionStage
 import java.util.function.{ Function => JFunc }
 
 import scala.compat.java8.FutureConverters.CompletionStageOps
 import scala.concurrent.{ ExecutionContext, Future }
+import scala.jdk.CollectionConverters._
 import scala.jdk.OptionConverters.RichOptional
 
 import com.google.protobuf.any.{ Any => ScalaPbAny }
@@ -41,7 +43,11 @@ import kalix.javasdk.workflow.Workflow.Effect
 import Workflow.AsyncCallStep
 import Workflow.CallStep
 import Workflow.WorkflowDef
+import com.google.api.HttpBody
+import kalix.javasdk.HttpResponse
+import kalix.javasdk.HttpResponse.STATUS_CODE_EXTENSION_TYPE_URL
 import kalix.javasdk.JsonSupport
+import kalix.javasdk.StatusCode
 import kalix.javasdk.impl.WorkflowExceptions.WorkflowException
 import kalix.protocol.workflow_entity.StepDeferredCall
 import kalix.protocol.workflow_entity.StepExecuted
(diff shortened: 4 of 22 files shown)


sdk/java-sdk-protobuf/src/main/java/kalix/javasdk/HttpResponse.java
```diff
@@ -0,0 +1,73 @@
...
+
+import javax.annotation.Nullable;
+import java.util.Optional;
+
+import static java.nio.charset.StandardCharsets.UTF_8;
+
+public class HttpResponse {
+  private final int statusCode;
+  private final Optional<String> contentType;
+  private final byte[] body;
```
aludwiko, line 28: @octonato I know that you suggested implementing the body as `Object`, but I'm not sure if this is a good idea, because of the deserialization. What should be under the body in that case? With an array of bytes, we are crystal clear, put your bytes here, and you will get them back. It's on you to ser/der them correctly. If you want auto-serarialization, then you are simply using low-level abstraction, go with Kalix default behavior.

sdk/java-sdk-protobuf/src/main/java/kalix/javasdk/HttpResponse.java
```diff
@@ -0,0 +1,73 @@
...
+
+  //TODO can't make it private because we use it internally for deserialization
+  public HttpResponse(int statusCode, Optional<String> contentType, byte[] body) {
+    //status code validation
+    assert StatusCode.Success.from(statusCode) != null;
+    assert body != null;
```
aludwiko, line 34: not sure we the empty body should be null or empty array?

sdk/java-sdk-protobuf/src/main/java/kalix/javasdk/HttpResponse.java
```diff
@@ -0,0 +1,73 @@
...
+
+import static java.nio.charset.StandardCharsets.UTF_8;
+
+public class HttpResponse {
+  private final int statusCode;
+  private final Optional<String> contentType;
+  private final byte[] body;
+
+  //TODO can't make it private because we use it internally for deserialization
+  public HttpResponse(int statusCode, Optional<String> contentType, byte[] body) {
```
aludwiko, line 31: I think it's safe to expose this constructor with a basic validation like the one below.

sdk/java-sdk-protobuf/src/main/java/kalix/javasdk/HttpResponse.java
```diff
@@ -0,0 +1,73 @@
...
+
+package kalix.javasdk;
+
+
+import javax.annotation.Nullable;
+import java.util.Optional;
+
+import static java.nio.charset.StandardCharsets.UTF_8;
+
+public class HttpResponse {
```
aludwiko, line 25: Should that be only the interface with a separate class with impl?

sdk/java-sdk-protobuf/src/main/java/kalix/javasdk/HttpResponse.java
```diff
@@ -0,0 +1,73 @@
...
+    this.statusCode = statusCode;
+    this.contentType = contentType;
+    this.body = body;
+  }
+
+  public static HttpResponse ok() {
+    return new HttpResponse(200, Optional.empty(), new byte[0]);
+  }
+
+  public static HttpResponse ok(byte[] body, String contentType) {
```
aludwiko, line 45: It could be replaced with a builder, `ok(body).contentType(...)`, but first we must agree on more important aspects.

sdk/java-sdk-protobuf/src/main/java/kalix/javasdk/HttpResponse.java
```diff
@@ -0,0 +1,59 @@
...
+
+import javax.annotation.Nullable;
+
+import static java.nio.charset.StandardCharsets.UTF_8;
+
+public interface HttpResponse {
+  StatusCode.Success getStatusCode();
+
+  @Nullable
+  String getContentType();
```
aludwiko, line 32: I removed the Optional<String> because I don't see a big advantage of it. Also, it will be easier to use this interface in the Scala code someday.

sdk/java-sdk-protobuf/src/main/java/kalix/javasdk/HttpResponse.java
aludwiko: Ok, check now.

sdk/java-sdk-protobuf/src/main/scala/kalix/javasdk/impl/HttpResponseImpl.scala
```diff
@@ -0,0 +1,51 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package kalix.javasdk.impl
+
+import kalix.javasdk.HttpResponse
+import kalix.javasdk.StatusCode
+
+private[kalix] class NoContent(statusCode: StatusCode.Success) extends HttpResponse {
```
aludwiko, line 22: Actually this type can be removed at this point.

sdk/java-sdk-spring/src/it/java/com/example/wiring/actions/echo/ActionWithHttpResponse.java
```diff
@@ -0,0 +1,46 @@
...
+  }
+
+  @GetMapping("/empty-text-body")
+  public Effect<HttpResponse> emptyCreated() {
+    return effects().reply(HttpResponse.created());
+  }
+
+  @GetMapping("/json-string-body")
+  public Effect<HttpResponse> jsonStringBody() {
+    return effects().reply(HttpResponse.bytes(StatusCode.Success.OK, "{\"text\": \"123\"}".getBytes(), "application/json"));
```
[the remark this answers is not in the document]
aludwiko: Yeah, I know, but I'm not sure how to block in an elegant way. Even if I remove this method and create:
```
static HttpResponse json(StatusCode.Success statusCode, String json) {
    return new JsonContent(statusCode, json);
  }
```
Users can use the `ByteContent` constructor directly and mess up the response. 

Having an enum with supported content types also seems problematic because there are so many of them.

sdk/java-sdk-protobuf/src/main/java/kalix/javasdk/HttpResponse.java
octonato: Thanks. I like the way you make it now.

sdk/java-sdk-protobuf/src/main/scala/kalix/javasdk/impl/HttpResponseImpl.scala
octonato: We should keep it, but have the status code fixed to 204

octonato: Maybe we should suffix all this types with `Response`, eg: `NoContentResponse`, `JsonContentResponse`. WDYT?

sdk/java-sdk-protobuf/src/main/java/kalix/javasdk/HttpResponse.java
```diff
@@ -0,0 +1,59 @@
...
+
+  byte[] getBody();
```
octonato, line 34: We can have a convenient method to get the body deserialized to a given type. Only for usage with Json payloads. 

Something like: 

```
public <T> T bodyAs(Class<T> clz) {...}
```

I have played a bit with an integration test returning HttpResponse and it's quite awkward because we need to take the byte array, go look for a ObjectMapper somewhere (I know where to find one, but users have no clue) and then deserialize the payload. 

We have already static methods to access our internal ObjectMapper so we don't really need to depend on an instance of JsonSupport.

Users can always decide to call `getBytes` and deserialize it the way they want if they are not interested in using our serializers.

aludwiko: Ok, check now. There is a direct dependency on JsonSupport, but I don't see any difference with JsonSuport vs ObjectMapper dependency.

sdk/java-sdk-protobuf/src/main/java/kalix/javasdk/HttpResponse.java
```diff
@@ -0,0 +1,59 @@
...
+
+  static HttpResponse ok() {
+    return new NoContent(StatusCode.Success.OK);
+  }
```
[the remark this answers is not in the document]
aludwiko: About the first comment, check now.

About the second. We can add the support later, not a big issue with the HttpBody.extension.

sdk/java-sdk-protobuf/src/main/java/kalix/javasdk/JsonSupport.java
```diff
@@ -39,6 +39,7 @@
 import com.google.protobuf.UnsafeByteOperations;
 import kalix.javasdk.annotations.Migration;
 import kalix.javasdk.impl.ByteStringEncoding;
+import org.jetbrains.annotations.NotNull;
```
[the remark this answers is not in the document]
aludwiko: removed, ale content type is once again optional.

sdk/java-sdk-protobuf/src/main/java/kalix/javasdk/StatusCode.java
```diff
@@ -46,6 +46,15 @@ enum Success implements StatusCode {
...
     }
+
+    public static Success from(int statusCode) {
+      for (Success s : Success.values()) {
+        if (s.value() == statusCode) {
+          return s;
+        }
+      }
+      throw new IllegalArgumentException("No matching status for [" + statusCode + "]");
+    }
```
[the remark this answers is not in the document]
aludwiko: Yes, check `kalix.javasdk.impl.workflow.WorkflowRouter#decodeInput` or `kalix.spring.impl.RestKalixClientImpl#toHttpResponse`.

sdk/java-sdk-protobuf/src/main/scala/kalix/javasdk/impl/HttpResponseImpl.scala
aludwiko: done.

sdk/java-sdk-protobuf/src/main/java/kalix/javasdk/JsonSupport.java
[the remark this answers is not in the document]
octonato: Using optional will be annoying for Scala and for users as well. 

Since the API is controlling the payload already, we can always set the content-type behind the scene. 

* when user passes `String` it will be `text/plain`
* when user passes `Object` it will be `application/json`
* when user passes `bytes[]` it will be `application/octet-stream`

The `NoContent` one can default to `application/octet-stream` which is not wrong since we are sending `byte[0]`.

sdk/java-sdk-protobuf/src/main/java/kalix/javasdk/HttpResponse.java
```diff
@@ -0,0 +1,165 @@
...
+    }
+
+    @Override
+    public byte[] getBody() {
+      return body;
+    }
+
+    public <T> T bodyAs(Class<T> clz) throws IOException {
+      return JsonSupport.parseBytes(body, clz);
+    }
```
octonato, line 123: I realise now that this method can only be useful if added to the HttpResponse. 

When we receive a response from the ComponentClient, we will take the internal http response from from the WebClient and map it to HttpResponse, no? 

Actually, it seems to me that the only public type should be `HttpResponse`. 

So, in that case, it must communicate in its name that it is only applicable with payload is json. Something like `jsonBodyAs(Class<T> clz)`

sdk/java-sdk-protobuf/src/main/java/kalix/javasdk/HttpResponse.java
```diff
@@ -0,0 +1,165 @@
...
+
+  static HttpResponse ok() {
+    return new NoContentResponse(StatusCode.Success.OK);
+  }
+
+  static HttpResponse created() {
+    return new NoContentResponse(StatusCode.Success.CREATED);
+  }
+
+  static HttpResponse empty(StatusCode.Success statusCode) {
```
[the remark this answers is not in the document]
aludwiko: done

sdk/java-sdk-protobuf/src/main/java/kalix/javasdk/HttpResponse.java
aludwiko: Having `HttpResponse` as the only public class would introduce another problems, for instance we allow to change the content-type only in case of `ByteContentResponse`. If we move this method to the interface, then other implementations will throw an exception. But I agree that in fact you will not be able to use this method, because I'm parsing the response always to `ByteContentReponse`.

sdk/java-sdk-protobuf/src/main/java/kalix/javasdk/JsonSupport.java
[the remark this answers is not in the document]
aludwiko: ok

sdk/java-sdk-protobuf/src/main/java/kalix/javasdk/HttpResponse.java
```diff
@@ -0,0 +1,165 @@
...
+    return new NoContentResponse(statusCode);
+  }
+
+  static ByteContentResponse text(String text) {
```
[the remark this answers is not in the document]
aludwiko: ok

sdk/java-sdk-protobuf/src/main/java/kalix/javasdk/HttpResponse.java
```diff
@@ -0,0 +1,165 @@
...
+    return new ByteContentResponse(StatusCode.Success.OK, "text/plain", text.getBytes(UTF_8));
+  }
+
+  static JsonContentResponse json(Object object) {
```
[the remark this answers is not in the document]
aludwiko: ok

sdk/java-sdk-protobuf/src/main/java/kalix/javasdk/HttpResponse.java
```diff
@@ -0,0 +1,165 @@
...
+    return new JsonContentResponse(StatusCode.Success.OK, object);
+  }
+
+  static JsonContentResponse json(StatusCode.Success statusCode, Object object) {
```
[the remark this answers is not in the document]
aludwiko: ok

sdk/java-sdk-protobuf/src/main/java/kalix/javasdk/HttpResponse.java
```diff
@@ -0,0 +1,165 @@
...
+    return new JsonContentResponse(statusCode, object);
+  }
+
+  static ByteContentResponse bytes(byte[] body) {
```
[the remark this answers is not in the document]
aludwiko: ok

sdk/java-sdk-protobuf/src/main/java/kalix/javasdk/HttpResponse.java
aludwiko: Maybe it's time to move one step back and remove all the specific implementations and have only a single class with factory methods? and then we could have `jsonBodyAs(Class<T> clz)`?

octonato: Yeah, I was having the same thought. We can keep the specific methods, but one could always change the content-type and status code later. Not a good idea, but it will make things much simpler. 

So, for example, if I really want to mess-up, I can do `HttpResponse.ok(Foo)`. That will return a 200 with json. Then `HttpResponse.ok(Foo).withContentType(whatever).withStatusCode(NotFound)`. 

I would be messing up, but then we can argue that the user chose to do so. 🤷

octonato: An alternative could be that we only offer a method to change the content-type, but not the status code nor the body. This is determined by the helper method `ok(T)`, `ok(byte[])`, `noContent()`, `notFound()`, etc.

aludwiko: I would have more faith in our users. This is low-level api, they shouldn't use it for normal cases, we just want to support very specific cases.

aludwiko: You can check the implementation now.

sdk/java-sdk-protobuf/src/main/java/kalix/javasdk/HttpResponse.java
```diff
@@ -0,0 +1,92 @@
...
+    return new HttpResponse(StatusCode.Success.CREATED);
+  }
+
+  public static HttpResponse noContent(StatusCode.Success statusCode) {
+    return new HttpResponse(statusCode);
+  }
+
+  public static HttpResponse ok(String text) {
+    if (text == null) throw new IllegalArgumentException("text must not be null");
+    return new HttpResponse(StatusCode.Success.OK, "text/plain", text.getBytes(UTF_8));
```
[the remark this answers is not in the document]
aludwiko: Not sure, null might indicate an implementation problem that an empty array will hide.

sdk/java-sdk-protobuf/src/main/java/kalix/javasdk/HttpResponse.java
```diff
@@ -0,0 +1,92 @@
...
+  }
+
+  public static HttpResponse ok(Object object) {
+    if (object == null) throw new IllegalArgumentException("object must not be null");
```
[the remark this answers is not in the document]
aludwiko: I would answer the same as above.

sdk/java-sdk-protobuf/src/main/java/kalix/javasdk/HttpResponse.java
```diff
@@ -0,0 +1,92 @@
...
+  }
+
+  public static HttpResponse ok(byte[] body) {
+    if (body == null) throw new IllegalArgumentException("body must not be null");
+    return new HttpResponse(StatusCode.Success.OK, "application/octet-stream", body);
+  }
+
+  public HttpResponse withStatusCode(StatusCode.Success statusCode) {
+    return new HttpResponse(statusCode, contentType, body);
+  }
```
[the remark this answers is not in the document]
aludwiko: ok, it was like that at the beginning, I will revert it

sdk/java-sdk-protobuf/src/main/java/kalix/javasdk/HttpResponse.java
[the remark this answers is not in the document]
octonato: I agree for byte[], but not for Object. Json spec accept defines `null` as a valid value (or should I say non-value?). 

Although we might find it a bad idea, it might be something that we should allow.

sdk/java-sdk-protobuf/src/main/scala/kalix/javasdk/impl/workflow/WorkflowRouter.scala
```diff
@@ -124,13 +129,30 @@ abstract class WorkflowRouter[S, W <: Workflow[S]](protected val workflow: W) {
...
   // in same cases, the Proxy may send a message with typeUrl set to object.
   // if that's the case, we need to patch the message using the typeUrl from the expected input class
-  private def decodeInput(messageCodec: MessageCodec, result: ScalaPbAny, expectedInputClass: Class[_]) =
+  private def decodeInput(messageCodec: MessageCodec, result: ScalaPbAny, expectedInputClass: Class[_]) = {
     if (result.typeUrl == JsonSupport.KALIX_JSON + "object") {
       val typeUrl = messageCodec.typeUrlFor(expectedInputClass)
       messageCodec.decodeMessage(result.copy(typeUrl = typeUrl))
+    } else if (result.typeUrl == "type.googleapis.com/google.api.HttpBody") {
+      val httpBody = HttpBody.parseFrom(result.value.newCodedInput())
+      httpBody.getExtensionsList.asScala.find(_.getTypeUrl == "Status-Code") match {
```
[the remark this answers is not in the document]
octonato: I think we have a place in the proto sdk where we have some other constants defined.

sdk/java-sdk-protobuf/src/main/java/kalix/javasdk/JsonSupport.java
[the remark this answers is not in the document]
octonato: Jetbrains dependency still being imported here.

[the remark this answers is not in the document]
aludwiko: removed

sdk/java-sdk-protobuf/src/main/java/kalix/javasdk/HttpResponse.java
```diff
@@ -0,0 +1,125 @@
...
+public class HttpResponse {
+
+  public static final String STATUS_CODE_EXTENSION_TYPE_URL = "Status-Code";
+
+  private final StatusCode.Success statusCode;
+  private final String contentType;
+  private final byte[] body;
+
+  private HttpResponse(StatusCode.Success statusCode, String contentType, byte[] body) {
+    if (contentType == null) throw new IllegalArgumentException("contentType must not be null");
```
octonato, line 36:
```suggestion
    if (statusCode == null) throw new IllegalArgumentException("statusCode must not be null");
    if (contentType == null) throw new IllegalArgumentException("contentType must not be null");
```

sdk/java-sdk-protobuf/src/main/java/kalix/javasdk/HttpResponse.java
octonato, line 1: I think you removed by accident the `jsonBodyAs` method, no?

aludwiko: Yes, it's back now

octonato (approved):
Let's ship it!

Thanks for the patience with all the change requests.

Issue #1907, opened 2023-12-12
docs links should be different depending on the SDK


aludwiko:
In the `DocLinks` class, we map error codes to documentation links. We should also consider the SDK type and choose links for code-first/schema-first documentation accordingly.

Issue #1313, opened 2022-12-06
Outdated links for errors


efgpinto:
Some of the links in https://github.com/lightbend/kalix-jvm-sdk/blob/283d07b7c00a411d8e9f286fe30dbcbd4d42d801/sdk/java-sdk/src/main/scala/kalix/javasdk/impl/DocLinks.scala are outdated and lead to 404. e.g. https://docs.kalix.io/java/topic-eventing.html

PR #1966, opened 2024-01-11
chore: update docs links on errors

efgpinto:
Close https://github.com/lightbend/kalix-jvm-sdk/issues/1907
Close https://github.com/lightbend/kalix-jvm-sdk/issues/1313

This PR:
- updates lists for error code and categories with updated doc links
- introduces different URLs depending on which sdk is in use

### DIFF
--- a/sdk/java-sdk-protobuf/src/main/scala/kalix/javasdk/impl/DocLinks.scala
+++ b/sdk/java-sdk-protobuf/src/main/scala/kalix/javasdk/impl/DocLinks.scala
@@ -16,34 +16,37 @@
 
 package kalix.javasdk.impl
 
-object DocLinks {
-
+case class DocLinks(sdkName: String) {
   private val baseUrl = "https://docs.kalix.io"
 
-  private val errorCodes = Map(
-    "KLX-00112" -> s"$baseUrl/java/views.html#changing",
-    "KLX-00402" -> s"$baseUrl/java/topic-eventing.html",
-    "KLX-00406" -> s"$baseUrl/java/topic-eventing.html",
-    "KLX-00414" -> s"$baseUrl/java/entity-eventing.html"
-    // TODO: docs for value entity eventing (https://github.com/lightbend/kalix-jvm-sdk/issues/121)
-    // "KLX-00415" -> s"$baseUrl/java/entity-eventing.html"
+  // sdkName is of format e.g. kalix-java-sdk-protobuf
+  private val sdkPath = if (sdkName.endsWith("-protobuf")) "java-protobuf" else "java"
+
+  val errorCodes = Map(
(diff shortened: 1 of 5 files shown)


sdk/java-sdk-protobuf/src/main/scala/kalix/javasdk/impl/DocLinks.scala
```diff
@@ -16,34 +16,37 @@
...
+    "KLX-003" -> "event-sourced-entities.html",
+    "KLX-004" -> "publishing-subscribing.html",
+    "KLX-005" -> "replicated-entity-crdt.html", // only pb sdks
+    "KLX-006" -> "writing-grpc-descriptors-protobuf.html#_transcoding_http", // only pb sdks
+    "KLX-007" -> "using-jwts.html",
+    "KLX-008" -> "timers.html",
+    "KLX-009" -> "access-control.html",
+    "KLX-010" -> "workflows.html", // only java sdk currently
+    "KLX-011" -> "actions.html#_actions_as_life_cycle_hooks",
+  )
```
efgpinto, line 43: I was a bit tempted to model this so that we would declare which codes are available for each sdk, but I've figured that e.g. java sdk will not be sending a `KLX-006` as it's only available for protocol-first sdks. So, ended up going for the simplest option. That would also be useful if we had the same category served from different docs pages, but htat is not the case for now.

sdk/java-sdk-protobuf/src/main/scala/kalix/javasdk/impl/DocLinks.scala
```diff
@@ -16,34 +16,35 @@
...
+    "KLX-002" -> "value-entity.html",
+    "KLX-003" -> "event-sourced-entities.html",
+    "KLX-004" -> "publishing-subscribing.html",
+    "KLX-005" -> "replicated-entity-crdt.html", // only pb sdks
+    "KLX-006" -> "writing-grpc-descriptors-protobuf.html#_transcoding_http", // only pb sdks
+    "KLX-007" -> "using-jwts.html",
+    "KLX-008" -> "timers.html",
+    "KLX-009" -> "access-control.html",
+    "KLX-010" -> "workflows.html", // only java sdk currently
+    "KLX-011" -> "actions.html#_actions_as_life_cycle_hooks")
```
efgpinto, line 41: I was a bit tempted to model this so that we would declare which codes are available for each sdk, but I've figured that e.g. java sdk will not be sending a KLX-006 as it's only available for protocol-first sdks. So, ended up going for the simplest option. That would also be useful if we had the same category served from different docs pages, but htat is not the case for now.

octonato (approved):
LGTM

PR #1969, opened 2024-01-12
docs: fix install kalix typos

beritou:


beritou:
Fun fact: while researching reasons why vale did not catch these typos, I learned that `instal` is British English way to spell the word, while `install` is the North American way. 🤷 

Regardless ,`install` is the industry standard way of spelling the word, so we should be consistent with that.

octonato (approved):
LGTM

I never noticed that we were writing `Instal`. Actuall, I'm learning now about this spelling. I never read it before.

PR #1971, opened 2024-01-15
feat: workflows support in java proto sdk

aludwiko:
References
-

### DIFF
--- a/codegen/core/src/main/scala/kalix/codegen/ModelBuilder.scala
+++ b/codegen/core/src/main/scala/kalix/codegen/ModelBuilder.scala
@@ -17,7 +17,9 @@
 package kalix.codegen
 
 import java.util.Locale
+
 import scala.jdk.CollectionConverters._
+
 import com.google.common.base.CaseFormat
 import com.google.protobuf.Descriptors
 import com.google.protobuf.Descriptors.ServiceDescriptor
@@ -27,6 +29,7 @@ import kalix.ReplicatedEntityDef
 import kalix.ServiceOptions.ServiceType
 import kalix.ValueEntityDef
 import kalix.View
+import kalix.WorkflowDef
 
 /**
  * Builds a model of entities and their properties from a protobuf descriptor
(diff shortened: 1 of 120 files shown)


sdk/java-sdk-protobuf/src/main/scala/kalix/javasdk/impl/workflow/WorkflowImpl.scala
```diff
@@ -303,6 +303,8 @@ final class WorkflowImpl(system: ActorSystem, val services: Map[String, Workflow
           Future.successful(toProtoEffect(effect, command.id))
 
         case Step(executeStep) =>
+          val context =
```
aludwiko, line 306: Context is required to call other components.

aludwiko:
Ready for the first round, just the codegen updates. Samples, and docs in the following PRs, because this one is big enough.

johanandren:
One good thing to try out with the refactoring, is to build a proto sample with current latest SDK, then bump to a local version including these changes and see if things break without deleting the old generated sources (user would have their own implementation in them so not so easy to just delete-regenerate)

johanandren (commented):
Looks good so far, one problem with the entity type change. 

Except for that I think next step is to add a sample with a workflow?

sdk/scala-sdk-protobuf/src/main/scala/kalix/scalasdk/eventsourcedentity/EventSourcedEntityProvider.scala
```diff
@@ -30,7 +29,7 @@ trait EventSourcedEntityProvider[S, E <: EventSourcedEntity[S]] {
 
   def serviceDescriptor: Descriptors.ServiceDescriptor
 
-  def entityType: String
+  def typeId: String
```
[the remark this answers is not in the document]
aludwiko: Is it really a problem to ask users to clean the repo after bumping? TBH, I can't imagine working with proto sdk without cleaning everything before compiling (relatively often). Otherwise, you can face many issues even now. 

In my first implementation, there was a deprecated `entityType` method, but this approach generates a lot of warnings, which can be fixed only by us, by releasing the next SDK version and a clean + compile will be required anyway to get rid of them.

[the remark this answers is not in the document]
johanandren: No, I had bad luck when thinking, provider is never extended by users themselves, expecting a clean+compile is fine 👍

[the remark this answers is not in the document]
aludwiko: Yeah, I asked @ceecer1 about this, I would leave it as it is because it shouldn't be a big issue to inform the client about the changes. We can revert this decision, but I would like to push this PR forward.

codegen/core/src/main/scala/kalix/codegen/ModelBuilder.scala
```diff
@@ -52,36 +55,37 @@ object ModelBuilder {
...
-    def lookupEntity(service: EntityService): Entity = {
-      entities.getOrElse(
+  case class Model(services: Map[String, Service], statefulComponents: Map[String, StatefulComponent]) {
+    def lookupEntity(service: EntityService): StatefulComponent = {
+      statefulComponents.getOrElse(
         service.componentFullName,
         throw new IllegalArgumentException(
-          "Service [" + service.messageType.fullyQualifiedProtoName + "] refers to entity [" + service.componentFullName +
-          s"], but no entity configuration is found for that component name. Entities: [${entities.keySet.mkString(", ")}]"))
+          "Service [" + service.messageType.fullyQualifiedProtoName + "] refers to stateful [" + service.componentFullName +
```
sebastian-alfers, line 70: `refers to stateful ["abc"]` vs `refers to component ["abc"]` or `refers to stateful component ["abc"]`?

Also, why using both concatenation (`+`) and string interpolation (`s`) together?

aludwiko: Fixed, I think that mixing was used to avoid very long line.

codegen/java-gen/src/main/scala/kalix/codegen/java/WorkflowSourceGenerator.scala
```diff
@@ -0,0 +1,285 @@
...
+          |public Effect<$outputType> ${lowerFirst(methodName)}($stateType currentState, $inputType ${lowerFirst(
+        cmd.inputType.name)}) {
+          |  return effects().error("The command handler for `$methodName` is not implemented, yet");
+          |}
+          |""".stripMargin
+    }
+
+    s"""package $packageName;
+       |
+       |${writeImports(imports)}
```
octonato, line 62: It's pity that we didn't refactored the java codegen to use our `c` interpolator. Check out the scala codegen for more details.

It's much easier to use the `c` interpolator and then we get the imports right out of the box.

We should do some gardening in the SDK to solve this kind of issues.

octonato (approved):
LGTM

We can merge as such, but we still need to add compilation tests for generated code. Can be added in follow-up as this PR is already massive.

codegen/java-gen/src/main/scala/kalix/codegen/java/WorkflowSourceGenerator.scala
aludwiko: https://github.com/lightbend/kalix-jvm-sdk/issues/1991

PR #1974, opened 2024-01-16
fix(java-sdk): enforce entity id presence on http calls

efgpinto:
Fix https://github.com/lightbend/kalix-runtime/issues/1848

With this PR:

- Instead of this (timeout + 500):
```shell
➜ curl -i -XGET -H "Content-Type: application/json" localhost:9000/cart/
HTTP/1.1 500 Internal Server Error
Access-Control-Allow-Origin: *
Server: akka-http/10.6.0
Date: Tue, 16 Jan 2024 16:29:27 GMT
Content-Length: 0
```

- We get:
```shell
➜ curl -i -XGET -H "Content-Type: application/json" localhost:9000/cart/
HTTP/1.1 400 Bad Request
Access-Control-Allow-Origin: *
Server: akka-http/10.6.0
Date: Tue, 16 Jan 2024 16:28:37 GMT
Content-Type: text/plain; charset=UTF-8
Content-Length: 69

INVALID_ARGUMENT: Entity key field(s) List(cartId) must not be empty.% 
```

Note that, this does NOT change current behavior where for non-entity id fields, one is able to do calls without providing all path variables. E.g for path defined as `"cart/{cart_id}/str/{msg}"`:

```
➜ curl -i -XGET -H "Content-Type: application/json" localhost:9000/cart/cart1/str/
HTTP/1.1 200 OK
Access-Control-Allow-Origin: *
Server: akka-http/10.6.0
Date: Tue, 16 Jan 2024 16:32:32 GMT
Content-Type: application/json
Content-Length: 2

""%
```

This means `{msg}` is considered to be `""`.

### DIFF
--- a/sdk/java-sdk-spring/src/it/java/com/example/wiring/SpringSdkIntegrationTest.java
+++ b/sdk/java-sdk-spring/src/it/java/com/example/wiring/SpringSdkIntegrationTest.java
@@ -108,6 +108,38 @@ public class SpringSdkIntegrationTest {
   }
 
   @Test
+  public void acceptRequestWithMissingPathParamIfNotEntityId() {
+
+    ResponseEntity<String> response =
+        webClient
+            .get()
+            .uri("/echo/message/") // missing param
+            .retrieve()
+            .toEntity(String.class)
+            .onErrorResume(WebClientResponseException.class, error -> Mono.just(ResponseEntity.status(error.getStatusCode()).body(error.getResponseBodyAsString())))
+            .block(timeout);
+
+    assertThat(response.getStatusCode()).isEqualTo(HttpStatus.OK);
+  }
+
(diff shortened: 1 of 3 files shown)


octonato (approved):
Fine to fix the most pressing issue for now. 

But I wonder if instead we should just return 404 if any path var is missing. At the end, accepting "" means that we are answering a request on a path that is not defined.

efgpinto:
> Fine to fix the most pressing issue for now.
> 
> But I wonder if instead we should just return 404 if any path var is missing. At the end, accepting "" means that we are answering a request on a path that is not defined.

Yeah, I have the code to do that on the runtime side. I will open a PR and we can follow up on that discussion. I still think this fix is worth it in any case.

PR #1977, opened 2024-01-18
bump: spring to 6.1.3 and spring boot to 3.2.1

efgpinto:
Close https://github.com/lightbend/kalix/issues/10420


octonato (approved):

octonato:
Oh, I approved but before merge there is something else we need to do.

The samples and templates also have the spring version in the pom.xml

octonato:
They import a parent pom, like here:
https://github.com/lightbend/kalix-jvm-sdk/blob/main/samples/java-spring-choreography-saga-quickstart/pom.xml#L7

efgpinto:
> They import a parent pom, like here: https://github.com/lightbend/kalix-jvm-sdk/blob/main/samples/java-spring-choreography-saga-quickstart/pom.xml#L7

Good call, thanks. Updated now.

PR #1992, opened 2024-01-24
feat: separate workflows step builder for different SDKs

aludwiko:


sdk/java-sdk-spring/src/it/java/com/example/wiring/workflowentities/WorkflowWithRecoverStrategy.java
```diff
@@ -29,7 +29,7 @@
 import java.util.concurrent.CompletableFuture;
 
 import static java.time.Duration.ofSeconds;
-import static kalix.javasdk.workflow.Workflow.RecoverStrategy.maxRetries;
+import static kalix.javasdk.workflow.AbstractWorkflow.RecoverStrategy.maxRetries;
```
aludwiko, line 32: This could potentially break the build for users, but I think no one is using workflows on prod at this moment.

aludwiko:
Ready for review, but also waiting for: https://github.com/lightbend/kalix-jvm-sdk/pull/1992/files

sdk/java-sdk-protobuf/src/main/java/kalix/javasdk/workflow/ProtoWorkflow.java
```diff
@@ -0,0 +1,44 @@
...
+ * <p>Workflow methods that handle incoming commands should return an {@link Effect} describing the
+ * next processing actions.
+ *
+ * @param <S> The type of the state for this workflow.
+ */
+@ApiMayChange
+public abstract class ProtoWorkflow<S extends GeneratedMessageV3> extends AbstractWorkflow<S> {
+  public ProtoStepBuilder step(String stepName) {
+    return new ProtoStepBuilder(stepName);
+  }
```
octonato, line 43: The nice thing about this design is that it puts the most relevant builder under our fingertips. 

But will it work if I use another builder? For example, from a code-first project calling a protobuf service. Can I choose a the ProtoStepBuilder?

sdk/java-sdk-spring/src/it/java/com/example/wiring/workflowentities/WorkflowWithRecoverStrategy.java
octonato: We do have a few people building PoCs with Workflows.

What about moving the RecoveryStategy to an interface that is then implemented by `Workflow` and `ProtoWorkflow`?

octonato (approved):
LGTM

I little bit in doubt about what to do with `RecoveryStrategy` though.

sdk/java-sdk-protobuf/src/main/java/kalix/javasdk/workflow/ProtoWorkflow.java
aludwiko: You can use proto-builder in the code-first sdk, but I'm not sure if it will give you any benefits. You can call the proto service with a default builder, but the response will be serialized as JSON anyway.

sdk/java-sdk-spring/src/it/java/com/example/wiring/workflowentities/WorkflowWithRecoverStrategy.java
aludwiko: I'm not sure if I follow you on that, what would that solve?

aludwiko: Ok, I'm merging this as it is, I would like to solve it in a separate PR once the scala part is done.

PR #1993, opened 2024-01-24
feat: java protobuf workflow samples and docs

aludwiko:


docs/src/modules/java-protobuf/pages/workflows.adoc
```diff
@@ -0,0 +1,200 @@
...
+include::ROOT:partial$include.adoc[]
+
+Workflows make it possible to implement long-running, multi-step business processes while focusing exclusively on domain and business logic. Workflows provide durability, consistency and the ability to call other components and services. Business transactions can be modeled in one central place, and the Workflow will keep them running smoothly, or roll back if something goes wrong.
+
+
+
+
+include::java-protobuf:partial$effects-workflow.adoc[]
+
+See also xref:java:development-process.adoc#_effects_api[Understanding what an Effect is]
```
franciscolopezsancho, line 14:
```suggestion
See also xref:java:development-process.adoc#_effects_api[Understanding what an Effect is].
```

franciscolopezsancho: It's the same in the `java` docs.

docs/src/modules/java/pages/workflows.adoc
```diff
@@ -51,7 +51,7 @@ include::example$java-spring-transfer-workflow/src/main/java/com/example/wallet/
 <3> An endpoint to deposit funds to the wallet.
 <4> An endpoint to get current wallet balance.
 
-Now we can focus on the workflow implementation itself. A workflow has state, which can be updated in command handlers and transition methods. During the state modeling we might consider the information is required for validation, running the steps, collecting data from steps or tracking the workflow progress.
+Now we can focus on the workflow implementation itself. A workflow has state, which can be updated in command handlers and transition methods. During the state modeling we might consider the information that is required for validation, running the steps, collecting data from steps or tracking the workflow progress.
```
franciscolopezsancho, line 54: Referring here to `transition methods` but not mentioned before. Maybe adding a link?

docs/src/modules/java-protobuf/pages/workflows.adoc
```diff
@@ -0,0 +1,200 @@
...
+
+The following `transfer_api.proto` file defines our `TransferWorkflow` component. In this file we instruct the Kalix code generation tooling (codegen) to generate all stubs for your workflow/service and corresponding tests, as well as an abstract class for your implementation to extend.
+
+[source,java,indent=0]
+.src/main/proto/com/example/transfer/transfer_api.proto
+----
+include::example$java-protobuf-transfer-workflow/src/main/proto/com/example/transfer/transfer_api.proto[tag=transfer-workflow]
+----
+<1> Any classes generated from this protobuf file will be in the Java package `com.example.transfer.api`.
+<2> Import the Kalix protobuf annotations or options.
```
franciscolopezsancho, line 48: This doesn't look correct referring to 
```
import "kalix/annotations.proto";
import "com/example/transfer/transfer_domain.proto"
```

franciscolopezsancho:
How about to add a link to download the example? I was looking for one. 

Didn't finish the review. Almost there. I'll finish tomorrow morning.

docs/src/modules/java-protobuf/pages/workflows.adoc
```diff
@@ -0,0 +1,200 @@
...
+
+[source,java,indent=0]
+.src/main/java/com/example/transfer/api/TransferWorkflow.java
+----
+include::example$java-protobuf-transfer-workflow-compensation/src/main/java/com/example/transfer/api/TransferWorkflow.java[tag=pausing]
+----
+<1> Schedules a timer as a Workflow step action. Make sure that the timer name is unique for every Workflow instance.
+<2> Pauses the Workflow execution.
+
+NOTE: Remember to cancel the timer once the Workflow is resumed. Also, adjust the Workflow timeout to match the timer schedule.
```
franciscolopezsancho, line 130: `Also, adjust the Workflow timeout to match the timer schedule. ` This is mentioned before showing Workflow's timeouts. Maybe add a link?

docs/src/modules/java-protobuf/pages/workflows.adoc
```diff
@@ -0,0 +1,200 @@
...
+
+By default, a workflow run has no time limit. It can run forever, which in most cases is not desirable behavior. A workflow step, on the other hand, has a default timeout of 5 seconds timeout. Both settings can be overridden at the workflow definition level or for a specific step configuration.
+
+[source,java,indent=0]
+.src/main/java/com/example/transfer/api/TransferWorkflow.java
+----
+include::example$java-protobuf-transfer-workflow-compensation/src/main/java/com/example/transfer/api/TransferWorkflow.java[tag=timeouts]
+----
+<1> Sets a workflow global timeout.
+<2> Sets a default timeout for all workflow steps.
```
franciscolopezsancho, line 156: Is the `global timeout` the total time the Workflow has to be completed? Meaning that the total sum of all the steps execution time has to be less than 5s. If so I would expect this Workflow only has 2 steps.

docs/src/modules/java-protobuf/pages/workflows.adoc
```diff
@@ -0,0 +1,200 @@
...
+----
+include::example$java-protobuf-transfer-workflow-compensation/src/main/java/com/example/transfer/api/TransferWorkflow.java[tag=compensation]
+----
+<1> Explicit deposit call result type `DepositResult`.
+<2> Finishes workflow as completed, in the case of a successful deposit.
+<3> Launches compensation step to handle deposit failure. The `"withdraw"` step must be reversed. Compensation step is just any other step, with the same set of functionalities.
+<4> Correct compensation can finish the workflow.
+<5> Any other result might be handled by a default recovery strategy.
+
+Compensating a workflow step (steps) might involve multiple logical steps and thus is part of the overall business logic that must be defined within the workflow itself. For simplicity, in the example above, the compensation is applied only to `withdraw` step. Whereas `deposit` step itself might also require a compensation. In case of a step timeout we can't be certain about step successful or error outcome.
```
franciscolopezsancho, line 198: "In case of a step timeout we can’t be certain about step successful or error outcome" Could you elaborate on that?

franciscolopezsancho:
Finished now.

docs/src/modules/java-protobuf/pages/workflows.adoc
aludwiko: why? because we import other things as well?

docs/src/modules/java-protobuf/pages/workflows.adoc
aludwiko: added link

docs/src/modules/java-protobuf/pages/workflows.adoc
aludwiko: Answering your question, yes. 5 sec is just a dummy value, so yes, only 2 slow steps.

docs/src/modules/java-protobuf/pages/workflows.adoc
aludwiko: For instance we are calling some service, in case of a timeout there are 2 options:
 - request failed
 - request was actually successful but take more time

docs/src/modules/java/pages/workflows.adoc
aludwiko: changed to `steps implementation`

ennru:
CI failure should be fixed with
- https://github.com/lightbend/kalix-docs/pull/2050

docs/src/modules/java-protobuf/pages/workflows.adoc
franciscolopezsancho: Right, right. 
I think is important the fact that it would happen when calling a service. Maybe some rephrasing? Here's one possibility.

```
When calling another service in a step, if the step times out, we can't be sure if the call was successful or not.
```

franciscolopezsancho (approved):
I left a last comment feel free to skip it. 
LGTM.

octonato (approved):
LGTM

PR #2026, opened 2024-02-08
docs: add tracing sample for java sdk

efgpinto:
Refs https://github.com/lightbend/kalix/issues/9415

This PR:
- ~adds a new `TraceContext` type that can be derived from the metadata (similar to `JwtClaims`)~
- ~adds support for automatic injection of an opentelemetry `Tracer` (similar to the injection of the `ComponentClient`) for easier custom child spans (i.e. tracking an external call or a specific local computation)~
- adds a new tracing sample with java sdk for exemplifying all things related with tracing in Kalix
  - the sample uses a ESE for User management. A User is created with an email and then 2 actions generate a name and a photo respectively using an external API. 
  - the idea would be for this sample to show the different bits of tracing, events, component calls, external calls, etc.

### DIFF
--- a/samples/java-spring-tracing/pom.xml
+++ b/samples/java-spring-tracing/pom.xml
@@ -1,0 +1,59 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+  <parent>
+    <groupId>io.kalix</groupId>
+    <artifactId>kalix-spring-boot-parent</artifactId>
+    <version>1.4.1</version>
+  </parent>
+
+  <groupId>com.example</groupId>
+  <artifactId>java-spring-tracing</artifactId>
+  <version>0.0.1-SNAPSHOT</version>
+  <packaging>jar</packaging>
+
+  <name>java-spring-tracing</name>
+  <properties>
+
+    <!-- For Docker setup see https://docs.kalix.io/projects/container-registries.html -->
+    <kalixContainerRegistry>kcr.us-east-1.kalix.io</kalixContainerRegistry>
+    <kalixOrganization>acme</kalixOrganization>
+    <mainClass>com.example.tracing.Main</mainClass>
+  </properties>
+
+  <dependencyManagement>
+    <dependencies>
+      <dependency>
+        <groupId>io.opentelemetry</groupId>
+        <artifactId>opentelemetry-bom</artifactId>
+        <version>1.34.1</version>
+        <type>pom</type>
+        <scope>import</scope>
+      </dependency>
+    </dependencies>
+  </dependencyManagement>
+
+  <dependencies>
+    <dependency>
+      <groupId>io.kalix</groupId>
+      <artifactId>kalix-spring-boot-starter</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>io.kalix</groupId>
+      <artifactId>kalix-spring-boot-starter-test</artifactId>
+      <scope>test</scope>
+    </dependency>
+
+    <dependency>
+      <groupId>org.junit.jupiter</groupId>
+      <artifactId>junit-jupiter-api</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.junit.jupiter</groupId>
+      <artifactId>junit-jupiter</artifactId>
+      <scope>test</scope>
+    </dependency>
+
+  </dependencies>
+</project>
--- a/samples/java-spring-tracing/src/it/resources/logback-test.xml
+++ b/samples/java-spring-tracing/src/it/resources/logback-test.xml
@@ -1,0 +1,17 @@
+<configuration>
+    <appender name="STDOUT" class="ch.qos.logback.core.ConsoleAppender">
+        <encoder>
+            <pattern>%date{ISO8601} %-5level %logger - %msg%n</pattern>
+        </encoder>
+    </appender>
+
+    <logger name="akka" level="INFO"/>
+    <logger name="kalix" level="INFO"/>
+    <logger name="com.github.dockerjava" level="INFO"/>
+    <logger name="io.grpc.netty" level="INFO"/>
+    <logger name="org.testcontainers" level="INFO"/>
+
+    <root level="DEBUG">
+        <appender-ref ref="STDOUT"/>
+    </root>
+</configuration>
--- a/samples/java-spring-tracing/src/main/java/com/example/tracing/api/GetRandomNameAction.java
+++ b/samples/java-spring-tracing/src/main/java/com/example/tracing/api/GetRandomNameAction.java
@@ -1,0 +1,70 @@
+package com.example.tracing.api;
+
+import com.example.tracing.domain.UserEvent;
+import io.opentelemetry.api.trace.*;
+import io.opentelemetry.context.Scope;
+import kalix.javasdk.action.Action;
+import kalix.javasdk.annotations.Subscribe;
+import kalix.javasdk.client.ComponentClient;
+import org.springframework.web.reactive.function.client.WebClient;
+
+import java.util.concurrent.CompletableFuture;
+
+
+@Subscribe.EventSourcedEntity(value = UserEntity.class, ignoreUnknown = true)
+public class GetRandomNameAction extends Action {
+
+  private static final org.slf4j.Logger log = org.slf4j.LoggerFactory.getLogger(GetRandomNameAction.class);
+
+  private final ComponentClient componentClient;
+
+  private final WebClient webClient;
+
+  public GetRandomNameAction(ComponentClient componentClient) {
+    this.componentClient = componentClient;
+    this.webClient = WebClient.create("https://randomuser.me/api/?inc=name&noinfo");
+  }
+
+  public Effect<String> handleAdd(UserEvent.UserAdded userAdded) {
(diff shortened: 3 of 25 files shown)


efgpinto:
I'm going to split this PR in 2:
1. the logic to correctly propagate tracing headers into the metadata available from `actionContext`
2. new specific sample for tracing demonstration

This is to allow for easier revision/merge of 1. because I think we might want to iterate a bit more on the sample.

This one will stay only with the sample. Will open 1. on the side.

samples/java-spring-tracing/src/main/java/com/example/tracing/api/GetRandomPhotoActionAsync.java
```diff
@@ -0,0 +1,85 @@
...
+          .put()
+          .uri(uriBuilder -> uriBuilder
+              .path("/user/{userId}/photo")
+              .queryParam("url", randomPhotoUrl)
+              .build(actionContext().eventSubject().get()))
+          .headers(h -> tracingMap.forEach(h::set))
+          .retrieve()
+          .bodyToMono(String.class)
+          .toFuture();
+    });
```
efgpinto, line 54: The idea was to have an example where we would inject the tracing headers manually... but I'm not sure using a local call for that was the best idea. Having another example for that might be less confusing.

samples/java-spring-tracing/pom.xml
```diff
@@ -0,0 +1,301 @@
...
+      <version>1.32.1-alpha</version>
+    </dependency>
+
+
+    <!-- TODO: TO REMOVE -->
+    <dependency>
+      <groupId>io.opentelemetry</groupId>
+      <artifactId>opentelemetry-sdk-extension-autoconfigure</artifactId>
+    </dependency>
+  </dependencies>
```
efgpinto, line 300: Auto discovery is to remove once internal OpenTelemtry instance is registered as bean

samples/java-spring-tracing/pom.xml
```diff
@@ -0,0 +1,301 @@
...
+      <groupId>io.opentelemetry.instrumentation</groupId>
+      <artifactId>opentelemetry-spring-web-3.1</artifactId>
+      <version>1.32.1-alpha</version>
+    </dependency>
+
+    <dependency>
+      <groupId>io.opentelemetry.instrumentation</groupId>
+      <artifactId>opentelemetry-spring-webflux-5.3</artifactId>
+      <version>1.32.1-alpha</version>
+    </dependency>
```
efgpinto, line 292: This is the auto-instrumentation for `RestTemplate` and `WebClient`.. it might be removed or not depending on what kind of examples we use.

efgpinto (commented):
I will leave this as a draft because I think we need to re-iterate when we have component to component tracing and we can se exactly what kind of sample we want to build. Main questions:
- how to build a custom span: keep external calls to real API vs replace with long local computations
- use real API for traces propagation example vs setup e.g. wiremock

efgpinto:
So, providing an update here, I've rebased the PR against the `release-1.5` branch and updated the samples to use the latest java version and new dependency repo.

I was starting to simplify the sample to be similar to the protobuf counterparts, however, I think there is value in having the also where tracing from a ES and a component call is also shown which is something the PB ones don't show (only focus creating a span). Thus, I wonder if I should keep this as is and then enrich the others similarly.

Additionally, I seem to have found a small issue with the automatic component tracing (seems to be creating a new trace instead of hooking as a child into the existing one). Maybe we should sync on this @octonato and see how to move forward.

sdk/java-sdk-spring/src/main/scala/kalix/javasdk/impl/action/ReflectiveActionRouter.scala
```diff
@@ -42,28 +42,24 @@ class ReflectiveActionRouter[A <: Action](
...
-                .invoke(action, invocationContext)
-                .asInstanceOf[Action.Effect[_]]
-          }
-        case None if ignoreUnknown => ActionEffectImpl.Builder.ignore()
-        case None =>
-          throw new NoSuchElementException(
-            s"Couldn't find any method with input type [$inputTypeUrl] in Action [$action].")
-      }
-    } finally {
-      componentClients.foreach(_.clearCallMetadata())
```
octonato, line 66: We should not clear this. There is one ComponentClient instance per call, so it's safe to keep the same call metadata. 

If someone uses the ComponentClient on a future call back, we risk to start a new span if we clear the metadata. 

In other works, all futures that are spawn from an Action call should and use its ComponentCall instance should use the same call metadata.

sdk/java-sdk-spring/src/main/scala/kalix/javasdk/client/ComponentCall.scala
```diff
@@ -57,15 +57,23 @@ final class ComponentCall[A1, R](
...
   }
 
-  def noParams[R](kalixClient: KalixClient, method: Method, ids: util.List[String]): DeferredCall[Any, R] = {
-    invoke(Seq.empty, kalixClient, method, ids.asScala.toList)
+  def noParams[R](
+      kalixClient: KalixClient,
+      method: Method,
+      ids: util.List[String],
+      callMetadata: Optional[Metadata]): DeferredCall[Any, R] = {
+    invoke(Seq.empty, kalixClient, method, ids.asScala.toList, callMetadata)
```
octonato, line 73: Actually, Those method should be private.

octonato: We can only make them private with `private[client]` and that is ignored by java, so not really useful. I will leave it as such.

The right thing to do is to move it to an `impl` package.

samples/java-spring-tracing/src/main/java/com/example/tracing/api/GetRandomNameAction.java
```diff
@@ -0,0 +1,70 @@
...
+  public GetRandomNameAction(ComponentClient componentClient) {
+    this.componentClient = componentClient;
+    this.webClient = WebClient.create("https://randomuser.me/api/?inc=name&noinfo");
+  }
+
+  public Effect<String> handleAdd(UserEvent.UserAdded userAdded) {
+    if (actionContext().eventSubject().isPresent()) {
+      var randomNameFut = getRandomNameAsync();
+
+      var updateNameCall = randomNameFut.thenCompose(name ->
```
aludwiko, line 32: maybe you could inline the method call, the `randomNameFut` is not used elsewhere

samples/java-spring-tracing/src/main/java/com/example/tracing/api/GetRandomNameAction.java
```diff
@@ -0,0 +1,70 @@
...
+  // gets random name from external API using a synchronous call and traces that call
+  private CompletableFuture<String> getRandomNameAsync() {
+    var otelCurrentContext = actionContext().metadata().traceContext().asOpenTelemetryContext();
+    Span span = actionContext().getTracer()
+        .spanBuilder("random-name-async")
+        .setParent(otelCurrentContext)
+        .setSpanKind(SpanKind.CLIENT)
+        .startSpan();
+
+    try (Scope ignored = span.makeCurrent()) {
```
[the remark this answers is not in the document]
efgpinto: thanks, changed now

samples/java-spring-tracing/src/it/resources/logback-test.xml
```diff
@@ -0,0 +1,17 @@
...
+        </encoder>
+    </appender>
+
+    <logger name="akka" level="INFO"/>
+    <logger name="kalix" level="INFO"/>
+    <logger name="com.github.dockerjava" level="INFO"/>
+    <logger name="io.grpc.netty" level="INFO"/>
+    <logger name="org.testcontainers" level="INFO"/>
+
+    <root level="DEBUG">
```
aludwiko, line 14: INFO?

samples/java-spring-tracing/src/main/java/com/example/tracing/api/UserEntity.java
```diff
@@ -0,0 +1,105 @@
...
+  }
+
+  @Override
+  public User emptyState() { // <2>
+    return new User(entityId, "", "", "");
+  }
+
+  @GetMapping
+  public Effect<User> get() {
+    if (currentState() == emptyState())
```
aludwiko, line 42: technically equals is safer

samples/java-spring-tracing/src/main/java/com/example/tracing/api/UserEntity.java
```diff
@@ -0,0 +1,105 @@
...
+
+    var updated = new UserEvent.UserNameUpdated(updateNameCmd.name);
+
+    return effects()
+        .emitEvent(updated)
+        .thenReply(newState -> "OK");
+  }
+
+  @PutMapping("/photo")
+  public EventSourcedEntity.Effect<String> updatePhoto(@RequestParam String url) {
```
aludwiko, line 76: same here

samples/java-spring-tracing/src/main/java/com/example/tracing/api/UserEntity.java
```diff
@@ -0,0 +1,105 @@
...
+    if (currentState() == emptyState()){
```
aludwiko, line 77: equals

samples/java-spring-tracing/src/main/resources/application.conf
```diff
@@ -0,0 +1,4 @@
+
+# //tag::snapshot-every[]
```
aludwiko, line 2: do we use this file in some docs?

efgpinto: removed, tks

samples/java-spring-tracing/src/it/resources/logback-test.xml
efgpinto: since this is for tests, I think DEBUG is fine.. right?

PR #2032, opened 2024-02-12
feat: (java-spring) support multiple subscriptions of the same stream

aludwiko:
waiting for the runtime release, but can be reviewed now

### DIFF
--- a/sdk/java-sdk-spring/src/main/java/kalix/javasdk/annotations/Subscribe.java
+++ b/sdk/java-sdk-spring/src/main/java/kalix/javasdk/annotations/Subscribe.java
@@ -153,6 +153,13 @@ public @interface Subscribe {
     String service();
 
     /**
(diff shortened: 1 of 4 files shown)


sdk/java-sdk-spring/src/main/java/kalix/javasdk/annotations/Subscribe.java
```diff
@@ -152,6 +152,13 @@
      */
     String service();
 
+    /**
+     * In case you need to consume the same stream multiple times, each subscription should have a unique consumer group.
+     * <p>
+     * Changing the consumer group will lead to starting over from the beginning of the stream.
+     */
+    String consumerGroup() default "";
```
aludwiko, line 160: I thought about a different name, but my conclusion was that it would be confusing to have the same semantics as Topic.consumerGroup, under another name.

johanandren (approved):

PR #2042, opened 2024-02-21
chore: GH built-in release notes formatting

ennru:
With this we should be able to remove the release-drafter workflow and still get the structured release notes.


johanandren (approved):
Neat

PR #2046, opened 2024-02-23
feat: direct access to component client

octonato:
Still draft because running it as an experiment for now. 

Workflow is expected to be used only from the code-first SDK. Therefore it moves into that package. 

A field for the `kalixClient` is added to it and set via reflection when we construct the class.

The `componentClient()` method is then directly accessible to the Workflow and we can pass in that method a call context with the tracing much like we do in other places.

The same can be done for Actions.

### DIFF
--- a/codegen/java-gen/src/test/resources/tests/action-service/named-new-style/generated-managed/org/example/service/MyServiceNamedActionRouter.java
+++ b/codegen/java-gen/src/test/resources/tests/action-service/named-new-style/generated-managed/org/example/service/MyServiceNamedActionRouter.java
@@ -3,7 +3,7 @@ package org.example.service;
 import akka.NotUsed;
 import akka.stream.javadsl.Source;
 import com.google.protobuf.Empty;
-import kalix.javasdk.action.Action.Effect;
+import kalix.javasdk.action.AbstractAction.Effect;
 import kalix.javasdk.action.MessageEnvelope;
 import kalix.javasdk.impl.action.ActionRouter;
 
--- a/sdk/java-sdk-spring/src/main/scala/kalix/spring/impl/KalixSpringApplication.scala
+++ b/sdk/java-sdk-spring/src/main/scala/kalix/spring/impl/KalixSpringApplication.scala
@@ -377,16 +377,24 @@ case class KalixSpringApplication(applicationContext: ApplicationContext, config
     ReflectiveActionProvider.of(
       clz,
       messageCodec,
-      context =>
-        wiredInstance(clz) {
-          case p if p == classOf[ActionCreationContext] => context
-          case p if p == classOf[KalixClient]           => kalixClient(context)
-          case p if p == classOf[ComponentClient]       => componentClient(context)
-          case p if p == classOf[WebClientProvider]     => webClientProvider(context)
-          case p if p == classOf[Tracer] =>
-            context.getOpenTelemetryTracer.orElseGet(() =>
-              throw new IllegalStateException("Tracer not available. Make sure to have enabled tracing."))
-        })
+      context => {
+        val action =
+          wiredInstance(clz) {
+            case p if p == classOf[ActionCreationContext] => context
+            case p if p == classOf[KalixClient]           => kalixClient(context)
+            case p if p == classOf[ComponentClient]       => componentClient(context)
+            case p if p == classOf[WebClientProvider]     => webClientProvider(context)
+            case p if p == classOf[Tracer] =>
+              context.getOpenTelemetryTracer.orElseGet(() =>
+                throw new IllegalStateException("Tracer not available. Make sure to have enabled tracing."))
+          }
+
+        val field = lookupSuperclassField("kalixClient", action.getClass.getSuperclass)
+        field.setAccessible(true)
+        field.set(action, kalixClient(context))
+
+        action
+      })
 
   private def workflowProvider[S, W <: Workflow[S]](clz: Class[W]): WorkflowProvider[S, W] = {
     ReflectiveWorkflowProvider.of(
--- a/sdk/java-sdk-spring/src/main/scala/kalix/spring/impl/KalixSpringApplication.scala
+++ b/sdk/java-sdk-spring/src/main/scala/kalix/spring/impl/KalixSpringApplication.scala
@@ -17,12 +17,16 @@
 package kalix.spring.impl
 
 import java.lang.reflect.Constructor
+import java.lang.reflect.Field
 import java.lang.reflect.Modifier
 import java.lang.reflect.ParameterizedType
+
+import scala.annotation.tailrec
 import scala.concurrent.Future
 import scala.jdk.CollectionConverters.CollectionHasAsScala
 import scala.jdk.FutureConverters.CompletionStageOps
 import scala.jdk.OptionConverters.RichOption
+
 import akka.Done
 import com.typesafe.config.Config
 import io.opentelemetry.api.trace.Tracer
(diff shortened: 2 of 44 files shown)


sdk/java-sdk-spring/src/main/scala/kalix/spring/impl/KalixSpringApplication.scala
```diff
@@ -373,16 +377,24 @@ case class KalixSpringApplication(applicationContext: ApplicationContext, config
...
+            case p if p == classOf[ActionCreationContext] => context
+            case p if p == classOf[KalixClient]           => kalixClient(context)
+            case p if p == classOf[ComponentClient]       => componentClient(context)
+            case p if p == classOf[WebClientProvider]     => webClientProvider(context)
+            case p if p == classOf[Tracer] =>
+              context.getOpenTelemetryTracer.orElseGet(() =>
+                throw new IllegalStateException("Tracer not available. Make sure to have enabled tracing."))
+          }
+
+        val field = lookupSuperclassField("kalixClient", action.getClass.getSuperclass)
```
franciscolopezsancho, line 392: Here's the magic sauce. Nice.

Then how would this work in https://github.com/lightbend/kalix-jvm-sdk/pull/2022
Let me try and fail. Each time the user's Action gets instantiated it will set the `kalixClient` with the `ActionCreationContext` so when calling `componentClient()` to create a deferred call, the `ComponentCall` can pick it and ...failing here ... because we need `ActionContext` not `ActionCreationContext` to pick the metadata to add to the deferred call. What do you think?

sdk/java-sdk-spring/src/main/scala/kalix/spring/impl/KalixSpringApplication.scala
```diff
@@ -419,10 +431,24 @@ case class KalixSpringApplication(applicationContext: ApplicationContext, config
...
         workflow
       })
   }
 
+  @tailrec
+  private def lookupSuperclassField(fieldName: String, clz: Class[_]): Field = {
+    try {
+      clz.getDeclaredField(fieldName)
+    } catch {
+      case e: NoSuchFieldException =>
```
franciscolopezsancho, line 447: Why superclass here? Isn't `kalixClient` in the new `Action` or `Workflow` both of which are the direct parent from the user's component? I mean, we already call `getSuperclass` when calling to this method in the two cases above.

octonato: We can only get the fields that are declared in a class. So, if I have `MyWorkflow extends Workflow`, to find the `kalixClient` field, we need to call `getDeclaredField` on the superclass of `MyWorkflow`, ie: `Workflow`.

So far so good. The first time we call this method we are already passing `workflow.getClass.getSuperclass`. 

But if we don't find it, we need to go one level down until we find it.

For example, image this case:
```java
abstract class MyBaseWorkflow extends Workflow {...}

class MyWorkflow extends MyBaseWorkflow {...}
```

The superclass of `MyWorkflow` is `MyBaseWorkflow` and therefore we won't find the `kalixClient` field on it. 
We need need to recover and dig deeper until we find the super-superclass `Workflow`. 

That's why this recursive lookup.

sdk/java-sdk-spring/src/main/scala/kalix/spring/impl/KalixSpringApplication.scala
octonato: I'm not sure I'm following you.

To build a trace enabled ComponentClient we need the `kalixClient` and the call context (not the `ActionCreationContext`).

Maybe the code is misleading because it gives the impression that `kalixClient` needs `ActionCreationContext`. That's not true. It needs the Akka materializer only. And we get it from the creation context. Actually, that's the same for all calls.

So here, I'm inject the `kalixClient` in a field. So now, when the user call the method `componentClient()` inside the command handler we have access to the call context and the `kalixClient`. As such, we can build a fresh `ComponentClient` that is tweaked to use the tracing from the call context. 

Each call to `componentClient()` would return a new `ComponentClient` and that's what we want because it will have a state now that is bounded to the call context.

To make it work with #2022, we will need to [remove this](https://github.com/lightbend/kalix-jvm-sdk/pull/2022/files#diff-0001cb59d6b3e4b2fe7d887f42329155d35299acff67b267a37f9f5db1b77e50R61) and add a optional `ActionContext` to the `ComponentClient` constructor.

sdk/java-sdk-spring/src/main/scala/kalix/spring/impl/KalixSpringApplication.scala
franciscolopezsancho: Right. I see. I never thought the users would create an abstract `Action` or `Workflow` themselves. Seems a bit contrived but two more lines of codes to cover that case I think it's a sale.

sdk/java-sdk-spring/src/main/scala/kalix/spring/impl/KalixSpringApplication.scala
franciscolopezsancho: Oh, Ok. Right, `componentClient()` it's a method now. Of course! Thanks for the explanation.

sdk/java-sdk-spring/src/main/scala/kalix/spring/impl/KalixSpringApplication.scala
octonato: `try { clz.getDeclaredField(fieldName) } catch { case e: NoSuchFieldException => lookupSuperclassField(fieldName, clz.getSuperclass) }`

I can make it one line 😄

efgpinto:
Seems like a good solution.

I'm assuming we would force users to move to the method-based instead of dependency injection.. but I'm wondering what can we do to make the life easier for our users, since this is a significant breaking change (although easy to migrate).

Maybe we can allow users to keep the old way as long as they are not using tracing? We could still try to warn them to migrate to the new way but let them use the old way (at least for some time) if they are not interested in tracing.

codegen/java-gen/src/test/resources/tests/action-service/named-new-style/generated-managed/org/example/service/MyServiceNamedActionRouter.java
```diff
@@ -3,7 +3,7 @@
 import akka.NotUsed;
 import akka.stream.javadsl.Source;
 import com.google.protobuf.Empty;
-import kalix.javasdk.action.Action.Effect;
+import kalix.javasdk.action.AbstractAction.Effect;
```
octonato, line 6: I just realised that we move Action to AbstractAction because the Effect changes its location and that can potentially break existing code. 

@franciscolopezsancho / @aludwiko

aludwiko: Wasn't that the same with workflows, and we did it anyway? Or maybe we just haven't spotted it?

franciscolopezsancho: Well spotted. Nice. 
Are there any other breaking changes we are aware we would like to do?
How about we group them all together and release 2.0? In a few weeks I mean.

Meanwhile we could release the original PR or just wait until releasing this breaking change, if we decide to do it, ofc.
WDYT guys?

franciscolopezsancho:
@octonato found a viable solution to this. If an instance of the ComponentClient exists in the constructor then by reflection find this is the case and assign one with the current context, when `_setInternalContext` is called. For Actions or Workflows.

octonato:
Closing this. We have an alternative approach.

Issue #1997, opened 2024-01-26
Make it easier to change runtime and service ports


octonato:
When running multi-services tests, users need to add a few changes to the docker-compose file.

* they need to add a env var for the runtime: `HTTP_PORT:9001` (that's runtime internal port)
* they need to change the port mapping to "9001:9001"
* they need to pick another port for the service USER_SERVICE_PORT:8081

The first two changes are not so obvious for the users. Someone experienced with Docker will think about only changing the exposed port "9001:9000", but then the user service running on host machine will try to talk to the runtime through port 9000 and it won't work. 

A possible solution (but also not perfect) is to use a `.env` file. 
Docker Compose reads this file when it starts and do variable substitution. 

The env file would look like this:

```
=❯ cat .env 
HTTP_PORT=9000
USER_SERVICE_PORT=8080
```

The compose file would look like this:

```
version: "3"
services:
  kalix-runtime:
    image: gcr.io/kalix-public/kalix-runtime:1.1.29
    container_name: java-spring-fibonacci-action
    ports:
      - "${HTTP_PORT}:${HTTP_PORT}"
    extra_hosts:
      - "host.docker.internal:host-gateway"
    environment:
      JAVA_TOOL_OPTIONS: >
      HTTP_PORT: ${HTTP_PORT}
      USER_SERVICE_PORT: ${USER_SERVICE_PORT}
```

Then, the runtime http port will be defined in a single place, as well as the USER_SERVICE_PORT. 
And we can document that `.env` is the file to edit whenever you want to have more than one service running on your local machine.

I don't find it an ideal solution, but if our samples and templates adopt this format, we can easily explain how to change the port number and then, if the user keep the original compose file, the mapping will be correct.

octonato:
Additionally, we can validate the ports inside the mvn/sbt plugin. If the final docker config says: 9001:9000 for example, then we know that this services will be broken and we can help the user in finding the issue.

octonato:
For reference:
https://docs.docker.com/compose/environment-variables/set-environment-variables/#compose-file

aludwiko:
Interesting, we can potentially use this .env file for other purposes, when it comes to the docker-compose I’m ok with the idea, I would just put a comment in the docker-compose file explaining where to look for values.

octonato:
I have a call with @hedefalk today to brainstorm once more about it. 

We settle for an explicit advertised port, but will will add the .env to the docker-compose. 

At the end, the file will look like:

[...]

PR #2048, opened 2024-02-27
chore: include ADVERTISED_HTTP_PORT

octonato:
Resolves https://github.com/lightbend/kalix-jvm-sdk/issues/1997


maven-java/kalix-maven-archetype-event-sourced-entity/src/main/resources/META-INF/maven/archetype-metadata.xml
```diff
@@ -18,6 +18,12 @@
                 <include>**/*</include>
             </includes>
         </fileSet>
+        <fileSet filtered="true">
+            <directory></directory>
+            <includes>
+                <include>.env</include>
+            </includes>
+        </fileSet>
```
octonato, line 26: This is the  trick to include hidden files in the archetype. 
There is an issue about .gitignore file as well. I will PR for the .gitignore separately.

maven-java/kalix-maven-archetype-event-sourced-entity/src/main/resources/archetype-resources/docker-compose.yml
```diff
@@ -8,15 +8,19 @@ services:
...
     extra_hosts:
       - "host.docker.internal:host-gateway"
     environment:
       JAVA_TOOL_OPTIONS: >
       # jvm -D properties can be added under this environment map (note: remove this comment when adding properties)
 
+      # those variables are defined in the .env file
+      # when running multiple services on your local machine, make sure that 
+      # the ports are unique by editing the corresponding .env files
+      ADVERTISED_HTTP_PORT: ${ADVERTISED_HTTP_PORT}
```
octonato, line 21: It's a pity that we need to repeat the env vars to pass it to the JVM. 

Also, I'm adding this text here because the service port is also defined here and it's better to get them all together. But `ADVERTISED_HTTP_PORT` shows up for the first time in the ports mapping above. 

Anyway, that might be fine an obvious enough for users.

samples/java-protobuf-eventsourced-customer-registry-subscriber/docker-compose-integration.yml
octonato, line 1: This docker-compose-integration are special. They have two runtimes. 

We could add other vars as well, but here I'm choosing to keep the one based on the .env and the main runtime (the one for the subscriber service) with explicit ports. 

In any case, that's mostly for our own tests.

octonato (commented):
Includes the advertised port to the docker compose and .env files.

maven-java/kalix-maven-archetype-event-sourced-entity/src/main/resources/archetype-resources/.env
```diff
@@ -0,0 +1,11 @@
...
+# when running multiple services on your local machine, make sure that the ports unique
+ADVERTISED_HTTP_PORT=9000
+
+# this is the port where the user services (your applicaiton) will open
+# when running multiple services on your local machine, make sure that the ports unique
+USER_SERVICE_PORT=8080
+
+# this variable defines the host where the kalix runtime (running in docker) 
+# will reach the user service in local development
+USER_SERVICE_HOST=host.docker.internal
```
octonato, line 11: This property is in most case irrelevant. Users can't simply change it. Maybe if using something like Colima or similar.

aludwiko (approved):
LGTM

PR #2051, opened 2024-02-28
feat: showcasing how to create spans

franciscolopezsancho:
Not ready for review. Just a place holder.


samples/java-protobuf-tracing/src/main/java/com/example/FactorialControllerAction.java
```diff
@@ -0,0 +1,103 @@
...
+
+  @Override
+  public Effect<MessageResponse> callSyncEndpoint(Empty empty) {
+    Span span  = actionContext().getOpenTelemetryTracer().get().spanBuilder("calculateFactorial").startSpan();
+    span.setAttribute("attribute1", "value1");
+    String result = "";
+    try {
+      result = callSyncService(span.getSpanContext().getTraceId()); 
+      span.setAttribute("result", result); //pick some value from it
+      span.end();
```
aludwiko, line 39: I would move end to the finally section

samples/java-protobuf-tracing/src/main/java/com/example/FactorialControllerAction.java
```diff
@@ -0,0 +1,103 @@
...
+import com.google.protobuf.Empty;
+
+
+// This class was initially generated based on the .proto definition by Kalix tooling.
+// This is the implementation for the Action Service described in your com/example/factorial_action.proto file.
+//
+// As long as this file exists it will not be overwritten: you can maintain it yourself,
+// or delete it so it is regenerated as needed.
+
+public class FactorialControllerAction extends AbstractFactorialControllerAction {
```
aludwiko, line 26: I don't understand the idea of naming this FactorialController, we just call some random service here.

aludwiko (commented):
I know it's a work in progress, but in the current form, do we need a separate sample for that. We can just add this action to existing samples.

samples/java-protobuf-tracing/src/main/java/com/example/FactorialControllerAction.java
franciscolopezsancho: That's outdated. Now is called `ControllerAction`.

franciscolopezsancho:
> I know it's a work in progress, but in the current form, do we need a separate sample for that. We can just add this action to existing samples.

I like that. Let's do it.

samples/java-protobuf-tracing/src/main/java/com/example/FactorialControllerAction.java
franciscolopezsancho: of course

franciscolopezsancho:
Although looking at the samples, the best candidate I can see it's Fibonacci action and adding it in there will make it harder to find. At least when looking at the `samples` folder. How about calling it like `java-protobuf-tracing-action` just like fibonacci one? 
On the other hand one can make the case that they are meant to be found (looked at) through the docs alone.

WDYT?

samples/java-protobuf-tracing/.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+	"java.configuration.updateBuildConfiguration": "interactive"
```
aludwiko, line 2: .gitignore?

aludwiko:
Ok, maybe we shouldn't mix things together too much. I don't have any strong preferences. Except that, I would go with `java-protobuf-tracing` for the name, maybe we will add other components to build more complex traces.

samples/java-protobuf-tracing/.vscode/settings.json
franciscolopezsancho: ups

franciscolopezsancho: removed. I'm not actually using the VS anymore

samples/java-protobuf-tracing/docker-compose.yml
```diff
@@ -0,0 +1,26 @@
+
+# If you're looking to use eventing with Google PubSub, to get an emulator running:
+# - add property "-Dkalix.proxy.eventing.support=google-pubsub-emulator" to the JAVA_TOOL_OPTIONS environment map under the kalix-runtime service
+# - uncomment the env var PUBSUB_EMULATOR_HOST and the section below for gcloud-pubsub-emulator service
+version: "3"
+services:
+  kalix-runtime:
+    image: gcr.io/kalix-public/kalix-runtime:1.1.31
```
franciscolopezsancho, line 8: I need to update this

samples/java-protobuf-tracing/README.md
```diff
@@ -0,0 +1,60 @@
...
+
+When running a Kalix service locally, we need to have its companion Kalix Runtime running alongside it.
+
+To start your service locally, run:
+
+```shell
+mvn kalix:runAll
+```
+
+It's worth noting that `application.conf` is passing `kalix.telemetry.tracing.collector-endpoint="http://localhost:4317"`
```
octonato, line 16: What happens if we deploy this application in Kalix? Will it read that property and start to call localhost:4317?

Maybe better to pass it to the mvn process as a `-D` property instead of having it in the `application.conf`. We can also pass `-D` properties in the pom.xm, I believe.

samples/java-protobuf-tracing/README.md
```diff
@@ -0,0 +1,60 @@
...
+```shell
+grpcurl -plaintext localhost:9000  com.example.Controller/CallAsyncEndpoint                                 
+```
+produces
+```
+{
+  "message": "{\n  \"userId\": 1,\n  \"id\": 1,\n  \"title\": \"sunt aut facere repellat provident occaecati excepturi optio reprehenderit\",\n  \"body\": \"quia et suscipit\\nsuscipit recusandae consequuntur expedita et cum\\nreprehenderit molestiae ut ut quas totam\\nnostrum rerum est autem sunt rem eveniet architecto\"\n}"
+}
+```
+or
```
octonato, line 35: This `or` looks strange. 
The next command and output is exactly the same as the first. 

Where you planning to add a REST example here or maybe the `CallSyncEndpoint`?

samples/java-protobuf-tracing/README.md
```diff
@@ -0,0 +1,60 @@
...
+For example, using [`grpcurl`](https://github.com/fullstorydev/grpcurl):
+
+```shell
+grpcurl -plaintext localhost:9000  com.example.Controller/CallAsyncEndpoint                                 
+```
+produces
+```
+{
+  "message": "{\n  \"userId\": 1,\n  \"id\": 1,\n  \"title\": \"sunt aut facere repellat provident occaecati excepturi optio reprehenderit\",\n  \"body\": \"quia et suscipit\\nsuscipit recusandae consequuntur expedita et cum\\nreprehenderit molestiae ut ut quas totam\\nnostrum rerum est autem sunt rem eveniet architecto\"\n}"
+}
```
octonato, line 33: Why this is full of `\` and `\n`?

samples/java-protobuf-tracing/src/main/java/com/example/ControllerAction.java
```diff
@@ -0,0 +1,137 @@
...
+
+    CompletableFuture<HttpResponse<String>> responseFuture = callAsyncService();
+
+    try (Scope scope = span.makeCurrent()) {// <4>
+      responseFuture.thenAccept(response -> {
+        String responseBody = response.body();
+        span.setAttribute("result", responseBody);// <5>
+      }).exceptionally(ex -> {
+        span.setStatus(StatusCode.ERROR, ex.getMessage());
+        return null;
```
octonato, line 125: This will be a problem. 

In case of an error, it will later cause a NPE because we are recovering the failed CompletionState with a successful one but if a null object. 

Instead, we should preserve the failure and later, in `public Effect<ControllerActionApi.MessageResponse> callAsyncEndpoint(Empty empty)`, we need to either return a `effects().reply` or a `effects().error`

samples/java-protobuf-tracing/src/main/java/com/example/ControllerAction.java
```diff
@@ -0,0 +1,137 @@
...
+      }
+    }
+    return effects().reply(ControllerActionApi.MessageResponse.newBuilder().setMessage(result).build());
+  }
+
+  private  HttpResponse<String> callSyncService() throws IOException, InterruptedException {
+    HttpRequest request = HttpRequest.newBuilder()
+            .uri(URI.create(url))
+            .build();
+    return httpClient.send(request, HttpResponse.BodyHandlers.ofString());
```
octonato, line 49: Maybe better not to promote sync calls. We could keep only the async one. 

The reason is that Kalix users are building distributed microservices and async must be central concept all the time.

octonato: Would be different if we were already on Java 21 and supporting Loom.

octonato (commented):
Looking good. 

The same has some issues that we need to address. 
The usage of CompletionStage needs some extra attention.

samples/java-protobuf-tracing/src/main/java/com/example/ControllerAction.java
franciscolopezsancho: Okay. This makes sense. I thought about this but failed to formulate the question. 
I'll delete the sync.

franciscolopezsancho: done

samples/java-protobuf-tracing/src/main/java/com/example/ControllerAction.java
franciscolopezsancho: Good catch. 
One thing. I've been looking at this and the only thing we can do is use `.exceptionally` and inside create a `MessageResponse` with the error or throw our own exception. I understand we can not use `effects().error` once we enter the monad-like `CompletableFuture`. That's why we use the `asyncReply` in the first place. I thought that we should create an `asyncError` but then I started to think that it will be just like `asyncReply`. WDYT?

samples/java-protobuf-tracing/README.md
franciscolopezsancho: Added as a `jvmArgs`. By the way there's not docs about that. How about we create some?

samples/java-protobuf-tracing/README.md
franciscolopezsancho: used now Jackson to parse it and show only the title

samples/java-protobuf-tracing/README.md
franciscolopezsancho: removed.
Yes but as we removed the `Sync` one it's the `or` also removed.

samples/java-protobuf-tracing/README.md
octonato: I don't know. It depends I guess. 

Passing jvmArgs to main class is part of the basic skills set that a developer building a jvm application needs to have. If we need to document each single bit, then the docs becomes bloated with a lot of non-kalix information.

We can add a comment on the pom explaining why we passing the args. Same as in the README file.

What might be useful is a short notice on how to add tracing spans in dev-mode pointing to the sample. I mean, a short notice in the docs section that talks about adding tracing spans. Similar to what we have about enabling ACL in dev-mode.

samples/java-protobuf-tracing/src/main/java/com/example/ControllerAction.java
octonato: So, just for the record. We got a call and things are now much clear.

I was wrong about the NPE. The `null` is never returned because this method is always returning the original `responseFuture`.

The code run inside the try, are just call backs for the successful and failure case.

samples/java-protobuf-tracing/src/main/java/com/example/ControllerAction.java
```diff
@@ -0,0 +1,94 @@
...
+    // tag::create-close-span[]
+    Span span  = tracer 
+            .spanBuilder("https://jsonplaceholder.typicode.com/posts/{}")
+            .setParent(actionContext().metadata().traceContext().asOpenTelemetryContext())// <1>
+            .startSpan(); // <2>
+    span.setAttribute("post", "1");// <3>
+
+    CompletableFuture<HttpResponse<Post>> responseFuture = callAsyncService();
+
+    try (Scope scope = span.makeCurrent()) {// <4>
```
octonato, line 75: @franciscolopezsancho, maybe something more to watch out. 

We will close the scope and end the span as soon as the try/finally block is executed.

But the `thenAccept` and `exceptionally` will happen later, when the future completes. So I guess this will produce unpredictable results.

franciscolopezsancho: So maybe this is way we didn't see the error marker in the span. Good idea. I'll take that into account.

franciscolopezsancho: That was the case yes. We need to move the `span.end()` outside the `finally`, putting in both callbacks.

samples/java-protobuf-tracing/README.md
franciscolopezsancho: I mean that the `jvmArgs` as a property is not a pom.xml standard AFAIK. You can add something similar (with a different name) to the compiler-maven-plugin or the failsafe one for example. As we are using the exec-maven-plugin under the covers I understand that plugin has not `jvmArgs` as such. With that exact name.

franciscolopezsancho: I exec-maven-plugin had such a param we could just state we are using that under the cover. On that I totally agree.

samples/java-protobuf-tracing/docker-compose.yml
franciscolopezsancho: done

samples/java-protobuf-tracing/src/main/java/com/example/ControllerAction.java
aludwiko: Guys, why, I have this feeling that it will not work as expected when we hit it with many concurrent requests, I'm worried about this part:
`try (Scope scope = span.makeCurrent()) {// <4>`

My implementation is a bit different: 
```
Optional<Span> span = startSpan("slack-api-post-new-cfp");
        return callAsyncService().thenApply(response -> {
            span.ifPresent(Span::end);
            return response;
          }
        );
```

```
private Optional<Span> startSpan(String spanName) {
    var otelCurrentContext = actionContext().metadata().traceContext().asOpenTelemetryContext();
    Optional<Span> span = actionContext().getOpenTelemetryTracer().map(tracer -> tracer
      .spanBuilder(spanName)
      .setParent(otelCurrentContext)
      .setSpanKind(SpanKind.CLIENT)
      .startSpan());
    return span;
  }
```

Could also include the `exceptionally` part, but here I don't have to worry about any current otel context.

aludwiko: I added a comment to a resolved conversation, unresolving it.

samples/java-protobuf-tracing/src/main/java/com/example/ControllerAction.java
octonato: What about the scope? Why are opening the scope in a try-with-resources? 

The resource is closed as soon as the try block ends. Are we possibly add the span on the wrong scope? 

Maybe the issue reported by Eduardo is similar to this. 
https://github.com/lightbend/kalix-runtime/issues/2369#issuecomment-2014894746

A span is added in the wrong scope because of some async execution order.

samples/java-protobuf-tracing/src/main/java/com/example/ControllerAction.java
aludwiko: I see the changes now; it's better, but still, I would remove the `try` statement and simplify this code.

samples/java-protobuf-tracing/README.md
octonato: It's not clear to me what you want to document exactly. Are you referring to the fact that someone can pass any `-D` args to the exec-maven-plugin? 

My point was more about the sample where we explain why we are passing this param in the pom.

And then, in the docs page about tracing, we should mention that one needs to add an extra container (jaeger) and let Kalix know about it.

By the way, I just realised that the same property needs to be added to the runtime container. So maybe we should not add anything to the pom.xml at all. 

The kalix plugin is already able to read the docker-compose file and pick the `-D` properties that we pass to the runtime and pass it to the jvm process as well. 
We don't do it for all the properties though, only for the service mappings. So instead, we could do the same for this new property and let the plugin solve it for the user automatically. 

This has the advantage that if you start kalix from inside the IDE (for example when debugging), it will also read the docker-compose file and pick the `-D` property.

octonato: This is the dev-mode class that reads the docker-compose and add extra properties to the jvm process.
https://github.com/lightbend/kalix-jvm-sdk/blob/027d68229b5ad9b0c2284397b2b7bb5102ff052c/devtools/src/main/scala/kalix/devtools/impl/DevModeSettings.scala#L64

This is also used when starting from inside the IDE

octonato: Oh, I'm wrong about it. It's not the same config. 

Hmm, the problem now is that when debugging the application inside the IDE, we won't have access to this property and telemetry will be out of the game. Unless the user explicitly pass it to their debug configuration inside the IDE.

franciscolopezsancho: darn. Well, we could provide this for now as suggested https://github.com/lightbend/kalix-jvm-sdk/pull/2051#discussion_r1538988745 and then  in another PR try to add this to the debug.

samples/java-protobuf-tracing/src/main/java/com/example/ControllerAction.java
franciscolopezsancho: Let me give it a try. No pun intended.

franciscolopezsancho: I'm gladly surprised. I did try this a couple of times without the `makeCurrent()` and couldn't make it work. Looked at the docs and didn't understand why. In any case it just works. So yeah, much better without it. Not sure it would have an impact because it was just writing but nevertheless, in the best possible scenario it was noise. 

Removed now.

samples/java-protobuf-tracing/README.md
octonato: That won't work as suggested because the property is not the same. Nor the host. 
I think we could go now with the property in pom.xml and then in another PR try to improve the DX by reading the docker-compose file and transforming the telemetry property we pass to the runtime to something the user service can consume. 

But that should be in another PR. 

And we don't need to document anything for now, because if we achieve good DX in follow-up PR, nothing needs to be documented. 😄

docs/src/modules/java-protobuf/pages/actions.adoc
```diff
@@ -449,6 +449,43 @@ include::example$java-protobuf-valueentity-counter/src/main/java/com/example/act
...
+
+[.tabset]
+Java::
++
+[source,java,indent=0]
+.src/main/java/com/example/ControllerAction.java
+----
+include::example$java-protobuf-tracing/src/main/java/com/example/ControllerAction.java[tag=create-close-span]
+----
+<1> setting the action's TraceContext as the parent of this span. Linking the action's trace with this span.
```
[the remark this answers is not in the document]
franciscolopezsancho: I decided for `Sets` in all of them.

samples/java-protobuf-tracing/docker-compose.yml
```diff
@@ -0,0 +1,26 @@
...
+    ports:
+      - "9000:9000"
+    extra_hosts:
+      - "host.docker.internal:host-gateway"
+    environment:
+      JAVA_TOOL_OPTIONS: >
+        -Dkalix.proxy.telemetry.tracing.enabled=true
+        -Dkalix.proxy.telemetry.tracing.collector-endpoint=http://jaeger:4317
+      USER_SERVICE_HOST: ${USER_SERVICE_HOST:-host.docker.internal}
+      USER_SERVICE_PORT: ${USER_SERVICE_PORT:-8080}
```
octonato, line 19: Check out how we are now making the docker-compose file. Variables go to .env file.

samples/java-protobuf-tracing/src/main/java/com/example/ControllerAction.java
```diff
@@ -0,0 +1,94 @@
...
+  HttpClient httpClient = HttpClient.newHttpClient();
+  public ControllerAction(ActionCreationContext creationContext) {}
+
+
+  @Override
+  public Effect<ControllerActionApi.MessageResponse> callAsyncEndpoint(Empty empty) {
+      // tag::get-tracer[]
+      Optional<Tracer> tracerOpt = actionContext().getOpenTelemetryTracer();
+      // end::get-tracer[]
+      CompletableFuture<HttpResponse<Post>> futureResponse;
```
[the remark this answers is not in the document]
franciscolopezsancho: It should be good now

docs/src/modules/java-protobuf/pages/actions.adoc
```diff
@@ -449,6 +449,43 @@ include::example$java-protobuf-valueentity-counter/src/main/java/com/example/act
...
+[.tabset]
+Java::
++
+[source,java,indent=0]
+.src/main/java/com/example/ControllerAction.java
+----
+include::example$java-protobuf-tracing/src/main/java/com/example/ControllerAction.java[tag=create-close-span]
+----
+<1> Sets the action's TraceContext as the parent of this span. Linking the action's trace with this span.
+<2> Creating and starting the span.
```
aludwiko, line 481: If Sets, and Creates etc.

docs/src/modules/java/pages/publishing-subscribing.adoc
```diff
@@ -241,7 +241,7 @@ include::example$java-spring-eventsourced-counter/src/it/java/com/example/Counte
 ----
 <1> Build a `CloudEvent` object with the 3 required attributes, respectively: `id`, `source` and `type`.
 <2> Add the subject to which the message is related, that is the `counterId`.
-<3> Set the mandatory header "Content-Type" accordingly.
+<3> Sets the mandatory header "Content-Type" accordingly.
```
aludwiko, line 244: If we decide to change it, I will do this in a separate PR and change all verbs in all bullet point lists.

samples/java-protobuf-tracing/src/main/java/com/example/ControllerAction.java
```diff
@@ -0,0 +1,87 @@
...
+
+  @Override
+  public Effect<ControllerActionApi.MessageResponse> callAsyncEndpoint(Empty empty) {
+    // tag::get-tracer[]
+    Optional<Tracer> tracerOpt = actionContext().getOpenTelemetryTracer();
+    // end::get-tracer[]
+    CompletableFuture<HttpResponse<Post>> futureResponse;
+
+    if(tracerOpt.isPresent()) {
+      futureResponse = callAsyncService(tracerOpt.get());
```
aludwiko, line 38: I know it's a matter of taste, but can't we just include the `CompletableFuture<HttpResponse<Post>> callAsyncService(Tracer tracer)` code here and remove `futureResponse`.  Instead of using `Optional<Trace>` we can use `Optional<Span>` as suggested in a different command and be more FP. WDYT?

aludwiko: A link: https://github.com/lightbend/kalix-jvm-sdk/pull/2051#discussion_r1538939313 creating the span might be in the same method.

docs/src/modules/java/pages/publishing-subscribing.adoc
franciscolopezsancho: Sure, I'll leave for to another PR.

docs/src/modules/java-protobuf/pages/actions.adoc
franciscolopezsancho: Sure, I'll leave for to another PR.

franciscolopezsancho: done

docs/src/modules/java/pages/publishing-subscribing.adoc
franciscolopezsancho: back to previous.

samples/java-protobuf-tracing/docker-compose.yml
franciscolopezsancho: .env added.

samples/java-protobuf-tracing/src/main/java/com/example/ControllerAction.java
franciscolopezsancho: Certainly is more elegant the way you propose. I like it very much. At least for a span without attributes. What I'm no so sure is how it will look when starting to add the attributes to the span. One `ifPresent` for each? On that note, I had a discussion with @octonato yesterday about providing a Tracer that can be a NoOpsTracer when tracing is disabled, so the user doesn't have to deal with all this mess of having two flows. `IfPresent` namely in any of the two flavours, the current one or the more functional you propose.
All that to say that I rather leave it the way it is until I add the NoOps Tracer.

aludwiko: No, you need a single if present, here my impl:
```
  @Override
  public Effect<ControllerActionApi.MessageResponse> callAsyncEndpoint(Empty empty) {
    // tag::get-tracer[]
    Optional<Tracer> tracerOpt = actionContext().getOpenTelemetryTracer();
    // end::get-tracer[]

    Optional<Span> span = tracerOpt.map(tracer -> {
      return tracer
        .spanBuilder("https://jsonplaceholder.typicode.com/posts/{}")
        .setParent(actionContext().metadata().traceContext().asOpenTelemetryContext())// <1>
        .startSpan() // <2>
        .setAttribute("post", "1");// <3>
    });

    CompletableFuture<ControllerActionApi.MessageResponse> asyncComputation = callAsyncService()
      .whenComplete((response, ex) -> {
        if (ex != null) {
          span.ifPresent(presentSpan ->
            presentSpan
              .setStatus(StatusCode.ERROR, ex.getMessage())// <6>
              .end());// <5>
        } else {
          span.ifPresent(presentSpan ->
            presentSpan
              .setAttribute("result", response.body().title)// <4>
              .end());// <5>
        }
      })
      .thenApply(response ->
        ControllerActionApi.MessageResponse.newBuilder().setMessage(response.body().title).build()
      );
    return effects().asyncReply(asyncComputation);
  }
```

Java Futures API is awful, but I think in this case `whenComplete` will do the job better.

I also like the idea of returning `NoOps Tracer` but this shouldn't block us with decisions about the current implementation.

franciscolopezsancho: I can't say no with this beautiful snippet. I'll put it in. 👍 

PS. Two `.ifPresent` :)

franciscolopezsancho:
For the record after the talk with Renato:
to change in the DevModeSettings
- Use only Port[Int]
- add rest to .else when no docker around
- More tests around TracingConfExctractor
- Picking the docker lines with a more performant algorithm

devtools/src/main/scala/kalix/devtools/impl/TracingConfExtractor.scala
```diff
@@ -0,0 +1,57 @@
...
+
+package kalix.devtools.impl
+
+import com.typesafe.config.{ ConfigFactory, ConfigParseOptions, ConfigSyntax }
+
+import scala.util.matching.Regex
+
+object TracingConfExtractor {
+
+  val PortPattern: Regex = """(\d{1,5})""".r
```
octonato, line 25: A better pattern would be, from 1 to 5 digits, occurring after a `:`.

devtools/src/test/scala/kalix/devtools/impl/TracingConfExtractorSpec.scala
```diff
@@ -0,0 +1,67 @@
...
+        |    image: gcr.io/kalix-public/kalix-runtime:1.1.34
+        |    container_name: tracing
+        |    ports:
+        |      - "9000:9000"
+        |    extra_hosts:
+        |      - "host.docker.internal:host-gateway"
+        |    environment:
+        |      JAVA_TOOL_OPTIONS: >
+        |        -Dmyproperty=false -Dkalix.proxy.telemetry.tracing.collector-endpoint=http://jaeger:4317
+        |        -Dkalix.proxy.telemetry.tracing.enabled=true
```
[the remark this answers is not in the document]
octonato: Some test suggesting to add

```scala
"extract none if tracing is enable, but port is not set" in {
  pending
}

"extract none if tracing is enable, but port is invalid (negative)" in {
  pending
}

"extract none if tracing is enable, but port out of range" in {
  pending
}
```

devtools/src/main/scala/kalix/devtools/impl/TracingConfExtractor.scala
```diff
@@ -0,0 +1,57 @@
...
+ * limitations under the License.
+ */
+
+package kalix.devtools.impl
+
+import com.typesafe.config.{ ConfigFactory, ConfigParseOptions, ConfigSyntax }
+
+import scala.util.matching.Regex
+
+object TracingConfExtractor {
```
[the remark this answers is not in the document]
franciscolopezsancho: makes sense. done.

devtools/src/test/scala/kalix/devtools/impl/TracingConfExtractorSpec.scala
```diff
@@ -0,0 +1,67 @@
...
+        |      # Comment to enable ACL check in dev-mode (see https://docs.kalix.io/services/using-acls.html#_local_development_with_acls)
+        |      ACL_ENABLED: "false"
+        |  jaeger:
+        |    image: jaegertracing/all-in-one:1.54
+        |    ports:
+        |      - 4317:4317
+        |      - 16686:16686
+        |""".stripMargin
+
+    "extract if tracing is enabled 2" in {
```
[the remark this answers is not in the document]
franciscolopezsancho: added proper name

devtools/src/main/scala/kalix/devtools/impl/TracingConfExtractor.scala
franciscolopezsancho: added

devtools/src/main/scala/kalix/devtools/impl/TracingConfExtractor.scala
```diff
@@ -0,0 +1,57 @@
...
+
+    dockerComposeUtilsLines.filter(_.contains("-D")).foreach { line =>
+      line.replace(System.lineSeparator(), " ").split("-D").foreach { segment =>
+        segment.split(" ").foreach {
+          case each if each.startsWith(DevModeSettings.tracingConfigEnabled) =>
+            enabled = ConfigFactory.parseString(each).getBoolean(DevModeSettings.tracingConfigEnabled)
+          case each if each.startsWith(DevModeSettings.tracingConfigEndpoint) =>
+            val url = ConfigFactory
+              .parseString(each, ConfigParseOptions.defaults().setSyntax(ConfigSyntax.PROPERTIES))
+              .getString(DevModeSettings.tracingConfigEndpoint)
```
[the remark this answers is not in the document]
franciscolopezsancho: `ConfigSyntax.PROPERTIES` is needed to be able to read the `collector-endpoint`. It has semicolon and as JSON that's not permitted if not between double quotes.

devtools/src/test/scala/kalix/devtools/impl/TracingConfExtractorSpec.scala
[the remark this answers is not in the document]
franciscolopezsancho: added

devtools/src/test/scala/kalix/devtools/impl/TracingConfExtractorSpec.scala
```diff
@@ -0,0 +1,67 @@
...
+      TracingConfExtractor.unapply(Seq(defaultFile)) shouldBe Some(4317)
```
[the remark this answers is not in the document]
franciscolopezsancho: done.

devtools/src/main/scala/kalix/devtools/impl/TracingConfExtractor.scala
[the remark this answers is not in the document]
franciscolopezsancho: In any case, you are right and `PortPattern.findFirstIn(line)` was enough. Nice.

devtools/src/main/scala/kalix/devtools/impl/TracingConfExtractor.scala
```diff
@@ -0,0 +1,57 @@
...
+
+  val PortPattern: Regex = """(\d{1,5})""".r
+
+  /**
+   * Extracts the port of the `collector-endpoint` from the docker file (after substituting any variable) if tracing is
+   * enabled.
+   * @param dockerComposeUtilsLines
+   * @return
+   */
+  def unapply(dockerComposeUtilsLines: Seq[String]): Option[Int] = {
```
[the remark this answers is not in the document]
franciscolopezsancho: done.

devtools/src/main/scala/kalix/devtools/impl/DevModeSettings.scala
```diff
@@ -16,17 +16,34 @@
...
+  val tracingConfigEndpoint = "kalix.proxy.telemetry.tracing.collector-endpoint"
 
   def fromConfig(config: Config): DevModeSettings = {
+    val devModeSettingsAddedPorts = addPortsFromConfig(config)
+    addTracingConf(config, devModeSettingsAddedPorts)
+  }
 
+  private def addTracingConf(config: Config, current: DevModeSettings): DevModeSettings = {
+    val tracingPort: Option[Int] =
+      if (config.hasPath(DevModeSettings.tracingConfigEndpoint)) {
```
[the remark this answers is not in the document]
franciscolopezsancho: Right. I see now. 

Question. Wouldn't it be better to pass to the `WebClientProviderImpl` the config that the `KalixRunner` already is aware of after using `addDevModeConfig` than extract the settings again in the `WebClientProviderImpl.devModeSettings`? 

I mean when creating the `webClientProvider` in  `KalixSpringApplication.webClientProvider` don't we have the host and the ports already because we have run `KalixRunner.loadPreparedConfig()`?

Not even sure I'm right, though.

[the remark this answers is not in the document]
franciscolopezsancho: removed.

samples/java-protobuf-tracing/.env
```diff
@@ -0,0 +1,2 @@
+USER_SERVICE_HOST=host.docker.internal
+USER_SERVICE_PORT=8080
```
octonato, line 2: Better to copy the .env from other samples. 

This file contains not only the main variables, but also explanations about each variable.

samples/java-protobuf-tracing/docker-compose.yml
octonato, line 1: Same here. Please use the most up-to-date docker-compose.yml we are using in the samples. Have a look at what we have in the other samples. 

It should be the same except for the jaeger extra settings.

octonato (approved):
LGTM

We need to align the `.env` and `docker-compose.yml` with what we have in the other samples. 

Feel free to merge as is and align the files in follow-up PR.

samples/java-protobuf-tracing/docker-compose.yml
franciscolopezsancho: duly noted

samples/java-protobuf-tracing/.env
franciscolopezsancho: sure

PR #2052, opened 2024-02-29
chore: use docker compose instead of docker-compose

girdharshubham:


franciscolopezsancho:
LGTM while I have an ask. 
Could you try removing your docker-compose binary? then go to the root of the `kalix-jvm-sdk` and `.publishLocally.sh`? So you build locally the SDK and later use that version in your project? Should take a handful of minutes. 
I did that on mac and it works, as expected.

octonato (approved):

octonato:
LGTM, but I didn't understand @franciscolopezsancho request (or its motivation), so I'm waiting before merging.

girdharshubham:
I think he means that I should test it locally first.

girdharshubham:
I tried it by swapping the kalix sbt plugin with one built locally and then a `runAll`. Looks to be running

efgpinto:
This will work out-of-the-box for anyone using Docker Desktop.

However, for someone like me using Colima and brew, I needed to follow [this advice](https://github.com/abiosoft/colima/discussions/874).

Just leaving this info here for future reference in case we get any user reporting a similar issue.

PR #2053, opened 2024-03-04
chore: update javadoc to reflect implementation

octonato:
Leftover from https://github.com/lightbend/kalix-jvm-sdk/pull/1562

We moved back to full-qualified class name but didn't change the javadoc for the TypeName annotation.

Also removed the word 'persisted' from the first line. It's not only about persisted events, but also events coming from a topic.

### DIFF
--- a/sdk/java-sdk-spring/src/main/java/kalix/javasdk/annotations/TypeName.java
+++ b/sdk/java-sdk-spring/src/main/java/kalix/javasdk/annotations/TypeName.java
@@ -22,10 +22,10 @@ import java.lang.annotation.RetentionPolicy;
 import java.lang.annotation.Target;
 
 /**
- * Annotation to assign a logical type name to persisted events.
+ * Annotation to assign a logical type name to events.
  *
  * Kalix needs to identify each event in order to deliver them to the right event handlers.
- * If a logical type name isn't specified, Kalix will use the non-qualified class name.
+ * If a logical type name isn't specified, Kalix will use the full-qualified class name.
  *
  * Once an event is persisted in Kalix, you won't be able to rename your class if no logical type name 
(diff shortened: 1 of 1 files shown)


sdk/java-sdk-spring/src/main/java/kalix/javasdk/annotations/TypeName.java
```diff
@@ -22,10 +22,10 @@
 import java.lang.annotation.Target;
 
 /**
- * Annotation to assign a logical type name to persisted events.
+ * Annotation to assign a logical type name to events.
  *
  * Kalix needs to identify each event in order to deliver them to the right event handlers.
- * If a logical type name isn't specified, Kalix will use the non-qualified class name.
+ * If a logical type name isn't specified, Kalix will use the full-qualified class name.
```
pvlugter, line 28: Usually `fully qualified`. Could be with or without the hyphen.

```suggestion
 * If a logical type name isn't specified, Kalix will use the fully qualified class name.
```

sdk/java-sdk-spring/src/main/java/kalix/javasdk/annotations/TypeName.java
```diff
@@ -37,7 +37,7 @@
 public @interface TypeName {
 
   /** Logical type name for the annotated type.
-   * If missing (or defined as Empty String), the non-qualified class name will be used.
+   * If missing (or defined as Empty String), the full-qualified class name will be used.
```
pvlugter, line 40:
```suggestion
   * If missing (or defined as Empty String), the fully qualified class name will be used.
```

pvlugter (approved):
LGTM, with `full` -> `fully`.

PR #2055, opened 2024-03-05
fix: cloud events - convert kafka prefix

octonato:
Convert events are using `ce_` prefix, we convert to `ce-`

### DIFF
--- a/sdk/java-sdk-protobuf/src/main/scala/kalix/javasdk/impl/MetadataImpl.scala
+++ b/sdk/java-sdk-protobuf/src/main/scala/kalix/javasdk/impl/MetadataImpl.scala
@@ -43,7 +43,7 @@ import java.util.Optional
 import scala.compat.java8.OptionConverters._
 import scala.jdk.CollectionConverters._
 
-private[kalix] class MetadataImpl(val entries: Seq[MetadataEntry]) extends Metadata with CloudEvent {
+private[kalix] class MetadataImpl private (val entries: Seq[MetadataEntry]) extends Metadata with CloudEvent {
 
   override def has(key: String): Boolean = entries.exists(_.key.equalsIgnoreCase(key))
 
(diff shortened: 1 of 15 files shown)


franciscolopezsancho:
I'm curious. Why so?

franciscolopezsancho:
Right. Now I know (after the standup)

sdk/java-sdk-protobuf/src/main/scala/kalix/javasdk/impl/MetadataImpl.scala
```diff
@@ -291,4 +302,17 @@ object MetadataImpl {
...
+  def of(entries: Seq[MetadataEntry]): MetadataImpl = {
+    val transformedEntries =
+      entries.map { entry =>
+        // is incoming ce key in one of the alternative formats?
+        // if so, convert key to our internal default key format
+        alternativeKeyFormats.get(entry.key) match {
+          case Some(defaultKey) => MetadataEntry(defaultKey, entry.value)
+          case _                => entry
+        }
+      }
```
octonato, line 314: I changes this implementation. 
Before, I was replacing any key starting with `ce_` with a key starting with `ce-`. 

This new implementation creates a Map for all know keys,like `ce_id` maps to `ce-id`. If we receive a `ce_id`, we transform it. 

If we receive keys starting with `ce_`, but not in the list of known keys, we don't touch them.

sdk/java-sdk-protobuf/src/test/scala/kalix/javasdk/impl/MetadataImplSpec.scala
```diff
@@ -159,10 +159,20 @@ class MetadataImplSpec extends AnyWordSpec with Matchers with OptionValues {
       val mdRedirect = md.withStatusCode(Redirect.MOVED_PERMANENTLY)
       mdRedirect.get("_kalix-http-code").toScala.value shouldBe "301"
     }
+
+    "support creationg with CloudEvents prefixed with ce_" in {
```
franciscolopezsancho, line 163: Shall we add a test with metadata prefixed with `ce-`?

franciscolopezsancho: oh, I see. It was already working for `ce-`.

franciscolopezsancho (approved):
LGTM

PR #2063, opened 2024-03-19
feat: ComponentClient tracing

octonato:
Alternative to https://github.com/lightbend/kalix-jvm-sdk/pull/2046


sdk/java-sdk-spring/src/main/scala/kalix/javasdk/impl/reflection/Reflect.scala
```diff
@@ -50,4 +54,32 @@ object Reflect {
     }
 
   }
+
+  private implicit val stringArrayOrdering: Ordering[Array[String]] =
+    Ordering.fromLessThan(util.Arrays.compare[String](_, _) < 0)
+
+  implicit val methodOrdering: Ordering[Method] =
+    Ordering.by((m: Method) => (m.getName, m.getReturnType.getName, m.getParameterTypes.map(_.getName)))
```
octonato, line 62: Orthogonal to main change in this PR. 
I merged ReflectionUtils with Reflect class.

sdk/java-sdk-spring/src/main/scala/kalix/javasdk/impl/reflection/Reflect.scala
```diff
@@ -50,4 +54,32 @@ object Reflect {
...
+
+  def lookupComponentClientField(instance: Any): List[ComponentClient] = {
```
octonato, line 64: This is the main utility method being added. 
For the moment in use when receiving Action commands, but not yet for Workflow

sdk/java-sdk-spring/src/main/scala/kalix/javasdk/impl/action/ReflectiveActionRouter.scala
```diff
@@ -50,23 +51,30 @@ class ReflectiveActionRouter[A <: Action](
...
-              .invoke(action, invocationContext)
-              .asInstanceOf[Action.Effect[_]]
-        }
-      case None if ignoreUnknown => ActionEffectImpl.Builder.ignore()
-      case None =>
-        throw new NoSuchElementException(
-          s"Couldn't find any method with input type [$inputTypeUrl] in Action [$action].")
-
+    // lookup ComponentClient
+    val maybeComponentClient = Reflect.lookupComponentClientField(action.getClass, action)
```
octonato, line 55: There is no test for this, but @efgpinto's [PR](https://github.com/lightbend/kalix-jvm-sdk/pull/2026/files) will add a sample that will then cover it.

sdk/java-sdk-spring/src/main/java/kalix/javasdk/client/ComponentClient.java
```diff
@@ -40,10 +42,20 @@ public class ComponentClient {
 
   private final KalixClient kalixClient;
 
+  private Optional<Metadata> callMetadata = Optional.empty();
+
   public ComponentClient(KalixClient kalixClient) {
     this.kalixClient = kalixClient;
   }
 
+  public void setCallMetadata(Metadata callMetadata) {
```
efgpinto, line 51: Should we document that this is intended for internal use? or is it ok for users to use this? if so, having this and the `.withMetadata` after building the deferred call might be confusing...

efgpinto (commented):
This LGTM with a small question. I will just try to run my sample against this to test before ✅ .

sdk/java-sdk-spring/src/main/java/kalix/javasdk/client/ComponentClient.java
octonato: Right. Actually this is a oversight of mine. I was planning to make it internal API. 

On the other hand, now I'm wondering what is the impact for using `withMetadata` in the `DeferredCall`. In both impls, we are setting the metadata, but we probably want to copy the tracing over, if any.

octonato:
@efgpinto, a few new things in this PR.

* made ComponentClient an interface and moved impl to ComponentClientImpl
* moved KalixClient to a `impl` package and undeprecated it
* integrated Cisco's PR (https://github.com/lightbend/kalix-jvm-sdk/pull/2022)

octonato:
Maybe good to review it commit by commit. 

First one is about setting the metadata on the client (bf5fe2e0e4f2ffe47b5ea67cae540638fc944719)
Second commit is integrating Cisco's PR (d4397b0d34f551536864bde0aba9680b8224569c)
Third is about moving the KalixClient to internal package (9f9d956ae4d322f071306a8087e8964fff17870a)

Last is just about formatting.

sdk/java-sdk-spring/src/main/java/kalix/javasdk/client/ActionCallBuilder.java
```diff
@@ -16,194 +16,178 @@
...
-import akka.japi.function.Function21;
-import akka.japi.function.Function22;
-import akka.japi.function.Function3;
-import akka.japi.function.Function4;
-import akka.japi.function.Function5;
-import akka.japi.function.Function6;
-import akka.japi.function.Function7;
-import akka.japi.function.Function8;
-import akka.japi.function.Function9;
+import akka.japi.function.*;
```
aludwiko, line 19: I think we should avoid them.

sdk/java-sdk-spring/src/main/scala/kalix/javasdk/impl/reflection/Reflect.scala
```diff
@@ -50,4 +55,32 @@ object Reflect {
...
 
   }
+
+  private implicit val stringArrayOrdering: Ordering[Array[String]] =
+    Ordering.fromLessThan(util.Arrays.compare[String](_, _) < 0)
+
+  implicit val methodOrdering: Ordering[Method] =
+    Ordering.by((m: Method) => (m.getName, m.getReturnType.getName, m.getParameterTypes.map(_.getName)))
+
+  def lookupComponentClientField(instance: Any): List[ComponentClientImpl] = {
```
aludwiko, line 65: why is this a list of ComponentClients? I would expect only a single one or none.

octonato: Me too, but we have not guarantee that how many component client a user will try to inject.

sdk/java-sdk-spring/src/main/java/kalix/javasdk/client/ActionCallBuilder.java
octonato: That was not intentional, I will revert it.

octonato: Done

sdk/java-sdk-spring/src/main/scala/kalix/javasdk/impl/action/ReflectiveActionRouter.scala
```diff
@@ -50,23 +51,33 @@ class ReflectiveActionRouter[A <: Action](
...
-        }
-      case None if ignoreUnknown => ActionEffectImpl.Builder.ignore()
-      case None =>
-        throw new NoSuchElementException(
-          s"Couldn't find any method with input type [$inputTypeUrl] in Action [$action].")
-
+    // lookup ComponentClient
+    val componentClients =
+      Reflect
+        .lookupComponentClientField(action.getClass, action)
```
efgpinto, line 57: Small bug found when trying this with the sample..

```suggestion
        .lookupComponentClientField(action)
```

sdk/java-sdk-spring/src/main/scala/kalix/javasdk/impl/action/ReflectiveActionRouter.scala
```diff
@@ -76,21 +87,31 @@ class ReflectiveActionRouter[A <: Action](
...
-        val response = methodInvoker.invoke(action, context).asInstanceOf[Flux[Action.Effect[_]]]
-        Source.fromPublisher(response)
-      case None if ignoreUnknown => Source.empty()
-      case None =>
-        throw new NoSuchElementException(
-          s"Couldn't find any method with input type [$inputTypeUrl] in Action [$action].")
+    // lookup ComponentClient
+    val componentClients =
+      Reflect
+        .lookupComponentClientField(action.getClass, action)
```
efgpinto, line 93: Small bug found when trying this with the sample..

```suggestion
        .lookupComponentClientField(action)
```

efgpinto (commented):
Partial review with small bug found and a typo

sdk/java-sdk-spring/src/main/scala/kalix/javasdk/impl/action/ReflectiveActionRouter.scala
octonato: I changed the signature from (Class, Any) to (Any), but how is it possible that the compiler didn't pick it?

efgpinto: The other method receives `Any` and this was considered a `Tuple2`.. so it worked. I was just not sure if this was because of some implicit lying around or if it doesn't even need that.. 🤔

octonato: Yeah, I was about to ask which error you saw. 
To be honest, I thought this kind of language quirk was already out of 2.13, but apparently not. 

Or we probably had a compiler warning, but didn't see because we have some many already.

By the way, I working on a branch that will make warnings fatal. Like we do in Akka and Kalix Runtime.

octonato:
We still had two tests using KalixClient directly. CI should pass now. 

I also changed the error message when someone tries to inject a 'illegal' dependency. Error was only showing the simple class name. Better to use a FQCN so users can easily find the offending class.

efgpinto:
Found another small issue.. The `traceparent` / `tracestate` don't seem to be injected automatically when executing (`execute()`) the deferred call.

octonato:
> Found another small issue.. The `traceparent` / `tracestate` don't seem to be injected automatically as headers when `.execute()`ing the deferred call.

@franciscolopezsancho, can you have a look at this? Seems to be something that was missing/broken in #2022.

franciscolopezsancho:
> Found another small issue.. The `traceparent` / `tracestate` don't seem to be injected automatically as headers when `.execute()`ing the deferred call.

@efgpinto Looking at the code seems fine but obviously you found something. Could you share your reproducer? if any?

franciscolopezsancho:
LGTM. I can create an example to test it all out if you'd like.

PR #2069, opened 2024-04-02
Protobuf scala span creation

franciscolopezsancho:
Adding example on how to create spans in the scala-protobuf

Depends on https://github.com/lightbend/kalix-jvm-sdk/pull/2051 . Specifically because `DevModeSettings` to pass the params to the user function needed to push traces to jaeger in a local environment.


samples/scala-protobuf-tracing/src/main/scala/com/example/ControllerAction.scala
```diff
@@ -0,0 +1,60 @@
...
+  implicit val formats: Formats = DefaultFormats
+
+  val url = "https://jsonplaceholder.typicode.com/posts"
+
+  // tag::create-close-span[]
+  override def callAsyncEndpoint(empty: Empty): Action.Effect[MessageResponse] = {
+    // tag::get-tracer[]
+    val tracerOpt = actionContext.getOpenTelemetryTracer
+    // end::get-tracer[]
+    val span = tracerOpt.map(_.spanBuilder(s"$url/{}")
```
aludwiko, line 25: Either `Opt` postfix for all or none.

samples/scala-protobuf-tracing/.env
```diff
@@ -0,0 +1,2 @@
+USER_SERVICE_HOST=host.docker.internal
+USER_SERVICE_PORT=8080
```
octonato, line 2: Same as in the other PR. We need to align the files.

samples/scala-protobuf-tracing/src/main/scala/com/example/ControllerAction.scala
```diff
@@ -0,0 +1,60 @@
...
+            .end())// <5>
+      }
+     effects.asyncReply(responseBody)
+  }
+  // end::create-close-span[]
+
+
+  private def callAsync(): Future[MessageResponse] = {
+    val request = basicRequest.get(uri"${url}/1")
+    val response: Future[Response[Either[String, String]]] = request.send(Main.backend)
```
octonato, line 48: is `Future[Response[Either[String, String]]]` being define by sttp?

sdk/scala-sdk-protobuf-testkit/src/main/scala/kalix/scalasdk/testkit/impl/TestKitActionContext.scala
```diff
@@ -33,5 +34,5 @@ final class TestKitActionContext(
 
   override def eventSubject = metadata.get("ce-subject")
   override def getGrpcClient[T](clientClass: Class[T], service: String): T = getComponentGrpcClient(clientClass)
-
+  override def getOpenTelemetryTracer: Option[Tracer] = None
```
octonato, line 37: So, this API doesn't exist yet in the Scala SDK? 
Maybe better to go for the Noops one we discussed the other day and not return Option.

samples/scala-protobuf-tracing/docker-compose.yml
octonato, line 1: same here

Needs to align with template and samples.

sdk/scala-sdk-protobuf-testkit/src/main/scala/kalix/scalasdk/testkit/impl/TestKitActionContext.scala
franciscolopezsancho: we have option in the java-proto. How about we use the same here and then we change to NoOps in both? or you think is better we change this now?

octonato: Better to not add an API that we plan to deprecate. 

But we need to decide if the NoOps is what we want. At first, it sounds like a good idea. But if we take that path, we will need to deprecate the method in the java sdk. Not a big issue, but better to do it now than later. 

If we decide to deprecate it, we need to come up with a new method name and then better to also use that new name here.

franciscolopezsancho: Ok. I'll go for the NoOps now. If you have any objections please let me know.

samples/scala-protobuf-tracing/.env
franciscolopezsancho: done

samples/scala-protobuf-tracing/src/main/scala/com/example/ControllerAction.scala
franciscolopezsancho: yeah, they have an  `implicit class sttp.client4.RichRequestTEither` which I imported

octonato: I ask because it feels super verbose. Shouldn't it be only `Future[Response[String]]`? 

I assume that `Response` is already a type that can express success or failure. Having it hold an `Either` feels overkill. But I might be wrong. I'm not familiar with sttp and I'm sure they added this API for a reason. 

On the other hand, if they have some other implicit that could simplify this return type, it will be better for a sample.

sdk/java-sdk-protobuf/src/main/scala/kalix/javasdk/impl/telemetry/Telemetry.scala
```diff
@@ -244,5 +236,5 @@ private object NoOpInstrumentation extends Instrumentation {
 
   override def buildSpan(service: Service, command: ActionCommand): Option[Span] = None
 
-  override def getTracer(): Tracer = null
+  override def getTracer: Tracer = OpenTelemetry.noop().getTracer("noop")
```
octonato, line 239: For the Noop one, it will be good if log from time to time that user is using a noops one. 

We should not log it each time, but let's say on every 100 calls to `getTracer`. If tracing is not configured, we write a log warning. As such users become aware that they are using a feature that is not enabled.

octonato (commented):
I will review more in detail later. 

I did a first pass though.

sdk/java-sdk-protobuf/src/main/scala/kalix/javasdk/impl/telemetry/Telemetry.scala
franciscolopezsancho: We have such a logging at the start of telemetry but is `.debug` level. I can put info there. Would that suffice you think?

sdk/java-sdk-protobuf/src/main/scala/kalix/javasdk/impl/telemetry/Telemetry.scala
```diff
@@ -235,7 +227,7 @@ private final class TraceInstrumentation(
   }
 
   // TODO: should this be specific per sdk?
-  override def getTracer(): Tracer = openTelemetry.getTracer("java-sdk")
+  override def getTracer: Tracer = openTelemetry.getTracer("java-sdk")
```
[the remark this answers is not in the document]
franciscolopezsancho: good point. I'll think about a better name.

docs/src/modules/java-protobuf/pages/actions.adoc
```diff
@@ -449,29 +449,59 @@ include::example$java-protobuf-valueentity-counter/src/main/java/com/example/act
...
++
+`actionContext()` gives you the link:{attachmentsdir}/scala-api/kalix/scalasdk/action/ActionContext.html[`ActionContext`{tab-icon}, window="new"]. That is, some methods that apply to the context of the action's request.
++
+[source,java,indent=0]
+.src/main/scala/com/example/ControllerAction.java
+----
+include::example$scala-protobuf-tracing/src/main/scala/com/example/ControllerAction.scala[tag=get-tracer]
+----
+
+IMPORTANT: If your tracing is enabled, you will get a tracer that actually creates span and exports data as expected.
```
octonato, line 477:
```suggestion
IMPORTANT: If tracing is enabled, you will get a tracer that actually creates span and exports data as expected.
```

docs/src/modules/java-protobuf/pages/actions.adoc
```diff
@@ -449,29 +449,59 @@ include::example$java-protobuf-valueentity-counter/src/main/java/com/example/act
...
+        -Dkalix.proxy.telemetry.tracing.collector-endpoint=http://jaeger:4317
+    ...
+  jaeger:
+    image: jaegertracing/all-in-one:1.54
+    ports:
+      - 4317:4317
+      - 16686:16686
+---
+
+Here the traces are pushed to a `jaeger` docker image, to the port `4317`. And you can check them out at `http://localhost:16686`.
```
octonato, line 501: We mentioned the JAVA_TOOL_OPTIONS, but we are saying nothing about the jaeger container. 

Maybe be enough. I think it's clear that we are adding another container. 

Anyway, dropping it here just to raise the question. Should we also mention that this extra container needs to be added or it is clear enough?

samples/java-protobuf-tracing/src/main/java/com/example/ControllerAction.java
```diff
@@ -30,29 +22,25 @@ public ControllerAction(ActionCreationContext creationContext) {}
...
               .startSpan() // <2>
               .setAttribute("post", "1");// <3>
-    });
 
     CompletableFuture<ControllerActionApi.MessageResponse> asyncComputation = callAsyncService()
       .whenComplete((response, ex) -> {
         if (ex != null) {
-          span.ifPresent(presentSpan ->
-            presentSpan
+          span
```
octonato, line 37: much cleaner with Noops tracer. 
💯

samples/scala-protobuf-tracing/project/build.properties
```diff
@@ -0,0 +1 @@
+sbt.version=1.9.9
```
octonato, line 1: I didn't even know that 1.9.9 existed. 

Maybe in follow-up PR we can upgrade the other samples.

samples/scala-protobuf-tracing/project/plugins.sbt
```diff
@@ -0,0 +1,4 @@
+addSbtPlugin("io.kalix" % "sbt-kalix" % "1.4.1")
```
octonato, line 1:
```suggestion
addSbtPlugin("io.kalix" % "sbt-kalix" % System.getProperty("kalix-sdk.version", "1.4.1"))
```
Like in the other samples.

sdk/java-sdk-protobuf/src/main/java/kalix/javasdk/action/ActionCreationContext.java
```diff
@@ -46,5 +46,8 @@ public interface ActionCreationContext extends Context {
    *
    * @return A tracer for the current action, if tracing is configured.
    */
+  @Deprecated(since = "1.4.2")
   Optional<Tracer> getOpenTelemetryTracer();
+
+  Tracer getTracer();
```
octonato, line 52:
```suggestion
/**
   * Get an OpenTelemetry tracer for the current action. This will allow for building and automatic
   * exporting of spans.
   *
   * @return A tracer for the current action, if tracing is configured. Otherwise, a noops tracer.
   */
  Tracer getTracer();
```

sdk/java-sdk-protobuf/src/main/java/kalix/javasdk/action/ActionCreationContext.java
```diff
@@ -46,5 +46,8 @@ public interface ActionCreationContext extends Context {
    *
    * @return A tracer for the current action, if tracing is configured.
    */
+  @Deprecated(since = "1.4.2")
```
octonato, line 49: In the java docs, we can add the following. Above the return.

```java
 * @deprecated use getTracer() instead.
 * 
 * @return A tracer for the current action, if tracing is configured.
```

sdk/java-sdk-protobuf/src/main/scala/kalix/javasdk/impl/telemetry/Telemetry.scala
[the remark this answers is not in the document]
octonato: what about `kalix`?

sdk/java-sdk-protobuf/src/main/scala/kalix/javasdk/impl/telemetry/Telemetry.scala
octonato: Fine as debug when we start telemetry. We should not make it a warning at that moment. 

But we should warn the user if they call the `getTracer` method when tracing is disable. Because in that case it's an indication that they are trying to use a disabled feature.

octonato: Btw, it can be on a follow up PR. Let's not feature creep this one.

octonato (approved):
LGTM

A few comments to address before merging. Some comments can go on follow-up PRs.

samples/java-protobuf-tracing/src/main/java/com/example/ControllerAction.java
franciscolopezsancho: :)

sdk/java-sdk-protobuf/src/main/scala/kalix/javasdk/impl/telemetry/Telemetry.scala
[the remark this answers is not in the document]
franciscolopezsancho: sounds about right!

sdk/java-sdk-protobuf/src/main/scala/kalix/javasdk/impl/telemetry/Telemetry.scala
franciscolopezsancho: added https://github.com/lightbend/kalix-jvm-sdk/issues/2074

samples/scala-protobuf-tracing/project/build.properties
franciscolopezsancho: https://github.com/lightbend/kalix-jvm-sdk/issues/2075

sdk/java-sdk-protobuf/src/main/scala/kalix/javasdk/impl/telemetry/Telemetry.scala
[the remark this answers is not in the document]
franciscolopezsancho: added

sdk/java-sdk-protobuf/src/main/java/kalix/javasdk/action/ActionCreationContext.java
franciscolopezsancho: added

samples/scala-protobuf-tracing/src/main/scala/com/example/ControllerAction.scala
franciscolopezsancho: you are right. I simplified now. using `quickRequest` that produces a `Request[String]` which in turn produces a `Future[Response[String]]` as you hinted. That is,

```
 val request = quickRequest.get(uri"${url}/1")
    val response = request.send(Main.backend)
    response.map { response =>
      val post = JsonMethods.parse(response.body).extract[Post]
      MessageResponse(post.title)
    }
```

octonato (approved):
LGTM

PR #2081, opened 2024-04-16
docs: match at least once in JWT static claim with pattern

efgpinto:
Refs https://github.com/lightbend/kalix-runtime/issues/2381


octonato (approved):
LGTM

Issue #695, opened 2021-11-10
Support Envoy Validation for Scala Protobuf projects


reid-spencer:
In my use of akka-grpc api development (besides Akka Serverless) we have become accustomed to using the ScalaPB validation library.  This really helps to communicate the semantics of the data fields in an API. Details on that are here: 
* https://scalapb.github.io/docs/validation

I tried to add this to my project but was only able to mess up the Scala code generation (generated code doesn't compile, missing references, duplicate code).  I believe that this should be pretty straightforward to provide in conjunction with Akka Serverless, but the existing installation instructions from ScalaPB (https://scalapb.github.io/docs/validation#installation) don't work directly.  It would be nice if this became an option on the Akka Serverless plugin or at least there was documentation about how to configure the ScalaPB validation feature with Akka Serverless.

johanandren:
I agree that being able to define declarative validation in the proto definition would be neat, I wonder though if it should rather be one step away from the SDK so that Akka Serverless already performs such validation of incoming requests without the SDK itself needing to generate any additional ScalaPB logic (this would also provide the same validation for all SDKs, not only Scala). 

@jroper @retgits anything we've considered before?

reid-spencer:
@johanandren - You wrote:
>  so that Akka Serverless already performs such validation of incoming requests without the SDK itself needing to
> generate any additional ScalaPB logic

Are you trying to say that Akka Serverless has some magical AI built-in so that when I have something like:
```protobuf
message Foo {
  string phoneNumber = 1;
}
```
that it can intuit that when a Foo is received that it parses the string in the phoneNumber value, it can ensure that the data value is a phone number with no further indication of its intended format?  ;->

I strongly doubt this and even if it did, this would be a reason to reject Akka Serverless as a solution because it is likely to interpret the domain of the solution incorrectly.  

Since the code generation is already handled by ScalaPB into separate files, simply enabling it doesn't seem like a lot of work.  

I am working on a PR, but my time is compromised by other things.  It seems to me like a small request and optional at that.

pvlugter:
> > so that Akka Serverless already performs such validation of incoming requests without the SDK itself needing to
> > generate any additional ScalaPB logic
> 
> Are you trying to say that Akka Serverless has some magical AI built-in

[...]

PR #2088, opened 2024-04-17
feat: Support ScalaPB generated envoy validation #695

johanandren:
References
- #695


johanandren:
Not sure if we should make it more official and add to docs or if this is enough for now. Setting it up was kinda tricky to get right so that could be an argument for showing it off in docs.

johanandren:
Curiously we generate all the message classes twice, once for main/prod and once for test (noticed when tests classes weren't validated), seems like something we should look into at some point.

Couldn't think of a reasonable way to trigger the validation failure from a test since the pb generated classes don't allow invalid message creation.

samples/scala-protobuf-validated/README.md
```diff
@@ -0,0 +1,70 @@
...
+
+```shell
+curl -XPOST -H "Content-Type: application/json" localhost:9000/com.example.validated.Validated/CallMeMaybe -d '{"email":"notanemail"}'
+```
+
+Which results in a `403 BadRequest` HTTP response with a text body with details about what was wrong
+
+Or a proper email:
+```shell
+curl -XPOST -H "Content-Type: application/json" localhost:9000/com.example.validated.Validated/CallMeMaybe -d '{"email":"notanemail"}'
```
efgpinto, line 39:
```suggestion
curl -XPOST -H "Content-Type: application/json" localhost:9000/com.example.validated.Validated/CallMeMaybe -d '{"email":"very.email@example.com"}'
```

samples/scala-protobuf-validated/src/test/scala/com/example/validated/ValidatedActionIntegrationSpec.scala
```diff
@@ -0,0 +1,40 @@
...
+
+  private val client = testKit.getGrpcClient(classOf[Validated])
+
+  "ValidatedAction" must {
+
+    "handle command CallMeMaybe" in {
+      val result = client.callMeMaybe(Request("valid.email@example.com")).futureValue
+      result should ===(Response("Hello valid.email@example.com, gosh what a valid email you sent me"))
+
+      // we can't test validation failure because we can't construct an invalid request
```
efgpinto, line 31: Did you consider enabling http transcoding and then having a http client here to make the invalid request? But then we would need to hardcode the host I guess.. 🤔 Probably not worth the hassle.

efgpinto (approved):
LGTM with suggested typo fix.

Had a note about test the failure but maybe not worth the trouble.

samples/scala-protobuf-validated/src/test/scala/com/example/validated/ValidatedActionIntegrationSpec.scala
johanandren: Yeah, would need to figure out the test container port, which seemed to tricky to be worth it.

PR #2094, opened 2024-04-18
Publish Maven plugin (Take 4)

ennru:
IIUC, the Cloudsmith repository doesn't support release staging.


octonato (approved):
Look ok, but why don't just delete it.

PR #2095, opened 2024-04-18
Publish sbt plugin (filter duplicate)

ennru:
Apparently sbt publishes two identical files with different names, Cloudsmith would consider them just one file.


octonato (approved):
LGTM

Never noticed we were publishing a 2.13 plugin.

ennru:
I wonder if the sbt scripted failure is related to this change.

> Never noticed we were publishing a 2.13 plugin.

It is not a 2.13 version it is just same same file with a different name.

PR #2098, opened 2024-04-18
chore: update scala version

octonato:


sdk/java-sdk-spring/src/main/scala/kalix/javasdk/impl/http/HttpEndpointMethodDefinition.scala
```diff
@@ -66,7 +66,8 @@ object HttpEndpointMethodDefinition {
     name = "ANY",
     safe = false,
     idempotent = false,
-    requestEntityAcceptance = RequestEntityAcceptance.Tolerated)
+    requestEntityAcceptance = RequestEntityAcceptance.Tolerated,
+    contentLengthAllowed = true)
```
efgpinto, line 70: Just curious, but why? What's the background here?

efgpinto (approved):

sdk/java-sdk-spring/src/main/scala/kalix/javasdk/impl/http/HttpEndpointMethodDefinition.scala
octonato: It also came as a surprise to me. 

I updated the scala version to 2.13.13 and now we get a deprecation error. What I suspect is that scalac now also detect methods deprecated with Java's `Deprecated` annotation. And that is the case of this method. 

It's a method from akka http, written in Scala, but deprecated with Java's annotation. It was not detected before as warning/error before.

About using `true` instead of `false`, according to the [PR in akka http](https://github.com/akka/akka-http/pull/4214) that deprecated this method, length should be 0 depending on the status code. 

Actually, what I did here is not good. I will update it. It should not simply use `true`, but we need to pass a function that decide it based on the status code. Hold on, I will update it.

ennru (approved):
LGTM.

PR #2100, opened 2024-04-18
Token to use git commands for docs publishing

ennru:
Docs publishing is about pushing docs to the `docs/kalix-current` branch.
https://github.com/lightbend/kalix-jvm-sdk/tree/docs/kalix-current

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,4 +1,4 @@
-name: Release
+name: Publish
 
 on:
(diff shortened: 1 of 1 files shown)


ennru:
Now I wonder if it is correct to do this on merge.

octonato:
> Now I wonder if it is correct to do this on merge.

Indeed, not on merge but on release. Wasn't it so before?

PR #2104, opened 2024-04-18
SDK version bump creation (transfer from Circle)

ennru:
The last bit that is still left in CircleCI.


efgpinto (approved):
LGTM. I don't see a change to disable the corresponding circle ci job.. you're keeping both?

ennru:
> I don't see a change to disable the corresponding circle ci job.. you're keeping both?

I don't trust myself. Want to see it work first.

PR #2117, opened 2024-04-23
feat: kalix-spring-parent pom

octonato:
References
-

### DIFF
--- a/publishLocally.sh
+++ b/publishLocally.sh
@@ -12,6 +12,17 @@ sbt 'publishM2; publishLocal'
 (
   cd maven-java
   mvn versions:set -DnewVersion=$SDK_VERSION
+
+  (
+    # special case for parent pom, version:set only changes project sharing the same parent
+    # kalix-spring-boot-parent is special because its parent is spring-boot-starter-parent
+    cd kalix-spring-boot-parent
+    mvn versions:set -DnewVersion=$SDK_VERSION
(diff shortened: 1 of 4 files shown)


octonato:
Hmm, let's not merge it yet. Publishing of the new parent pom won't work yet.

publishLocally.sh
```diff
@@ -12,6 +12,17 @@ sbt 'publishM2; publishLocal'
...
+
+  (
+    # special case for parent pom, version:set only changes project sharing the same parent
+    # kalix-spring-boot-parent is special because its parent is spring-boot-starter-parent
+    cd kalix-spring-boot-parent
+    mvn versions:set -DnewVersion=$SDK_VERSION
+
+    # also needs to change kalix-sdk.version in parent pom
+    sed "s/<kalix-sdk.version>\(.*\)<\/kalix-sdk.version>/<kalix-sdk.version>$SDK_VERSION<\/kalix-sdk.version>/" pom.xml
+  )
```
octonato, line 24: this is almost the same as in publish-maven.sh

The only difference is that one is doing local install and the other is deploy (publishing) the artifacts.

octonato (commented):
Scripts are fixed now. After merge, it should correctly publish the new parent pom.

ennru (approved):
LGTM.

PR #2119, opened 2024-04-23
fix: kalix-spring-boot-parent publishing

octonato:
kalix-spring-boot-parent was not being published and that was related to the fact that it was not sharing the same pom parent pom as the other maven project we have. 

This PR fix it by making full member of the Kalix maven projects and the Spring Boot Starter pom is removed and we only pull in the Spring BOM instead. 

Did some local tests and working as expected. In follow up PR I will change all the samples to use the new parent pom and then we will have more test running and building with it.

### DIFF
--- a/.github/publish-maven.sh
+++ b/.github/publish-maven.sh
@@ -31,18 +31,14 @@ cd maven-java
 
 # update poms with the version extracted from sbt dynver
 mvn --quiet --batch-mode versions:set -DnewVersion=${SDK_VERSION}
-  (
+
+  ( # also needs to change kalix-sdk.version in parent pom
     cd kalix-java-protobuf-parent
-    # also needs to change kalix-sdk.version in parent pom
     sed -i.bak "s/<kalix-sdk.version>\(.*\)<\/kalix-sdk.version>/<kalix-sdk.version>$SDK_VERSION<\/kalix-sdk.version>/" pom.xml
   )
 
-  ( # special case for parent pom, version:set only changes project sharing the same parent
-    # kalix-spring-boot-parent is special because its parent is spring-boot-starter-parent
+  ( # also needs to change kalix-sdk.version in parent pom
     cd kalix-spring-boot-parent
-    mvn versions:set -DnewVersion=$SDK_VERSION
-
-    # also needs to change kalix-sdk.version in parent pom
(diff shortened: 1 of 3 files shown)


maven-java/kalix-spring-boot-parent/pom.xml
```diff
@@ -5,10 +5,9 @@
     <modelVersion>4.0.0</modelVersion>
 
     <parent>
-        <groupId>org.springframework.boot</groupId>
-        <artifactId>spring-boot-starter-parent</artifactId>
-        <version>3.2.4</version>
-        <relativePath />
```
octonato, line 11: not the parent anymore

publishLocally.sh
```diff
@@ -13,20 +13,14 @@ sbt 'publishM2; publishLocal'
...
-    # also needs to change kalix-sdk.version in parent pom
     sed -i.bak "s/<kalix-sdk.version>\(.*\)<\/kalix-sdk.version>/<kalix-sdk.version>$SDK_VERSION<\/kalix-sdk.version>/" pom.xml
     rm pom.xml.bak
   )
 
-  ( # special case for parent pom, version:set only changes project sharing the same parent
-    # kalix-spring-boot-parent is special because its parent is spring-boot-starter-parent
+  ( # also needs to change kalix-sdk.version in parent pom
     cd kalix-spring-boot-parent
-    mvn versions:set -DnewVersion=$SDK_VERSION
```
octonato, line 27: idem as above

.github/publish-maven.sh
```diff
@@ -31,18 +31,14 @@ cd maven-java
...
     cd kalix-java-protobuf-parent
-    # also needs to change kalix-sdk.version in parent pom
     sed -i.bak "s/<kalix-sdk.version>\(.*\)<\/kalix-sdk.version>/<kalix-sdk.version>$SDK_VERSION<\/kalix-sdk.version>/" pom.xml
   )
 
-  ( # special case for parent pom, version:set only changes project sharing the same parent
-    # kalix-spring-boot-parent is special because its parent is spring-boot-starter-parent
+  ( # also needs to change kalix-sdk.version in parent pom
     cd kalix-spring-boot-parent
-    mvn versions:set -DnewVersion=$SDK_VERSION
```
octonato, line 43: not needed anymore, since now part of the family

PR #2120, opened 2024-04-24
fix: isolate parent pom plugins

octonato:


maven-java/pom.xml
```diff
@@ -49,52 +49,25 @@
...
+      <build>
+        <plugins>
+          <plugin>
+            <!-- 
+              this plugin allow us to change the version of this projects and submodules
+              we use it in CI to align the project with the current SDK version
+            -->
+            <groupId>org.codehaus.mojo</groupId>
+            <artifactId>versions-maven-plugin</artifactId>
+            <version>2.8.1</version>
```
octonato, line 65: Moved into profile so nobody inherits this. Not a real issue, but we have it only for our builds.

.github/publish-maven.sh
```diff
@@ -30,7 +30,7 @@ fi
 cd maven-java
 
 # update poms with the version extracted from sbt dynver
-mvn --quiet --batch-mode versions:set -DnewVersion=${SDK_VERSION}
+mvn --quiet --batch-mode --activate-profiles patch-version versions:set -DnewVersion=${SDK_VERSION}
```
octonato, line 33: active `patch-version` profile to be able to change the version. 

See below

maven-java/pom.xml
```diff
@@ -123,6 +96,26 @@
...
+          </plugin>
+          <plugin>
+            <groupId>org.apache.maven.plugins</groupId>
+            <artifactId>maven-resources-plugin</artifactId>
+            <version>3.2.0</version>
+          </plugin>
+          <plugin>
+            <groupId>org.sonatype.plugins</groupId>
+            <artifactId>nexus-staging-maven-plugin</artifactId>
+          </plugin>
```
octonato, line 118: all publishing related plugin moved to `release` profile

octonato (commented):
@ennru, I'm not sure if this is what we need. Some local test are Ok, we need to test it again after we publish it.

octonato:
@ennru, I can confirm that I can run now `mvn deploy kalix:deploy` on a code-first and protobuf projects built with the new parent poms. 

Both are being built correctly and deployed without interference from our own release settings. 

The code-first variant seems not to start, but that's due to another issue related with the assembly of fatjar. Independent from this PR.

.github/publish-maven.sh
ennru: Doesn't the `release` profile also need enabling?

.github/publish-maven.sh
```diff
@@ -62,6 +62,7 @@ cat <<EOF >~/.m2/settings.xml
         <activeByDefault>true</activeByDefault>
       </activation>
       <properties>
+        <altSnapshotDeploymentRepository>snapshots::default::https://s01.oss.sonatype.org/content/repositories/snapshots</altSnapshotDeploymentRepository>
```
ennru, line 65: Does this need a `altDeploymentRepository`, as well?

octonato: We only had a alt for snapshot. Actually, I don't know why we had this settings.

.github/publish-maven.sh
octonato: No, we are only patching the version in the pom. 
Later we run the release and then we activate `release`

ennru (approved):
LGTM.

PR #2121, opened 2024-04-25
fix: remove bogus plugin re-declaration

octonato:
Release didn't work because we were adding the plugin twice and the second addition was overwriting the settings we needed.


octonato:
Approving and merging myself to get the snapshot release tested.

PR #2122, opened 2024-04-25
fix: enable all parent pom plugins

octonato:
Parent poms were not working because we need to explicitly enable the plugins.

### DIFF
--- a/.github/publish-maven.sh
+++ b/.github/publish-maven.sh
@@ -28,20 +28,7 @@ then
 fi
 
 cd maven-java
-
-# update poms with the version extracted from sbt dynver
-mvn --quiet --batch-mode --activate-profiles patch-version versions:set -DnewVersion=${SDK_VERSION}
-
-  ( # also needs to change kalix-sdk.version in parent pom
-    cd kalix-java-protobuf-parent
-    sed -i.bak "s/<kalix-sdk.version>\(.*\)<\/kalix-sdk.version>/<kalix-sdk.version>$SDK_VERSION<\/kalix-sdk.version>/" pom.xml
-  )
-
-  ( # also needs to change kalix-sdk.version in parent pom
-    cd kalix-spring-boot-parent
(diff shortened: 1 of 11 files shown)


updateSdkVersions.sh
```diff
@@ -39,7 +39,7 @@ updateScalaSamples() {
 
 updateMavenPlugin() {
   echo ">>> Updating maven plugin to $SDK_VERSION"
-  cd maven-java && mvn versions:set -DnewVersion="$SDK_VERSION" && cd ..
+  cd maven-java && mvn --activate-profiles patch-version versions:set -DnewVersion="$SDK_VERSION" && cd ..
```
octonato, line 42: hmm, we neeed to call the path-maven-version here as well.

.github/workflows/ci.yml
```diff
@@ -316,6 +316,7 @@ jobs:
       - name: Value Entity Archetype
         run: |-
           export SDK_VERSION=$(cat ~/kalix-sdk-version.txt)
```
ennru, line 318: Is this version file still there? Alternatively, this should reference `steps.build_sdk.outputs.sdk_version`, right?

.github/publish-maven.sh
```diff
@@ -75,4 +75,4 @@ echo "${PGP_SECRET}" | base64 -d | gpg --import --batch
 
 # Maven deploy with profile `release`
 # mvn --quiet --batch-mode --activate-profiles release deploy
-mvn --batch-mode --activate-profiles release deploy
+mvn --batch-mode --activate-profiles release -Dskip.docker=true deploy
```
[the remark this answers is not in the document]
ennru: I understand it is backward in the Docker plugin already. The negation can become confusing, but I don't see what we could do about that.

ennru (approved):
LGTM.

octonato:
@ennru, I merged as is now and will check your suggestions next. I will send a follow-up PR to improve.

PR #2123, opened 2024-04-25
feat: ease build for multi-platform images

octonato:
On top of #2122 

Make it easy to build arm64 images. 

This will become more common when using multi-service tests with user service containers. 
Not yet documenting it, because my main goal for now is to have it in place to ease my own debugging.

### DIFF
--- a/maven-java/kalix-java-protobuf-parent/pom.xml
+++ b/maven-java/kalix-java-protobuf-parent/pom.xml
@@ -32,6 +32,7 @@
         <dockerTag>${project.version}-${build.timestamp}</dockerTag>
 
         <docker.base.image>docker.io/library/eclipse-temurin:21.0.2_13-jre-jammy</docker.base.image>
+        <docker.platform>linux/amd64</docker.platform>
 
         <maven.build.timestamp.format>yyyyMMddHHmmss</maven.build.timestamp.format>
         <java.version>21</java.version>
@@ -172,7 +173,7 @@
                             <!-- Base Docker image which contains jre-->
                             <from>${docker.base.image}</from>
                             <createImageOptions>
-                            <platform>linux/amd64</platform>
+                                <platform>${docker.platform}</platform>
                             </createImageOptions>
                             <tags>
                             <!-- tag for generated image -->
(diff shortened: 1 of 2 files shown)


octonato:
Not so sure if best solution though. Build for standard image (linux/amd64) seems to be much slower now.

octonato:
Some more information about this. 

The idea of providing a multi-platform build is complicated. 
The end goal is to build a linux/amd64 image as this is the platform required in production. 

But users may want to build a linux/arm64 to run some local tests in a multi-service setup where one service is run in a container. That applies only for users using new generation Apple processors. 

The problem is that if one builds an image using `mvn install -Ddocker.platform=linux/arm64`, they will download an arm64 image for JDK temurin. 

Later, users will need to manually delete their local arm64 temurin image before they build a image for prod. This can be quite confusing and error prone. 

For the moment, that's the only solution we have and we need to think further how to best deal with it. 
It still useful for us to have the option to pass `docker.platform=linux/arm64`. That will make it easier for us to do some local tests, but end users will need a better solution.

octonato:
I did some more experiments, and one possible recommendation is to create the image with:

```
mvn install -Ddocker.platform=linux/arm64 -Ddocker.base.image=docker.io/library/eclipse-temurin@sha256:94e6272fa873d796f277635350b640ee4ec169086f184b3af29d2a23986e4d32
```

Instead of passing a tag, we can pass the sha256 for the platform specific image. As a result, when building a arm64 image, we end up with 

```
eclipse-temurin                                             <none>                          a938dfab2f73   4 weeks ago     270MB
``

That's an untagged arm64 image. 

Then, if you run it again without specifying a platform, you end up with two eclipse-temurin images.

```
eclipse-temurin                                             <none>                          a938dfab2f73   4 weeks ago     270MB
eclipse-temurin                                             21.0.2_13-jre-jammy             97848f48a194   4 weeks ago     279MB
```

Now we get two temurin images for two different platforms and they don't interfere with each other. 

Still complicated way of building arm64 images for local tests. We can consider adding it to a profile, but not a priority right now. We can give more time to think about it.

ennru (approved):
LGTM.

PR #2125, opened 2024-04-30
feat: expose underlying Akka version

ennru:
For sbt users that want to build Action tests relying on the Akka Stream testkit, this exposes the underlying Akka core version for reference in their build file.


octonato (approved):
LGTM

So the goal here is to make it easier to add extra Akka dependencies, but still align with the current version. And the only extra dependency that a user could possibly want to add is a testkit. All other dependencies are already provided, correct?

PR #2127, opened 2024-04-30
fix: correctly configure spring-boot plugin

octonato:
To have the spring infra correctly packaged, we need the to correctly configure the spring-boot plugin.

This was previously done by the spring-parent pom. We need to do the same in our new parent pom.


ennru (approved):
LGTM.

Issue #2133, opened 2024-05-13
Scala SDK - Scala 3 support 


aklikic:
Adding offical support for Scala 3.

efgpinto:
I still need to update the g8 template before marking this as done.

PR #2135, opened 2024-05-16
feat: support scala 3

efgpinto:
Refs #2133 

This PR:
 - adds cross-compilations for the modules required to offer scala 3 support for scala protocol-first sdk 
 - bumps samples to scala 3.3.3 by default
 - configs CI to run all samples both with scala 2.13 and scala 3 so we keep them compatible with both versions

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -757,35 +757,19 @@ jobs:
           cd
           tar -xf dependencies.tar.gz
 
-      - name: sbt Test/compile Scala 2.13
+      - name: sbt Test/compile
         env:
           DIR: ${{ matrix.sample }}
         run: |-
           export SDK_VERSION=$(cat ~/kalix-sdk-version.txt)
           cd samples/${DIR}
           echo "Running sbt on ${DIR} with SDK version: '$SDK_VERSION'"
+          echo "==== Testing with Scala 2.13 ===="
           sbt --client -Dkalix-sdk.version=$SDK_VERSION '++2.13.14! Test/compile'
-
-      - name: sbt test Scala 2.13
-        env:
-          DIR: ${{ matrix.sample }}
-          PRE_CMD: ${{ matrix.pre_cmd }}
-        run: |-
-          export SDK_VERSION=$(cat ~/kalix-sdk-version.txt)
-          cd samples/${DIR}
-          ${PRE_CMD}
-          sbt  --client -Dkalix-sdk.version=$SDK_VERSION '++2.13.14! test'
-
-      - name: sbt Test/compile Scala 3
-        env:
-          DIR: ${{ matrix.sample }}
-        run: |-
-          export SDK_VERSION=$(cat ~/kalix-sdk-version.txt)
-          cd samples/${DIR}
-          echo "Running sbt on ${DIR} with SDK version: '$SDK_VERSION'"
+          echo "==== Testing with Scala 3 ===="
           sbt --client -Dkalix-sdk.version=$SDK_VERSION 'clean; ++3.3.3! Test/compile'
 
-      - name: sbt test Scala 3
+      - name: sbt test
         if: matrix.test
         env:
           DIR: ${{ matrix.sample }}
@@ -794,6 +778,9 @@ jobs:
           export SDK_VERSION=$(cat ~/kalix-sdk-version.txt)
           cd samples/${DIR}
           ${PRE_CMD}
+          echo "==== Testing with Scala 2.13 ===="
+          sbt  --client -Dkalix-sdk.version=$SDK_VERSION '++2.13.14! test'
+          echo "==== Testing with Scala 3 ===="
           sbt  --client -Dkalix-sdk.version=$SDK_VERSION 'clean; ++3.3.3! test'
 
       - name: rm & sbt Test/compile
@@ -802,7 +789,8 @@ jobs:
         run: |-
           export SDK_VERSION=$(cat ~/kalix-sdk-version.txt)
           cd samples/${DIR}
-          echo "==== Verifying that generated unmanaged sources compile ===="
+          echo "==== Verifying that generated unmanaged sources compile with Scala 2.13 ===="
           rm -rf src/main/scala src/test/scala src/it/scala
-          sbt  --client -Dkalix-sdk.version=$SDK_VERSION '++2.13.14! Test/compile'
+          sbt  --client -Dkalix-sdk.version=$SDK_VERSION +Test/compile
+          echo "==== Verifying that generated unmanaged sources compile with Scala 3 ===="
           sbt  --client -Dkalix-sdk.version=$SDK_VERSION 'clean; ++3.3.3! Test/compile'
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -770,7 +770,7 @@ jobs:
           export SDK_VERSION=$(cat ~/kalix-sdk-version.txt)
           cd samples/${DIR}
           echo "Running sbt on ${DIR} with SDK version: '$SDK_VERSION'"
-          sbt --client -Dkalix-sdk.version=$SDK_VERSION '++${SCALA_213_VERSION}! Test/compile'
+          sbt --client -Dkalix-sdk.version=$SDK_VERSION '++$SCALA_213_VERSION! Test/compile'
 
       - name: sbt test Scala 2.13
         if: matrix.test
@@ -781,7 +781,7 @@ jobs:
           export SDK_VERSION=$(cat ~/kalix-sdk-version.txt)
           cd samples/${DIR}
           ${PRE_CMD}
-          sbt  --client -Dkalix-sdk.version=$SDK_VERSION '++${SCALA_213_VERSION}! test'
+          sbt  --client -Dkalix-sdk.version=$SDK_VERSION '++$SCALA_213_VERSION! test'
 
       - name: sbt Test/compile Scala 3
         env:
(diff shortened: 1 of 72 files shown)


build.sbt
```diff
@@ -384,7 +394,8 @@ def githubUrl(v: String): String = {
 lazy val devTools = devToolsCommon(
   project
     .in(file("devtools"))
-    .settings(name := "kalix-devtools", scalaVersion := Dependencies.ScalaVersion))
+    .settings(name := "kalix-devtools", scalaVersion := Dependencies.ScalaVersion,
+      crossScalaVersions := Seq(Dependencies.ScalaVersion, Dependencies.Scala3Version)))
```
octonato, line 398:
```suggestion
      crossScalaVersions := CrossScalaVersions))
```

build.sbt
```diff
@@ -323,6 +332,7 @@ lazy val scalaSdkProtobuf = project
       "protocolMinorVersion" -> Kalix.ProtocolVersionMinor,
       "scalaVersion" -> scalaVersion.value,
       "akkaVersion" -> Dependencies.AkkaVersion),
+    crossScalaVersions := Seq(Dependencies.ScalaVersion, Dependencies.Scala3Version),
```
octonato, line 335:
```suggestion
    crossScalaVersions := CrossScalaVersions),
```

build.sbt
```diff
@@ -85,6 +93,7 @@ lazy val javaSdkProtobuf = project
       "scalaVersion" -> scalaVersion.value,
       "akkaVersion" -> Dependencies.AkkaVersion),
     buildInfoPackage := "kalix.javasdk",
+    crossScalaVersions := Seq(Dependencies.ScalaVersion, Dependencies.Scala3Version),
```
octonato, line 96:
```suggestion
    crossScalaVersions := CrossScalaVersions),
```

build.sbt
```diff
@@ -58,6 +65,7 @@ lazy val coreSdk = project
     crossPaths := false,
     Compile / javacOptions ++= Seq("--release", "11"),
     Compile / scalacOptions ++= Seq("-release", "11"),
+    crossScalaVersions := Seq(Dependencies.ScalaVersion, Dependencies.Scala3Version),
```
octonato, line 68:
```suggestion
    crossScalaVersions := CrossScalaVersions,
```

project/Dependencies.scala
```diff
@@ -14,8 +14,10 @@ object Dependencies {
   }
 
   // changing the Scala version of the Java SDK affects end users
-  val ScalaVersion = "2.13.14"
+  val CrossScalaVersions = Seq("2.13.14", "3.3.3")
+  val ScalaVersion = CrossScalaVersions.head
   val ScalaVersionForTooling = "2.12.19"
+  val Scala3Version = "3.3.3"
```
octonato, line 20:
```suggestion
```
Can be removed if we use `CrossScalaVersions`

sdk/java-sdk-protobuf/src/main/java/kalix/javasdk/TraceContext.java
```diff
@@ -17,7 +15,7 @@ public interface TraceContext {
    *
    * @return the trace context as an OpenTelemetry context.
    */
-  Context asOpenTelemetryContext();
+  io.opentelemetry.context.Context asOpenTelemetryContext();
```
octonato, line 18: any particular reason for not using the import?

sdk/java-sdk-protobuf/src/main/scala/kalix/javasdk/impl/DiscoveryImpl.scala
```diff
@@ -85,8 +83,9 @@ class DiscoveryImpl(
 
     // possibly filtered or hidden env, passed along for substitution in descriptor options
     val env: Map[String, String] = system.settings.config.getAnyRef("kalix.discovery.pass-along-env-allow") match {
-      case false => Map.empty
-      case true  => sys.env
+      // FIXME: how to do this better?
+      case obj if !obj.asInstanceOf[Boolean] => Map.empty
+      case obj if obj.asInstanceOf[Boolean]  => sys.env
```
octonato, line 88: This is indeed weird. 

What about this? 

```suggestion
      case bool: Boolean => if (bool) sys.env else Map.empty
```

It still feels awkward in pattern matching, but more concise.

octonato (commented):
Looking good

Left some minor nitpicking and suggestion.

efgpinto:
Ok, this is looking good now but before I look at the last bits and do some clean-up.. question: should we cross-compile all samples to scala 2.13 and 3? should we have them on scala 3 by default? or should we have only a separate scala 3 sample? I'm inclined to have them all on scala 3 by default but still compile with 2.13 on CI, to make sure we stay compatible. Thoughts?

ennru:
> I'm inclined to have them all on scala 3 by default but still compile with 2.13 on CI, to make sure we stay compatible.

SGTM.

build.sbt
efgpinto: see current version

sdk/java-sdk-protobuf/src/main/java/kalix/javasdk/TraceContext.java
efgpinto: not warranted after all, reverted

project/Dependencies.scala
efgpinto: Not sure I fully understood your suggestion.. you would rather have the version directly on the `CrossScalaVersions` val? see what you think of current setup anyway.

.github/workflows/ci.yml
```diff
@@ -786,4 +811,5 @@ jobs:
           cd samples/${DIR}
           echo "==== Verifying that generated unmanaged sources compile ===="
           rm -rf src/main/scala src/test/scala src/it/scala
-          sbt  --client -Dkalix-sdk.version=$SDK_VERSION Test/compile
+          sbt  --client -Dkalix-sdk.version=$SDK_VERSION '++$SCALA_213_VERSION! Test/compile'
```
octonato: Instead of calling sbt twice, we can use +Test/compile.

octonato: Above, we have to compile steps for each scala version, here we have only one. Not big issue, but we could eventually merge the two other compile steps into one as well. 

The advantage to have then apart is that we make it easier for us to see if it's a Scala 2.13 or 3 compilation error.

project/Dependencies.scala
octonato: Looks good now. 

Either

```scala
val ScalaVersion = "2.13.14"
val Scala3Version = "3.3.3"
val CrossScalaVersions = Seq(ScalaVersion, Scala3Version)
```

or

```scala
val ScalaVersion = CrossScalaVersions.head
val CrossScalaVersions = Seq("2.13.14", "3.3.3")
```

I don't remember how it was before, but looks good now.

samples/scala-protobuf-eventsourced-counter/src/test/scala/com/example/CounterServiceIntegrationSpec.scala
```diff
@@ -96,7 +96,7 @@ class CounterServiceIntegrationSpec
       counter.value shouldBe 15
 
       // verify message published to topic
-      val Message(decEvent, md): Message[Decreased] = eventsTopic.expectOneTyped
+      val Message(decEvent, md) = eventsTopic.expectOneTyped[Decreased]
```
octonato, line 99: Is this changed due to Scala 3? If so, then it's a good plan to move all samples to Scala 3. Scala 3 compiler being more strict, we want to catch possible issues earlier.

samples/scala-protobuf-transfer-workflow-compensation/src/main/scala/com/example/transfer/api/TransferWorkflow.scala
```diff
@@ -40,7 +40,7 @@ class TransferWorkflow(context: WorkflowContext) extends AbstractTransferWorkflo
   override def definition: AbstractWorkflow.WorkflowDef[TransferState] = {
 
     val withdraw = step("withdraw")
-      .asyncCall { withdraw: Withdraw =>
+      .asyncCall { (withdraw: Withdraw) =>
```
octonato, line 43: Are the `( )` needed?

octonato (approved):
LGTM

Left a few comments, but nothing major. Fine to merge as is.

samples/scala-protobuf-transfer-workflow-compensation/src/main/scala/com/example/transfer/api/TransferWorkflow.scala
efgpinto: yes, scala 3 requires those

samples/scala-protobuf-eventsourced-counter/src/test/scala/com/example/CounterServiceIntegrationSpec.scala
efgpinto: it is indeed required by scala 3.

.github/workflows/ci.yml
efgpinto: yeah, that's the reason I kept those apart.. and I thought this one was less likely to present issues so kept it together. But maybe I can also consider merging the other ones.

efgpinto: actually I think I need to call sbt twice here..  or otherwise would need to set `crossScalaVersions := ...` on the same sbt command so that the `+Test` has multiple versions to go through.

PR #2141, opened 2024-05-27
feat: include sdk deps in parent pom

efgpinto:
Close #2140 

This PR:
- simplifies user poms by removing direct dependency on sdk modules and having them included directly from the new parent pom
- update maven templates and samples accordingly

PR #2142, opened 2024-05-27
chore: remove deps also from archetypes

efgpinto:
Good catch, tks @ennru 

Follow up https://github.com/lightbend/kalix-jvm-sdk/pull/2141


ennru (approved):
Less is more.
LGTM.

PR #2148, opened 2024-05-28
chore: explain how to change ports

octonato:


docs/src/modules/java-protobuf/partials/env-file.txt
octonato, line 1: This is a bit unfortunate. Antora is not being able to include the original file `.env` because, I guess, it's a hidden file. 

So I have to make a copy of the env file in the docs folder and include it instead. 

Also, we will have to remember to keep it in sync and we can't add message in the original .env. First because it ends in the user project, but also because there is no master .env file. They are all copies of each other spread over the templates and samples.

efgpinto (approved):

Issue #2139, opened 2024-05-23
Release Kalix Java/Scala SDKs v1.5.0


efgpinto:
# Release Kalix Java/Scala SDKs v1.5.0

### Prepare

- [x] Make sure all important PRs have been merged
- [x] Check that the [latest build](https://github.com/lightbend/kalix-jvm-sdk/actions?query=branch%3Amain) successfully finished
- [x] Make sure a version of the Kalix Runtime that supports the protocol version the SDK expects has been deployed to production

You can see the Kalix Runtime version on prod [on grafana](https://grafana.sre.kalix.io/d/b30d0d8e-3894-4fbf-9627-9cb6088949ee/prod-kalix-metrics?orgId=1) or using [various other methods](https://github.com/lightbend/kalix/wiki/Versioning-and-how-to-determine-what-version-is-running).

### Cutting the release 

- [x] Update the "Change date" on [the license](../blob/main/LICENSE#L9) to release date plus three years
- [x] Use the "Generate release notes" button to create [a new release](https://github.com/lightbend/kalix-jvm-sdk/releases/new) with the appropriate tag.
    - Review the generated notes and "Publish release"
    - CI will automatically publish to the repository based on the tag
    - CI will update the docs/kalix-current branch

### Check availability

- [x] Check that [`docs/kalix-current`](https://github.com/lightbend/kalix-jvm-sdk/commits/docs/kalix-current) has been updated
- [x] Check the release on [Maven Central](https://repo1.maven.org/maven2/io/kalix/kalix-jvm-core-sdk/)

### Fix and publish docs

- [x] In case there were important updates, update the [supported version in the main docs](https://github.com/lightbend/kalix-docs/blob/main/docs/modules/ROOT/partials/include.adoc#L21) (affects [Supported Languages](https://docs.kalix.io/sdk-support/supported-languages.html))
- [x] Add an item to the [Release Notes](https://github.com/lightbend/kalix-docs/blob/main/docs/modules/release-notes/pages/index.adoc) in the documentation
- [x] Release the Kalix documentation to get the SDK docs updates published

### Update to the latest version
 
- [x] Review and merge PR created by bot (should appear [here](https://github.com/lightbend/kalix-jvm-sdk/pulls?q=is%3Apr+is%3Aopen+auto+pr+)). While reviewing confirm the release version is updated for:
    - `kalix-sdk.version` in the `samples/*/pom.xml` files
[...]

PR #2149, opened 2024-05-28
build: fix duplicate publishing

efgpinto:
Refs https://github.com/lightbend/kalix-jvm-sdk/issues/2139

Don't think his solves all but we should need this to avoid duplication for the 2.12_1.0 packages.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -58,6 +58,12 @@ def disciplinedScalacSettings: Seq[Setting[_]] = {
   } else Seq.empty
 }
 
+lazy val dontPublishTwice = Seq(
+  publishSignedConfiguration := publishSignedConfiguration.value.withArtifacts(
+    // avoid publishing the plugin jar twice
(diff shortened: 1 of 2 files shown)


johanandren (commented):
The problem should only be for sbt-plugins which sbt publishes twice for some reason (same if you look at some sbt plugin on maven central)

efgpinto:
You're right. I have doubled check and only for sbt-kalix it tries to write twice with a different name:
```
[info] Wrote /home/runner/work/kalix-jvm-sdk/kalix-jvm-sdk/sbt-plugin/target/scala-2.12/sbt-1.0/sbt-kalix_2.12_1.0-1.5.0-1-a5094919.pom
[info] Wrote /home/runner/work/kalix-jvm-sdk/kalix-jvm-sdk/sbt-plugin/target/scala-2.12/sbt-1.0/sbt-kalix-1.5.0-1-a5094919.pom
```

For the remaining (codegenJava, codegenScala, etc), although it tries to publish twice as well, it's with the same name, so this filter would not even work. Closing.

PR #2156, opened 2024-05-31
fix: keep .env during repo sync

ennru:
The new `.env` file need to be explicitly added to the PR.

Also bumps the pull request action to the current version.

References
- https://github.com/peter-evans/create-pull-request/releases


octonato (approved):
We learn every single day!

PR #2158, opened 2024-05-31
fix: keep .env during repo sync 3

ennru:
Explicitly copy `.env`.

References
- #2156
- #2157


octonato (commented):
PR'ing as there is not tomorrow!

octonato (approved):

PR #2160, opened 2024-06-24
feat:  adding tracing id to UF logs

franciscolopezsancho:
References https://github.com/lightbend/kalix-runtime/issues/2444#issuecomment-2175215532

Looking into adding tests for it.

### DIFF
--- a/sdk/java-sdk-protobuf/src/main/scala/kalix/javasdk/impl/action/ActionsImpl.scala
+++ b/sdk/java-sdk-protobuf/src/main/scala/kalix/javasdk/impl/action/ActionsImpl.scala
@@ -22,7 +22,7 @@ import kalix.javasdk.impl.telemetry.{ ActionCategory, Instrumentation, Telemetry
 import kalix.protocol.action.{ ActionCommand, ActionResponse, Actions }
 import kalix.protocol.component
 import kalix.protocol.component.{ Failure, MetadataEntry }
-import org.slf4j.{ Logger, LoggerFactory }
+import org.slf4j.{ Logger, LoggerFactory, MDC }
 
 import java.util.Optional
 import scala.compat.java8.OptionConverters.RichOptionForJava8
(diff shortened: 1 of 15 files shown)


sdk/java-sdk-protobuf/src/main/scala/kalix/javasdk/impl/action/ActionsImpl.scala
```diff
@@ -192,7 +193,10 @@ private[javasdk] final class ActionsImpl(_system: ActorSystem, services: Map[Str
               Future.successful(handleUnexpectedException(service, in, ex))
           }
         fut.andThen { case _ =>
-          span.foreach(_.end())
+          span.foreach { s =>
+            mdc.remove(Telemetry.TRACE_ID)
```
pvlugter, line 197: This is removing from a copy of the MDC, so doesn't affect the actual MDC.

It's also removing in a future callback, which may not be the same thread as where the trace id was added to the MDC. This means that the trace id can be 'leaked' — other logs when something runs on that same original thread will be incorrectly logged with the trace id.

sdk/java-sdk-protobuf/src/main/scala/kalix/javasdk/impl/eventsourcedentity/EventSourcedEntitiesImpl.scala
```diff
@@ -234,7 +236,12 @@ final class EventSourcedEntitiesImpl(
                         serializedSnapshot,
                         delete))))
             }
-          } finally { span.foreach(_.end()) }
+          } finally {
+            span.foreach { s =>
+              mdc.remove(Telemetry.TRACE_ID)
```
pvlugter, line 241: Removing trace id in a finally block on the same thread is good, but needs to remove from the actual MDC rather than the copied map.

franciscolopezsancho: done

sdk/java-sdk-protobuf/src/main/scala/kalix/javasdk/impl/action/ActionsImpl.scala
franciscolopezsancho: done

pvlugter (approved):
LGTM

PR #2162, opened 2024-06-27
Adding mdc with traceid to components II

franciscolopezsancho:
Copying adding_mdc_with_traceid_to_components to check the CI pipeline. It's failing in a test I can reproduce. Not in local not in an Ubuntu22.04.


franciscolopezsancho:
just checked that the error was being reproduce in the pipeline.

PR #2163, opened 2024-06-27
Adding mdc with traceid to components 0.2

franciscolopezsancho:
same as https://github.com/lightbend/kalix-jvm-sdk/pull/2162


franciscolopezsancho:
Another test that proves that with the same commits the `sbt fails`

PR #2165, opened 2024-06-27
forcing order in Components[X]Calls codegen

franciscolopezsancho:
WIP. Trying to sort the order in the codegen to pass tests in CI


johanandren (approved):
Strange, LGMT I guess.

PR #2172, opened 2024-07-06
feat: passing through metadata traceparent

franciscolopezsancho:
feat to automatially passing the traceparent in the metadata on 'reply' or 'asyncReply'

In the java-spring-eventsourced-counter (slightly modified, see modification* below) we were getting this if we didn't pass explicitly the metadata 

Screenshot 2024-07-06 at 15 49 13

instead of the expected

Screenshot 2024-07-06 at 15 39 48

*the `modification` involved to make the action`CounterCommandFromTopicAction` consume from the same topic that `CounterJournalToTopicAction` was publishing to, such when consuming events that would reply to the ESE that originated the event `CounterJournalToTopicAction` was consuming in the first place. There's no circularity here because the ESE was initially called with `increase` while the second call to the ESE was `multiply`.

### DIFF
--- a/sdk/java-sdk-protobuf/src/main/scala/kalix/javasdk/impl/action/ActionEffectImpl.scala
+++ b/sdk/java-sdk-protobuf/src/main/scala/kalix/javasdk/impl/action/ActionEffectImpl.scala
@@ -4,20 +4,18 @@
 
 package kalix.javasdk.impl.action
 
-import kalix.javasdk.{ DeferredCall, Metadata, SideEffect }
-import kalix.javasdk.action.Action
-import java.util
-import java.util.concurrent.CompletionStage
-
 import io.grpc.Status
 import kalix.javasdk.StatusCode.ErrorCode
+import kalix.javasdk._
+import kalix.javasdk.action.Action
 import kalix.javasdk.impl.StatusCodeConverter
-import scala.concurrent.ExecutionContext
-import scala.concurrent.Future
+import kalix.javasdk.impl.telemetry.Telemetry
+
+import java.util
+import java.util.concurrent.CompletionStage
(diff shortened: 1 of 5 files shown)


sdk/java-sdk-protobuf/src/main/scala/kalix/javasdk/impl/action/ActionEffectImpl.scala
```diff
@@ -101,13 +100,26 @@ object ActionEffectImpl {
...
+            metadata.add(Telemetry.TRACE_PARENT_KEY, traceparent)
+          case _ => metadata
+        }
+      }
+    }
+
   }
 
-  def builder(): Action.Effect.Builder = Builder
+  def builder(context: Metadata): Action.Effect.Builder = new Builder(context)
```
octonato, line 123: Is this method in use? It seems not, because signature changed and there is no call to this method in this diff. So probably it's never called at all.

On the other hand, we can use it instead of `new ActionEffectImpl.Builder(Metadata.EMPTY)`.

octonato (approved):
LGTM

But there is some clean-up to do, I think.

sdk/java-sdk-protobuf/src/main/scala/kalix/javasdk/impl/action/ActionEffectImpl.scala
franciscolopezsancho: Yes, this is used [here](https://github.com/lightbend/kalix-jvm-sdk/pull/2172/files#diff-42fba786a8bdebbedae909afdf95687a254b108eb0a425e2c21814b05ee0fd53R84)

and yes (good catch:)

octonato (approved):
LGTM

PR #2174, opened 2024-07-12
fix: removing infinite subscription loop

aludwiko:


octonato (approved):
LGTM

It seems that we are using this in the docs, but the purpose of the sample is to demonstrate ValueEntities with Views. 

So we kind of piggybacked it with this DoubleCounterAction because of the docs.

PR #2182, opened 2024-08-14
fix: adding missing tracing MDC for views

franciscolopezsancho:
When testing realized that there wasn't MDC from tracing in Views.


sdk/java-sdk-protobuf/src/main/scala/kalix/javasdk/impl/telemetry/Telemetry.scala
```diff
@@ -36,6 +37,14 @@ object Telemetry extends ExtensionId[Telemetry] {
   val TRACE_ID: String = "trace_id"
   override def createExtension(system: ExtendedActorSystem): Telemetry =
     new Telemetry(system)
+
+  private val traceParentPattern: Regex = """^[0-9a-f]{2}-([0-9a-f]{32})-[0-9a-f]{16}-[0-9a-f]{2}$""".r
+
+  def extractTraceId(traceParent: String): String =
+    traceParent match {
+      case traceParentPattern(traceId) => traceId
+      case _                           => ""
```
johanandren, line 46: What is the scenario where this would not be passed a string matches? Isn't that an error rather than something that should return an empty trace id?

franciscolopezsancho: Good point. Are you suggesting to throw an exception here? If we throw an exception and at any point we add the wrong data to the metadata, as `traceparent` we could break the application. 
I would rather log no trace found in this `traceparent` and return the blank. WDYT?

johanandren: Or return option if you _think_ it could be present but not contain trace id, but then would that ever be valid? Also, is there no built in opentelemetry thing for extracting it rather than doing our own string manipulation?

franciscolopezsancho: it seems they do this when invalid.  returning `00000000000000000000000000000000`
https://github.com/open-telemetry/opentelemetry-java/blob/dcdc6727f78638557ccc65742cb7898e306ea227/api/all/src/main/java/io/opentelemetry/api/trace/TraceId.java#L78

That makes more sense

franciscolopezsancho: so how about returning "00000000000000000000000000000000"?

johanandren: From TraceId.java
> A class that represents a trace identifier. A valid trace identifier is a 16-byte array with at least one non-zero byte.

Isn't it possible to reach it via `Metadata#traceContext#asOpenTelemetryContext#something`?

franciscolopezsancho: you're right. 
```
Span.fromContext(context).getSpanContext.getTraceId
```

franciscolopezsancho: As suggested, we can do the same with the following:
```
  def extractTraceId(metadata: Metadata): String = {
    val context = W3CTraceContextPropagator.getInstance().extract(Context.root(), metadata, otelGetter)
    Span.fromContext(context).getSpanContext.getTraceId
  }
``` 

When no traceparent return is `00000000000000000000000000000000`

franciscolopezsancho: I have mixed feelings. I think this is more obscure (the first line of `extractTraceId` specifically)  but I like the signature more. `Metadata` as input instead of a `String`. That said I could keep this signature with the other implementation. 
In any case I see some benefits in this implementation in that there is less risk I got something wrong. Nor that there was much to get wrong in the RegEx but anyways. How you find it?

johanandren: The SpanContext is already available from MetadataImpl, so no need to parse the w3c format again just for logging?

`def traceId(): String = Span.fromContext(asOpenTelemetryContext()).getSpanContext.getTraceId` in the SDK `MetadataImpl#traceContext` seems reasonable to me.

franciscolopezsancho: You mean, use it in the `Telemetry` method? or to add this method to our `TraceContext` 
I like the latter.

sdk/java-sdk-protobuf/src/main/scala/kalix/javasdk/impl/view/ViewsImpl.scala
```diff
@@ -93,6 +101,9 @@ final class ViewsImpl(system: ActorSystem, _services: Map[String, ViewService])
               val commandName = receiveEvent.commandName
               val msg = service.messageCodec.decodeMessage(receiveEvent.payload.get)
               val metadata = MetadataImpl.of(receiveEvent.metadata.map(_.entries.toVector).getOrElse(Nil))
+              if (TraceId.isValid(metadata.traceContext.traceId())) {
```
johanandren, line 104: Still feels a bit weird to not have it `None` and not do anything on missing rather than return invalid and then validate that to know that we don't have one, seems cheap enough though I guess.

sdk/java-sdk-protobuf/src/main/scala/kalix/javasdk/impl/view/ViewsImpl.scala
```diff
@@ -102,6 +113,8 @@ final class ViewsImpl(system: ActorSystem, _services: Map[String, ViewService])
                   case e: ViewException => throw e
                   case NonFatal(error) =>
                     throw ViewException(context, s"View unexpected failure: ${error.getMessage}", Some(error))
+                } finally {
+                  MDC.remove(Telemetry.TRACE_ID)
```
johanandren, line 117: This means we always touch a thread local, even if there was no trace to start with. Maybe keep a boolean around from setting it or not to know if it needs clearing out or not?

johanandren (commented):
Almost good to go

sdk/java-sdk-protobuf/src/main/scala/kalix/javasdk/impl/view/ViewsImpl.scala
franciscolopezsancho: So you think is better` Optional<String> traceId();` in `TraceContext`. And not follow their style, but more and Scala-ish.

sdk/java-sdk-protobuf/src/main/scala/kalix/javasdk/impl/view/ViewsImpl.scala
franciscolopezsancho: done

sdk/java-sdk-protobuf/src/main/scala/kalix/javasdk/impl/view/ViewsImpl.scala
franciscolopezsancho: done

johanandren (approved):

PR #2205, opened 2024-09-04
fix(docs): fail deploy.sh on error in script

sebastian-alfers:
Attempt to fail the bash script if `git push` fails if no permissions.


ennru (approved):
Unrelated build failure.

LGTM.

PR #2208, opened 2024-09-10
chore: adding docs about how to add headers to grpcClient

franciscolopezsancho:
https://github.com/lightbend/kalix-runtime/issues/2708

### DIFF
--- a/docs/src/modules/java-protobuf/pages/call-another-service.adoc
+++ b/docs/src/modules/java-protobuf/pages/call-another-service.adoc
@@ -70,6 +70,34 @@ include::example$scala-protobuf-doc-snippets/src/main/scala/com/example/Delegati
 <4> Once we get a response for the `getCurrentCounter` we transform that into a `Response` - the response type of this action method.
 <5> We can reply with a `Future[Result]` using `effects.asyncReply`. Once the `Future` completes, the client will get the response back.
 
+To pass headers to this client you need to `lift` (by casting) the service interface to the specific client.
+In this example, casting the service interface `com.example.CounterService` to the client class `com.example.CounterServiceClient`.
+Once this is done you can call the desired method to create a request builder, add the headers, and then `invoke` the builder with the desired input.
+
+In our delegating implementation:
+
+Java::
++
+[source,java,indent=0]
+.src/main/java/com/example/DelegatingServiceAction.java
+----
+include::java-protobuf:example$java-protobuf-doc-snippets/src/main/java/com/example/DelegatingServiceAction.java[tag=delegating-action-lifted]
+----
++
+<1> Casting, creating the request builder with `increase()`, and adding the headers to it.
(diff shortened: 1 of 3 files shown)


samples/java-protobuf-doc-snippets/src/main/java/com/example/DelegatingServiceAction.java
```diff
@@ -37,4 +38,29 @@ public Effect<DelegatingServiceApi.Result> addAndReturn(DelegatingServiceApi.Req
...
+    // tag::delegating-action-lifted[]
+    SingleResponseRequestBuilder<CounterApi.IncreaseValue, Empty> counterServiceClientBuilder
+            = ((CounterServiceClient) counterService).increase().addHeader("key", "value"); // <1>
+
+
+    CounterApi.IncreaseValue increaseValue = CounterApi.IncreaseValue.newBuilder()
+            .setCounterId(request.getCounterId())
+            .setValue(1)
+            .build();
+    CompletionStage<Empty> increaseCompleted = counterServiceClientBuilder.invoke(increaseValue);  // <2>
```
johanandren, line 53: I'd say the canonical use is to never catch the builder, avoiding the type soup of the builder

```
CompletionStage<Empty> increaseCompleted = 
    ((CounterServiceClient) counterService).increase()
        .addHeader("key", "value") // <1>
        .invoke(increaseValue); // <2>

File: docs/src/modules/java-protobuf/pages/call-another-service.adoc [hunk 2]
```diff
@@ -70,6 +70,34 @@ include::example$scala-protobuf-doc-snippets/src/main/scala/com/example/Delegati
...
++
+[source,scala,indent=0]
+.src/main/scala/com/example/DelegatingServiceAction.scala
+----
+include::example$scala-protobuf-doc-snippets/src/main/scala/com/example/DelegatingServiceAction.scala[tag=delegating-action-lifted]
+----
++
+<1> Casting, creating the request builder with `increase()`, and adding the headers to it.
+<2> Invoke the builder with the desired `increaseValue` parameter.
+
```
johanandren, line 100: Only documenting this for Java, what about Scala?

johanandren (approved):
LGTM (but we should maybe consider making it possible to look up a client using the client class directly)

samples/java-protobuf-doc-snippets/src/main/java/com/example/DelegatingServiceAction.java
franciscolopezsancho: I wasn't sure. I wanted the user to know about that class. I think it's clearer in that respect, what's happening underneath. But it's true that it's a bit messy. 
I can use the canonical/cleaner way, np.

franciscolopezsancho: done

PR #2219, opened 2024-09-27
refactor: change format for logMessage mkString so it shows with tabs…

girdharshubham:
… instead of new lines which doesn't get shipped through log aggregators as a complete message and can lead to confusion

References
-

### DIFF
--- a/sdk/java-sdk-protobuf/src/main/scala/kalix/javasdk/impl/DiscoveryImpl.scala
+++ b/sdk/java-sdk-protobuf/src/main/scala/kalix/javasdk/impl/DiscoveryImpl.scala
@@ -189,7 +189,7 @@ class DiscoveryImpl(
     val detail = if (in.detail.isEmpty) Nil else List(in.detail)
     val seeDocs = DocLinks(sdkName).forErrorCode(in.code).map(link => s"See documentation: $link").toList
     val messages = message :: detail ::: seeDocs ::: sourceMsgs
-    val logMessage = messages.mkString("\n\n")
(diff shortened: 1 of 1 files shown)


sdk/java-sdk-protobuf/src/main/scala/kalix/javasdk/impl/DiscoveryImpl.scala
```diff
@@ -189,7 +189,7 @@ class DiscoveryImpl(
     val detail = if (in.detail.isEmpty) Nil else List(in.detail)
     val seeDocs = DocLinks(sdkName).forErrorCode(in.code).map(link => s"See documentation: $link").toList
     val messages = message :: detail ::: seeDocs ::: sourceMsgs
-    val logMessage = messages.mkString("\n\n")
+    val logMessage = messages.mkString("\t|\t")
```
octonato, line 192: is the pipe `|` intentional?

ennru (approved):
LGTM.

sdk/java-sdk-protobuf/src/main/scala/kalix/javasdk/impl/DiscoveryImpl.scala
girdharshubham: Yes, I thought I should put a delimiter just in case.

ennru: It makes sense where tabs aren't rendered.

PR #2222, opened 2024-10-04
chore: handle UpdateState workflow entity protocol message

aludwiko:


sdk/java-sdk-protobuf/src/main/scala/kalix/javasdk/impl/workflow/WorkflowImpl.scala
```diff
@@ -339,6 +339,11 @@ final class WorkflowImpl(system: ActorSystem, val services: Map[String, Workflow
 
         case Empty =>
           throw ProtocolException(init, "Workflow received empty/unknown message")
+
+        case _ =>
+          // ignore, UpdateState from the runtime and update this case when introducing replication to Kalix
+          // currently added to satisfy the compiler
+          Future.successful(WorkflowStreamOut(WorkflowStreamOut.Message.Empty))
```
johanandren, line 346: So existing services would fail with the latest runtime, that's unfortunate

johanandren (approved):

PR #2239, opened 2024-11-08
fix: deserializing json.akka.io

franciscolopezsancho:
When updating to 1.2.0

References
- https://github.com/lightbend/kalix-jvm-sdk/actions/runs/11687496895/job/32546106625#step:8:1283
- https://app.circleci.com/pipelines/github/lightbend/kalix-runtime/8522/workflows/0787e9b2-1692-4f0a-8944-98c417b20e4c/jobs/70684?invite=true#step-118-102946_94

PR #2240, opened 2024-11-08
fix: Support new json type url

johanandren:
Continuation of #2239 

For compatibility with runtime 1.2.0

Also bumps runtime to 1.2.0

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -10,7 +10,7 @@ object Dependencies {
     val RuntimeVersion = System.getProperty(
       "kalix-runtime.version",
       // temporarily accept the old system property name
-      System.getProperty("kalix-proxy.version", "1.1.46"))
+      System.getProperty("kalix-proxy.version", "1.2.0"))
   }
 
   // changing the Scala version of the Java SDK affects end users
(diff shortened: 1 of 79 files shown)


project/Dependencies.scala
```diff
@@ -10,7 +10,7 @@ object Dependencies {
     val RuntimeVersion = System.getProperty(
       "kalix-runtime.version",
       // temporarily accept the old system property name
-      System.getProperty("kalix-proxy.version", "1.1.46"))
+      System.getProperty("kalix-proxy.version", "1.2.0"))
```
efgpinto, line 13: If we are bumping here, should we also bump the remaining places? Or maybe we do on a separate PR?

efgpinto (approved):
LGTM

project/Dependencies.scala
johanandren: Ah, I didn't notice that from Ciscos original PR, hmm. Might be fine to do here, to make sure tests are run against the new version. We anyway want that bump.

efgpinto:
I jsut remembered we have some hardcoded ones [here](https://github.com/lightbend/kalix-jvm-sdk/blob/f46439ce1b7180caba9fe7cc7025248aa2e310de/sdk/java-sdk-protobuf-testkit/src/main/scala/kalix/javasdk/testkit/impl/EventingTestKitImpl.scala#L384)

johanandren:
I think that's fine in that direction, the important part here is to accept the json.akka.io typeurl when it shows up

PR #2249, opened 2024-12-10
Update release-issue-template.md

sebastian-alfers:
Manual verification to ensure that the pom is at least one time resolved through Cloudsmith, which means it will appear on our repo index UI.


efgpinto (approved):
Seems reasonable. CI complains though. If you can have a look, fine. Otherwise, I can look at it as well.

PR #2260, opened 2025-01-30
chore: avoid wrong deprecation message

franciscolopezsancho:
To avoid `[WARNING] Using 'entity_type: "counter"' is deprecated, replace it with 'type_id: "counter"` on `mvn compile`

on the following .proto without `entity_type` defined.
```
service CounterService {
  option (kalix.codegen) = {
    value_entity: {
      name: "com.westpac.domain.Counter"
      type_id: "counter"
      state: "com.westpac.domain.CounterState"
    }
  };
  option (kalix.service).acl.allow = { principal: ALL };

  rpc Increase(IncreaseValue) returns (google.protobuf.Empty);
  rpc Decrease(DecreaseValue) returns (google.protobuf.Empty);
  rpc Reset(ResetValue) returns (google.protobuf.Empty);
  rpc GetCurrentCounter(GetCounter) returns (CurrentCounter);
}
```

### DIFF
--- a/codegen/core/src/main/scala/kalix/codegen/ModelBuilder.scala
+++ b/codegen/core/src/main/scala/kalix/codegen/ModelBuilder.scala
@@ -18,6 +18,7 @@ import kalix.View
 import kalix.WorkflowDef
 
 import scala.annotation.nowarn
+import scala.annotation.nowarn
 
 /**
  * Builds a model of entities and their properties from a protobuf descriptor
@@ -709,6 +710,7 @@ object ModelBuilder {
       State(resolveMessageType(workflowDef.getState, protoPackageName, additionalDescriptors)))
   }
 
+  @nowarn("msg=deprecated")
   private def extractValueEntity(
       serviceProtoDescriptor: ServiceDescriptor,
       entityDef: ValueEntityDef,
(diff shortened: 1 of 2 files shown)


codegen/core/src/main/scala/kalix/codegen/ModelBuilder.scala
```diff
@@ -709,21 +710,23 @@ object ModelBuilder {
       State(resolveMessageType(workflowDef.getState, protoPackageName, additionalDescriptors)))
   }
 
+  @nowarn("msg=deprecated")
```
octonato, line 713: I'm confused. Calling `getEntityType` gives a deprecation and therefore we are excluding with `@nowarn`? 

Why we are doing this?

franciscolopezsancho: The `@nowarn` to allow compilation.

efgpinto: Got it now.. had to look at https://github.com/lightbend/kalix-jvm-sdk/pull/2260/files#diff-a2f65419f287e3a9648619b8f1f6b70e1da8169481f66d2aae60ccb0b84c43fbR793 to understand it better.

Btw, same fix was applied previously [here](https://github.com/lightbend/kalix-jvm-sdk/pull/2167) but we missed these 2 cases I guess.

efgpinto (approved):
LGTM

PR #2270, opened 2025-03-03
bump: Broad sweep of dependency bumps

johanandren:
* Akka 2.10.2
 * Akka HTTP 10.7.0
 * Akka gRPC 2.5.3
 * Logback 1.5.17
 * Commons IO 2.18.0
 * Drop explicit dep on slf4j

Drop the one sample of a spring client so that we don't have to bother with keeping spring and the multitude of deps updated.

Also bumps other direct dependencies from in the tracing sample.

References
- https://github.com/lightbend/kalix-runtime/issues/3398


patriknw (approved):
LGTM

Issue #2280, opened 2025-04-04
Grpc Client Provider for Workflow in Kalix SDK


max8github:
Currently there is no way to get hold of a grpc client provider from Workflows in the Scala SDK.
For Actions it is available through the `action context`, but not for Workflows.
(Request coming from [Visma, Laurin](https://typesafe.slack.com/archives/C04SL082Q75/p1742468429793339))

johanandren:
In Kalix JVM SDK 1.5.8, released just now.

PR #2281, opened 2025-04-07
feat: Access to gRPC clients in through workflow context

johanandren:
References
- https://github.com/lightbend/kalix-jvm-sdk/issues/2280


johanandren:
For the record the problem was that we were still using artifacts in repo.scala-sbt.org which had outtage or was shut down today: https://floss.social/@scala_sbt/114295724150536181

efgpinto (approved):

PR #2283, opened 2025-04-07
feat: Run user logic on virtual threads

johanandren:
Waiting for upstream Akka core release.

### DIFF
--- a/sdk/java-sdk-protobuf/src/main/scala/kalix/javasdk/impl/action/ActionsImpl.scala
+++ b/sdk/java-sdk-protobuf/src/main/scala/kalix/javasdk/impl/action/ActionsImpl.scala
@@ -25,6 +25,7 @@ import kalix.protocol.component.{ Failure, MetadataEntry }
 import org.slf4j.{ Logger, LoggerFactory, MDC }
 
 import java.util.Optional
+import scala.concurrent.ExecutionContext
 import scala.concurrent.Future
 import scala.jdk.CollectionConverters.SeqHasAsJava
 import scala.jdk.OptionConverters._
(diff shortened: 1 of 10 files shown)


johanandren:
Same build problem as #2281 (I'll hold off doing anything more here until that is merged)

sdk/java-sdk-protobuf/src/main/scala/kalix/javasdk/impl/action/ActionsImpl.scala
```diff
@@ -258,25 +265,32 @@ private[javasdk] final class ActionsImpl(_system: ActorSystem, services: Map[Str
...
-          service.factory
-            .create(context)
-            .handleStreamedOut(in.name, MessageEnvelope.of(decodedPayload, context.metadata()), context)
-            .asScala
-            .mapAsync(1)(effect => effectToResponse(service, in, effect, service.messageCodec))
-            .recover { case NonFatal(ex) =>
-              // user stream failed with an "unexpected" error
-              handleUnexpectedException(service, in, ex)
+        // Note: invocation in future to guarantee create and invocation is running on sdk dispatcher with virtual thread support
+        Source
```
franciscolopezsancho, line 269: question: without this we would run the `catch` bit on the default pool?

johanandren: Without this we would call `service.factory.create(context)` and then `handleStreamedOut` method on the returned instance, both invokes user logic which could be blocking, on the thread that called the surrounding `handleStreamedOut`, on default dispatcher (unless you tweak how the service is bound to an actual server listener, which we don't).

PR #2296, opened 2025-04-11
feat: Access to component gRPC clients in actions and workflows

johanandren:
Follow up to #2281


johanandren:
No specific test coverage, but double checked by changing the scala-protobuf-transfer-workflow like this and then running integration test which exercises workflow:

```scala
 val withdraw = step("withdraw") // <1>
      .asyncCall { (withdraw: Withdraw) =>
        val withdrawRequest = WithdrawRequest(withdraw.from, withdraw.amount)
        context.getComponentGrpcClient(classOf[WalletService]).withdraw(withdrawRequest)
      } // <2>
```

aludwiko (approved):
LGTM

PR #2305, opened 2025-04-29
feat: workflow delete effect

aludwiko:


project/Dependencies.scala
```diff
@@ -9,7 +9,7 @@ object Dependencies {
     val RuntimeVersion = System.getProperty(
       "kalix-runtime.version",
       // temporarily accept the old system property name
-      System.getProperty("kalix-proxy.version", "1.2.15"))
+      System.getProperty("kalix-proxy.version", "1.2.16-3-9f15c59e-dev-SNAPSHOT"))
```
efgpinto, line 12: friendly reminder to revert before merge

docs/src/modules/java-protobuf/pages/workflows.adoc
```diff
@@ -179,6 +179,36 @@ IMPORTANT: For simplicity purposes, we are returning the internal state directly
 
 A full transfer workflow source code is available https://github.com/lightbend/kalix-jvm-sdk/tree/main/samples/java-protobuf-transfer-workflow[here (Java Protobuf), {tab-icon}, window="new"] or https://github.com/lightbend/kalix-jvm-sdk/tree/main/samples/scala-protobuf-transfer-workflow[here (Scala Protobuf), {tab-icon}, window="new"]. Follow the `README` file to run and test it.
 
+
+== Deleting state
+
+If you want to delete the workflow state, you can use the `delete` method from the effect API. This will remove the workflow from the system.
```
efgpinto, line 185: Seems more clear when we say that we are deleting the workflow rather than deleting the workflow state (as if something would still remain).

```suggestion
If you want to delete the workflow, you can use the `delete` method from the effect API.
```

efgpinto (approved):

PR #2382, opened 2025-11-12
fix: Concurrency issue in build plugin docker integration

johanandren:
Saw this fail a couple of times yesterday in CI for #2365


octonato (approved):
LGTM, but still think StringBuffer is better. 

It's not that it's hard to follow. It's just that StringBuffer is simpler and delivers the same. 

Anyway, no strong opinions on this. Feel free to merge as is.

PR #2396, opened 2026-02-02
feat: Start views and actions from snapshot

patriknw:


samples/java-protobuf-eventsourced-customer-registry/src/main/proto/customer/view/customer_view.proto
```diff
@@ -73,6 +73,18 @@ service CustomerByName {
...
+  rpc ProcessSnapshot(domain.CustomerState) returns (api.Customer) {
+    option (kalix.method).eventing.in = {
+      event_sourced_entity: "customers"
+      start_from_snapshots: true
+    };
+    option (kalix.method).view.update = {
+      table: "customers"
+    };
+  }
+  // end::snapshot-handler[]
```
patriknw: I tested more combinations with snapshots in this sample:
- plain action
- s2s to the subscriber service, which has a view

PR #2420, opened 2026-04-08
update release template to remove old repo

JustinPihony:
Updated the Maven dependency check instruction to include a placeholder for the resolver token.


johanandren:
Better to drop the entire maven.repo property and it will just work copy-paste on our machines since we have the repo set up.

JustinPihony:
Changed

octonato (approved):

PR #2431, opened 2026-06-05
chore: Improved event sourced and value entity replay/startup error reporting

johanandren:
Currently a bit opaque with a "Terminating entity due to unexpected failure".

This adds which entity type and what entity id had the error along with the type of the event that made it fail. Also moves such ERROR logging to a logger for the user entity implementation class.

Bom file updates is something out of sync in main, this PR does not change any dependencies.


pvlugter (approved):
LGTM. Could make sense to put the bom updates in a separate PR.

PR #2433, opened 2026-06-08
feat: exposing jvm metrics

aludwiko:
part of: https://github.com/lightbend/kalix/issues/11440

### DIFF
--- a/sdk/java-sdk-protobuf/src/main/resources/reference.conf
+++ b/sdk/java-sdk-protobuf/src/main/resources/reference.conf
@@ -168,6 +168,23 @@ kalix {
         collector-endpoint = ""
         collector-endpoint =  ${?COLLECTOR_ENDPOINT}
      }
+
+     metrics {
+        # When enabled, an embedded Prometheus exporter HTTP server is started exposing
+        # metrics on the `/metrics` endpoint (scrapeable by Prometheus).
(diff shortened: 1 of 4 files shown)


sdk/java-sdk-protobuf/src/main/resources/reference.conf
```diff
@@ -168,6 +168,23 @@ kalix {
         collector-endpoint = ""
         collector-endpoint =  ${?COLLECTOR_ENDPOINT}
      }
+
+     metrics {
+        # When enabled, an embedded Prometheus exporter HTTP server is started exposing
+        # metrics on the `/metrics` endpoint (scrapeable by Prometheus).
+        enabled = true
```
johanandren: Doesn't enabled by default in reference.conf mean also enabled in dev mode?

johanandren (approved):
LGTM with one question

sdk/java-sdk-protobuf/src/main/resources/reference.conf
aludwiko: Yes, is that a problem for dev mode?

aludwiko: Ok, I see this might be problematic for IT, ok I will set this to false by default.

PR #2453, opened 2026-08-07
fix: Toggle view state access from ViewRouter

johanandren:
This could, in some special circumstances, lead to an exception with "Current state is only available when handling updates" when View#viewState was called while handling a view update.


sdk/java-sdk-protobuf/src/test/scala/kalix/javasdk/impl/view/ViewRouterSpec.scala
```diff
@@ -0,0 +1,78 @@
...
+  }
+
+  class TestViewRouter(view: TestView) extends ViewRouter[String, TestView](view) {
+    override def handleUpdate(commandName: String, state: String, event: Any): View.UpdateEffect[String] =
+      view.handleEvent(state, event)
+  }
+
+  "ViewRouter" should {
+
+    "make the current view state available via View.viewState() while handling an update" in {
```
aludwiko, line 49: So that wasn't possible at all in the current implementation?

johanandren: Looks like that, View#viewState is marked internal API though, and we don't generally use it from generated code. One customer generates their own code that seems to access it in some circumstances.

aludwiko (approved):

